PR #56, opened 2015-04-13
Auto column width based on overall grid width

bkirk1012:


ceolter:
Thanks for the suggestion. The plan in my mind for fitting the columns is
as follows:

+) Keep it separate to the process of setting up the columns.
+) Have an API method that the client can call to auto-fix the columns into
the grid.
+) This API will see what cols are present, how much size is left /
overflowed, then size each col accordingly.

So if 300px is leftover, it would add the 300px to the columns in a ratio
equal to the current columns size.

That way, the user would call this method after the cols are set into the
grid, or after the data is set in (as the data can change the width,
depending on whether there are scroll bars or now).

Hope this makes sense!

Thanks,
Niall.

On 13 April 2015 at 17:02, Billy Kirk <notifications@github.com> wrote:

> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/ceolter/angular-grid/pull/56
> Commit Summary
>
>    - Auto column width based on overall grid width
>
> File Changes
>
>    - *M* src/angularGrid.js
>    <https://github.com/ceolter/angular-grid/pull/56/files#diff-0> (3)
>    - *M* src/colModel.js
>    <https://github.com/ceolter/angular-grid/pull/56/files#diff-1> (16)
>    - *M* src/constants.js
>    <https://github.com/ceolter/angular-grid/pull/56/files#diff-2> (1)
>    - *M* website/dist/angularGrid-latest.js
>    <https://github.com/ceolter/angular-grid/pull/56/files#diff-3> (20)
>
> Patch Links:
>
>    - https://github.com/ceolter/angular-grid/pull/56.patch
>    - https://github.com/ceolter/angular-grid/pull/56.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ceolter/angular-grid/pull/56>.
>



-- 
-- 

The information transmitted is intended only for the person or entity to
which it is addressed and may contain confidential and/or privileged
material. Statements and opinions expressed in this e-mail may not
represent those of the sender. Any review, retransmission, dissemination or
other use of, or taking of any action in reliance upon, this information by
persons or entities other than the intended recipient is prohibited. If you
received this in error, please contact the sender immediately and delete
the material from any computer.

bkirk1012:
Thanks Niall!  This makes sense to me. 

I am trying to implement this - I am looking for the best way to get the "how much size is left" number.  I have the total width of the columns in the grid (through this.eBodyContainer.style.width ) and need a standard way to get the total width of the eBody. eBody.style.width = "" , and previously I was calling eBody.clientWidth to get this "overall width" value.  Is there a more robust way to do this?

Separately - one additional point that came to mind: if autoFitColumns is true, and the total width of the columns > the overall width, then this function should size the columns regularly and toggle the horizontal scrolling.

What are your thoughts?

ceolter:
hi Billy,

I'd hold off on the auto-resize for another day . . . . I've done some major rework with the columns . . . . in particular the column model is gone, replaced with a columnController. this is in support of expanding / collapsing groups of columns. i'll hopefully be merging into the master tonight.

in the new version, there is a method on the columnController for getting the pinned container width and the body container width.

bkirk1012:
Sounds good to me - in that case, when should I revisit this auto-size feature?

ceolter:
Tomorrow!!! (if all goes to plan this evening for me)

ceolter:
This is implemented now. I haven't updated website yet, however if you check out the latest from git, there is now an API method api.sizeColumnsToFit().

Issue #66, opened 2015-04-16
Custom Renderers to accept a promise that resolves template


tannerlinsley:
It would be great for custom renderers to accept a promise that resolves the cell/header content.  Some things just needs to be async sometimes ;)

ceolter:
That's actually already possible!! The infinite scrolling does exactly that
when you scroll to a page that isn't loaded yet. What you do is create an
empty object. Then when the cell is rendered (you get notified of this by
the cell renderer getting called), you then kick off you async process.
When the process comes back, get the table re refresh the required row!

On 16 April 2015 at 23:56, Tanner Linsley <notifications@github.com> wrote:

> It would be great for custom renderers to accept a promise that resolves
> the cell/header content. Some things just needs to be async sometimes ;)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ceolter/angular-grid/issues/66>.
>



-- 
-- 

The information transmitted is intended only for the person or entity to
which it is addressed and may contain confidential and/or privileged
material. Statements and opinions expressed in this e-mail may not
represent those of the sender. Any review, retransmission, dissemination or
other use of, or taking of any action in reliance upon, this information by
persons or entities other than the intended recipient is prohibited. If you
received this in error, please contact the sender immediately and delete
the material from any computer.

tannerlinsley:
Got an example of this?

ceolter:
Just the code behind virtualization I'm afraid. Would make a good example
for the documentation alright.

tannerlinsley:
So your saying I can return a promise inside of the cellRenderer on a
column Def?

ceolter:
No you can't return a promise. Return blank from the renderer (or whatever
to display when waiting). Resolve the promise yourself, and when the result
comes back, get the grid to refresh that row - second time around rendering
the data will be there.

tannerlinsley:
I see. It feels cumbersome though.

tannerlinsley:
Are there other reasons eg. performance, etc, that the cellRenderer function could not simply return a promise?

ceolter:
Hi Tanner,

No performance on the grid side. Just from a client point of view it
doesn't make sense to me. For example, if only one column requires this,
then showing n rows would result in n simultaneous network calls (assuming
the promise is to get the data remotely). Would of though such would be
better handled in batch, then get the grid to refresh the whole view (which
is practically instantaneous). Is your use case different?

tannerlinsley:
[...]

PR #71, opened 2015-04-19
Support for simple cellRenderer promises

tannerlinsley:
Simply return a promise that resolves to your template.
Keep in mind, if you are using this to decorate or add meta data that
is not in the grid data model already, it will not be filterable or
obey aggregations…. for now.


tannerlinsley:
Fixes #66

ceolter:
Have commented on #66, don't think this should be done.

PR #78, opened 2015-04-20
fix bower.json references

zeitiger:
I use an automatic wiring logic based on bower.json meta information. It would be nice if you could merge this fix.
Thanks


ceolter:
Thanks @zeitiger  - shouldn't the style css files go in here also? If not, then the user won't be able to use the bundled styles?

PR #83, opened 2015-04-21
Fix pagination error when rowCount is null

sunilbandla:
When pagination is enabled and `rowCount` is null, `calculateTotalPages()` blows up because of the check in this line. This fix makes sure that `rowCount` is always a number.


ceolter:
Thanks Sunil

PR #121, opened 2015-05-08
Consume Unused Column Widths

pinnacleag-dev:
Uses the defined column widths to determine the remaining grid width for
columns without widths defined and


ceolter:
thanks for the PR @pinnacleag-dev  . . . . have had a look and looks nice. however a) i want to think about this for a while and i'm very busy this week (just back from holidays and work are sending me off-site in the morning, so will have little time to myself and b) i want to make sure the solution is generic enough to cater for other scenarios, as am afraid it might be to specific to your scenario only. thanks :)

ceolter:
ok i merged this in, so @pinnacleag-dev you are registered as a contributor!! but then in rewrote it in a different way. now, if you mark a column with suppressSizeToFit, it doesn't partake in the resize. this is more generic than your proposal, covers more scenarios!!

pinnacleag-dev:
Excellent, thank you!

jovana:
@pinnacleag-dev or @ceolter can you please explain how I can use this? I also want the grid fit over the total div size. Now it does not align over the rest of the space.

jovana:
Got it:

 ready: function(api) {
                    api.sizeColumnsToFit();
                }

Issue #98, opened 2015-04-26
Support multi-column sorting


sfescape:
Extend sorting to multi-column by allowing the selection of additional sort columns with shift-click.

ceolter:
Fair enough :)

ceolter:
In latest release.

sfescape:
Thanks!

semko2m:
Dear Celoter,
I am using your grid. I use the custom header. How can I get sorting number, on multi sorting? Do you have API for that? 

Kind Regards

PR #134, opened 2015-05-15
Allow sorting by multiple columns (by holding shift)

Robinson7D:
Related to issue #98 

Let me know if there are any requested changes (any functions that you want split, if `shiftKey` (as in, e.shiftKey) should be a constant/configurable, etc.)


ceolter:
just reviewed your code, all looks fine. thanks for this. i didn't realise multi-column sorting would be that easy!!

Robinson7D:
No problem. The code base is wonderfully clean, so it was easy to figure out and add. Thanks for the excellent grid library!

ceolter:
wonderfully clean!!! indeed!!! something that's pretty rare these days... a forgotten art in today's world... :)

PR #152, opened 2015-05-22
Package.json update for browserify compatibility

nathanmarks:
Can you add this in so I can properly `require('angular-grid')` using browserify?

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "angular-grid",
   "version": "1.5.0",
   "description": "High performance and feature rich data grid for AngularJS",
-  "main": "main.js",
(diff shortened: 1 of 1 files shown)


ceolter:
thanks nathan. i have a question - when you require('angular-grid'), shouldn't it be using what's in the dist folder??

nathanmarks:
@ceolter Total oversight on my part. I had originally tried it with the non-minified version which didn't work. This created issues due to browserify trying to follow the requires. However in the minified version the require calls are renamed and subsequently ignored by browserify.

nathanmarks:
@ceolter submitted another request changing this, nice catch.

Issue #137, opened 2015-05-18
Scroll to Column


ceolter:
Currently we have scroll to row index. Also need scroll to column index.

ceolter:
fixed in 1.8.0

PR #179, opened 2015-06-04
Scroll to Column #137

froakiemon:
I take a quick look at how you implemented the row scroll via ensureIndexVisible(index), and provide a similar implementation for the column scroll. I modify ensureIndexVisible method to pass in an optional colIndex parameter. I tested and verify that this change works well for me, so I just want to share the code changes for you to review (and to include in next release if you are ok with it).

I did not change the original "index" parameter name to "rowIndex", so please update the parameter name as you see fit.

### DIFF
--- a/src/js/grid.js
+++ b/src/js/grid.js
@@ -381,7 +381,7 @@ Grid.prototype.ensureNodeVisible = function(comparator) {
     }
 };
 
-Grid.prototype.ensureIndexVisible = function(index) {
+Grid.prototype.ensureIndexVisible = function(index, colIndex) {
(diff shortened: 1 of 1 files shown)


ceolter:
I fixed it up as follows:
+ didn't take into account visible vs hidden columns (used in header groups)
+ didn't take into account cols of different widths, or cols that are resized after first shown
+ gridOptions.colWidth is optional, code broke if missing
+ didn't work with pinned columns
+ broken into separate method, so now different method for col and row - makes API less confusing when only requiring one or the other

PR #285, opened 2015-07-02
UnSort Icon

lucasvall:
Hi. I implemented an optional icon when a column is unsorted (activated at a column level). This is popular with [other grids](https://www.datatables.net/examples/basic_init/table_sorting.html)

Summary of changes:
- Add class to Asc & Desc Icons containers (to style them better)
- Add unSortIcon option to columns: if true an unSort Icon is shown when the column is not sorted
- Icon can be customised with gridOptions.icon.sortUnSort (excellent to use something like `<i class="fa fa-sort"></i>` from font-awesome
- Add SVG factory for the unSort Icon
- Adjusted Documentation (including one example)

### DIFF
--- a/docs/angular-grid-column-definitions/index.php
+++ b/docs/angular-grid-column-definitions/index.php
@@ -93,6 +93,10 @@ include '../documentation_header.php';
             <td>Set to true if no sorting should be done for this column.</td>
         </tr>
         <tr>
+            <th>unSortIcon</th>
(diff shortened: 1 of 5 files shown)


ceolter:
thanks. merged and changed as follows:
+ new class names following ag- standard
+ added option to gridOptions to set for all columns

lucasvall:
+1. Thanks

PR #299, opened 2015-07-09
Pivot Groups use valueGetters when available

tannerlinsley:


ceolter:
code didn't behave as expected!

the value getter was passed the node. however to be consistent with how value getters work, it should of been passed a params object with node, data, api, context etc.

changed to be consistent and reuse utils class (where this logic is shared). correct fix now checked.

thanks!

PR #313, opened 2015-07-14
Issue 279 ensureIndexVisible

SteveQiu:
Hi Ceolter

This pull request may fix several issue related to the use of incorrect rowModel

SDQ


ceolter:
method setDatasource() already set the row model in the services, further down in the method. however one was missing, the gridPanel, which I've now fixed. thanks for pointing me in this direction.

PR #314, opened 2015-07-15
Some typescript definitions

AkosLukacs:
Hi!

Created TypeScript type definition for ColDef based on the documentation and the code itself. In two or three cases TypeScript compiler couldn't figure out what to use from union types, so had to add a few explicit casts. Or had to reorder type checking so tsc can figure out the actual type used with union types. This union type handling should improve in TS 1.5.

Also added some  generics to the array/object iteration functions the utils.ts. Therefore some `any` typedefs could be could removed from the usage of those functions, but getting actual type safety based on the type of the array.

The generated javascript is pretty much the same as before, "dist" file not changed.


ceolter:
thanks for that. i intend moving all the standard classes to proper types in time. you bed me to it for this part!

PR #316, opened 2015-07-15
294 - Column change callbacks on gridOptions.

wagonburger:
Added columnResized, columnVisibilityChanged and columnOrderChanged callbacks to gridOptions.


ceolter:
thanks. merged after fixing the 'visiblity', what you had was getting called whenever the item was clicked. not just when it was marked as visible.

PR #317, opened 2015-07-15
Adding isInverted parameter to the signature of the column comparator

wagonburger:
We have a use case where we are flattening a parent-child hierarchy in the grid. The custom sort treats all children as if they have the same values as their parent so when sorting children are always next to their parents. To do this we need to know if the sort is ASC or DESC as children need to always be below their parent no matter the direction of the sort

### DIFF
--- a/docs/angular-grid-sorting/index.php
+++ b/docs/angular-grid-sorting/index.php
@@ -41,7 +41,7 @@ include '../documentation_header.php';
     </p>
 
     <pre>
-colDef.comparator = function (value1, value2, data1, data2) {
+colDef.comparator = function (value1, value2, data1, data2, isInverted) {
     return value1 - value2;
 }</pre>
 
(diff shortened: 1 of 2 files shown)


docs/angular-grid-sorting/index.php
```diff
@@ -41,7 +41,7 @@
     </p>
 
     <pre>
-colDef.comparator = function (value1, value2, data1, data2) {
+colDef.comparator = function (value1, value2, data1, data2, isInverted) {
```
wagonburger: Should this be a documented parameter? This may confuse users into thinking they have to take into direction in their custom sort which in 99% of cases you don't.

ceolter: everything needs to be documented!!!! with regards the PR itself, i will take a close look when i'm not in work :)

ceolter:
to be honest, i didn't understand what it was you are trying to do. but i merged anyway cos' it was a very small change, and having the extra info in is no harm.

PR #325, opened 2015-07-17
283 minWidth for sizeColumnsToFit

SteveQiu:
283 minWidth for sizeColumnsToFit
283 web page update

Hi Ceolter,

I made some modification on the ag-grid regarding issue 283 - minimum width for sizeColumnsToFit. I hope you like it.

SDQ

### DIFF
--- a/docs/angular-grid-api/index.php
+++ b/docs/angular-grid-api/index.php
@@ -51,6 +51,10 @@ include '../documentation_header.php';
                 open, will stay.</td>
         </tr>
         <tr>
+            <th>sizeColumnsToFit()</th>
+            <td>Automatically resize the columns to fill the screen</td>
+        </tr>
+        <tr>
             <th>selectAll()</th>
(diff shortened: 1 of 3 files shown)


ceolter:
almost works, but problem i see with your implementation is, if the last column has minWidth (but the others don't), then the grid can end up with empty space on the right hand side.

for it to work correctly, all the columns with minWidth should be calculated first, and then the remaining spread across the cols with no minWidth.

on top of minWidth, maxWidth should also be applied. makes sense to do both. not sure how the algorithm would work for that.

and then in addition to sizeColsToFit, minWidth and maxWidth should be applied when resizing the columns normally! but i guess that's a separate issue. getting it working with sizeColsToFit can be done as a first step.

SteveQiu:
Thanks for pointing that out. I am gonna work on it.

JasonZough:
Have this feature been released ? I have same demand.

PR #331, opened 2015-07-21
issue 283 min width for sizeColumnToFit

SteveQiu:
Hi Ceolter

Here is the version for issue 283

SDQ

### DIFF
--- a/docs/angular-grid-api/index.php
+++ b/docs/angular-grid-api/index.php
@@ -51,6 +51,10 @@ include '../documentation_header.php';
                 open, will stay.</td>
         </tr>
         <tr>
+            <th>sizeColumnsToFit()</th>
+            <td>Automatically resize the columns to fill the screen</td>
+        </tr>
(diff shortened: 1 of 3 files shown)


ceolter:
thanks. may be a day or to before i get a chance to review, but i'll get back to you once i do.

SteveQiu:
Hi, how's everything going? Have you had a chance to look at the code?

ceolter:
funny, i am actually looking at it right now!!! spent about 20min looking at it. i'm going to have think about it more over the weekend (currently in work).

the implementation is tricky. i'm going to see if i can come up with a cleaner algorithm!! maybe i won't be able to.

ceolter:
hi, after thinking about it a lot, and taking inspiration from your code, i've decided to go with a different implementation. i've just checked it in. the code should be easier to maintain and also caters for min and max widths. i've also updated column resize to use the min and max widths also.

PR #333, opened 2015-07-22
fixes #332

### LINKED ISSUE #332: First column's cells don't get the `ag-cell-focus` class
There is a bug that doesn't allow cells in the first column to be focused. Apparently an empty string is assigned to the col attribute (`col=""`) instead of 0 (`col="0"`). That leads to the following line not finding any nodes to set the class:

```js
_.querySelectorAll_replaceCssClass(this.eParentOfRows, '.ag-cell-focus', 'ag-cell-focus', 'ag-cell-no-focus');
```

The bug seems to be due to type coercion from Number to Boolean in [renderedCell.ts](https://github.com/ceolter/angular-grid/blob/032516b0c796a2fa8a99f0f9c2edf4ef522baed3/src/ts/rendering/renderedCell.ts#L106):

```js
this.vGridCell.setAttribute("col", this.column.index ? this.column.index.toString() : '');
```

JannesMeyer:


ceolter:
thank you

JannesMeyer:
You are very welcome! Thank you for the great library!

PR #335, opened 2015-07-24
fix ( updateGroupParentsIfNeeded ) was iterating 1 more time than needed

christoffee:
It was iterating one more time then it needed to. As last row was the count.


ceolter:
are you sure this was wrong???

this is the loop that inserts the rows (line 225 of rowRenderer):
for (var rowIndex = this.firstVirtualRenderedRow; rowIndex <= this.lastVirtualRenderedRow; rowIndex++) {

ceolter:
no comment back, closing

PR #345, opened 2015-07-28
Adding the `suppressGroupRow` option

ocombe:
I also updated the docs accordingly.
I didn't include the docs/dist/angular-grid.js since you said that you will generate this yourself, but you'll need to generate it to test the example that I added into the documentation.

### DIFF
--- a/docs/angular-grid-grouping/example6.html
+++ b/docs/angular-grid-grouping/example6.html
@@ -1,0 +1,13 @@
+<html>
+<head>
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.3.8/angular.min.js"></script>
+    <!-- you don't need ignore=notused in your code, this is just here to trick the cache -->
+    <script src="../dist/angular-grid.js?ignore=notused2"></script>
+    <link rel="stylesheet" type="text/css" href="../dist/angular-grid.css?ignore=notused2">
+    <link rel="stylesheet" type="text/css" href="../dist/theme-fresh.css?ignore=notused2">
+    <script src="example6.js"></script>
+</head>
+<body ng-app="example" ng-controller="exampleCtrl">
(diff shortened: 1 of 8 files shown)


ceolter:
PR accepted. I made some changes:

Changed docs a bit.

Changed suppressGroupRow to groupSuppressRow - to keep consistent with all the other group properties starting with 'group'.

The change is fine, and it works, however I think it doesn't properly demonstrate what this feature does. The example (appreciated btw) doesn't really get across what this feature is doing. To get it across, then would need to have the grid render somehow showing this.

One option would be to only populate the group cell for the first entry in the group. To do this, the node would need to have the 'child index'.

Another option would be to have a border at the bottom of each group?? (again, node would need a flat 'first in group' and / or 'last in group'.

Or could have a different background color for the group (would need node group index, then cell class could have odd / even colors for the group).

Anyway, I'm sure you have it implemented, exactly as you need it!!! So maybe this is for me to do :(

ocombe:
ahah in fact I'll be implementing a different color for the group & a border around it, I didn't think that it was something that you might need and I was starting to make a complicated solution on my own.
I can make a new PR if you want ?
Do you have suggestions on how to do that ? By adding new parameters ? Or do you think that I can do this we what we have now ?

ceolter:
ok i've just checked a headstart in.

now in the rowController, i tag each node with:
node.firstChild
node.lastChild
node.childIndex

because this is on the node, it is accessable to the callbacks that do the styling.

i updated your example, to highlight where the row is the first child.

i believe that's all that is needed FROM THE GRID. the rest of the work is up to the example to apply appropriate style based on what is presented here. do you want to work with that and improve the demo?

right, i've got to get to bed, nite.

ocombe:
Sure I'll do that tomorrow, thanks ! :)

PR #347, opened 2015-07-28
Add in default sort direction for a columnDef

321ckatz123:
I could find no way to tell the column to sort by descending by default. In our use case, we are showing a series of numbers and the higher the number is the more important the number is, so we want to show it first.

I was going to create a bug, but saw how easy this would be to add it via a columnDef, so just created the pull request first instead of a bug. 

I'm also not 100% sure why my commits came across as unknown :/

### DIFF
--- a/docs/angular-grid-column-definitions/index.php
+++ b/docs/angular-grid-column-definitions/index.php
@@ -101,6 +101,10 @@ include '../documentation_header.php';
             <td>Set to true if no menu should be shown for this column header.</td>
         </tr>
         <tr>
+            <th>defaultSort</th>
+            <td>Set to "descending" to sort this column descending by default, else 'ascending'. If this isn't specified, it will default to 'ascending'.</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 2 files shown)


ceolter:
i'm in the middle of large refactoring and my latest change has caused conflicts with this PR :(

however, looking at the code, i don't think it works.

the logic in getNextSortDirection, if default is desc, then goes from no sort, to descending, to no sort (ie skips the ascending). just visual inspection, i didn't test (due to the conflicts).

if you fix the conflicts before i hit the code again, i'll check it out and verify that (or you can fix if what i say makes sense and i'm right)

321ckatz123:
Updated. It should work a lot better now. Additionally, it would probably make sense to add a isSuppressAscSort() method so both ASC and DESC work the same way, but as this won't get merged due to your refactoring anyway...

321ckatz123:
With the refactoring, I will create a new pull request (if needed, I haven't looked yet).

PR #354, opened 2015-07-31
remove values in a filter when don't influence showing data in grid like in Excel

orolin:
Hello,

this feature is inspirated by a behavior in Excel connected with checkbox filters(SetFilter in grid). For example, if you have a grid with countries and continents and turn off Europa in a filter, setFilter in countries filter will not include countries situated in Europe.

To turn on this feature, just set enableFilterExcel in gridOption as true.

If you want to create your own filter with supporting this feature of filtering, you have to implement method setFilteredDisplayValues.


ceolter:
tnks, will have to look at it at a later point, the coming few days are very busy for me.

orolin:
Should you look at this patch? Because we use it in our project and it seems it works

ceolter:
sorry for the delay, i have been extremely occupied the last two weeks. i will try and look at it this weekend if i don't get to it sooner.

src/ts/filter/filterManager.ts
```diff
@@ -140,11 +140,13 @@ module awk.grid {
             return filterPresent;
         }
 
-        private doesFilterPass(node: any) {
+        private doesFilterPass(node: any, skipColumnFilter: any) {
```
ceolter: why do you have skipColumnFilter? it looks like nothing is ever passed for this when called from inMemoryRowController??

ceolter:
i've checked this out locally and made a lot of changes. i want to do more changes, but am afraid of getting conflicts, so am going to merge, and then make more changes against another branch (probably straight to master).

Some issues:
1) Didn't work with quick filter.
2) Didn't work when loading new rows and newRowsAction=keep
3) Sould use gridOptionsWrapper and not gridOptions directly

I'm doing refactoring to allow fixing some of the above.

The general idea I liked (and kind of impressed you got that far), but need to clean it up as above, and also restructure the code so logic belongs where I think it should be.

PR #356, opened 2015-08-03
Add in default sort direction for a columnDef and suppressAscSort for gridOptions

321ckatz123:
Adding in a default sort direction for a column and adding in a suppressAscSort for the grid options (to match the suppressDescSort, which helps clean up the logic).

In the use case this is needed for, we are showing a series of numbers and the higher the number is the more important it is, so our users want us to show it first.

Tested against the examples and everything appears to work. This is a duplicate of pull request #347, but done against the refactoring effort (which, by the way, I rather like more than it used to be. It was quite easy to find and make this change).

### DIFF
--- a/docs/angular-grid-column-definitions/index.php
+++ b/docs/angular-grid-column-definitions/index.php
@@ -105,6 +105,10 @@ include '../documentation_header.php';
             <td>Set to true if no sorting should be done for this column.</td>
         </tr>
         <tr>
+            <th>defaultSort</th>
+            <td>Set to ASC, if the column should sort ascending by default, or DESC, if the default should be descending. Columns sort ASC by default.</td>
+        </tr>		
+        <tr>
             <th>unSortIcon</th>
             <td>Set to true if you want the unsorted icon to be shown when no sort is applied to this column.</td>
         </tr>
--- a/docs/angular-grid-grid-options/index.php
+++ b/docs/angular-grid-grid-options/index.php
@@ -84,6 +84,10 @@ include '../documentation_header.php';
             <td>Set to true or false. If true, sorting descending is disabled.</td>
         </tr>
         <tr>
+            <th>suppressAscSort</th>
+            <td>Set to true or false. If true, sorting ascending is disabled.</td>
+        </tr>
+        <tr>
             <th>unSortIcon</th>
             <td>Set to true to show the 'no sort' icon.</td>
         </tr>
--- a/src/ts/entities/colDef.ts
+++ b/src/ts/entities/colDef.ts
@@ -69,6 +69,9 @@ module awk.grid {
         /** Set to true if no sorting should be done for this column. */
         suppressSorting?: boolean;
 
+        /** Set to ASC, if the column should sort ascending by default, or DESC, if the default should be descending. Columns sort ASC by default. */
+        defaultSort?: string;
+
         /** Set to true if you want the unsorted icon to be shown when no sort is applied to this column. */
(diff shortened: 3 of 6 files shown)


321ckatz123:
I apologize for the goofy formatting as well. Locally I don't see that...

ceolter:
what is defaultSort used for? i don't see any reference to it in the code apart from the definition.

btw, i've just added in 'sort' and 'sortedAt' as colDefs attributes, to allow having a col sorted by default. i don't know if this is what you intended with defaultSort???

321ckatz123:
It's used in src/ts/rendering/renderedHeaderCell.ts on line 256 in my changed file. 

As for the intention, I was not adding this so that something could be sorted by "default" (so clearly it is poorly named), but more, when I go to sort this column AFTER the grid is rendered and I click the header, should the initial sort be DESC or ASC. The grid currently assumes everything should be sorted ASC first. As such, I don't think this is a duplicate of 'sort' and 'sortedAt'.

For an actual case study (I'll stick to the Olympic swimmers you use in the examples), say you had a grid that contained swimmer, country name, and # of golf medals and the data was initially sorted by swimmer. When I wanted to sort by gold medals, the current grid would start showing the swimmers in order from 0 through most medals THEN most through 0. It would be MORE ideal if we showed Phelps at the top of the grid when this column was sorted instead of the hundreds of swimmers who never got a gold medal (so the default sort direction would be DESC, not ASC) as it is much more likely that people want to see the "most" of something first. This would make sense for pretty much any type of statistical column, money, etc.

I can change it to "initialSort" or "initialSortDirection" or "sortDirection" or ... if that helps (as I agree that "defaultSort" would confuse people when you can also have a default sort on the grid).

ceolter:
cool understood. maybe 'initialSort' would be a better name.

don't worry about changing it though, i can change when i merge the PR.

i also want to think about it for a few hours . . . . because we have so many options now with sorting, the getNextSortDirection method is getting confusing.

ceolter:
i will be accepting this pr, with some changes, just bear with me!!! busy few days . . . .

surjikal:
Would be great if this could be applied table-wide as well (i.e. in `grid.options`). Not crucial though.

ceolter:
> Would be great if this could be applied table-wide

yea i agree. i'll put this in table wide.

ceolter:
After thing about this a lot, I've decided to go with a different approach, as there were to many options. We had suppressSortDesc, suppressSortDesc, supporessUnSort, defaultSort, and possibly more in the future. Easy to understand if you just want to use one, but tricky and possibly conflicting if you want to use many.

So the solution is sortingOrder. I've updated the documentation (see the page on sorting) to describe what I have done. Basically you say something like sortingOrder = [descending, ascending] to say you want to sort first by descending, then ascending. This one property replaces the need for all the other properties.

321ckatz123:
Quickly looked it over. I understand where you are coming from and your approach makes sense to me. It does make what I was asking for a bit more complicated (as I believe you would have to pass in ['descending', 'ascending']), but I have also given it some thought and have come up with no better solution for all the things you were juggling with sorting. When do you think you will be pushing the refactored version to bower?

ceolter:
> I believe you would have to pass in ['descending', 'ascending']

almost, pass in ['desc','asc'] or if you want it to 'unsort' when you click the same col while it's asc then ['desc','asc',null]

> When do you think you will be pushing the refactored version to bower?

this weekend

surjikal:
Nice solution!

Issue #363, opened 2015-08-05
SetFilter with newRowsAction as 'keep' hides new rows even though selectAll checkbox selected


wagonburger:
This occurs only when the new row contains a value for the filtered column that is not in the current set of unique values for that filter.

Example:
* rowdata = [{status:'A'}, {status:'B'}, {status:'C'}]
* Click filter menu on status column
* Uncheck select all
* Check select all
* Add new row to dataset {status:'D'}
* New row is not shown
* clicking on the filter menu shows status A.B,C selected and D unchecked

I have a fix which i'll add a PR for soon but will need some feedback as i'm not familiar with all the use cases and don't want to f*ck it up for everyone else!

ceolter:
you wont phuck it up for anyone, as i won't be accepting any PR without doing a full review of the code!!!!

wagonburger:
That goes without saying. If i wanted to wear a cowboy hat i would would contribute to ui-grid

ceolter:
lol

ceolter:
sorry if i'm getting mixed up, but did i accept a PR form you fixed this the other day? and hence we can close this issue?

wagonburger:
Yes mate. All merged in. Thank you

PR #364, opened 2015-08-06
#363 Bug with SetFilters when newRowsAction as keep and selectAll checkbox selected

wagonburger:
SetFilter's with a newRowsAction of 'keep' should always select all filter values when the select All checkbox is checked.

To recreate the problem:
* rowdata = [{status:'A'}, {status:'B'}, {status:'C'}]
* Click filter menu on status column
* Uncheck select all
* Check select all
* Add new row to dataset {status:'D'}
* New row is not shown
* clicking on the filter menu shows status A.B,C selected and D unchecked

### DIFF
--- a/src/ts/filter/setFilter.ts
+++ b/src/ts/filter/setFilter.ts
@@ -116,8 +116,9 @@ module awk.grid {
 
         public onNewRowsLoaded(): void {
             var keepSelection = this.filterParams && this.filterParams.newRowsAction === 'keep';
+            var isSelectAll = this.eSelectAll && this.eSelectAll.checked && !this.eSelectAll.indeterminate;
             // default is reset
-            this.model.refreshUniqueValues(keepSelection);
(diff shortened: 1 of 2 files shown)


src/ts/filter/setFilter.ts
```diff
@@ -116,8 +116,9 @@ module awk.grid {
 
         public onNewRowsLoaded(): void {
             var keepSelection = this.filterParams && this.filterParams.newRowsAction === 'keep';
+            var isSelectAll = this.eSelectAll && this.eSelectAll.checked && !this.eSelectAll.indeterminate;
```
wagonburger: Wasn't sure whether to add a isSelectAllChecked to the SetFilterModel and update as state changes. Went for just passing it into the model as is only used in one place

ceolter:
review complete, looks fine :)

Issue #380, opened 2015-08-16
Escape key should close popup (in addition to clicking the body)


scotthovestadt:
Currently, PopupService automatically closed if the body is clicked. It should also listen for the ESCAPE key.

Let me know your thoughts or if you'd like me to submit a pull request.

ceolter:
PR sounds good - should be a small enough change. It should go into the popup service as an option somehow 'close on ESC pressed'.

ceolter:
The PR worked, so closing.

PR #382, opened 2015-08-16
#380 - Add option for escape key to close popup

scotthovestadt:
Option added: ````filterClosedOnEscPressed````

Off by default to preserve existing behavior.

### DIFF
--- a/docs/angular-grid-grid-options/index.php
+++ b/docs/angular-grid-grid-options/index.php
@@ -245,10 +245,6 @@ include '../documentation_header.php';
                 over the column header.</td>
         </tr>
         <tr>
-            <th>filterClosedOnEscPressed</th>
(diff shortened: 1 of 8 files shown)


scotthovestadt:
Resolves issue #380

Please let me know if you're willing to merge it in.

ceolter:
looks almost fine, but when i mentioned having it as an option, i meant as an option to the popup service. so it would be used always for the filters, but the option would be there not not have it on if the popup service was to be used someplace else in the future. if you can fix, that would be welcome, however i'm happy to fix it also. i think the gridOptions options is overkill, possibly confusing the gritOptions unnecessarily. my apologies if i confused things trying to explain the optional property.

scotthovestadt:
@ceolter I've applied your feedback. Do you think it's OK to merge in now?

ceolter:
cool. thanks.

PR #416, opened 2015-09-03
Expose renderedRows on gridApi

wagonburger:
Add getRenderedNodes to grid api. Use case for this is to update volatile columns only when their row is shown in the grid

http://www.angulargrid.com/forum/showthread.php?tid=2734


ceolter:
cool, thanks :)

PR #422, opened 2015-09-07
addition of getRowsAfterFilter to the api.getModel

christoffee:
##### the change is 'src/ts/rowControllers/inMemoryRowController.ts' (the other changes are from the build)

So i added it to model so that i could get the rows that hadn't been filtered out as the array already exists.

I was using it for exporting data from the table.

Is this right?


christoffee:
the change is 'src/ts/rowControllers/inMemoryRowController.ts'

ceolter:
ah, you want to get the rows after the filter, where'as 'getVirtualRowCount' and 'getVirtualRow' show you what's displayed (so can have rows missing if your grouping and some groups are closed).

christoffee:
yh exactly... is adding getRowsAfterFilter the best way to do it?

ceolter:
am not sure . . . . as the data you get back will be grouped . . . . . might be better if i provide way that returned a flat list - like what you give the grid.

ceolter:
closing this as you can now achieve this with the api.forEachNodeAfterFilter() method.

PR #426, opened 2015-09-09
Minor changes to accept string in definition file

S-YOU:
just one line changes - you may want to do it yourself or drop if you disagree.

`headerClass?: string | string[] | ((params: any) => any);`

note: last line ending was from my vscode IDE; you may ignore.

### DIFF
--- a/src/ts/entities/colDef.ts
+++ b/src/ts/entities/colDef.ts
@@ -37,7 +37,7 @@ module awk.grid {
         headerCellRenderer?: Function | Object;
 
         /** CSS class for the header */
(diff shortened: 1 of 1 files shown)


ceolter:
this isn't exactly right:
headerClass?: string | string[] | ((params: any) => any);
should be:
headerClass?: string | string[] | ((params: any) => string | string[]);

doing the second required some cleanup in the core code, so i've fixed it up and checked it in outside of this PR.

thanks for the suggestion.

S-YOU:
:thumbsup:

PR #430, opened 2015-09-11
Added rowDeSelected option and method for row deselection

ashishpanchal:
I have added a feature for getting the node of the row when a row gets deselected as per having my requirement. for this feature we have to provide a option  rowDeSelected in gridoption and a callback function so once user deselect any row that callback will get called and user can get object of that deselected row.


ceolter:
i have just implemented rowDeselection. sorry for taking so long to get back to you on this.

ashishpanchal:
thank you ceolter for your updates.

PR #434, opened 2015-09-16
Enable declaration of localeTextFunc in gridOptions

thallgren:
It is currently not possible to integrate a translation service with
the ag-grid because the ag-grid assumes a translation table in the
form of a key/value map. A component like the angular-translate
provides a function that will allow a key to be translated but it
does not expose its internal map.

This commit makes it possible to declare a localeTextFunc function
in the gridOptions. When declared, this function will be returned
by the GridOptionsWrapper.getLocaleFunc() method.


ceolter:
makes sense. thanks :)

PR #455, opened 2015-09-30
Added option and functionality for comma separating

TFrascaroli:
First of all, I don't know if this is worth a pull request, but I guess
you'll be the judge of that. I'll try to explain the 'why' behind my
(small) change:
It appears that Microsoft Excel sometimes doesn't open CSV files just
right. It appears it might have another character as a separator in the
default options (or if you change them). This is easily worked around
defining, in the first line of the file:
sep=<your_separator>
which in this case translates to:
sep=,

More info:
http://kb.paessler.com/en/topic/2293-i-have-trouble-opening-csv-files-with-microsoft-excel-is-there-a-quick-way-to-fix-this

I also checked on other spreadsheet software, but the few I checked
(OpenOffice and GoogleDocs) aren't suffering from this. Therefore I
added another option to params by which you can tell the generator if
the implicit declaration of the separator is needed or not. The option
is called `forceCommaSeparator`

Thanks for reading this and a million thanks for your awesome product.
We're including it as part of our software, and we will try to
contribute as much as we can.

### DIFF
--- a/src/ts/csvCreator.ts
+++ b/src/ts/csvCreator.ts
@@ -1,12 +1,13 @@
 module ag.grid {
 
-    var LINE_SEPERATOR = '\r\n';
+    var LINE_SEPARATOR = '\r\n';
 
     export interface CsvExportParams {
         skipHeader?: boolean;
         skipFooters?: boolean;
         skipGroups?: boolean;
         fileName?: string;
+		forceCommaSeparator?: boolean;
     }
 
     export class CsvCreator {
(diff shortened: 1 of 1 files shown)


ceolter:
thanks for suggestion. will have a deeper look when i'm home from work this evening.

TFrascaroli:
Perfect, that's all I'm asking.

ceolter:
finally have had a chance to look at this. i've never had this problem, however it's such a minor change, and the link you provided makes sense, so am about to click the merge button!!

ceolter:
actually, i'm going to change it, to have 'first line' and 'last line' as parameters. then this call be used in a much more generic sense.

ceolter:
instead, i've added in customHeader and customFooter, to give more flexibility in this feature.

TFrascaroli:
Nice, that's perfect. I'll update and start using it right away. Also I
hope i'll be able to provide more help, since this grid is perfect us, and
it would be our way of saying "thank you".
Best regards

PR #480, opened 2015-10-06
Added the option for the header class when being filtered

### LINKED ISSUE #440: add class to header for when a filter is set
http://www.angulargrid.com/forum/showthread.php?tid=2805

TFrascaroli:
fixes #440. Turns out I needed this one too, so I found the time to implement it :smile_cat: Tested and it works.


ceolter:
noted. will check when i get  time. thanks.

S-YOU:
not really tested myself, but looks like classList is not supported in IE9 - http://caniuse.com/#feat=classlist
also not sure IE9 is under support list.

ceolter:
i'm supporting IE 9 in theory, but i'm not desperately trying to, i only test on the versions installed on my laptop, which are all latest versions of the browsers.

TFrascaroli:
hum... haven't tested there. The thing is I never test in IE, but I can change the code if it must support it.

TFrascaroli:
I mean, we can always check with regex and add/remove with string concat/replace functions. This way (the old way) it's going to support IE9.

ceolter:
don't worry to much, when i go through it, l'll use the same code techniques i'm using everywhere else, so no new 'ways of doing things' will be introduced.

TFrascaroli:
Allright, I'm just sorry that I make you do a double effort (check my code AND change it).

ceolter:
did something similar, but changed things a bit and used my internal utility class for managing css (which has my 'tired and tested' ways of working with css!!)

TFrascaroli:
Cool, thanks.

PR #481, opened 2015-10-06
Added custom selection checkbox id provider feature

noherczeg:
Automated UI tests could be simplified by this. Since if the grid is configured with:
rowSelection: 'simple'

then ag-grid provides a built in way how to manage checkbox toggling mechanics. We previously used a custom cell renderer, but the above example gave us too big of a deal to handle manually.

Hope you'll like this! :)

### DIFF
--- a/docs/angular-grid-column-definitions/index.php
+++ b/docs/angular-grid-column-definitions/index.php
@@ -110,6 +110,10 @@ include '../documentation_header.php';
             <td>Set to true to render a selection checkbox in the column.</td>
         </tr>
         <tr>
+            <th>checkboxSelectionIdGetter</th>
+            <td>Function callback, sets the id attribute of the rendered checkbox. Provides rowIndex, and the node in it's parameter.</td>
+        </tr>
+        <tr>
             <th>suppressMenu</th>
             <td>Set to true if no menu should be shown for this column header.</td>
         </tr>
--- a/docs/angular-grid-selection/example5.html
+++ b/docs/angular-grid-selection/example5.html
@@ -1,0 +1,15 @@
+<html>
+<head>
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.3.8/angular.min.js"></script>
+    <!-- you don't need ignore=notused in your code, this is just here to trick the cache -->
+    <script src="../dist/ag-grid.js?ignore=notused12"></script>
+    <link rel="stylesheet" type="text/css" href="../dist/ag-grid.css?ignore=notused12">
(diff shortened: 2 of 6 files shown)


ceolter:
thanks for the PR.

each cell in the grid is marked with the column id, eg from the example:
```
<div col="2" colid="language" /*items omitted for clarity*/ >English</div>
```

and then each row has the row 
```
<div row="1" /*items omitted for clarity*/>
```

why can't you use these in your css selector for testing?

noherczeg:
Hey!

You are right about this, but I think that since this checkbox component is pretty much auto generated users might want to have a little bit of control on it. I mean aside from turning it on, and off, we can't really do much with it and I think that this might be worthy of a slight improvement.

On the other hand, this particular pull request makes it much easier to test.

Again, You are absolutely right, that the particular reason I gave in the starting point is not much for a reason to do this, but I'd like to ask you to evaluate my points in this comment a little bit more :)

Cheers!

ceolter:
i think i'd prefer something more generic. instead of asking for 'checkbox id', it should ask for 'row business key', which it would then use to tag each row (primarily) and then it could be used by the cellRenderers also for tagging stuff inside the cells.

current the cells have a business key for the cols (it's the col id).

so the row would be something like:

```
<div row="1" row-id="pk_223" /*items omitted for clarity*/>
```

then the checkbox would use this, like any other cell, to mark the checkbox, et:
```
<input selection-id="pk_223" ....
```

that would work right? and would be much more generic?

noherczeg:
This could be good as well.

BTW, one more thing: Marking the checkbox with a unique attribut (value) is necessary since that is the only thing which guarantees a html structure agnostic way of selecting the given checkbox.

Let's say you release a version of 2.3 where you modify the internal div structure of the cells. If a e2e test was written in way that it used those divs, all tests could break.

If I'll have time for it, then I'll try to implement your suggestion since it looks nice as well, and would do the job for us too.

ceolter:
i have just checked in getBusinessKeyForNode(node) callback.

Documentation is::
```
Return a business key for the node. If implemented, then each row in the dom will have an attribute
                <i>row-id='abc'</i> where abc is what you return as the business key. This is useful for automated
            testing, as it provides a way for your tool to identify rules based on unique business keys.
```

noherczeg:
Is this expected in 2.4? :)

ceolter:
yes

ceolter:
closing this as i've implemented getBusinessKeyForNode, this will be in a release i'm about to do today.

PR #487, opened 2015-10-09
avoid global declaration of `module` variable

S-YOU:
instead this PR make it as local variable


S-YOU:
closing because of extra commits

ceolter:
don't think it would of worked anyway. you declared a new variable, what is needed is to use the variable that the commonjs framework provides.

PR #488, opened 2015-10-09
fix: avoid global declaration of module variable

S-YOU:


ceolter:
same here, this wont work. it needs to reference the variable provided by commonjs, not declare a new one.

these mean different things:
declare var i; // tell typescript about a variable that already exists, so no compile errors.
var i; // declare a new variable

S-YOU:
re-declaration does not overwrite the origin value in javascript, 
but you are right, it isn't usable inside closure. closing.

ceolter:
i've just checked in a fix for this, i've moved to using node.d.ts

S-YOU:
thank you, much better.

PR #491, opened 2015-10-10
fix: `this` scope issue in headerClass

S-YOU:
This PR should fix the issue on http://www.ag-grid.com/forum/showthread.php?tid=2921


ceolter:
thanks!!

PR #500, opened 2015-10-14
Added a renderingContext field for valueGetter functions

gastonelhordoy:
This allows different cell contents depending on where the value is rendered. As an example, I can render some HTML with a link displaying the table online and plain text for exporting to CSV.

### DIFF
--- a/docs/angular-grid-export/index.php
+++ b/docs/angular-grid-export/index.php
@@ -49,6 +49,7 @@ include '../documentation_header.php';
         <li>Value getters are used to work out the value to export (the 'Group' col in the example below uses a value getter to take the first letter of the country name)</li>
         <li>Aggregated values are exported.</li>
         <li>For groups, the first exported value (column) will always have the group key.</li>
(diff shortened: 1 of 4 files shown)


gastonelhordoy:
I had this very same requirement and I didn't see any other way to achieve the goal. I tested it also in my project and it works as expected.

ceolter:
hi @gastonelhordoy, thanks for the PR, if i understand you correctly, then i don't agree with the implementation.

the valueGetter is responsible for returning the value.

for rendering, that's where the cellRenderer comes into play.

so what i would expect, is your valueGetter to return a raw value (not formatted in any way) which gets used in csv. if you want it rendered a certain way in your grid, then you should use cellRenderer.

or maybe i misunderstood something, happy to have you explain if i did :)

thanks,

gastonelhordoy:
Hey @ceolter, thanks for pointing me out to the cellRenderer feature, somehow I completely overlooked it. Since that's exactly what I was looking for, I'm discarding this PR.

Cheers

PR #504, opened 2015-10-15
Added ability to suppress the hiding a column

killyosaur:
Currently set to just the individual columns, could easily implement this on the grid level if requested. I know there may be situations where this would be helpful, such as with a check box column, (if you are giving users the ability to save settings or change the view you may not want them to be able to hide the check box column or other similar columns). Also added a few references to the .ts files so that Visual Studio and MS Visual Studio Code's intellisense picks up those file references properly.

Hopefully I did not miss anything,


killyosaur:
I did miss one thing in my code, I have some thoughts on column pinning, I would like the ability to to pin a column directly, which would remove that column from the moveable columns in the tool panel. I accidentally left that property in the coldef.ts file. That does not need to be merged over yet as that suggested feature is not done yet.

ceolter:
thanks for the pr. i'm going to have to put this on ice for a while until i figure out the plan for the toolPanel. thanks.

killyosaur:
no worries. this is just one of 2 things I wouldn't mind seeing on this grid to get to feature parity with the features I'm currently using out of UI-Grid (which I'd like to replace as it is making my application slow :) ).

ceolter:
and the other thing is?

killyosaur:
The other thing is the ability to Pin specific columns to the left of the grid and preventing them from being moved. In your current implementation, I can pin columns but the columns that are pinned are based on their location and the number of pinned columns. If I give a user the ability to move columns, they could feasibly move those pinned columns as well changing what columns are pinned and which ones are not. In UI-Grid, I can set a property called "Pinned" on the column definition and pick which side to pin the column to which moves that column and pins it permanently there (assuming you don't give the user the ability to unpin the column). That would be the only other feature I'd need to see.

Ideally I would also like to have the ability to drag and drop the columns directly as well (not through the tool panel) but that change can be taught to the users, just as the fact that the column menus will go away, and that multi-sort is still there, it's just done a bit differently.

ceolter:
cool. i'll get all this done in time.

PR #505, opened 2015-10-16
fix(csvExport): large data export need createObjectURL

S-YOU:
This should fix - https://github.com/ceolter/ag-grid/issues/501
Same method as ng-csv is doing.


ceolter:
i merged, and it doesn't compile :(

src\ts\csvCreator.ts(36,44): error TS2339: Property 'URL' does not exist on type 'Window'.
[08:37:07] 
Error: getaddrinfo ENOTFOUND
    at errnoException (dns.js:37:11)
    at Object.onanswer [as oncomplete] (dns.js:124:16)

ceolter:
fixed it.

S-YOU:
ah, I didnt check d.ts, thank you.

PR #522, opened 2015-10-23
Read template from responseText also

lucaalex87:
Read template from responseText in case response is undefined.
In IE9 the template is on responseTest, not on response.
As a result there was an infinite loop that would crash IE9

There are more details in this StackOverflow post: 
http://stackoverflow.com/questions/33295932/ag-grid-templateurl-crashes-in-ie9/33299054#33299054

And this issue:
https://github.com/ceolter/ag-grid/issues/521

### DIFF
--- a/dist/ag-grid.js
+++ b/dist/ag-grid.js
@@ -3449,7 +3449,8 @@ var ag;
                     return;
                 }
                 // response success, so process it
-                this.templateCache[url] = httpResult.response;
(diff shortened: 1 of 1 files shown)


lucaalex87:
Code already pushed by ceolter in commit: 
https://github.com/ceolter/ag-grid/commit/255ee2f1e252dce2936f9fb980c0e57e02bc4f88

PR #544, opened 2015-11-02
2 new options : Lock the first column + Forbid to hide a column

renotidus:
I try to add 2 new options to this amazing grid :
- The possibility to forbid to hide a column (with a new parameter noHide in ColumnDef)
- The possibility to lock the first column (Fix it on first position in the table) (with the firstColumnLocked parameter in gridOptions).

I'm not an expert in this kind of development, I purpose my options with an example of code, but the I'm pretty sure than the community can easily improve my code.


renotidus:
Hi,

Is there any chance to have the 2 options I propose on the Enterprise edition of the grid ? I could be interested by the enterprise version but I really need these 2 options.
To explain, sometimes, I use the grid with checkbox selection in order to let user select lines in the grid. In that case, I don't want user can hide the first column of the grid (the one with the checkbox) and I also don't want he can "reorder" this column. It always must be the first one of the grid.
What I propose on this pull request is a little more general with an option to forbid to hide a column and another one to pin left a column and don't permit to "reorder" it.
I know this pull request is no more compliant with actual version of the grid and badly coded because I directly updated the js file but maybe you can provide these 2 options in a future release.

If necessary we can discuss about this by another way (email, private message).

Thanks in advance.
Renaud

rmc-software:
Closing this old PR as it has now been implemented. See docs: https://www.ag-grid.com/javascript-grid-pinning/

PR #546, opened 2015-11-02
fix(rowGroup): rename `parent` to `$parent` to avoid deep-watchers

### LINKED ISSUE #533: Grouped Rows (tree structure) throws 'circular structure' error
Hi! After modeling my data model accordingly, setting `rowsAlreadyGrouped='true'` and loading the `rowData` (as suggested in https://github.com/ceolter/ag-grid/issues/483), **I can successfully render a multi-level tree structure**:

mul

Unfortunately after the data model is processed by ag-grid it throws these errors:

```
Error: $interpolate:interr
Interpolation Error
Can't interpolate: {{rowData}}
TypeError: Converting circular structure to JSON
```

**Failing JSBin Example: http://jsbin.com/baliqa/2/edit?js,console,output**

Any hint would be greatly appreciated :smile: Cheers!

benoror:
Fix #533


ceolter:
the issue this relates to is no closed. does this mean this PR is no longer needed?

benoror:
Hi @ceolter,

As stated in the issue, this is trying to solve an specific angular $digest cycle issue.

[I temporarily solved it in my project by using my own fork.][1]

Nevertheless [I noted][2] that this needs to be fixed in a more robust way

If you want to keep your lib 100% framework-agnostic and you think renaming a var is not worth the hazzle, then we might need to come up with a different approach to solve this.

Otherwise i think this is a simpler workaround imho

[1]: https://github.com/panaxit/panaxui-angularjs/blob/master/package.json#L29
[2]: https://github.com/panaxit/panaxui-angularjs/issues/74

ceolter:
yes it is a simple workaround, that is true. it just doesn't make sense to me that the grid should have to use angular convention inside it.

the grid doesn't use this parent value, it's just there for the clients sake, if it wants to navigate the tree. what about a property 'suppressParentInNode', which when true, makes the nodes a unidirectional tree with no cyclic dependencies?

benoror:
@ceolter fair enough :sweat_smile:

I will try to elaborate on a cleaner fix, thanks for your feedback and patience :smile: 

Closing the PR

ceolter:
well if you don't actually need the parent reference, i think that's the cleanest. i can appreciate sometimes it is better to have a non-cyclic tree.

ceolter:
i have just implemented a property 'suppressParentsInRowNodes', if set to true, the parent variable will not be set in the row nodes. the grid didn't use this property, it was just for the client to use if it needed to navigate the tree, so having it optional is not a problem for the grid.

benoror:
@ceolter wow thanks! gonna try it asap

benoror:
@ceolter I set `suppressParentsInRowNodes: true` to `gridOptions` but apparently `parent` is still added to children nodes.

File Browser Example: http://jsbin.com/naviji/1/edit?html,js,console,output

**disclaimer: I'm using `dist/` files from my own repository since ag-grid's is not updated*

ceolter:
i just tested locally, and it's working in my test drive. can you try after the release? easier for me than look at your repository (sorry, everyone is trying to get some of my time!!)

benoror:
Absolutely, I appreciate your help! a lot :tada: 

Unfortunately I'm still getting `parent` references with latest release (3.1.0): http://jsbin.com/naviji/3/edit?html,js,console,output

benoror:
I think the clue is that `rowsAlreadyGrouped` is set to true `true` ?

benoror:
@ceolter I found thi issue, gonna update my PR asap :smile:

benoror:
new PR: https://github.com/ceolter/ag-grid/pull/657

ceolter:
ah, missed that. cool, thanks :)

PR #547, opened 2015-11-03
Add documentation for params.node

sarod:
I found out from angular-grid-angular-compiling/index.php example that params object exposes a node parameter that was not documented.
So this is a proposal to fix the documentation.


ceolter:
cool thanks. updated in git, but won't be live until i get back from hols.

PR #548, opened 2015-11-03
Fix typo

sarod:
Fix a typo "more complex HTML thanks just a string" => "more complex HTML than just a string".


ceolter:
again, thanks.

PR #552, opened 2015-11-05
deselectNode should pass down suppressEvents parameter

jkorrek:
The deselectNode function of the gridApi did not pass down the suppressEvents parameter to the selectionController.


ceolter:
Thanks for pr. Will review when back from holidays.

GustavSt:
If you could merge this PR I would be really happy!

PR #478, opened 2015-10-06
Add ag-hover when row is hovered

MikaAK:
I know there is definitely a better way to achieve this to maintain the style of this code repo, but I wasn't quite sure how so, I put a quick fix for work!. This solves #223.

If you could provide guidance on how to fix up the code so it fits the correct style I would be happy to change it up.

Issue #223, opened 2015-06-15
Add class to whole row to indicate hover (required when columns are pinned to highlight whole row)


jonsamwell:
Hi,

If you have a pinned column and you want to say change the background colour when any of the row is hovered over it currently will not work due to the pinned column(s) and non-pinned column(s) not being under the same parent.  If you could add a class to the pinned and non-pinned row/columns to indicate the mouse is over them this would solve the problem.

Here is a Plunker which highlights the issue.

http://plnkr.co/edit/vXJNaoDgsRRGJbwVR9Lx?p=preview

Thanks,

Jon

ceolter:
noted. thanks.

MikaAK:
+1 Is this still non-existant? We cannot use this at work without hover functionality in place.

Sergeykp:
+1

chiptus:
can't you just use .ag-row:hover? that's how we do it

Sergeykp:
.ag-row:hover will not work correctly, when you have pinned columns

mpetkov:
+1

antru:
+1

ceolter:
already done. class is ag-row-hover

jlarso11:
Is there a way to remove the default color for this?  I want to control the color myself in CSS

ceolter:
@jlarso11 it's css - so if you lay your css out correctly, your style will get used.

LinnykS:
ag-row-hover class working properly only when you hover pinned column but what about highlighting the pinned rows when the main rows hover?

PR #563, opened 2015-11-12
Add ag-hover when row is hovered

MikaAK:
Resynced and updated the pr from #478, github auto closed issue so I'm creating a new one. Solves #223. I've also added in the fix from @antur84
Hopefully @ceolter if you could provide some guidance on where to put or refactoring that needs to be done this we can get this merged in so I don't have to keep the fork up to date at work :).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "ag-grid",
-  "version": "2.3.6",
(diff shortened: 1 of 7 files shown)


ceolter:
I won't be making any relevent code changes before looking at this pr, which may not be until the end of the year!

Issue #577, opened 2015-11-23
deselecNode doesn't take suppressEvents parameter


GustavSt:
According to the docs the api function deselectNode is supposed to take a second parameter, suppressEvents, it does not. The other select/deselect api functions have this parameter, only deselectNode that doesn't.

GustavSt:
saw that it was already a pull request for this one, #552

senthanal:
:+1: 
Hi, Yes, suppress events doesn't work at the moment. Fix would be great :) 
Thanks for the awesome grid :)

ceolter:
suppressEvents is no longer supported, stop listening for the event if you no longer want it

PR #578, opened 2015-11-23
Added optional suppressEvents boolean.

GustavSt:
Api function deselectNode is supposed to have a suppressEvents boolean as a second parameter.
Issue #577


GustavSt:
Second one to be to hasty with the pull request regarding this issue..

PR #586, opened 2015-12-07
Fixed: field expression now can be a complex javascript expression!

FHeNuS:
Hello :).

Your code is awesome. I just made a small change to allow complex field expressions at ColDef's using eval so that the field expression can be a complex field or a method: object.field.name or object.field().nameOfMethod()


ceolter:
I've changed colDef.field so now you can have deep values eg colDef.field = 'person.firstName'.

This cannot call functions like the proposed, however this is exactly what valueGetter is for, so not sure of the point in doing it again here.

FHeNuS:
I've changed the implementation, so that if the field is a complex expression, I create a valueGetter internally. The point of doing things like this, is that some part of your code only executes if a field is supplied (sorting and etc). With this, the function calls are supported even on the field expression. What do you think?

ceolter:
All parts of code should use valueGetter if it's there. Sorting and Filtering use the value getter. This is something I added along the way, maybe it has changed since you last saw the code??

PR #590, opened 2015-12-08
Recursive Header Groups

russell-miller:
I wrote a [gist](https://gist.github.com/russell-miller/aec10d6b5d2aff6af6f1) proposing a new format of column definitions that allows for groups within groups. I also reached out via email to discuss this approach.

I've reviewed every page of the documentation to ensure this new functionality plays nicely with all other features.

I've had my colleagues at Rentrak review my commits in our own [pull request](https://github.com/rentrak/ag-grid/pull/1).


ceolter:
am looking at this for a few hours now. thought i'd give an update on thoughts:
+ on a whole, very happy with the quality. i'm not decided yet if i'll work with the branch, or code from scratch myself and take influence from the branch. i like coding myself because then my brain fully understands everything that's going on, where i would copy/paste items from the branch. if i do this, it is no way in disrespect of the provided quality code, just me taking extra time to make sure every line of code is scrutinised!!
+ some reference to colDef.pinned in the code, but colDef has no such property!
+ the code to work out the pinning is a bit complex. it is on my plan to change how pinning works (to have a pin attribute on the colDef as on of {left,right}. i am considering implementing the change in pinning first before the new grouping of cols.
+ i think the ability to reorder the cols is now gone when grouping the cols. is this right? i'm thinking about how this would/should work.
+ the grid as a whole needs refactoring how it sets itself up. the dependencies between the modules are complex when grid is initialising. i'm thinking of doing some refactoring around this area. this may help (or impact) how the columnController is implemented.

russell-miller:
> i'm not decided yet if i'll work with the branch, or code from scratch myself and take influence from the branch

I totally get this. Let me know if there's anything I can explain more or help out with.

> the code to work out the pinning is a bit complex

Yeah I wasn't super thrilled about the complexity of the recursive pinning code. I'm curious to see how it would look with your new pinning method.

> i think the ability to reorder the cols is now gone when grouping the cols. is this right?

I'm not sure, I don't think I ever tried it. It's not a use case Rentrak has really needed to consider so I forgot to test it.

russell-miller:
Apparently I misused the word "depth" when dealing with column def items. What I was really computing was height (and it's supposed to be zero-based not one-based). This [stack overflow](http://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height) has a good illustration that explains the difference.

russell-miller:
Okay I think I figured this one out:

> some reference to colDef.pinned in the code, but colDef has no such property!

In the column controller, when I wrote `processColDef ` I was looking at the old method `buildGroups` and when it created a new `ColumnGroup` it passed in `column.pinned`. I goofed and where I am now looking at column *defs*, I thought I could use `colDef.pinned` the same way. I think the best fix for this would be to just initialize `pinned = false` when creating the groups, since `updateVisibleColumnGroupsAndPinning` will set the correct value of `pinned` for each group.

ceolter:
i've finished this now in the v3 branch. the interface is very similar to what you have done, however i have done a LOT of refactoring around the column controller. basically, i've reworked all the foundations, to allow bringing the new functionality to be easier. in particular:
+ columns and groups are now both 'ColumnGroupChild' objects, which means the column controller now works with trees of ColumnGroupChild objects. the tree dept can be 1 or more (1 meaning no column groups), which means there is now just one set of code, regardless of grouping columns or not.
+ introduced concept of 'OriginalColumns' (similar to what was in this PR) and DisplayedColumns. The displayed columns take into account ordering of the cols, and also break the groups up if column movement means the group needs to be split. the logic of knowing 'do these cols beside each other belong in the same group' took me a long time to get right.

i've done pretty much nothing else apart from this over the past few days. it's taken me a lot longer than i hoped!! but the result now has it coded elegantly (i was constantly refactoring) and works with all parts of the grid (moving columns, pinning, master/slave etc).

thanks for the time on the original PR. it was helpful for my learning and pushing the grid in the right direction. i'll be releasing v3 in the coming days (if not this week, deffo next week).

russell-miller:
@ceolter that's exciting, I can't wait to see v3!

PR #607, opened 2015-12-19
Pinning columns on the right

### LINKED ISSUE #31: Column Pinning left and right
It would be great to pin the columns also on the right side of the table.

[An example](http://embed.plnkr.co/5Ag6gN8xaSyDnKNlowpK/).

cezarykluczynski:
Fixes #31.

### DIFF
--- a/docs/angular-grid-column-api/index.php
+++ b/docs/angular-grid-column-api/index.php
@@ -125,7 +125,11 @@ include '../documentation_header.php';
         </tr>
         <tr>
             <th>isPinning()</th>
-            <td>Returns true if pinning, otherwise false.</td>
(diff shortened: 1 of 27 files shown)


ceolter:
thanks for the pr. i'm going to have to put time aside to go through this and think it through.

ceolter:
pining columns to the right is new functionality i've introduced in v3. you can check out the latest in that branch, or wait for v3 release in the coming days.

by the way, i didn't run this PR, i had a quick look at the code. when i implemented, i got into problems getting it working with floating header and footer, and also grouping (plus the new multi-level grouping). in addition, the columnController has been totally re-factored to be easier to understand and support the multi-level of grouping - you might be interested in taking a look!

PR #608, opened 2015-12-20
Update pagingServerSide.js

pacificeyes:
Added below code at line 88 for setting the rowData which is essential if user wants a default row selection feature after loading data to grid:
params.rowData = rowsThisPage;


ceolter:
i guess this is related to the issue here:
https://github.com/ceolter/ag-grid/issues/601

but as with comment on that issue, i don't see what this does or why it's needed.

PR #615, opened 2015-12-25
Cancelable columnBeforeResize event

cezarykluczynski:


ceolter:
what are you trying to achieve here? it is possible to suppress resize on each column via the columnDef, so i don't see what the advantage is here. if no clear advantage, i'd rather keep it simple.

cezarykluczynski:
@ceolter I'd like to be able to decide on the fly whether the resize should be prevented or not, based on the current column width. Without that, our custom CSS doesn't look right.

ceolter:
thanks for info. i'll have to think about it. the implementation of 'preventDefault' i don't think is correct. the grid generates the event after it has finished processing. similar to 'click' evet, that's created by the dom, you cannot pretend the click never happened or undo the click, you can just stop other consumers from consuming, however the grid isn't a consumer of the event.

there is a part of the code that checks the new size fits within the allowed min and max sizes. maybe something can be put in there.

rmc-software:
Closing this PR as the code base has evolved too much at this point. This feature request has been raised and can be tracked on our pipeline. https://www.ag-grid.com/ag-grid-pipeline/

PR #616, opened 2016-01-01
Change for angular2@2.0.0-beta.0

TeeNak:
Made it work both with angular2-alpha and angular2-beta.
Note that we have to change the markup when working with angular2-beta like this
[From]
```
    [column-defs]="columnDefs"
    [row-data]="rowData"
```
[To]
```
    [columnDefs]="columnDefs"
    [rowData]="rowData"
```
because of the kebab-case removal in angular2.
https://github.com/angular/angular/commit/da9b46a

### DIFF
--- a/src/ts/components/agGridNg2.ts
+++ b/src/ts/components/agGridNg2.ts
@@ -25,34 +25,34 @@ module ag.grid {
         private columnApi: ColumnApi;
 
         // core grid events
-        public modelUpdated = new _ng.EventEmitter();
-        public cellClicked = new _ng.EventEmitter();
-        public cellDoubleClicked = new _ng.EventEmitter();
-        public cellContextMenu = new _ng.EventEmitter();
-        public cellValueChanged = new _ng.EventEmitter();
-        public cellFocused = new _ng.EventEmitter();
-        public rowSelected = new _ng.EventEmitter();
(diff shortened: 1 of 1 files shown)


ceolter:
Thanks for the info. I've just checked in these changes. I took influence from you comments and your pull request, however I made the change myself as I wanted to go through the trial and error of learning Angular 2 - also I don't want to support older versions. All along, as different alphas were coming out, I was programming to the latest as changes happened. If I was to support all the alphas and betas, the code would be just to messy for no real gain.

TeeNak:
I just worried about npm semver rules which requires backword compatibility.
Yes but it's alpha, I think you don't have to support alpha any longer.
Keep moving :-)

ceolter:
npm rules, really? what is the rule?

so if i want to make a change that is not backward compatible, which part of the version number am i allowed to change?

TeeNak:
This is the semver rules adopted by npm.
http://semver.org/
```
8. Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.
```
So strictly saying, you have to change the major version number when you introduce a backwards incompatible change.

Issue #627, opened 2016-01-08
have option to autosize grid, ie call api.sizeColsToFit automatically


ceolter:
grid should have option to fit cols by default, and also option to do this when grid size changes.

gabrielenosso:
If the option would be set, it should update the cols width also when the columns are resized, sown/hidden and when the columns panel is toggled.

DESURE:
Thank you, I will consider these events

gabrielenosso:
Is it merged in the current version?
It doesn't seem to work yet.

And also: is the "gridSizeChanged" event invoked also after the first rendering?

ceolter:
"Is it merged in the current version?"-> no

"And also: is the "gridSizeChanged" event invoked also after the first rendering?" -> yes

armuell:
How is this coming along? I would really like this functionality and I am having problems hacking this to get it to work. I have tried setting a timeout/using onGridReady like described in #269 or #596 . Should I try and continue to hack it or is there a fix for this in the works?

ceolter:
@armuell you should use the 'gridReady' event, and put the call in there.

see second example on this page:
https://www.ag-grid.com/javascript-grid-refresh/index.php

dtslvr:
Thanks for your input, @ceolter. This works as expected (distributing columns equally) in ag-grid v5.1.2:
`
onGridReady: event => {
  event.api.sizeColumnsToFit();
}
`

But this does not change anything
`
onGridReady: event => {
  event.columnApi.autoSizeColumns();
}
`

Any ideas?

ceolter:
@dtslvr 
autoSizeColumns(keys) takes a list of columns keys to autosize. you need to provide this list.

note it will only autosize columns that are rendered, so if using column virtualisation (the default), only cols visible by the horizontal scroll will be autosized.

dtslvr:
Using the list of columns keys works, but only if I wrap it in a timeout like this:
`
onGridReady: event => {
  $timeout(function() {
    event.columnApi.autoSizeColumns(keys);
  }, 2000);
}
`

It also works if I add a button as in the example of [Resizing](https://www.ag-grid.com/javascript-grid-resizing). For me it seems that `onGridReady` fires too early.

ceolter:
[...]

PR #629, opened 2016-01-11
issue #627 added option enableColSizeFit for call api.sizeColsToFit automatically

DESURE:


ceolter:
I have implemented the new event onGridSizeChanged. This allows you now to do this without needed a custom ag-Grid. I'm purposely keeping the logic out of the grid as I want the grid to be extendible more than I want it to support every single feature.

Issue #630, opened 2016-01-11
Add drag and drop reorder of columns (directly rather than via the tool panel)


vladimir-ivanov:
As mentioned (incorrectly) in https://github.com/ceolter/ag-grid/issues/593

Can create a pull request if the feature is acceptable?

the difference is in https://github.com/ceolter/ag-grid/blob/0fa265e8f876c4a99683d14839f3c3ab40f90ff5/src/ts/headerRendering/renderedHeaderCell.ts <https://github.com/ceolter/ag-grid/blob/0fa265e8f876c4a99683d14839f3c3ab40f90ff5/src/ts/headerRendering/renderedHeaderCell.ts>

inside private setupComponents() method, added this code  e.g. line 160 (after this.addMenu() -> self = this))

            if (self.gridOptionsWrapper.gridOptions.enableColReorder) { //TODO - find the best place for this
                headerCellLabel.setAttribute('draggable', 'true');
                headerCellLabel.addEventListener('dragstart', ev => sourceColumn = self.column);

                headerCellLabel.addEventListener('drop', ev => {
                    ev.preventDefault();

                    self.gridOptionsWrapper.gridOptions.columnApi.moveColumn(sourceColumn.index, self.column.index);
                });

                headerCellLabel.addEventListener('dragover', ev => ev.preventDefault());
            }

ceolter:
if you create a pull request, it will stand a better chance of been included. however no promises. i plan on implementing draggable columns in the coming weeks and will be taking my time to implement a full solution.

vladimir-ivanov:
Thank you, will create one hopefully tonight.

On another note, would be nice some methods to be made protected (rather than private) so methods can be extended without typescript complaining of extending private methods/ properties. It will allow overwriting classes to add custom behaviour. E.g. want to add setState / getState pair to add sort property to the state. 
Typescript still compiles if we ignore the errors.
Let me know your thoughts.

thanks

gabrielenosso:
Will the commit be merged anytime?

ceolter:
the latest relesat 3.3.3 has it

gabrielenosso:
I searched for "suppressMovableColumns" in the code (documented on ag-grid website), but I couldn't find anything that uses it. Am I drunk? :)

ceolter:
used here
https://github.com/ceolter/ag-grid/blob/master/src/ts/headerRendering/renderedHeaderCell.ts

xfight0:
[...]

PR #634, opened 2016-01-11
Feature: #630 added columns drag and drop inline option

vladimir-ivanov:


scotthovestadt:
+1

vladimir-ivanov:
the implementation is broken, does not take into consideration hidden columns, will fix and update the pull request today

ceolter:
dragging columns is now supported.

PR #655, opened 2016-01-20
Small changes in csv export to smoothen usage in Excel

dannylagrouw:
- Include special UTF-8 marker so Excel recognizes non-ASCII characters.
- Allow user to specify a custom column separator.

### DIFF
--- a/src/ts/csvCreator.ts
+++ b/src/ts/csvCreator.ts
@@ -25,7 +25,7 @@ module ag.grid {
             var csvString = this.getDataAsCsv(params);
             var fileNamePresent = params && params.fileName && params.fileName.length !== 0;
             var fileName = fileNamePresent ? params.fileName : 'export.csv';
-            var blobObject = new Blob([csvString], {
+            var blobObject = new Blob(["\ufeff", csvString], {
                 type: "text/csv;charset=utf-8;"
             });
             // Internet Explorer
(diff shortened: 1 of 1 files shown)


ceolter:
any side effect of including "\ufeff" ?

any webpage reference i can read to understand why this is needed?

dannylagrouw:
Yes, sorry, could have told you a bit more.
Excel doesn't recognize the CSV file as UTF-8, resulting in incorrect display of non-ASCII characters (e.g. ü). If you add this marker, all is well. If found out about this here:
- http://stackoverflow.com/questions/17879198/adding-utf-8-bom-to-string-blob
- http://stackoverflow.com/questions/18925210/download-blob-content-using-specified-charset/18925211#18925211

There is one catch however. If you specify a separator character in the CSV export, it doesn't work well with the UTF-8 marker. I'm talking about adding "sep=," as a header. In the US you don't need this, but (at least) in the Netherlands, we use a semicolon as standard separator character. If you have commas as separator in your CSV, you can put "sep=," in the header to inform Excel.

As both hacks don't work well together (see https://social.msdn.microsoft.com/Forums/office/en-US/ed67b453-ec8b-4384-9833-b51835a720af/csv-file-encoding-in-excel-2007?forum=exceldev) we decided to use the UTF-8 marker and set semicolon as column separator. That's where the other change comes from that I've submitted to your.

I hope all this makes sense and can be of some use to you!

ceolter:
i'm afraid if i put it in, it will cause issues for other people!!

i think i'll stick it in as an option :)

thanks.

PR #662, opened 2016-01-21
Fixed displaying cell content with number zero.

tbielcik:


ceolter:
i also put a check in against undefined.

PR #670, opened 2016-01-22
fix error in rowRenderer - prevent unbinding invalid row

russell-miller:
Hey @ceolter I was hoping you could help me debug an error I was getting in `src/ts/rendering/rowRenderer.ts`.

I'm unable to reproduce this on the Virtual Paging / Infinite Scroll example page, but when I implemented virtual paging in my project I got an error any time I scrolled the grid quickly (for some reason scrolling very slowly prevented the error from occurring).

I traced where it was coming from:

`drawVirtualRows` makes 2 private method calls:
 * `workOutFirstAndLastRowsToRender` - uses `rowModel.getRowAtPixel` to find first/last rendered rows
 * `ensureRowsRendered` - uses first/last rendered rows as a range, then tries to remove some virtual rows by index by calling:
   * `removeVirtualRow` - gets an array of row indexes that are **not** in `renderedRows`! ends up calling:
     * `unbindVirtualRow` - error: "Cannot read property 'destroy' of undefined" (from `renderedRow.destroy()`)

I'm not sure if my solution is the right one, because it seems like `workOutFirstAndLastRowsToRender` is producing the wrong first/last row indexes. But I wasn't sure where to go next. I think the check I've added is reasonable, though.

### DIFF
--- a/src/ts/rendering/rowRenderer.ts
+++ b/src/ts/rendering/rowRenderer.ts
@@ -313,6 +313,9 @@ module ag.grid {
 
         private unbindVirtualRow(indexToRemove: any) {
             var renderedRow = this.renderedRows[indexToRemove];
+            if (!renderedRow) {
+                return;
+            }
             renderedRow.destroy();
 
             var event = {node: renderedRow.getRowNode(), rowIndex: indexToRemove};


ceolter:
i do not want to fix unless i know what's going on.

in the datasource, there is a method getRows(), which should return asynchronously by calling the successCallback() function. Can you advise, are you returning asynchronously? If not, can you change so you are (put the callback into a timeout).

this is the only way in can think that the grid is getting confused.

let me know!

russell-miller:
@ceolter 

That fixed it!!!! I didn't know the timeout was a requirement. 

On this page - [Virtual Paging](http://www.ag-grid.com/angular-grid-virtual-paging/index.php):

```
  // To make the demo look real, wait for 500ms before returning
```

I assumed that was the only reason for the timeout.

This page - [Datasource](http://www.ag-grid.com/angular-grid-datasource/index.php) - also doesn't seem to mention needing a timeout/promise.

Perhaps a sentence could be added about that requirement?

Thank you!

ceolter:
yup, i'll fix the documentation, actually better, i'll fix the code, so that if someone tries to 'callback' before the getRows has returned, i'll throw an error.

russell-miller:
+1

thanks much :grinning:

PR #691, opened 2016-02-02
fix property for column definitions in documentation overview

rweng:
Hi,

only a small adjustment because I was wondering why the columns didn't render. In the documentation overview `colDef` is used, which does not work for me in version 3.2.2. Using `columnDefs` however, does.


ceolter:
thanks. i've made this change locally as i'm in the middle of refactoring and don't want to risk having to do a merge!!

PR #704, opened 2016-02-09
update release task (clean up and lib without styles support)

vitvad:
- add clean up task before build
- build 2 versions for lib (with inline styles and without)
- add editor config


package.json
```diff
@@ -29,26 +29,27 @@
   },
   "homepage": "http://www.ag-grid.com/",
   "devDependencies": {
+    "css-loader": "0.23.1",
```
vitvad: @ceolter just to clarify:

I add only "del": "^2.2.0",
but if you use ```npm install del --save-dev``` it will automatically reorder dependency list in alphabetic order

gulpfile.js
```diff
@@ -136,7 +144,7 @@ function stylusTask() {
 }
 
 function copyToDocsTask() {
-    gulp.src('./dist/*')
+    gulp.src('./dist/**')
```
vitvad: @ceolter, 
fixed to coping lib files and styles

not sure maybe ```['./dist/*.js', './dist/styles*']``` will be better - to exclude lib folder files ?

gulpfile.js
```diff
@@ -86,21 +93,22 @@ function tscTask() {
     ])
 }
 
-function webpackTask(minify) {
+function webpackTask(minify, isStyleIncluded) {
```
vitvad: not the best way to do this, but i'm new to webpack so maybe little bit later will suggest better solution

ceolter:
i've taken 2 of 3 ideas:
 - now have no-style version of the bundled files
 - now clean the dist folder

i didn't take the formatter, it's not something i'm familiar with.

vitvad:
@ceolter, formatter ? you mean .editorconfig ? 
http://editorconfig.org/ - is just a file that could be in project root folder
if IDE open project and see this file it will apply settings from file, I add same settings that you use for development.

Config need because different devs like different default formatting in IDE so this file just bring some consistency and tell to all devs "my project use 4 spaces for those files. This is my rules GTFO" ^_^

P.S. by the way did you think about create chat room in gitter ?

ceolter:
re chat room, yes, but it was taking up to much of my time! with github and the forum, i dedicate about 30min each morning on issues (however today it was more like 2hr!!!)

ceolter:
acually there is still a chatroom, i just don't go into it anymore.

PR #705, opened 2016-02-09
clean up styles and move copy-paste from themes to common file

vitvad:
little bit cleaned up styles 
will try to rewrite themes with stylus variables usage and so on

### DIFF
--- a/src/styles/ag-grid.styl
+++ b/src/styles/ag-grid.styl
@@ -16,6 +16,9 @@ ag-grid-ng2
     /* was getting some 'should be there' scrolls, this sorts it out */
     overflow: hidden
 
+.ag-column-moving .ag-cell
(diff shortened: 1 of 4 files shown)


ceolter:
this is more than just formatting, eg the following i see introduced in ag-grid.css
```
.ag-column-moving .ag-cell
     transition: left 0.2s;
 
```
the core grid css is not meant to have any styling, just layout or the core grid to work. styling belongs in the themes.

vitvad:
@ceolter like for me it's much better to redefine few styles if you don't use them, then look through theme file and try to find all this ```.ag-column-moving ``` but anyway up to you.

vitvad:
at least I advice to add stylus variables for colors, paddings etc. in themes

PR #707, opened 2016-02-09
Shift Select Rows

NisStrom:
Implementation for using shift select on rows.


ceolter:
i have implemented a version of shift selection that will be in the next release. the version that is checked in btw works with groups, which this pr didnt' do. thanks :)

Issue #713, opened 2016-02-11
aggFunc: first


ghost:
for grouping it would be convenient to get the first row whatever type it is.
Introduction of aggFunc "first"

ceolter:
done.

PR #714, opened 2016-02-11
#713 Feature: aggFunc "first", show the first available value

ghost:


ceolter:
'first' and 'last' was introduced in a release a few weeks ago

PR #730, opened 2016-02-22
columnApi.getColumnState, setColumnState rename

AkosLukacs:
A documentation fix...


ceolter:
I've changed this in the next version. also changed resetState to resetColumnSate. Also updated the docs.

PR #733, opened 2016-02-22
Fixed horizontal scroll on trackpads.

Elecash:
Apple trackpads (and I suppose that Win/Linux trackpads) can't scroll horizontally on the grid.

I've added a check to know if scroll is horizontal or vertical and applied scrollLeft to body container when it's necessary.

Added on `Utils` class a mouseWheel normalizer authored by Facebook.


Elecash:
Tested on Mac OS X with Chrome and works. This fix doesn't works on Firefox but I'll try to add a new PR fixing Firefox.

ScottTrenda:
Why not add the deltaX to scrollLeft and the deltaY to scrollTop at the same time? That way, it still responds correctly to diagonal scrolls instead of trying to pick out which scroll is more important.

Elecash:
I've tried that but controlling was weird and scrolling in just one direction seemed to me more clear than scrolling in both directions.

ceolter:
i will need to put some time aside to look at this. interesting function you pulled in. i will go over it in the coming days (weeks?) as i've now put this PR on my 'todo' list (which I keep on my desktop).

Elecash:
No problem, I've added `dist` files to the PR, I know that you suggest to not add them but I need them for now because we're using this in production and I'm pointing NPM dependencies to my repo until you merge this.

Thanks again for reviewing and keep up the good work.

ceolter:
dist files doesn't matter in this case as i won't be merging, i'll be just copy / pasting. the next release is in a branch and i don't want conflict hassles as i've done massive changes. also, copy / pasting puts me in the driving seat, no code slips in unless i put it in explicitly!

ceolter:
i've checked this into trunk. i didn't merge as to many conflicts, just copy / paste. i'll be doing a release this evening.

Elecash:
Thanks!!

PR #736, opened 2016-02-23
fix context

vitvad:
filter instance doesn't have access to `gridOptionsWrapper`
so need save link to context object


ceolter:
thanks. i've fixed this in my 'next release' branch.

PR #743, opened 2016-02-26
fix issue when filter value is "0"

vitvad:
type coercion will lead to that this change will be ignored
I suppose that value could be also set to `null` so best way check only for `undefined` values

P.S. could you release only this fix in 3.3.4 ?

### DIFF
--- a/src/ts/filter/filterManager.ts
+++ b/src/ts/filter/filterManager.ts
@@ -98,7 +98,7 @@ export default class FilterManager {
                 return;
             }
             var model = filterApi.getModel();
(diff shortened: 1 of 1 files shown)


ceolter:
sorry, i will only be doing releases for major bugs, for something small like this, will have to wait for next version. you can make the change locally?

vitvad:
@ceolter , I use workaround right now with value = "0" (string)

vitvad:
@ceolter, congratulations with 4.0.0

I see same issue in new version, should I make PR to master ?

PR #753, opened 2016-03-02
add dynamic row attributes

### LINKED ISSUE #751: Angular compile element after it's inserted into the DOM
http://stackoverflow.com/questions/26791597/append-child-directive-within-parent-directive
https://www.ag-grid.com/forum/showthread.php?tid=3478

### LINKED ISSUE #752: [FEAT] Dynamic Row Attributes
It would be great if there were not only the possibility to dynamically add classes or styles but also attributes to the rows.

I need it to assign a drag and drop directive (angular) to each row so I can drag the row out of the grid for further use.

I already implemented it and will send you a pull request.

ftorghele:
closes #752
closes #751

### DIFF
--- a/docs/javascript-grid-properties/index.php
+++ b/docs/javascript-grid-properties/index.php
@@ -180,7 +180,12 @@ include '../documentation_header.php';
         <tr>
             <th>rowStyle</th>
             <td>The style for a particular row. Provide an array of CSS key / value pairs eg {color: 'red'}.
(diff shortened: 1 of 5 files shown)


ceolter:
reading my issues and working on some as i go, i implemented the solution in your issue before seeing this! anyway, i think my solution is a wider 'catch all' as you can change anything on the row, not just attributes, and requires less code on the grid side (easier to maintain). so i'll stick with my implementation. thanks :)

PR #756, opened 2016-03-02
Fix CsvExportParams import

AkosLukacs:
Already fixed in v.3.4 branch, don't know your the planned release date...
CsvExportParams is not the default export of csvCreator, and breaks typescript compilation...


ceolter:
next release will be from the 3.4 branch.

PR #764, opened 2016-03-08
Google material design datatable

sasivarnan:
I have done a new theme for google material design specification. The new class is ag-material.
The google material design for datatable gives specification for following, and out of those, I have implemented the below to begin with.
1. the font , size of data and header
2. row hover
3. row selection
4. sorted column font with icon

The following can be implemented only by custom cell rendering for now:
1. cell edit for material design specification
2. Inline text editing (slightly different from cell edit overlay)

The following is not yet done:
1. Col Name shortening for long columns and tooltip to show full name


jogibear9988:
Could this be included?

ceolter:
i am planning to get this done next week.

pml984:
I want to first thank you for doing this. It really saved me a lot of time. 

I did run into a problem with the checkbox styling. You added the before pseudo-element directly to the input. This works in Chrome but both Firefox & IE don't support pseudo-elements on inputs as they say it's not to standard.

I was able to get it to work in both, but it requires some styling fixes and a few configurations. Here is what I did

```
.ag-material .ag-root .ag-selection-checkbox + .ag-cell-value:before {
  width: 18px;
  height: 18px;
  border: 2px solid #757575;
  -webkit-border-radius: 2px;
  border-radius: 2px;
  content: '';
  background: #fff;
  position: absolute;
  -webkit-box-sizing: border-box;
  -moz-box-sizing: border-box;
  box-sizing: border-box;
  -webkit-transition: 0.2s all ease-in;
  -moz-transition: 0.2s all ease-in;
  -o-transition: 0.2s all ease-in;
  -ms-transition: 0.2s all ease-in;
  transition: 0.2s all ease-in;
}

.ag-material .ag-root .ag-selection-checkbox:checked + .ag-cell-value:before {
  background: #00BCD4;
  border-color: #00BCD4;
}

.ag-material .ag-root .ag-selection-checkbox:checked + .ag-cell-value:after {
  border-bottom: 2px solid #fff;
  border-left: 2px solid #fff;
  border-bottom-left-radius: 1px;
  content: '';
  position: absolute;
  -webkit-box-sizing: border-box;
  -moz-box-sizing: border-box;
  box-sizing: border-box;
  width: 12px;
  height: 6px;
  margin: 5px 3px;
  -webkit-transform: rotate(-45deg);
  -moz-transform: rotate(-45deg);
  -o-transform: rotate(-45deg);
  -ms-transform: rotate(-45deg);
  transform: rotate(-45deg);
}:
```

The default input that is provided by ag-grid doesn't have a label, which is usually required to style a checkbox. There is a span with class  *ag-cell-value* that I'm using above to style it properly, but the checking of the checkbox did not work correctly. To fix that, I was able to use the selection capability of the grid to get it to work correctly.

```
var gridOptions = {
          suppressRowClickSelection : true,
          rowSelection : 'multiple',
          onRowClicked : function(event){
            var rowNode = event.node
        
            rowNode.setSelected(!rowNode.isSelected())
          }
      };
```

sasivarnan:
@pml984 Thanks for the observation and suggestion. I had made some changes along with your above suggestion. This will work without using the selection capability of the grid.

Works fine with Mozilla Firefox, Microsoft Edge and it will not work in IE-9 / IE-10.

@ceolter I have updated my [plunker](http://plnkr.co/1PmRtTC3TGLysmEZolRi) for the above changes.
Kindly check it out.

jogibear9988:
Any news to this?

ceolter:
we took influence from this PR and created another material design theme.

we started with this PR, but then went back to the drawing board as it was to awkward maintaining the themes the way we were. instead we decided to put all the common items into one theme that we now call ag-common - and then each them extends this.

so with regards Material Design, can you suggest changes to the Material Design that we introduced in the last release?

DzmitryShylovich:
@sasivarnan your theme doesn't respect `headerHeight` property. Example: http://plnkr.co/edit/N3MIaYctOdMLxKoJc1iT?p=preview
as u can see I set it to 100px but it still 35

sasivarnan:
@DzmitryShylovich. I guess, you will be getting the header height as 56px. 
The theme is made with exact dimensions specified in `Google Material Design Standard` for [Data Table](https://material.google.com/components/data-tables.html#data-tables-tables-within-cards), which states the height of the header to be `56`.

ceolter:
@sasivarnan any advice for us on our material design packed with ag-grid?

we did look at yours when we designed ours, however we wanted to keep ours in line with the other themes provided out of the box with ag-Grid, ie keep with the shared parts and build process we use with stylus. but if there are small changes we can make to our one to bring in line, we'd appreciate the feedback.

DzmitryShylovich:
@sasivarnan sorry to bother you. Do u have any plan to update your material theme to work with the latest grid changes? After 5.x release something broke [before](http://plnkr.co/edit/l63Tcs2EM8Of8TPB3Peg?p=preview) vs [after](http://plnkr.co/edit/rQUktiZopyG4MsrXJypL?p=preview)

PR #774, opened 2016-03-11
Add startPage parameter to datasource.

attilakocsis78:
Hello @ceolter,

We would like to set start page parameter to the table to help users with them works.
The use case is simple:
User processes records in table row by row with drill drawing.  (So there is a list of tasks and you can navigate to task detail screen to modify some data.) After modification user can step back to the table (screen) and the table should stay that page from where the navigation happen. So if user stay at the 10. page and process the task and step back then he/she should arrive at page 10. 

Maybe there is a simpler solution what i didn't recognize. If yes pls tell me.

Thanks!

### DIFF
--- a/src/ts/rowControllers/paginationController.ts
+++ b/src/ts/rowControllers/paginationController.ts
@@ -15,25 +15,25 @@ import {Events} from "../events";
 import {FilterManager} from "../filter/filterManager";
 
 var template =
-        '<div class="ag-paging-panel ag-font-style">'+
-            '<span id="pageRowSummaryPanel" class="ag-paging-row-summary-panel">'+
-                '<span id="firstRowOnPage"></span>'+
-                ' [TO] '+
-                '<span id="lastRowOnPage"></span>'+
-                ' [OF] '+
-                '<span id="recordCount"></span>'+
-            '</span>'+
-            '<span class="ag-paging-page-summary-panel">'+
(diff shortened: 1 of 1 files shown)


attilakocsis78:
any news @ceolter ?

ceolter:
thanks for contributing. the reason i've put this on hold is i want to come back and revisit the paging at some point and create a proper api into it. so want to wait until i have the time to do it proper rather than taking delta changes.

ssashok10:
The paging panel is very basic with ag-grid. It would be nice to have a paging panel with other user friendly features. Here are my suggestions,

1. The paging panel has "First", "Next", "Previous" and "Last" buttons. If we have 50 pages on the server side pagination, the user needs to click the "Next" button 30 times if he wants to go to the 30th page.  The paging panel should provide easy options to navigate between different pages.

2. It would be nice to see the "Page Size" select option on the Paging panel itself. 

3. We can't have a check box on the header level to select or un-select all rows on a page with the current version. The header check box would be a very useful feature. 

The "startPage" parameter on the data source is a very  useful option. Thanks to "attilakocsis78" to initiate it. whenever we refresh the page using F5, the grid is re-setting to the first page. There is no way we can be on the same page during the page refresh. 

Please update us if you make any progress on improving pagination. Thanks in advance.

seanlandsman:
There is a now an option to specify start page/index with the row models. Can you please verify that this covers your scenario?

PR #791, opened 2016-03-17
Rename IMenuFactory files to iMenuFactory

kampdomoj:
otherwise the build fails


ceolter:
i've changed this manually. only the ts file should be changed. the .d.ts is generated by typescript. thanks.

PR #794, opened 2016-03-17
Fix bug in virtual page row

### LINKED ISSUE #793: Crash when trying to change the sortModel before setting the datasource (infinite scroll mode)
This fix : https://github.com/ceolter/ag-grid/commit/42b60a3477b6e097042ee7f8a7d7edf79737bfc3
Needs to be applied to virtualPageRowController as well.

Reproduced here : http://plnkr.co/edit/uo1O4GjDbZyFl1YqLPTw?p=preview
(see line 72 in index.html)

`TypeError: Cannot read property 'rowCount' of undefined`

Wykks:
See #793


ceolter:
thanks

PR #821, opened 2016-04-05
Rename PopupService.js to popupService.js

kayue:


ceolter:
i don't understand what's happening here. i have already changed the file from upper case to lower case.

see here:
https://github.com/ceolter/ag-grid/tree/master/src/ts/widgets

maybe you are working off an old version?

kayue:
@ceolter But the file in `dist` directory aren't being updated.

I got the following error when compiling it.

```
Unhandled rejection Error on fetch for github:ceolter/ag-grid@4.0.5/dist/lib/widgets/popupService at file:///home/hype/hypebeast/releases/20160405102258/src/Hypebeast/Bundle/WarehouseBundle/Resources/public/jspm_packages/github/ceolter/ag-grid@4.0.5/dist/lib/widgets/popupService.js
```

kayue:
@ceolter Please consider merging this, otherwise people could not build their app...

ceolter:
i have just fixed this and released 4.0.6.

advise if the problem is not fixed. thanks.

Issue #822, opened 2016-04-06
deselectAll() aren't triggering selectionChanged event


kayue:
As title. 
It was working in previous version.

EndarValuk:
Confirmed bug. Would be nice to solve asap.

ceolter:
will be in next release.

PR #831, opened 2016-04-12
Dispatch 'selectionChanged' on deselectAll

GustavSt:
When calling deselectAll 'selectionChanged' event is not triggered, documentation states it should. Issue noted in #822.


ceolter:
this is now fixed.

Issue #812, opened 2016-03-28
autoSizeColumns() doesn't account for cell padding


mpetkov:
I was trying to use autoSizeColumns() to get the cell width to auto size to the cell content and I found an issue.  It turns out that the padding of the cell is not taken into account so the content is being cut off.  Can this functionality be updated to behave correctly?

Here is example:
http://plnkr.co/edit/wNSPC2Gnagf2Z6u6TdXg?p=preview

mpetkov:
One solution could be to just expose a new padding parameter for each column where the user can specify the padding to add to the width of content?

ceolter:
noted.

ceolter:
new property autoSizePadding was put in last year

PR #852, opened 2016-04-21
Made autoSizeColumns() to work with cells that have padding

mpetkov:
My suggested fix for issue #812


ceolter:
thanks. will take a look when i get time.

arkon:
Why not just get the container's `clientWidth` instead of its `offsetWidth`? Wouldn't that include the padding and eliminate the need for these calculations?

ceolter:
i went to work on this, and as far as i can work out, it's not needed.

in your example, i changed the overridden style to the following and it worked
```
.ag-fresh .ag-cell-not-inline-editing { padding: 0 40px; }
```

PR #858, opened 2016-04-25
Adding sortable class to sortable headers

zalohcin:
If a header is sortable this will add a class "ag-header-cell-sortable" to it, and then in the styles I added cursor: pointer.


ceolter:
thanks. will take a look when i get a chance.

ceolter:
i have added the class, but i have not taken the suggestion of changing the mouse to pointer. this is easy to add yourself if you want. reason is that no other parts of the header (eg menu) change when mouse hovers over them.

PR #869, opened 2016-04-30
textCellEditor now ignores case when initial value is undefined

MichaelMcKechnie:
I noticed that if the underlying object for a row has no property to match the 'field', switching to edit mode for the cell using the default editor TextCellEditor initialises the displayed text to the word 'undefined.' This pull request skips assignment of the initial value to the <input> element if the underlying value is undefined.


ceolter:
thanks for posting. i've implemented this manually (not a merge) for consistency, uses the ag-grid utils class (arguable with one is better, however now it's consistent)

PR #870, opened 2016-05-02
This PR adds the gridOptions.context to the angular scope as scope.co…

swalters:
…ntext.

For small grids, using angular is super convenient.  I need the ability to reference my controller in the angular template.

With this PR, I can reference my controller in a template using context.myControllerReference.controllerFunction().


ceolter:
thanks for pr. accepting, however fyi, i plan to revisit how angular is used in a few months from now, so this may change as i refactor.

PR #879, opened 2016-05-04
New material style

seanlandsman:


mikeerickson:
@ceolter wow, i was reading this ticket and it refreshed before my very eyes as you merged.  Never seen that before.   

So my question is,,, when will this be available.  I have created my own theme (along lines of Bootstrap) but would love to try material theme.

ceolter:
if you fancy building the latest, you can check out now?

we haven't scheduled a release yet, so i can't say when it will be officially available. we are just off the back of 4.1.x, the next release, emm, well it would be great to get this out. maybe end of next week... maybe not!

ceolter:
@mikeerickson seanlandsman works with me btw, this is not just a random PR

PR #906, opened 2016-05-13
Adding ability to suppress the cell focus piece

killyosaur:
So a little checking around found that the issue appears to be that you force the focus back to the page upon refresh. This may be all fine and good for most applications but if creates an issue on IE 11. So in order to allow for a user to manually choose at time of refresh whether to do this or not, I set an optional parameter on all refresh calls that call the "restoreFocusedCell" function. if that parameter is true, the function does not restore focus.


killyosaur:
Fixed my pull request, guess I should compile the typescript to ensure no errors :P

ceolter:
added property 'suppressFocusAfterRefresh' - didn't use PR as it has to many files changed - maybe you checked in the dist folder?

killyosaur:
That's fine, the PR was modified too many times. It may have been updated with a newer version of the grid as I kept checking the issues. Thanks for the add though!

PR #915, opened 2016-05-19
Fix issue with rowClass not being considered when autoSizing a column.

jonbcard:
Currently `rowClass` is not being considered when auto-sizing on a column happens. This means that changes to font or emphasis done at the row-level do not end up being considered when sizing the column. This fixes that issue.

_Note_: this does not copy `rowStyle` properties over to the cloned cell's parent, so there is still a gap there for those that use that feature.


ceolter:
interesting. i'll have a think about this. thanks.

ceolter:
i mean, i'm happy that there is a problem that should be fixed!!! i mean think about the solution.

rmc-software:
This issue has since been fixed so closing PR.

PR #977, opened 2016-06-21
eGui to afterGuiAttached as parameter

ua9msn:
Sometimes it's required to set a position of popup editor relative to the cell. 
But the afterGuiAttached has no access to the cell DOM to get the bounding rect. 
I propose the changes to pass eGui to the afterGuiAttached function for that

### DIFF
--- a/src/ts/rendering/cellEditors/iCellEditor.ts
+++ b/src/ts/rendering/cellEditors/iCellEditor.ts
@@ -21,7 +21,7 @@ export interface ICellEditor {
     init?(params: ICellEditorParams): void;
 
     /** Gets called once after GUI is attached to DOM. Useful if you want to focus or highlight a component (this is not possible when the element is not attached)*/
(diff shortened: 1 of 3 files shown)


ua9msn:
Hello Niall!
Yes, putting both eGridCell and eGui into init params would be even better, who knows where it may be used further.

ceolter:
couldn't do what i wanted to do. by having init(params) call getGui(), it then had a cyclic dependency, as getGui() has a dependency on init(prams) getting called first.

i don't want to change the afterGuiAttached() method without fully thinking on the popup service - i think what really needs to change is the popup service needs rethinking.

however, you have your own element - when afterGuiAttached is called, you can use element.parentNode to climb the dom tree, from your element to the popup parent. this would be my preferred way for now, as changing the afterGuiAttached() takes this part of the code out of sync with other parts where popup is used.

PR #1000, opened 2016-07-01
selectCellEditor stop mousedown propagation in IE

bogdanandreica:
In Internet Explorer mousedown event on select editor is passed to the next sibbling cell, and triggers cellValueChanged() callback before selectEditor update his value.
Also the selectEditor should let developer set different fields for value and text in the cellEditorParams object.


bogdanandreica:
wrong branch

PR #1035, opened 2016-07-24
bugfix - Dummy container for the purpose of width calculation, was be…

Shrulik:
…ing created for columns that weren't being rendered, and therefore aren't part of the width calculation, which is a waste, but the bug is this dummy container also wasn't being removed from the DOM.

To recreate this bug, one only needs to call columnApi.autoSizeColumns on grid with columns that aren't being rendered. There will be extra DOM elements as siblings to the rows. 

Tried to keep with the code style. Left putHeaderCellsIntoDummyContainer as a single line function for the comment.


Shrulik:
Oops, forgot to change the if condition. I will make a new one.

PR #1043, opened 2016-07-29
Add support for a default aggregation func

tloten:
An example of what I meant in this post. 
https://ag-grid.com/forum/showthread.php?tid=3988

See https://embed.plnkr.co/egDvBp/ for a demo of this PR (might need to try loading it a few times - plunkr doesn't like the large JS file).

I'm not expecting this to be merged by any means, as it is quite a specific feature - though it does integrate fairly well, and feels quite intuitive. But keep it in mind in case you know of a nicer/more generic way to allow for this functionality.

Also needs a few ag-grid-enterprise changes - see this commit: https://github.com/tloten/ag-grid-enterprise/commit/d680650235e3901f714ebf93dd4aee03313e5003

### DIFF
--- a/src/ts/columnController/columnController.ts
+++ b/src/ts/columnController/columnController.ts
@@ -519,14 +519,20 @@ export class ColumnController {
     }
 
     public addValueColumns(keys: (Column|ColDef|String)[]): void {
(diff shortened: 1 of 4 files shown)


ceolter:
i'm sticking with my forum comment :)

tloten:
I agree - doing it as a post-render step (as you suggested) was much nicer.

PR #1074, opened 2016-08-11
set default focus when the grid lost the focus

jare-talend:
I define a default focus when the grid lost the focus.


rmc-software:
Thanks for this PR but this change is no longer required. Closing.

PR #1084, opened 2016-08-16
Fix to theme-common making theming more flexible

tomek-d:
I would like to propose small change to make theming more flexible. 

Right now theming works as follow: 
- there is one file `theme-common.styl` that defines all common styles but requires some variables to be defined
- every theme defines all variables required by `theme-common.styl` and then includes this file
- as a result css file is generated with root class name `.ag-common`
- this magic class name is replaced on the fly by gulp task to match theme name eg. `.ag-blue`

This works fine but this is also a limitation. It is hard to create another theme by importing `theme-common.styl` one need to remember to replace this static class name.

As a solution to that problem I would like to propose dynamic class name generation. Every theme file has already variable named `ag-style` with theme name. To make it working all that need to be done is to define root class name in `theme-common.styl` as dynamic class name: `ag-${ag-style}` instead of giving it static name. Thanks to that workaround with replacing class name on the fly can be removed, and creating new themes is no longer problematic.

Thanks.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -150,17 +150,22 @@ function webpackTask(minify, styles) {
 function stylusTask() {
     // Uncompressed
     gulp.src(['src/styles/*.styl', '!src/styles/theme-common.styl'])
-        .pipe(foreach(function(stream, file) {
-            var currentTheme = path.basename(file.path, '.styl');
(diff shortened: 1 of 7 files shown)


ceolter:
cool! thanks :)

Issue #1090, opened 2016-08-19
5.x breaks horizontal scroll 


connorblack:
![grid-fail-generic](https://cloud.githubusercontent.com/assets/1903789/17808779/099d05b2-65e0-11e6-977d-fd064470430b.gif)

Chrome Version 52.0.2743.116 (64-bit)
Mac OSX 10.11.6 (15G31)

ceolter:
i don't see anything wrong with this on my win 10 using either chrome of safari. i don't have a mac :(

connorblack:
@ceolter it's pretty easy to replicate. Downgrading to 4.x fixes the issue. 

What can I do to help you fix this?

loverajoel:
I have the same issue on Chrome in Windows 7.
I can add efforts to resolve it

billyvg:
This is also broken for us, we are using a virtualized grid with many columns.

It looks like `ag-header-cell` cells are not being positioned properly, the `left` properties have the correct value, but it's not being rendered the way you want.

ceolter:
@connorblack @loverajoel @billyvg i did a lot of style changes with the headers between v4 adn v5, as now the columns are virtualised, the 'left' style is used to place the header cells, whereas before they were just laid out in a flow layout (ie one after the other, starting at the start). i will see if i can reproduce again - however to fix, if you can help, i either need to get it reproduced here, or if you figure out what css is needed to fix it - then i can trust you and just apply it - as long as it's not breaking anything else then i'll release based on what you suggest.

this has got to be a rare issue - something about your environments, maybe some clashing css in your apps? i say this as it has not been widely reported. 25k+ downloads in the last month since v5 was released.

ceolter:
is it on any of the online examples? or just happening inside your apps?

jeffshaver:
In my case, it was a custom theme I was using. The problem being that the css set `.ag-header-cell` to `position: relative`. Removing that fixed the issue for me

connorblack:
I still have the issue even when `.ag-header-cell` isn't `position: relative`

connorblack:
@ceolter I unfortunately can't give you access since my project is private

billyvg:
@ceolter You're right, we were using ag-grid from npm, but using ag-grid CSS from a different source so it was outdated.

light049:
I also meet this issue, I use ag-grid in react.

light049:
Sorry, fixed after .ag-header-cell adding position: absolute

[...]

PR #1096, opened 2016-08-23
Refactor stylus themes to separate theme variable definitions

billyvg:
This allows users to easily extend themes and gives them an easier upgrade path as well. Inspired by our mistakes with #1090 

Example:

``` stylus
@import 'ag-grid/src/styles/vars/fresh';

ag-background-1 = pink

@import 'ag-grid/src/styles/theme-common';
```


billyvg:
I'm not super familiar with Stylus, but SCSS has a `!default` property (https://robots.thoughtbot.com/sass-default) that would make this PR unnecessary.

After a quick glance, I couldn't find the equivalent in Stylus.

ceolter:
interesting. i'll need to revisit this when i have more time, as need to make double sure no side effects.

flybayer:
I'd love to have this feature for customizing a theme

PR #1141, opened 2016-09-18
isRowModelDefault() doesn't need the !(...)

rchipman:


ceolter:
not true - the row model is default if it is NOT one of the following: viewport, pagination, virtual pagination

:)

PR #1265, opened 2016-11-15
Prevent errors in SSR setups

kaanon:
This is in response to issues like https://github.com/ceolter/ag-grid-react/issues/24
I use a server-side rendering setup, and even though I know ag-grid won't support server side rendering, the reference to `HTMLElement` is preventing me from even including the file.

This naive polyfill won't add support, but it will allow my setup to build, so that I can use ag-grid in the client side where HTMLElement exists.


ceolter:
why can you not put this into your code, before you include the ag-Grid library?

crumblix:
Just joining in on this PR ... in favour.  Firstly thanks Ceolter for a great library :)

If there is a better way of fixing this in our code I'd be happy to, but if I understand correctly you're suggesting we should manually edit these files every install?  That doesn't sound ideal to me I must admit.

With the latest release I am now finding I need to also edit:

headerGroupComp.js
headerComp.js
dateFilter.js
headerWrapperComp.js

In addition dateFilter.js also needs:
var HTMLSelectElement = typeof HTMLSelectElement === 'undefined' ? function(){} : HTMLSelectElement;

And (although it may not be related) I also need to swap icons in headerGroupComp.js:
        this.addInIcon('columnGroupOpened', 'agClosed', svgFactory.createGroupExpandedIcon);
        this.addInIcon('columnGroupClosed', 'agOpened', svgFactory.createGroupContractedIcon);

Otherwise I get a plus when I should get a minus when expanding/collapsing header groups.  Not sure what's going on there, but since I've had to edit other changes my suspicion is I've inadvertently broken something.  And of course if that's the case, what else have I broken?  Nothing I can see at the moment, but you just can't be sure.  It leaves you with an uneasy feeling.

If the changes really don't hurt or break anything for people not getting this problem, and there isn't a better way to fix it, then I can't see what is to be lost by merging the PR?  Except of course it obviously needs updating now.

Anyway thanks again ... love the new features, especially the checkbox feature in the header :)

manuxio:
@ceolter 
> why cant you not put this into your code, before you include the ag-Grid library?

Strictly speaking, imports should come above any other code (linter says...), so you can't do that if you want to remain compliant with major rules.

Plus, I have the feeling that the context of "import"s is always different, so you can't really define anything that will pass to the "imported" code, except, maybe, for global.*

Besides, in another thread, you asked what was the reason of this code...
I believe you understand the lexical meaning of the code, so can you explain a little more your concerns?

I am developing a SSR application with your grid, so I feel the need to be helpful in that matter, if I can.

Will check the context issue in a couple of minutes (maybe a little more...) and eventually update the post.

Thanks

manuxio:
I have created a new PR for this...

seanlandsman:
I'll add these declarations in a future release. I'll add them to our TS files so I wont accept the PR as it stands, but thank you for raising this.

As I say, it will be in the available next release of ag-grid

PR #1335, opened 2016-12-13
removes unused `ag-background-2` variable

dimitropoulos:
At least, it appears to be unused, if I'm not mistaken.


ceolter:
thanks!

Issue #1386, opened 2017-01-09
sizeColumnsToFit() bug with horizontal scrollbar


uvtzxpm:
Reproduction:
1. Go to https://www.ag-grid.com/javascript-grid-resizing/
2. In the example at the bottom, keep clicking the "Size to Fit" button.
3. The grid cannot decide on a constant size for the columns between calls to `sizeColumnsToFit()`. The problem might be related to the width of the vertical scrollbar.

GIF showing problem: http://i.imgur.com/Z8CajVQ.gif

I'm using the most recent Chrome (55.0.2883.87 m).

Probably related to: https://github.com/ceolter/ag-grid/issues/1333

I came across this problem because I'm trying to implement something like: https://github.com/ceolter/ag-grid/issues/627

Ruud-cb:
+1 of course, indeed the same issue as #1333 , but then I try to do it automatically.

Ruud-cb:
Found the issue, see PR

uvtzxpm:
@Ruud-cb I can confirm that your PR seems to fix the problem for me. Thanks!

nischi:
@Ruud-cb I found out that the issue only happens if you have a display with high dpi and the width is a floating value and not an integer. can you confirm that?

Ruud-cb:
@nischi I have one iiyama proLite E2208HDD with a dpi of 100 and resolution of 1980x1080, and one old samsung syncMaster 940bw with a dpi of 96 and resolution of 1440x900. So if you have a screen with a lower dpi or face issues, please provide some more info on what is happening at yours.
Also, please check for yourself if it makes a difference for a floating value or integer. As far as I've seen, it was always an integer during debugging.
Perhaps take on this discussion in the PR itself.

nischi:
@Ruud-cb I have a ASUS PB287Q with 3840x2160 and 157dpi. I have only the problem if the div in which the grid is has a float number like 2507.2. If i change that to 2507 or 2508 it works without blinking. So i think something with calculate the width is wrong if you have a width with a float-number.

Ruud-cb:
@nischi Hmm.. well I've tried to change the container div of the grid with floating values, but this seems to have no real effect on the calculation, because it uses integer values. It eventually uses `clientWidth` which returns an int.
About the dpi, hmm maybe, but a phone has a dpi of somewhat 300, where I don't have the issue now.
[...]

Issue #1333, opened 2016-12-13
size columns to fit causes scrollbar flipping


Ruud-cb:
Hi,

I'm having trouble with the `sizeColumnsToFit()` method. I'm executing this with a timeout after a column is resized, like so:
```
this.columnResizedEvent = function (col) {
            clearTimeout(self.timeoutResize1);
            self.timeoutResize1 = setTimeout(function () {
                self.gridTableOptions.gridOptions.api.sizeColumnsToFit();
                clearTimeout(self.timeoutResize1);
            }, 1000);
    };
```

where `self` just refers to an object so we have a global variable.
The idea is to resize the columns after a change in width has happened, but only execute it after a minute, so it is not happening instantly.

Best to illustrate the problem is by providing some pictures.
First, this is the state of the grid, everything is correct. (The height is static).

![image](https://cloud.githubusercontent.com/assets/10037514/21137888/e4a56022-c12b-11e6-98d8-7211a7b4da54.png)

Then, I drag the middle column and start resizing larger, when I do this the scroll bars appear, that's obvious because it doesn't fit anymore.

![image](https://cloud.githubusercontent.com/assets/10037514/21137934/182dd7b2-c12c-11e6-967f-4158fed8c9a1.png)

Now when I stop resizing, the event kicks in and the timeout executes which resizes all columns.
The result is some what surprising:

![image](https://cloud.githubusercontent.com/assets/10037514/21138120/fc3fd7e8-c12c-11e6-822d-02b1f12f3b5a.png)

As you see, there is enough space below it so there is no scroll bar needed, yet it still keeps showing up.
Now at that time, the column resize event gets triggered again, eventually after a second causing another call to `sizeColumnsToFit()`. Then, this happens:

![image](https://cloud.githubusercontent.com/assets/10037514/21138195/3c3d0320-c12d-11e6-8637-8d3594a3afe2.png)

The horizontal scrollbar shows up again, while I can't scroll with it.

It is very hard to re-produce, it requires a specific height with a specific amount of items. Worst-case scenario is happening when the column resized event gets triggered again, causing an infinite loop in flipping scroll bars.
I think there is a little bit going wrong in the calculation when columns get resized.

Could you please advise me on a work around or fix this bug?

AnaSofiaGrilo96:
+1

Piccirello:
I'm experiencing this issue as well.

[...]

PR #1399, opened 2017-01-11
Fix calculation for sizeColumnsToFit

Ruud-cb:
Fix for #1333 and #1386 

Hi @ceolter,
I had the issue described in 1333. When I debug through the `sizeColumnsToFit` function, I've noticed that, when having both a vertical and horizontal scrollbar, `this.isBodyVerticalScrollShowing();` is true and thus reduces the `availableWith`. When only having a vertical scrollbar, `this.isBodyVerticalScrollShowing();` is false and used the full client witdh. Using the below function seems to fix the problem. It also works correctly when having a column pinned right.

I was thinking about changing `this.isBodyVerticalScrollShowing()`, perhaps the `else` function should be changed.
Tested on chrome v 55.0.2883.87

### DIFF
--- a/src/ts/gridPanel/gridPanel.ts
+++ b/src/ts/gridPanel/gridPanel.ts
@@ -847,8 +847,8 @@ export class GridPanel extends BeanStub {
         // if pinning right, then the scroll bar can show, however for some reason
         // it overlays the grid and doesn't take space. so we are only interested
         // in the body scroll showing.
-        var removeScrollWidth = this.isBodyVerticalScrollShowing();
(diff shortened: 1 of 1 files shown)


Ruud-cb:
https://github.com/ceolter/ag-grid/issues/1386#issuecomment-272374664
perhaps take a look at this comment, a Todo

ceolter:
thanks.

Ruud-cb:
Hi, Thanks for merging, one did tell me that it was fixed in v7.2.2, I didn't check that version before submitting this PR. (FYI)

PR #1473, opened 2017-02-14
Server side rendering, defining fallback HTMLElement and HTMLSelectEl…

manuxio:
…ement

Changes filter/dateFilter to allow server side rendering...


seanlandsman:
I'll add these declarations in a future release. I'll add them to our TS files so I wont accept the PR as it stands, but thank you for raising this.

As I say, it will be in the available next release of ag-grid

luanped:
Ag grid currently still don't support SSR right?

AlejandroFlorin:
Same question, just confirming the SSR is not yet supported, correct?

PR #1496, opened 2017-02-25
For information only: Implement an RxJS Observable-supporting row model.

mrsheepuk:
I have no expectation that you guys will merge this one, but I wanted to share what we've done with you for your information / reference - if there's anything useful from it for the main ag-Grid, you're welcome to adopt or ignore as you see fit (any changes I have made can be considered MIT licensed) :) 

If there's anything we've done that screams "terrible idea!!!" please let me know!

The purpose of this change is to allow an RxJS Observable data source to power the row data, with intelligent RowNode re-use when that observable emits an array containing identical objects to previous, and to re-use those nodes but update their data when the object is different but has the same 'ID' (identified from the data by a configured property rowDataSourceKeyProperty).

The reason for doing this at all was that we found in our use case for the grid (a data table with ~30 columns, ~50 rows displayed, data set of up to 2,000 rows, updating from a data source every 5 seconds), on Internet Explorer (11) in particular, there was a jarring pause while the browser processed the data dump each time we refreshed the data. We're using a unified top-down read-only immutable state for our application (controlled by NgRx Store), and our state update process to refresh the data ensures that where the data is identical, the objects get re-used, and where it changes, new objects are returned.

When using this observable in memory row model, we now get great performance for this use case.


seanlandsman:
Hi,

Thanks for the PR - some interesting ideas here. 

We've done a number of changes around how you can update the grid, as well as providing RxJS examples here https://www.ag-grid.com/javascript-grid-rxjs/?framework=all#gsc.tab=0

Please have a look and see if this helps. It's unlikely we'll incorp RxJS support directly into the grid, but the api and examples above should hopefully help

thanks for the PR!

mrsheepuk:
Hi @seanlandsman - just got round to moving from our fork back to the main ag-grid and I'm happy to report excellent performance now with updates flowing from an observable, using your examples as a guide.

Thanks!

PR #1498, opened 2017-02-28
Only prevent scroll when gridPanel can scroll

kindy:


rmc-software:
Closing as this PR is no longer required. Thanks.

PR #1538, opened 2017-03-20
Added no-dependencies badge to README

AmitMY:
A very interesting thing about ag-grid is that it is dependencies free.
I think it would be nice to show it on README


AmitMY:
That shield does not work anymore.
I still think the idea is solid, and you should add a shield for it @ceolter 
Closing

PR #1546, opened 2017-03-23
Supporting cellRendererFramework in group cell renderer

### LINKED ISSUE #1545: AutoGroupColumn doesn't support cell renderer framework
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
```

**Current behavior**
Currently, if I use aurelia ag-grid (or any other framework!) and using my own cell render template for a column which I want to group - when I group the column - it displayed the original data (plain text) instead of using the column's cell template.
(This BTW, is because the groupCellRenderer code only checks if a cellRenderer was deined on the column, instead of looking AS WELL if a cellRendererFramework was defined - before returning the plain data.)

**Expected behavior**
That grouped columns cell renders will correctly display the framework template I defined for that column.

**Minimal reproduction of the problem with instructions**
Define an ag-aurelia grid (or any other framework grid) with a column that has its own cell template (e.g. with red background) and group it.
You will see that when grouped - the plain text data is presented - and not with the red background.

THIS IS A BUG.

N/A

* **ag-Grid version:** 8.2.0

YES - I'm suing latest version.

* **Browser:**
All

* **Language:**
all | TypeScript X.X | ES6/7 | ES5

bondib:
fixes #1545
It would be nice BTW the receive a free dev. license for the help...


rmc-software:
Closing this PR as this issue has now been fixed. Thanks.

PR #1587, opened 2017-04-11
fix(gridOptions): set type of paginationAutoPageSize to boolean

PatrickEifler:
Please take the following issue as reference: https://github.com/ceolter/ag-grid/issues/1563

According to the documentation the paginationAutoPageSize field should be a boolean.
In the ComponentUtil Class it is already part of the BOOLEAN_PROPERTIES
Enum. Thats why this PR only consists of the gridOptions update.


kylecannon:
+1

kylecannon:
@ceolter @seanlandsman can we please get this looked at? This is causing issues with Angular's build process.

ceolter:
will be in next release.

PR #1593, opened 2017-04-15
Fix: cell should not be selected in range on row selection when also using ag-grid-enterprise

zealoushacker:
When using ag-grid with ag-grid-enterprise, and enabling both range selection and row selection, the grid creates a focused cell, which also becomes one of the selected ranges. This is highly undesirable, because when selecting a row, the intent of a user is *not* to *also* select that row's 0th cell as one of their selected ranges. 

This fixes that bug.


zealoushacker:
For this to work properly with ag-grid-enterprise cell range selection, here's the relevant PR on ag-grid-enterprise: https://github.com/ceolter/ag-grid-enterprise/pull/123

rmc-software:
This is working as expected when both row selection and range selection is enabled, so closing this PR.

PR #1635, opened 2017-05-15
#1634 issue -  creating single-click-edit option in colDef object

saidganim:
Added single-click-edit field in colDef interface . Also added checking this field in click-handler.


rmc-software:
This has been added to the latest branch and will be merged into master once we do our next release. Thanks.

PR #1689, opened 2017-06-05
[bugfix] custom floating filters

jeffskelton3:
**Summary**

Partial fix the issue mentioned in this issue: https://github.com/ceolter/ag-grid/issues/1528 despite it being closed, it still appears to be happening as of version 10.0.1 

Heres a [plunkr](https://plnkr.co/6FFPWITofzjz7kffJ2qf?p=info) using version 10.0.1 that demonstrates the issue still happens.

**Details**

`onParentModelChanged` appears to be be missing from the `mandatoryMethodList` array in the component provider. Any time one tries to update a filter using a custom floating floating filter type, it breaks and throws a `this.floatingFilterComp.onParentModelChanged is not a function` error. This PR adds them to the array and appears to prevent the error from happening. That said, the filter model does not appear to be getting updated (to be fair, I could be missing something there). I'm still submitting this as the custom floating filters are simply not functional as is and this allows one to work around since we have access to the grid api directly in versions 10+ (thank you soooo much for adding that btw. So useful!!).

### DIFF
--- a/dist/lib/componentProvider.js
+++ b/dist/lib/componentProvider.js
@@ -50,42 +50,42 @@ var ComponentProvider = (function () {
                 defaultComponent: headerGroupComp_1.HeaderGroupComp
             },
             setFloatingFilterComponent: {
-                mandatoryMethodList: [],
+                mandatoryMethodList: ['onParentModelChanged'],
                 optionalMethodList: [],
                 defaultComponent: floatingFilter_1.SetFloatingFilterComp
             },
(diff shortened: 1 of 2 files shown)


bcandullo:
This fixes the errors I was experiencing using angular4 + ag grid 10.0.1. The error is no longer thrown, and filter works correctly.

rmc-software:
This issue has now been fixed in master so closing this PR. Thanks.

PR #1711, opened 2017-06-15
Fix default filter type selection on initial render

jakeboone02:
If `filterParams.defaultOption` is set for a colDef, the choice is not reflected in the filter type drop down menu on the first load.

Example:

```typescript
import {ColDef, BaseFilter} from 'ag-grid';
const cd = <ColDef> {
  filter: 'number',
  filterParams: {
    defaultOption: BaseFilter.LESS_THAN
  }
};
```

The first time the filter menu is opened for this column, the drop down for filter type will be set to "Equals" even though the actual filter used is `lessThan`.

This PR should resolve the issue.


solvemon:
Here is a reproduction of this bug: https://plnkr.co/edit/PcnTlVHqMdQzib8IyH4N?p=preview

PR #1714, opened 2017-06-16
Add IViewportDatasourceParams to main exported type definitions

Amerzel:
https://github.com/ceolter/ag-grid/issues/1713

```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
```

I would like to see ```IViewportDatasourceParams``` added to the exported types in ```main.d.ts``` so that I can easily import and use this type in my code. With the recent restructuring of the exported types list I now have to import ```IViewportDatasourceParams``` directly from ```ag-grid/dist/lib/interfaces/iViewportDatasource``` in order to use the interface.

**ag-Grid version:** 10.1.0


amirburbea:
YES!!!
Currently in my app I have to do this

import { IViewportDatasourceParams } from 'ag-grid/dist/lib/interfaces/iViewportDatasource';

which is silly

seanlandsman:
Hi

IViewportDatasourceParams should now be exported - please upgrade and this should resolve the problem for you

thanks

PR #1771, opened 2017-07-20
ag-grid 11

pillaiv:
upgrade from version 6 to 11


seanlandsman:
Hi - thanks for your contribution but this PR will not be accepted. It touches far too many files and it's not clear to me what you're actually trying to do here.

thanks

PR #1817, opened 2017-08-09
Fix CSV headers with grouping columns

shirosaki:
CSV headers are not correctly displayed with grouping columns.
When 3 level nested headers, header position of second level is
wrong.
Header `undefined` is shown above no grouping columns.

For example grouping header of CSV export looks wrong with the following header.

header_example

CSV export shown in Excel
csv_export_example

### DIFF
--- a/src/ts/gridSerializer.ts
+++ b/src/ts/gridSerializer.ts
@@ -332,18 +332,18 @@ export class GridSerializer {
     }
 
     recursivelyAddHeaderGroups<T> (displayedGroups:ColumnGroupChild[], gridSerializingSession:GridSerializingSession<T>):void{
-        let directChildrenHeaderGroups:ColumnGroupChild[];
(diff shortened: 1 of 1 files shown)


shirosaki:
`recursivelyAddHeaderGroups` was added in this commit.

https://github.com/ag-grid/ag-grid/commit/f149c7da863590b3e78cb05806352ff783aced8c

I think that `AG-428 Allowing more than two grouped headers` is not completed.

makinggoodsoftware:
Hi,

Thanks for this, since we develop in latest and not in master, I could not merge easily your changes in, so I have done it manually, this will be part of master from the next release.

Thanks!

shirosaki:
Thanks!
I tried Version 13.0.0, but `undefined` is still shown.
Another fix is #1847.

PR #1825, opened 2017-08-13
fix cell grouping for columns that are null, undefined or empty string

shmuely:
fix cell grouping for columns that are null, undefined or empty string


rmc-software:
This is by design so we are rejected it. As a workaround users supply an empty string rather than null.

PR #1847, opened 2017-08-28
Fix group header name `undefined` in CSV/Excel export

shirosaki:
If an empty column group is built, `undefined` is shown in
CSV/Excel export.


shirosaki:
This seems to be fixed by AG-749.

PR #1848, opened 2017-08-28
Fix not using innerRenderer for non group row

shirosaki:
If group renderer with innerRenderer option is used, innerRenderer
is not used for non group row with version 13.

innerRenderer works for non grouping row with version 12.


ceolter:
we have fixed this locally. thanks for the patch, however it doesn't cover all the use cases.

PR #1862, opened 2017-08-30
Fix problem when running embedded in gwt

StephanWald:
This fixes our integration problems with our GWT. If I use the getDocument() that we set when constructing the document, both popup menus open and close correctly, _and_ D&N works. If document is used, the popups do not close correctly when clicking outside them somewhere else on the grid.

Maybe that's not everything that's needed to build all your versions, but changing this statement in the dist files made it work for us in all integrations.


ceolter:
this is now in
https://github.com/ag-grid/ag-grid/commit/29e2d9cd1ac422ba01d6e56a25650b9415f698e3

didn't accept the PR as we program against branch 'latest', not 'master'

PR #1876, opened 2017-09-02
Rename method calls for propagating changes to pinned row data

holmari:
Without this change, the pinned rows would now appear, at least in conjunction with ag-grid-react.


holmari:
Looks like this is fixed.

PR #1881, opened 2017-09-04
[bugfix] Fixing 'pivotPanelShow' option type

rafaelaudy:
pivotPanelShow was considered a boolean, so you were not able to pass 'always' or 'onlyWhenPivoting' as an option.
It would always default to false.

Pretty similar to this error which was about the property 'rowGroupPanelShow':
ag-grid/ag-grid-enterprise#128


rafaelaudy:
Pretty similar to this error:

rmc-software:
This has been added to latest branch and will be merged into the master branch when we do our next release. Thanks.

Issue #1894, opened 2017-09-06
Clear button doesn't respect with filterOptions


cdumas22:
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see 

When you have the specific settings for a column
{
  filter: 'text',
  filterParams: {
    clearButton: true,
    filterOptions: ['equals'],
    defaultOption: 'equals'
  }
}

Issue 1 - The dropdown starts out as empty if you dont put defaultOption, and the filter defaults to contains. 
Issue 2 -  when you click the clear button it will switch the filter to contains even though that is not in the list of available values.

I would expect the filterOptions to only allow options specified in the passed in parameter and to default to the correct default filter type on clear.

* **ag-Grid version:** 13.1.0

makinggoodsoftware:
Hi,

We have created AG-759 to track this issue this should be fixed in the next one or two releases.

Thanks for raising this.

killyosaur:
As an FYI, if I change the data in the grid, the filter options are not being respected either.

killyosaur:
A further dive into this bug (at least as far as I am seeing in our situation, but it might help determine what the underlying issue is), when the filter initializes, the filterParams are applied correctly. When a change is made to the grid, these are ignored on subsequent requests. So in our situation, if I haven't opened the filter yet, then changing the grid data does not affect the way the parameters are applied. Only after initialization is this an issue.

seanlandsman:
Thanks for the follow up - I've raised it as a comment to the issue above

makinggoodsoftware:
Hi,

Just to let you know that we have just delivered 13.3.1 which includes fixes related to this thread, for the full change log please check:

https://www.ag-grid.com/change-log/changeLogIndex.php

Thanks

PR #1908, opened 2017-09-12
Fix for the default options bug

killyosaur:
This corrects the bug where when a grid's data changes (or the clear button is called as that is probably related to this issue as well considering the problem is related to the reset function hard coding the defaultFilter that it resets to). Test out the clear as well (my app doesn't do that). Also, fixed the date filter's default to use the BaseFilter.EQUALS instead of the text.


killyosaur:
this is a possible fix for https://github.com/ag-grid/ag-grid/issues/1894 in case it wasn't clear :P

Issue #1889, opened 2017-09-05
Class 'BaseFilter<T, P, M>' incorrectly implements interface 'IFilterComp'.


katasup:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

```
[at-loader] Using typescript@2.4.2 from typescript and "tsconfig.json" from /Users/username/projects/myProject/tsconfig.json.

[at-loader] Checking started in a separate process...

[at-loader] Checking finished with 1 errors
Time: 41624ms
chunk    {0} app.min.js, app.css (app) 484 kB {1} [initial] [rendered]
chunk    {1} vendor.js, vendor.css (vendor) 4.11 MB [entry] [rendered]

ERROR in [at-loader] ./node_modules/ag-grid/dist/lib/filter/baseFilter.d.ts:21:31 
    TS2420: Class 'BaseFilter<T, P, M>' incorrectly implements interface 'IFilterComp'.
  Types of property 'afterGuiAttached' are incompatible.
    Type '(params: IAfterGuiAttachedParams) => void' is not assignable to type '((params?: IFilterAfterGuiAttachedParams | undefined) => void) | undefined'.
      Type '(params: IAfterGuiAttachedParams) => void' is not assignable to type '(params?: IFilterAfterGuiAttachedParams | undefined) => void'.
        Types of parameters 'params' and 'params' are incompatible.
          Type 'IFilterAfterGuiAttachedParams | undefined' is not assignable to type 'IAfterGuiAttachedParams'.
            Type 'undefined' is not assignable to type 'IAfterGuiAttachedParams'.
```

**Expected behavior**

**Minimal reproduction of the problem with instructions**

webpack@3.5.4, awesome-typescript-loader@3.2.2

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

MacOS Sierra, VS Code, Node.js 8.4.0

* **ag-Grid version:** 13.1.0

* **Browser:**

* **Language:** TypeScript@2.4.2

makinggoodsoftware:
Hi,

This seems to be compiling fine in our end, can you check if you are not using mixed versions of ag-grid in your modules (ie all are 13.1.0)

Thanks

katasup:
[...]

PR #1909, opened 2017-09-12
Fix incorrect signature of `afterGuiAttached`

manu-st:
Change signature to have an optional parameter `params` to match the specification in interfaces/iComponent.d.ts. Otherwise it causes TypeScript with strictNullChecks to throw an error. See issue #1889

### DIFF
--- a/dist/lib/widgets/component.d.ts
+++ b/dist/lib/widgets/component.d.ts
@@ -19,7 +19,7 @@ export declare class Component extends BeanStub implements IComponent<any, IAfte
     private compId;
     constructor(template?: string);
     setTemplateNoHydrate(template: string): void;
-    afterGuiAttached(params: IAfterGuiAttachedParams): void;
+    afterGuiAttached(params?: IAfterGuiAttachedParams): void;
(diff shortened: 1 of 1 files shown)


mast4461:
I'd appreciate it if this pull request was evaluated as soon as possible. 🙂

nmdanny:
you should probably modify the source as well as dist

sylvanaar:
Yes, you want to change it in `src`, and the build `dist` normally.

UselessPickles:
Something worth considering (requires someone knowledgable of this codebase, intent, etc)...

There are 2 "solutions" that fix the compiler error:

Update the signature of IComponent.afterGuiAttached to make "params" NON-optional.
Update the signature of the method Component.agterGuiAttached to make "params" OPTIONAL, and update the implementation to handle the fact that "params" may be undefined.
Which is actually the correct solution? Is "params" truely optional? Or will a non-undefined/null value always be passed to the method?

rmc-software:
Closing PR as this issue was fixed since v14.1.1.

PR #1966, opened 2017-10-05
Fix issue #1889: incorrect signature for 'afterGuiAttached'

nmdanny:


src/ts/widgets/component.ts
```diff
@@ -42,7 +42,7 @@ export class Component extends BeanStub implements IComponent<any, IAfterGuiAtta
         this.template = template;
     }
 
-    public afterGuiAttached(params: IAfterGuiAttachedParams): void {
+    public afterGuiAttached(params?: IAfterGuiAttachedParams): void {
         if (!this.eHtmlElement && params.eComponent) {
```
UselessPickles: "params" is now potentially undefined in this code. A null-check should be necessary now.

Or is it more correct to change the signature of this method in IComponent to make the param non-optional? Is it really optional?

UselessPickles: Copying my comment from other related places. This requires a decision from someone that understands the codebase and the intended design of this interface method:

There are 2 "solutions" that fix the compiler error:

1. Update the signature of IComponent.afterGuiAttached to make "params" NON-optional.
2. Update the signature of the method Component.agterGuiAttached to make "params" OPTIONAL, and update the implementation to handle the fact that "params" may be undefined.

Which is actually the correct solution? Is "params" truly optional? Or will a non-undefined/null value always be passed to the method?

rmc-software:
Closing PR as this issue was fixed since v14.1.1.

PR #1985, opened 2017-10-16
fix #1961

### LINKED ISSUE #1961: AG Grid does not sanitize data displayed on the grid!
```
[ ] bug report => search github for a similar issue or PR before submitting
[X] feature request
```

**Current behavior**

If a string value containing HTML comes into the grid, the grid just displays it without sanitizing the data, In some instances, this causes the grid to break

**Expected behavior**

Preventing the grid from breaking due to unsanitized data in the grid (per example)
prevent possible XSS issues

OS: Win 10, IDE VS 2017, Bower

* **ag-Grid version:** X.X.X

13.2.0

* **Browser:**

Both IE and Chrome

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

dominikg:


dominikg:
this is just a quick patch, for performance reasons you might want to consider 

a) store document reference on params change
b) use cloneNode on a template span instead of createElement

i imagine the default cellrenderer gets used a lot, so optimizations here may be beneficial

dominikg:
fixed by the authors in a different way

PR #1990, opened 2017-10-18
Delay the layout math until the styles are present

petyosi:
Guys, please review this. It improves the way layout happens when using the scripts which bundle css. 

It basically waits to do the layout until the loaded styles actually kick in. There is also a "I give up" clause. 

The CSS property I use should have no effect on the element.


petyosi:
Related to AG-942

petyosi:
There are several cases where this just won't work and break existing setups. Closing.

Issue #1995, opened 2017-10-19
`SerializedDateFilter` not exported


WxWatch:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

`SerializedNumberFilter` and `SerializedTextFilter` are being exported by `ag-grid/main` but `SerializedDateFilter` is not. This is preventing the creation of a custom date filter component.

**Expected behavior**

Without the bug, I would be able to import `SerializedDateFilter` and use it to create a custom filter component, similarly to what I can do with `SerializedNumberFilter` and `SerializedTextFilter`

**Minimal reproduction of the problem with instructions**

Simply try to `import { SerializedDateFilter } from 'ag-grid/main';` in any `TS` file

**What is the motivation / use case for changing the behavior?**

I'm creating custom filters, including a date filter, and need this exported in order to do so.

**Please tell us about your environment:**

OSX, npm, webpack, angular cli

* **ag-Grid version:** X.X.X

13.3.1

* **Browser:**

N/A

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TS 2.3.4

petyosi:
Hi, 

I logged this as  AG-957, our dev team will review it and update the status here.

WxWatch:
Thanks for the update.

Just FYI, PR #1996 resolves this if you guys want to just review that and merge it. :)

makinggoodsoftware:
Hi,

The way we see this file is as a minimum entry point of dependencies that we export, we actively try to export as less as possible.

If you want to use our internals at your own risk, then you can, but you will have a different sintax for the import ...

ie 'dist/lib/...'

Thanks

===============

If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

Bugs Raised Here:

We will do a periodic review of GitHub issues for any bugs raised and these can be tracked on our pipeline page: https://www.ag-grid.com/ag-grid-pipeline

Feature Requests Raised Here:

We are reviewing how we process Feature Requests that come from GitHub. We currently have a very full pipeline of work which will be our priority for the coming months.

As a result, we don’t have capacity to take feature requests from GitHub.

We will review this as we move through our current backlog.

[...]

PR #1996, opened 2017-10-19
Add export SerializedDateFilter

WxWatch:
This resolves issue #1995


rmc-software:
This has been added to latest (main.ts is used now rather than exports.ts). It will be added to master in our next release. Thanks.

PR #1998, opened 2017-10-22
Added full column editing as a beta-feature

mourtisma:
Hello,

I know you rarely accept feature PRs, but I thought that one might be useful : full-column editing.

## How does it work ?
To enable it, you just have to set `gridOptions.editType = 'fullColumn'`.

## Description of the changes

- Added the new option in `gridOptionsWrapper.ts`

- Added `CellComp[]` as an attribute of `Column` (a list, not a JSON for the moment)

- Renamed `StartRowOrCellEdit` of class `CellComp` to `StartComponentEdit`, as we distinguish now between rows and columns. Did the same for `StopRowOrCellEdit`

- Added methods in the `Column` class for starting and stopping editing. For the moment, they do not have any listener (we use the Cell's listener instead). To keep some coherence, maybe we should create `ColComp` and `ColRenderer` classes, like for the rows

## Improvements to be done if we want to continue this way

- As said before, treat a Column like a Row, and add the corresponding Component and Renderer classes.

- Identify the colum's CellComps by the RowIndex

### DIFF
--- a/src/ts/entities/column.ts
+++ b/src/ts/entities/column.ts
@@ -15,6 +15,8 @@ import {IEventEmitter} from "../interfaces/iEventEmitter";
 import {ColumnEvent} from "../events";
 import {ColumnApi} from "../columnController/columnController";
 import {GridApi} from "../gridApi";
+import {CellComp} from "../rendering/cellComp";
+import {Beans} from "../rendering/beans";
 
 
 // Wrapper around a user provide column definition. The grid treats the column definition as ready only.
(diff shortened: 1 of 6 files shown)


rmc-software:
Thanks for the feature PR however we can't accept it as is unfortunately.

We envisage an API where you can edit and cells you want concurrently - so this API could be used to edit all cells in a column. That means that the 'editing' would not be coded against the column, editing the column would just be a particular usage of the general 'edit this bunch of cells' API.

We do have a plan to enable multi-cell editing - or whole grid editing -  when we do, we would then look at this PR for inspiration.

Thanks

PR #2026, opened 2017-10-31
Performance improvements

chawilso:
Speed of rendering a table was getting quite slow. With the below changes I was able to have a 50% perf increase in table rendering speed by removing expensive checks when they don't need to be made.


chawilso:
@ceolter  @seanlandsman  Any chance you can take a look at my two perf PRs. They're fairly simple, and at least for us, sped up the initialization quite a bit.

ceolter:
sorry for the long wait on this, we are hiring people and trying to catch up on everything. we regards this issue, we no longer use this mechenism for calculating the grids dimensions. instead we use CSS flex. so the suggest code change is no longer relevant. thanks for the input.

PR #2027, opened 2017-10-31
auto height perf

chawilso:
We saw a major slowdown in the time it takes to do a row expansion when auto-height was turned on.

We have a table with 20 rows, and one of the columns has a custom angular component rendered in the cells. Every time a row was expanded this would cause a redraw of every cell, including these custom angular components. 

The cost of this was quite high, and on Edge we noticed it was taking around 3 seconds to expand. I was able to get that time down to 150ms by NOT redrawing the rows and just ensuring the dom order. I think there could be other fixes here instead, perhaps just putting the new expanded rows in the correct location instead of ensuring order afterwards.

### DIFF
--- a/src/ts/rendering/rowContainerComponent.ts
+++ b/src/ts/rendering/rowContainerComponent.ts
@@ -38,7 +38,7 @@ export class RowContainerComponent {
 
     @PostConstruct
     private postConstruct(): void {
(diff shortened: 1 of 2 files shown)


ceolter:
auto-height no longer works in the same way. now auto-height works like the other rows and the whole grid is not refreshed.

Issue #2061, opened 2017-11-16
incorrect reference in updated sass theming


outofgamut:
**I'm submitting a ...**
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Angular build breaks, giving the following message `Undefined variable: "$ag-mat-font-primary"`

**Expected behavior**
Build succeeds

**What is the motivation / use case for changing the behavior?**
incorrect import in sass file

**Please tell us about your environment:**
Mac OS High Sierra, VSCode, NPM

* **ag-Grid version:** 14.2.0

* **Browser:** "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36"

* **Language:** Sass (.scss)

<img width="1361" alt="build error" src="https://user-images.githubusercontent.com/3232357/32915342-2354d7e6-cade-11e7-8ab0-0dc49a74e896.png">

aciccarello:
Here's a workaround

```scss
$ag-mat-icons-path: "~ag-grid/src/styles/material-icons/";
$ag-icons-path: "~ag-grid/src/styles/icons/";
$ag-mat-primary: mat-color($primary);
$ag-mat-accent: mat-color($accent);

// Define this variable using the primary color. Should be removed once #2062 is merged
$ag-mat-font-primary: $ag-mat-primary;

@import '~ag-grid/src/styles/ag-theme-material';
```

petyosi:
Hey, 

Thanks for that, https://github.com/ag-grid/ag-grid/commit/d4afd5aab4b80e919ec3ed63b197c06c0c76ed6c is addressing it. You can also update your override to `$primary-color` meanwhile.

PR #2062, opened 2017-11-16
fix var reference

outofgamut:
**Fix:** Change sass variable reference

Resolve incorrect sass referencing breaking certain builds.

- Change sass variable reference from `ag-mat-font-primary` to `ag-mat-primary`
- Resolves Issue #2061


petyosi:
Hi, 

thank you, I pushed the same fix before checking what you have submitted.

PR #2073, opened 2017-11-22
fix type for ColDef.headerComponentFramework

miller-time:
When setting the `headerComponentFramework` of a colDef to an Angular component, there is a type error because it expects the component class to have a `new` function. 

When comparing the types of `headerComponentFramework` and `cellRendererFramework` (which I'm currently using without getting this type error), I noticed that `cellRendererFramework` has type of simply `any`.


rmc-software:
Hi, 

This fix has been applied to the latest branch (tracked by AG-1916) and will be merged into master when we do our next release.

Thanks

PR #2081, opened 2017-11-29
Add groupRemoveSingleChildren for columns

shirosaki:
Add groupRemoveSingleChildren option of colDef to remove groups
which only has one child for specified column.


rmc-software:
Hi,

We are going to hold off on accepting the PR as it's very edge case and there are workarounds available.

Thanks

Issue #2054, opened 2017-11-14
Typescript types for "sortModel" and "filterModel" are unhelpful.


UselessPickles:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
```

**Current behavior**
Typescript types for "sortModel" and "filterModel" throughout the code are unhelpful.

Most of the methods/interfaces that deal with "sortModel" and "filterModel" simply have the type as "any". This is unhelpful in Typescript.

Examples: 

in IEnterpriseGetRowsRequest, RowNodeCacheParams, IGetRowsParams:
```ts
    sortModel: any;
    filterModel: any;
```

in SortController:
```ts
    getSortModel(): { // WHY ISN'T THIS A WELL-DEFINED/REUSED INTERFACE?
        colId: string;
        sort: string;
    }[];
    setSortModel(sortModel: any): void;
```

in GridApi: 
```ts
    setSortModel(sortModel: any): void;
    getSortModel(): { // WHY ISN'T THIS A WELL-DEFINED/REUSED INTERFACE?
        colId: string;
        sort: string;
    }[];
    setFilterModel(model: any): void;
    getFilterModel(): any;
```
etc...

**Expected behavior**
There should be well-defined interfaces/types for sortModel and filterModel so that I get useful IDE assistance and compiler assistance to ensure I am using sortModels and filterModels correctly.

**Minimal reproduction of the problem with instructions**
Look at the code.

**What is the motivation / use case for changing the behavior?**
Improve type safety when developing in TypeScript.

**Please tell us about your environment:**
Irrelevant

* **ag-Grid version:** 
14.1.1

* **Browser:**
Irrelevant

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TypeScript 2.6

UselessPickles:
I created a pull request to fix the "sortModel" portion of this issue. The structure of a "filterModel" is less obvious to me right now, so I though it was best to tackle it separately later.

makinggoodsoftware:
Hi,

Thanks for the Pull Request, I have created the ticket AG-1231 that you can track from our pipeline

https://www.ag-grid.com/ag-grid-pipeline/

Thanks

UselessPickles:
@makinggoodsoftware I see that AG-1231 is currently in the "backlog" of "Complex Feature Requests"

[...]

PR #2092, opened 2017-12-04
Introduce ISortModel to clearly define the structure of a sortModel.

UselessPickles:
Fixes the "sortModel" portion of this issue: https://github.com/ag-grid/ag-grid/issues/2054

* Replaced static constants Column.SORT_ASC and Column.SORT_DESC with an enum named Column.SortDir (better organization).
* Introduced ISortModel to define the structure of a "sort model".
* Updated as many places as I could find to use ISortModel[] and Column.SortDir (or Column.NullableSortDir) instead of any and string.
* Explicitly included "| null" for types where a null value is allowed (as far as I can tell) so that these changes should be ready for compatibility with stricter compiler settings.

NOTE FOR REVIEWERS: Please pay special attention to my use of NullableSortDir vs SortDir. Did I use NullableSortDir where null should never be allowed? Did I use SortDir where null should be allowed?


src/ts/interfaces/iSortModel.ts
```diff
@@ -0,0 +1,9 @@
+import { Column } from "../entities/column";
+
+/** Sort model, defining the sort direction of a single column. */
+export interface ISortModel {
```
UselessPickles: This may be a bad name for this interface. A "sort model" is technically an array of these. Any suggestion for a better name for this interface? Maybe ISortDef?

UselessPickles:
Closing this PR because it clearly is never being reviewed by anyone or merged.

PR #2169, opened 2018-01-26
Set default auto group column position

shirosaki:
* Set the default position of the auto group column using option `autoGroupColumnIndex`
* Keep the auto group column position using getColumnState() and setColumnState()


rmc-software:
Hi,

Unfortunately we can't accept this PR as is because it doesn't work when 'groupMultiAutoColumn: true'.

If you drag reposition one of the multi group columns it doesn't preserve position upon saveColumnState.

Also now that we have added 'lockedPosition' we probably don't want to add autoGroupColumnIndex at this time.

We'll keep the FR open however in our backlog.

Thanks.

PR #2189, opened 2018-02-05
Fix full width cell renderer not removed

shirosaki:
Full width cell renderer of pinned rows is not removed after grid is destroyed.


rmc-software:
This issue has already been fixed and the code this fix is for has been removed from the codebase. Thanks

PR #2201, opened 2018-02-08
Update question link to Stack Overflow

datascape:
Old MyBB question forum is discontinued...
https://www.ag-grid.com/forum/


rmc-software:
This has been fixed in latest branch. When we do our next release it will appear in master. Thanks.

PR #2261, opened 2018-03-19
speed up scrolling

katasup:
Hello!
Removing of unused rows takes a half time of scrolling process.
`Array.prototype.forEach` is too much slow than `for` iterate. So i fixed it.


rmc-software:
I did some profiling and didn't observe any speed up.

Note that rowsToRemove is typically only has around 50 items, due to row virtualization, so this optimization is unlikely to make a difference over the native for loop.

Closing as a result.

Thanks.

PR #2329, opened 2018-04-19
Fix failCallback comment

arnas:


rmc-software:
Comment fixed in branch latests. Will be in master branch when we do our next release. Thanks

PR #2354, opened 2018-04-30
Fix onTouchUp bug in DragService

mikejune:
`onTouchUp` will call the `getFirstActiveTouch` with `touchEvent.targetTouches`, which is an empty `TouchList`, and will default to assigning `touch = this.touchLastTime`. This is problematic because using the `onRowDragEnd` prop in React, if trying to save the row position using the `TouchEvent` data, the data corresponds to the `touchLastTime` data instead of where the last touch event occurred.

This makes use of `changedTouches`, which does include the last touch event.


rmc-software:
Hi,

This change has already been applied to master. Thanks

Issue #1201, opened 2016-10-14
[touch] Cell doubleclick does not work on iphone/ipad


jogibear9988:
On IPhone/IPad a doubleclick zooms in and does invoke the doubleclick callback.

harishgadiya:
By default mobile Safari does not support the javascript doubleclick event. It's interpreted by Safari as a "zoom".

```
var agent = navigator.userAgent.toLowerCase();
if(agent.indexOf('iphone') >= 0 || agent.indexOf('ipad') >= 0){
       // implement double-tap
} else{
       // implement double click
}
```

And you can also add viewport meta tag in your html file inside the head tag. It will prevent user to zoom in or out the page
<meta name="viewport" content="width=device-width; initial-scale=1.0; maximum-scale=1.0; user-scalable=0;">

jogibear9988:
the meta tag does not work. and i want a cell doubleclick, so i think it needs to be implemented in aggrid?

ceolter:
thanks, i've added this to my list for next time i do a pass on touch devices. i see what you mean, you can't enter edit mode.

ceolter:
tracked by AG-146
https://www.ag-grid.com/ag-grid-pipeline/

Keki91:
Don't suppose there's an estimated release date for this feature?

R4DIC4L:
This issue doesn't appear in the pipeline anymore. Has it been dropped?

makinggoodsoftware:
Released in v15.0

GeorgeKnap:
Double tap doesn't work on iPhone in version 19. Why is this closed as resolved?

ratheeps:
Row double tap doesn't work on iPod in Safari, Please reopen this issue.

codemonkey047:
Row double tap was not working, so I switched to cell double tap. It had all of the event information I needed, but it would be great to fix the row double tap as well.

PR #2360, opened 2018-05-02
Add iPhone check to regex for user agent.

mikejune:
Check for iPhone user agent to deal with double-tap zoom issues https://github.com/ag-grid/ag-grid/issues/1201


ceolter:
thanks. i've updated this locally and checked in against our 'latest' branch (for current dev). will be in next release.

mikejune:
@ceolter Should this regex also check for android devices?

Issue #2373, opened 2018-05-10
groupSelectsFiltered bug in RowNode.selectChildNodes


jessicashin:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

Selecting a group after filtering with `[groupSelectsChildren]="true"` and `[groupSelectsFiltered]="true"` selects child rows that are filtered out.

**Expected behavior**

Selecting a group should select only filtered rows when `[groupSelectsFiltered]="true"`

**Minimal reproduction of the problem with instructions**

https://plnkr.co/edit/3Ll0gMwYpnMwfvmY0sDl?p=preview
Click "Filter Only Swimming" and then click the first checkbox for "United States". The selected row count is 899, which does not match the count of filtered rows for United States (145).

**What is the motivation / use case for changing the behavior?**

`groupSelectsFiltered` should only select filtered rows.

**Please tell us about your environment:**

* **ag-Grid version:** 17.1.1

makinggoodsoftware:
This was actually fixed in v18.1. Thanks.

PR #2374, opened 2018-05-10
Pass groupSelectsFiltered param when selecting child nodes

jessicashin:
Fixed a bug with groupSelectsFiltered where selecting a row with child nodes does not properly select only filtered nodes.
https://github.com/ag-grid/ag-grid/issues/2373


rmc-software:
This was actually fixed in v18.1. Thanks.

PR #2393, opened 2018-05-18
Fixed SetFloatingFilter to skip keyboard input on readonly input

artiz:
Issue is reproduced on official demo https://www.ag-grid.com/javascript-grid-filter-set/
To reproduce it click on some readonly input in grid column header (Athlete, Country) and enter some letters. Grid will be cleared and if you open minifilter then all check there will be unchecked.


rmc-software:
I assume you are testing this in IE as we were using readonly html attribute which isn't supported by IE.
This issue was fixed in v18.1 by using disabled.

Thanks

Issue #2435, opened 2018-06-14
TextFilter always filters out 0 when the backing field is a number


bursauxa:
Hello,

Here is some context: we have a scenario where we have some numeric data in a column, but we want users to be able to filter that data by typing a few digits, rather than typing full numbers as is required by NumberFilter. To that end, we use TextFilter, although the backing field is of type `number` and not `string`.

Things work perfectly for all numbers, except for 0. When users type "0" in the filter box, rows that contain the digit 0 are visible (:heavy_check_mark:), rows that do not contain the digit 0 are not visible (:heavy_check_mark:), but rows that are exactly equal to `0` are not visible (:x:).

Here is a plunker that reproduces the issue : https://plnkr.co/szT2CKfRARDbpTt48gR0. You can verify that filtering with 0 in the "string" column works as expected, but not in the "number" column.

Technically, this is related to #2266, although it is different since in our case we do not use any custom comparator. I will have a PR coming your way.

bursauxa:
PR: #2436

rmc-software:
Thanks for this PR. This has been fixed as part of AG-1897 and will be included in our next release. For details see: https://www.ag-grid.com/ag-grid-pipeline/

PR #2436, opened 2018-06-14
Fix for issue #2435

bursauxa:
Restrict automatic pass/faill of TextFilter to `null` and `undefined`, instead of all falsy values. Typically useful for number `0` to match string `"0"`.


rmc-software:
Thanks for this PR. This has been fixed as part of AG-1897 and will be included in our next release. For details see: https://www.ag-grid.com/ag-grid-pipeline/

bursauxa:
Thanks for the feedback @rmc-software :+1:

PR #2461, opened 2018-06-24
Exporting data which size is greater than 2mb on chrome throws an error

### LINKED ISSUE #2460: [Bug] Exporting data which size is greater than 2mb on chrome throws an error
Hi,
When trying to download a file which size is more than 2mb in chrome the download gives a network error.

This can be easily fixed by fixing the downloader.ts code to be as so:
```
import {Bean} from "./context/context";

@Bean('downloader')
export class Downloader {
    download (fileName:string, content:string, mimeType:string){
        // for Excel, we need \ufeff at the start
        // http://stackoverflow.com/questions/17879198/adding-utf-8-bom-to-string-blob
        let blobObject = new Blob(["\ufeff", content], {
            type: mimeType
        });
        // Internet Explorer
        if (window.navigator.msSaveOrOpenBlob) {
            window.navigator.msSaveOrOpenBlob(blobObject, fileName);
        } else {
            // Chrome
            const element = document.createElement('a');
            const blob = new Blob([content], {type: 'octet/stream'});
            const url = window.URL.createObjectURL(blob);
            element.setAttribute('href', url);
            element.setAttribute('download', fileName);
            element.style.display = 'none';
            document.body.appendChild(element);
            element.click();
            window.URL.revokeObjectURL(url);
            document.body.removeChild(element);
        }
    }
}
```

this one will also fix the potential memory leak in the current code.

vlio20:
resolves #2460

### DIFF
--- a/packages/ag-grid/src/ts/downloader.ts
+++ b/packages/ag-grid/src/ts/downloader.ts
@@ -1,8 +1,8 @@
 import {Bean} from "./context/context";
 
-@Bean('downloader')
+@Bean("downloader")
 export class Downloader {
-    download (fileName:string, content:string, mimeType:string){
+    download(fileName:string, content:string, mimeType:string) {
         // for Excel, we need \ufeff at the start
         // http://stackoverflow.com/questions/17879198/adding-utf-8-bom-to-string-blob
         let blobObject = new Blob(["\ufeff", content], {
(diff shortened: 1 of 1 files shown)


rmc-software:
This PR has been added to the latest branch and will be merged into master in the next release. Thanks.

Issue #2521, opened 2018-07-23
stopEditingWhenGridLosesFocus not working on angularjs


gsubiran:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

I have set stopEditingWhenGridLosesFocus = false 
but when click outside grid popup cell editor hides

**Expected behavior**

stopEditingWhenGridLosesFocus = false and no hide cell editor

**Minimal reproduction of the problem with instructions**

I'm using angularjs 1.5.8. sniffing in ag-grid.js I saw that "hidePopupOnMouseEvent" function is fired and doesn't take care about stopEditingWhenGridLosesFocus property set in gridOptions. I guess should be any kind of check here to prevent popup to hide.

When function addStopEditingWhenGridlosesFocus is called (place where take care about below property in function isStopEditingWhenGridLosesFocus()) the popup is already closed.

Make sense?

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **ag-Grid version:** 

It's happening in v17.1.1 and checked in latest v18.1.1 and still the same problem

* **Browser:**
 Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.99 Safari/537.36

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
ES6

RodSmithAG:
Hi,

Can you provide a reproducible scenario using one of our examples or your source code in a plunker and we will investigate?

Accepted reproducible scenarios are
- A description of the detailed steps to reproduce your behaviour in one of our examples in the docs.
- A plunker

If you decide to send us a plunkr, from any example in our website use the plunkr button in there to fork your own code by following the steps below:

- Select the framework that is appropriate to you from the drop-down
- Open it in plunker. (Use the button plunker in our example)
- Add your changes so that the behaviour is reproduced
- Save and Freeze the plunker(On the top left corner)
- Send us the link to the plunker(You can copy the URL from the browser)

If reporting a bug make sure to state.

Actual behaviour.
Expected behaviour. If possible back this up with our docs/examples if possible
***

Note that this issue is getting closed, but we will reopen it if we are provided with a reproduction scenario as specified above

[...]

PR #2525, opened 2018-07-24
Prevents stopEditing for popup cellRender mode, when grid loses focus…

gsubiran:
… and stopEditingWhenGridLosesFocus = false.

#2521 

I saw that "hidePopupOnMouseEvent" function is fired and doesn't take care about stopEditingWhenGridLosesFocus property set in gridOptions. I guess should be any kind of check here to prevent popup to hide.

When function addStopEditingWhenGridlosesFocus is called (place where take care about below property in function isStopEditingWhenGridLosesFocus()) the popup is already closed.

To fix that I did those changes so now when addStopEditingWhenGridlosesFocus = false is consistent with cell editing as popup too. The way to close edit is with ESC key as usual and I have added ENTER key too cause with ESC changes are theorically discarded.

Make sense?

### DIFF
--- a/packages/ag-grid/src/ts/widgets/popupService.ts
+++ b/packages/ag-grid/src/ts/widgets/popupService.ts
@@ -274,7 +274,8 @@ export class PopupService {
 
         let hidePopupOnKeyboardEvent = (event: KeyboardEvent) => {
             let key = event.which || event.keyCode;
-            if (key === Constants.KEY_ESCAPE) {
+            if (key === Constants.KEY_ESCAPE 
+                || key === Constants.KEY_ENTER) {
                 hidePopup(null);
             }
         };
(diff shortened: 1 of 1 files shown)


smile2014:
Which version will be resolved ？？？

smile2014:
Which version will be resolved ？？？ @gsubiran

smile2014:
@ceolter  Which version will be resolved ？？？

ceolter:
thanks for the PR, however we cannot accept.

the problem is with how our richSelect works - which is an alternative to the browser Select widget.

normally for HTML Select, when you open the select drop down, then when you click anywhere on the screen, the select closes. so by accepting this PR, then the select popup will stay open even when the user clicks somewhere else on the screen.

in hind sight what we should of done is 'clicking outside the grid stops editing' should be the default. or even clicking anywhere in the grid (not just another cell, eg clicking a floating filter or on tool panel) should also stop the editing. however this would be a bigger change, and a breaking change, so appetite to do this at this point is low.

PR #2534, opened 2018-07-27
Reverted changes that caused deprecation warning

TheDoctor0:
MutationObserver use in resizeObserver has been removed in release 18.1.2. which caused "Use of Mutation Events is deprecated. Use MutationObserver instead." for DOMSubtreeModified event listener. 
This PR reverts those changes.


ceolter:
we have completely changed direction with the resizeObserver - I believe the code this PR is referring to is no longer part of latest ag-Grid.

PR #2645, opened 2018-09-20
Remove debug logs

thebergamo:
I was testing the dnd feature and I face a suspicious log on my console, after investigations I found that it comes from here.

In my understanding it should not be shipped in the production version of ag-grid. Let me know if I can improve the PR anyway :D


gportela85:
Issue has been patched https://github.com/ag-grid/ag-grid/commit/f59546da9f026a35061599b3cc742c96321fdbaa#diff-4161de7f0f325784406834e4707b72bcL208

Issue #2649, opened 2018-09-21
Utils.isBrowserSafari Check causing namespace error


soorajshankar:
I'm submitting a issue which is happening inconsistently especially when i use it with amchart export plugin.
Bug Report 
**Uncaught (in promise) TypeError: Cannot read property 'toString' of undefined
    at eval (utils.js?211a:919)
    at Function.Utils.isBrowserSafari (utils.js?211a:920)
    at HeaderGroupComp.setupLabel (headerGroupComp.js?696e:95)
    at HeaderGroupComp.init (headerGroupComp.js?696e:43)
    at ComponentResolver.initialiseComponent (componentResolver.js?9a28:290)
    at ComponentResolver.createAgGridComponent (componentResolver.js?9a28:241)
    at ComponentRecipes.newHeaderGroupComponent (componentRecipes.js?10e5:52)
    at HeaderGroupWrapperComp.appendHeaderGroupComp (headerGroupWrapperComp.js?ffd5:108)
    at HeaderGroupWrapperComp.postConstruct (headerGroupWrapperComp.js?ffd5:58)
    at eval (context.js?450d:215)
(anonymous) @ utils.js?211a:919
Utils.isBrowserSafari @ utils.js?211a:920
HeaderGroupComp.setupLabel @ headerGroupComp.js?696e:95
HeaderGroupComp.init @ headerGroupComp.js?696e:43
ComponentResolver.initialiseComponent @ componentResolver.js?9a28:290
ComponentResolver.createAgGridComponent @ componentResolver.js?9a28:241
ComponentRecipes.newHeaderGroupComponent @ componentRecipes.js?10e5:52
HeaderGroupWrapperComp.appendHeaderGroupComp @ headerGroupWrapperComp.js?ffd5:108
HeaderGroupWrapperComp.postConstruct @ headerGroupWrapperComp.js?ffd5:58
(anonymous) @ context.js?450d:215
(anonymous) @ context.js?450d:215
Context.postConstruct @ context.js?450d:212
Context.wireBeans @ context.js?450d:75
Context.wireBean @ context.js?450d:69
HeaderRowComp.createHeaderComp @ headerRowComp.js?beeb:207
(anonymous) @ headerRowComp.js?beeb:178
HeaderRowComp.onVirtualColumnsChanged @ headerRowComp.js?beeb:154
HeaderRowComp.onDisplayedColumnsChanged @ headerRowComp.js?beeb:143
[...]

PR #2650, opened 2018-09-21
Utils.isBrowserSafari p undefined issue #2649

soorajshankar:
Since p sometimes return undefined and tries to convert it into string 

https://github.com/ag-grid/ag-grid/blob/master/dist/lib/utils.js#L1003

added condition before tries to do p.toString()


soorajshankar:
addng issue 
https://github.com/ag-grid/ag-grid/issues/2649

soorajshankar:
adding issue https://github.com/ag-grid/ag-grid/issues/2649

ceolter:
Thanks for raising. I've applied this to our 'latest' branch which will be in our next release.

PR #2766, opened 2018-11-12
Add vuePluginProps to whitelist specific vue plugins in ag-grid-vue

paulpdaniels:
In a previous commit a3d0c506836cc63c47e6e1d88596015b2112012c, support was added for a vuex store component. This PR adds the ability for the user of ag-grid to whitelist a set of plugins which will be passed to the cell components when they are created.

cc/ @mccxiv


packages/ag-grid-vue/src/vueComponentFactory.js
```diff
@@ -149,18 +152,20 @@ export class VueComponentFactory {
...
             },
-            parent,
-            router: parent.$router,
-            store: parent.$store
+            parent
         };
 
+        for (let key of whitelist) {
+            details[key] = parent['$' + key]
+        }
```
paulpdaniels: Main change here is that instead of hard coding the components we use a whitelist provided by the user.

packages/ag-grid-community/src/ts/propertyKeys.ts
```diff
@@ -15,7 +15,8 @@ export class PropertyKeys {
 
     public static ARRAY_PROPERTIES = [
         'slaveGrids', 'alignedGrids', 'rowData',
-        'columnDefs', 'excelStyles', 'pinnedTopRowData', 'pinnedBottomRowData'
+        'columnDefs', 'excelStyles', 'pinnedTopRowData', 'pinnedBottomRowData',
+        'vuePluginProps'
```
paulpdaniels: I'm open to naming suggestions as well, I chose this because it seemed to be in line with other lib specific variables.

seanlandsman:
@paulpdaniels I'm implementing a mechanism that effectively does what you've provided, but in my testing I don't believe that this is necessary anymore now that the dynamic vue components have their parent set correctly.

Thanks for the PR - we appreciate it

PR #2839, opened 2018-12-16
fix(agGridReact): implement portals in the same tree to fix context support

### LINKED ISSUE #2838: reactNext / Portals implementation does not work with React 16.3 Context
This is the same as #2578, but it was closed; unfortunately the issue is still persisting in the latest ag-grid version. 

```
[] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[x] support request => see 'Requesting Community Support'
```

**Reproducible Scenario**
https://codesandbox.io/s/x74wn4mmow

**Current behavior**
- MakeCell returns `{}` from MakeContext.Consumer

**Expected behavior**
- MakeCell prints `{green: true}` from MakeContext.Consumer

* **ag-Grid version:** 19.1.4, ag-grid-react 19.1.2 
* **Language:** all

--- 

Investigation / Potential Fix
---

screen shot 2018-12-15 at 11 38 06 pm

I have created a PR with a possible fix in #2839 - but I don't know if there's missing test cases that I need to look out for - e.g. async as an example.

https://codesandbox.io/s/xo22qp6yl4 demonstrates that the fixes are working, and the expected behavior is showing
screen shot 2018-12-16 at 12 18 19 am

screen shot 2018-12-16 at 12 14 31 am
_This shows that it is possible to put all the portals in the same tree even though AgGrid has the full control of the DOM (due to ShouldComponentUpdate => false)_ 

React-Redux v6 and Legacy React Context
----

Since React-Redux now uses Context, passing down store as a prop is no longer allowed. With this fix, take out the `store={store}` prop and it should fix the issue.

See example here: https://plnkr.co/edit/IRuM34IDxZH6yUUHPOWO?p=preview

Legacy context also no longer needs `store={store}` prop as well.

hrgui:
Closes #2838 

This PR fixes the usage of `reactNext` by implementing all the portals as children of `<AgGridReact />`, and by doing so, this will make Ag-Grid work with React 16.3 Context. 

Any component under `*framework` in react will still render in its the spot the user desires due to portals.

---
**Some tests of this PR that I am going through* 
- [x] destroy needs to tell parent to unmount and update - Fixed - added a batch updater mechanism
- [x] unmountComponentAtNode - since the parent component is now rendering the element, there is no need to do unmountComponentAtNode - the table will handle it

### DIFF
--- a/packages/ag-grid-react/src/agGridReact.ts
+++ b/packages/ag-grid-react/src/agGridReact.ts
@@ -25,6 +25,8 @@ export class AgGridReact extends React.Component<AgGridReactProps, {}> {
     gridOptions: AgGrid.GridOptions;
     api: AgGrid.GridApi;
     columnApi: AgGrid.ColumnApi;
+    portals = [];
+    hasPendingPortalUpdate = false;
 
     protected eGridDiv: HTMLElement;
 
@@ -38,7 +40,7 @@ export class AgGridReact extends React.Component<AgGridReactProps, {}> {
             ref: e => {
                 this.eGridDiv = e;
             }
-        });
+        }, this.portals);
     }
 
     createStyleForDiv() {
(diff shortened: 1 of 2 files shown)


seanlandsman:
@hrgui thanks very much for both this PR and the issue you raised - it'll definitely help

This is great timing as I'm literally looking at this side of things at the moment!

I'm close both this PR and the issue but will be incorporating the ideas manually as I have other changes in play at the moment.

Thanks again, this will def accelerate the final solution implemented. Once done feedback would be great too once done.

seanlandsman:
In the end I've taken what you've provided as is - it looks like it does the job well. Good work and thanks again!

hrgui:
Awesome, thanks!

PR #2864, opened 2019-01-10
correctly type getValueCallback field to accept Columns

tjallingt:
`ColumnController.getPrimaryColumn` takes `string | Column` as its key, so `ValueService.getValueCallback` can take the same as field.


gportela85:
fixed in `latest`

PR #2888, opened 2019-01-17
http to https

orenyakobi:
Causing a mix-content error when using Mcaffee plugin


gportela85:
@orenyakobi Had to revert this commit and change `https` back to `http`. It caused the SVG to stop working. For more info see: https://stackoverflow.com/questions/39573020/requested-https-www-w3-org-2000-svg-found-http-www-w3-org-2000-svg

PR #2892, opened 2019-01-18
fix(cell-stop-editing): add try-catch to prevent node.removeChild error from breaking the grid in Chrome

nnguyenleap:
Context: `stopEditingWhenGridLosesFocus: true`, more than 1 column with custom cellRenderer and keyboard navigation cause `Failed to execute 'removeChild' on 'Node'` error.

Environment: Angular 7.1.4 | Aggrid 20.0.0 | Chrome (IE is fine)

image

### DIFF
--- a/packages/ag-grid-community/src/ts/utils.ts
+++ b/packages/ag-grid-community/src/ts/utils.ts
@@ -611,7 +611,11 @@ export class Utils {
     static removeAllChildren(node: HTMLElement) {
         if (node) {
             while (node.hasChildNodes() && node.lastChild) {
-                node.removeChild(node.lastChild);
+                try {
+                    node.removeChild(node.lastChild);
+                } catch (e) {
+                    return;
+                }
             }
         }
     }


kyg9823:
I have the same issue. This PR will be worked for me. 
Please check this PR or resolve this issue.

Environment : Angular 6.1.10 | Aggrid 19.1.2 | Chrome

zorrme:
When can we merge this ?

gportela85:
@nnguyenleap I created this test case and couldn't see the problem https://plnkr.co/edit/SvUu01N3JnRjznOBVp2U?p=preview
Added more than 1 custom renderer +  `stopEditingWhenGridLosesFocus` + angular + agGrid v20 + keyboard navigation.

Can you give some more details or provide an example that reproduces the issue you are trying to patch here?

DanielYKPan:
@gportela85 
This issue happens when you are using custom editor components.
If you use only one custom editor, it works fine. But if you are using more one custom editors in a row, it would generate this issue.

I also tried to build a small project to reproduce this issue, but unfortunately nothing happened. This issue seems to be relative to javascript time frame. If your project is medium or large, this issue happens.

nnguyenleap:
@gportela85, I got the same problem with @DanielYKPan

The fact that IE doesn't produce error indicates that it's not a problem in our code. Besides, I believe try-catch DOM manipulation is a good thing to have, which causes no harm to the logic of aggrid

gportela85:
@nnguyenleap I agree that adding a try/catch is harmless but not the correct thing to do. Understanding what is causing the issue is essential.
Also, the problem may not affect IE because it fires `blur` / `focus` asynchronously which is the source of a lot of problems but apparently helps here.

DanielYKPan:
@gportela85 
I think the post explains quite clear what causes this issue in Chrome. https://stackoverflow.com/questions/21926083/failed-to-execute-removechild-on-node/22934552

nnguyenleap:
Fixed in 20.1.0. Closing pull request.

PR #2894, opened 2019-01-18
Fix object declarations in JS Getting Started guide

RobertHerhold:
There are several instances of things like
```js
var columnDefs = [
    {headerName: "Make", field: "make", sortable=true },
];
```
where it should be `sortable: true`


gportela85:
fixed on `latest` by https://github.com/ag-grid/ag-grid/commit/62ec35c9597271d7014d7354b6a844e94a233368

PR #2906, opened 2019-01-23
Added `singleClickEdit` to `ColDefUtil.BOOLEAN_PROPERTIES`

wenkepaul:
Added `singleClickEdit` to `ColDefUtil.BOOLEAN_PROPERTIES` in order to fix `invalid colDef property 'singleClickEdit' did you mean any of these`


gportela85:
fixed in `latest`

PR #2908, opened 2019-01-23
Add the api param to NavigateToNextCellParams

dmytro-krekota:
Hi ag-grid team, I use your table in my project and params of the method `navigateToNextCell` are not convenient for me. I need to access the grid api somehow, so I need to cash it somewhere.

In the pull request I've added the grid api to `NavigateToNextCellParams` and after it using the method `navigateToNextCell` became easier. Also, your grid has a similar approach in other places, like `GetMainMenuItemsParams`, `ProcessRowParams`.


ralfisalhon:
This would've been pretty helpful to have right now

StephenCooper:
As an FYI, this has been added and should be included in the next release which should be 26.1

Issue #2861, opened 2019-01-07
StatusPanelDef type is not exported


tjallingt:
**I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

**Current behavior**

The following code:
```js
/**
 * @type {import("ag-grid-community").StatusPanelDef}
 */
const statusBarDef = {
  statusPanels: [
    { statusPanel: 'agTotalRowCountComponent', align: 'left' },
    { statusPanel: 'agFilteredRowCountComponent', align: 'left' },
  ],
};
```

causes the typescript error:
```
Namespace '"c:/Users/tjalling/Dev/projects/project/node_modules/ag-grid-community/main"' has no exported member 'StatusBarDef'. ts(2694)
```
because it is not exported from the `./dist/lib/interfaces/iStatusPanel` file where it is declared.

see https://github.com/ag-grid/ag-grid/blob/master/main.d.ts#L62 and https://github.com/ag-grid/ag-grid/blob/master/dist/lib/interfaces/iStatusPanel.d.ts

**Expected behavior**

That I can import and use the StatusBarDef type to type my statusBarDef variable.

**Please tell us about your environment:**

Windows 10, VS Code, Yarn, webpack-dev-server, Typescript version 3.2.2

* **ag-Grid version:**

"ag-grid-community": "^19.0.0"
"ag-grid-enterprise": "^19.1.1"

* **Browser:**

N/A

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
ES6/7
using JsDoc to type and Typescript version 3.2.2 to typecheck

tjallingt:
Seems that the change to make this work would be:
```diff
- export { IStatusPanel, IStatusPanelComp, IStatusPanelParams } from "./dist/lib/interfaces/iStatusPanel";
+ export { StatusPanelDef, IStatusPanel, IStatusPanelComp, IStatusPanelParams } from "./dist/lib/interfaces/iStatusPanel";
```

makinggoodsoftware:
Hi,

ag-grid staff does not provide support through github. If you are a customer please contact us through our official support channel. Note that this issue has been flagged as managed-by-the-community, we also recommend you to check stackoverflow, note that if there is no activity on this issue after a while we will close it.

Thanks

tjallingt:
I am a customer but I don't need **support**...
I just want to know whether there is a good reason to not export the type and/or whether you will accept a PR to export the `StatusPanelDef` type.

PR #2912, opened 2019-01-25
added export for StatusPanelDef

tjallingt:
fixes: https://github.com/ag-grid/ag-grid/issues/2861


gportela85:
fixed in `latest`

PR #2935, opened 2019-02-06
Custom Tooltips

gportela85:


packages/ag-grid-community/src/ts/widgets/tooltipManager.ts
```diff
@@ -0,0 +1,135 @@
...
+        this.showTimer = 0;
+        this.hideTimer = window.setTimeout(this.hideTooltip.bind(this), 10000);
+
+        if (!registeredComp.comp) {
+            const colDef = cmp.getComponentHolder();
+            const cell = (cmp as CellComp);
+            let rowIndex;
+            let column;
+
+            if (cell.getGridCell) {
```
seanlandsman: Should this be within the if(cell) block?

PR #2943, opened 2019-02-07
Delete orphaned nodes from allNodesMap in NodeManager

stefanruijsenaars:
When using `deltaRowDataMode=true`, and streaming row updates of 1000+ rows with unique IDs every ~250ms, performance gets gradually worse (and the grid eventually becomes unusable) as `ClientSideNodeManager.getCopyOfNodesMap()` performs a clone of the nodes map on every update, taking several 100ms to do so whenever the map has 100k+ orphaned entries all set to `undefined`:
```typescript
    public getCopyOfNodesMap(): {[id:string]: RowNode} {
        const result: {[id:string]: RowNode} = _.cloneObject(this.allNodesMap);
        return result;
    }
```

I don't see anywhere in the codebase that relies on the keys being set on the object (i.e. `for`/`in`/`hasOwnProperty`/`Object.keys`), so an immediate delete should do the same as setting the item to `undefined`, while fixing the performance issue with the cloning.

Example benchmark: https://repl.it/@stefanruijsenaars/MediumTurbulentFirm

### DIFF
--- a/packages/ag-grid-community/src/ts/rowModels/clientSide/clientSideNodeManager.ts
+++ b/packages/ag-grid-community/src/ts/rowModels/clientSide/clientSideNodeManager.ts
@@ -233,7 +233,7 @@ export class ClientSideNodeManager {
             rowNode.clearRowTop();
 
             _.removeFromArray(this.rootNode.allLeafChildren, rowNode);
-            this.allNodesMap[rowNode.id] = undefined;
+            delete this.allNodesMap[rowNode.id];
         }
     }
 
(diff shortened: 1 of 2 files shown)


ceolter:
I've accepted this change, however I updated locally and checked in rather than merging the PR.

I didn't accept the change to file immutableService - this needs to allow undefined to be set as we set the items to undefined later down the list. this will not degrade over time as this particular map is a copy that only lasts the duration of the transaction. deletes are slower to execute than setting something to undefined hence avoiding deletes where we can.

PR #2971, opened 2019-02-23
fix ag-grid-react main.d.ts

BenGladman:
After upgrading to the latest ag-grid I get the following type errors (using ts 3.3.1): 
```
node_modules/ag-grid-react/main.d.ts:13:5 - error TS2666: Exports and export assignments are not permitted in module augmentations.
13     export {ICellEditorReactComp} from './lib/interfaces';
       ~~~~~~
node_modules/ag-grid-react/main.d.ts:13:40 - error TS2307: Cannot find module './lib/interfaces'.
13     export {ICellEditorReactComp} from './lib/interfaces';
```
(error is repeated for each of the exports from lib/interfaces)

Removing the `declare module` fixes the issue. I don't think the declare is needed, there is nothing similar in any of the other packages.

### DIFF
--- a/packages/ag-grid-react/main.d.ts
+++ b/packages/ag-grid-react/main.d.ts
@@ -4,24 +4,19 @@ export * from './lib/agGridColumn';
 import {AgGridReactProps} from "./lib/agGridReact";
 import {AgGridColumnProps, AgGridColumnGroupProps} from "./lib/agGridColumn";
 import {Component} from "react";
-declare module "ag-grid-react" {
-    export class AgGridReact extends Component<AgGridReactProps, {}> {
-    }
-    export class AgGridColumn extends Component<AgGridColumnProps | AgGridColumnGroupProps, {}> {
-    }
 
-    export {ICellEditorReactComp} from './lib/interfaces';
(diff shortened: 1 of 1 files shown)


mikejune:
The suggestion made by @BenGladman seems to do the trick for us (we're using TypeScript 3.0.1). Hope it gets resolved soon!

seanlandsman:
@BenGladman thanks for this - we already have a fix for this (apologies!) and will be pushing it out soon

Can you please share your tsconfig and sample usage of ag-grid-react?  The example I use (and reference) doesn't catch this and I'd like to add yours as a test case to prevent this sort of thing going forward

BenGladman:
Thanks, that's good news!

I've created a minimal reproduction case in this repo https://github.com/BenGladman/ag-grid-typing-issue - just `npm run typecheck` to see the error.

seanlandsman:
Hi @BenGladman - we've released a patch to fix this

The longer term fix will take the form of alone the lines of what you have. Apologies for the hassle and thanks for the repo. Weirdly react-create-app didnt catch this, but I've added test cases to prevetn this sort of error happening going forward

PR #2988, opened 2019-03-01
Fix IE11 height issue

kgb-financial-com:
We noticed that in some cases, IE11 will not heed the flexbox rule that flexbox items do not shrink below
their minimum content size, and will use flex-basis instead. With `flex: 1;` flex-basis is set to 0%, which then leads to grids with a height of zero. Setting explicitly `flex: 1 1 auto;` as done previously for another element in https://github.com/ag-grid/ag-grid/commit/6bd222946c0d26a7a42a40f38bf4dff94bb3b30d works around this.

In our current project, we are successfully working around this issue with the following entry in our css:

```
.ag-root-wrapper-body.ag-layout-normal {
  flex: 1 1 auto;
}
```

We noticed that `flex: 1;` is used in 8 more files of ag-grid. We are not sure whether `1 1 auto;` is a good choice for them all, but maybe a short review would be a good idea.

### DIFF
--- a/packages/ag-grid-community/src/styles/ag-grid.scss
+++ b/packages/ag-grid-community/src/styles/ag-grid.scss
@@ -138,7 +138,7 @@ $icons: aggregation arrows asc checkbox-checked-readonly checkbox-checked checkb
     flex-direction: row;
 
     &.ag-layout-normal {
-        flex: 1;
+        flex: 1 1 auto;
         overflow: hidden;
     }
 }


kgb-financial-com:
Could somebody please look into this before the robot sets it to WontFix due to timeout? It's a trivial fix; would be a pity if this bug remained in the code.

mlembke1:
I can confirm that @kgb-financial-com is correct here and his fix is working great for us.

We are using:
**"ag-grid-community": "^20.1.0",
 "ag-grid-react": "^20.1.1",**

Interestingly enough, a colleague of mine is implementing the same community version of Ag-grid using Angular and has had no such issue in ie11. So, this may have something to do with _ag-grid-react_...?

THANK YOU @kgb-financial-com for sharing your fix!

gportela85:
This was patched in the latest branch and will be available in the next release.
Thanks for reporting it.

kgb-financial-com:
Thanks for the report that this has been seen elsewhere, and thanks for incorporating this fix into ag-grid code. For future reference, this has happened in commit https://github.com/ag-grid/ag-grid/commit/c922387e27bc6d882ddd08394805d8e808fdd359 .

Issue #3053, opened 2019-04-03
refData type definition is narrower than its intended us.


LMFinney:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[] support request => see 'Requesting Community Support'
```

**Current behavior**
refData on ColDef is defined like this:
`refData?: {
    [key: string]: string;
};`

However, elements of refData can be anything, not just strings. The expanded use is documented in the documentation: https://www.ag-grid.com/javascript-grid-reference-data/#example-reference-data-with-ref-data-prop

I can work around it by asserting as any, but the types should be loosened.

**Expected behavior**
refData should be defined as 
`refData?: {
    [key: string]: any;
};` or `refData?: Record<string, any>;`

**Please tell us about your environment:**
Windows, vsCode, Angular. Doesn't really matter.

**ag-Grid version:** X.X.X
We're on 19.1.2, but it's also in the [current version](https://github.com/ag-grid/ag-grid/blob/master/dist/lib/entities/colDef.d.ts).

**Browser:**
Chrome, but irrelevant

**Language:** TypeScript 3.1.6

makinggoodsoftware:
Hi,

Thank you for raising this issue for us.

As I am sure you are aware, we offer two products:

ag-Grid - this is entirely free to use in line with our MIT license. (issues reported via GitHub)
ag-Grid Enterprise - our commercial version which funds the development of both versions. (issues reported via zendesk)

If you are a customer, you should be using ag-Grid Enterprise in which case you should have access to our customer support system (via zendesk), and that is the only channel where we accept feature requests.

If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

IT-CASADO:
I need this too

We are also an enterprise customer, but why does it matter here!? This is a basic bug that should be solved in both products.

PR #3058, opened 2019-04-04
Loosen type for refData

LMFinney:
As mentioned in https://github.com/ag-grid/ag-grid/issues/3053, properties of refData should not be restricted to strings. ag-grid's documentation shows examples of using arrays and other objects instead of strings (https://www.ag-grid.com/javascript-grid-reference-data/#example-reference-data-with-ref-data-prop).

Loosen the typing to match actual behavior.


IT-CASADO:
Why is this labeled fix wontfix!? I have the same issue with that...

PR #3064, opened 2019-04-06
Add the possibility to override odd-row-background-color with Material Theme

CyrilSiman:
When we follow the sample on this page : https://www.ag-grid.com/react-getting-started/?utm_source=ag-grid-readme&utm_medium=repository&utm_campaign=github#customize-the-theme-look

We can't override **odd-row-background-color** with Material Theme.

odd-row-background-color isn't set in $params map variable used to init ag-grid-theme 
`@include ag-grid-theme($params)`


CyrilSiman:
Up to prevent stale state.

PR #3109, opened 2019-04-30
replace duplicated setting of datasource to serverSideDatasource

nikdatsko:
I need to set the serverSideDatasource after the initialization of the grid.
I could see that it does not get updated in onChanges. But datasource gets updated twice.
Please, check if this is a bug.

Thanks!
Nik


packages/ag-grid-community/src/ts/components/componentUtil.ts
```diff
@@ -164,8 +164,8 @@ export class ComponentUtil {
             api.setColumnDefs(changes.columnDefs.currentValue, "gridOptionsChanged");
         }
 
-        if (changes.datasource) {
```
nikdatsko: `datasource` is already being set once on line 203.
`serverSideDatasource` is never updated

PR #3115, opened 2019-05-03
cancel bubble on blob download

kaleguy:
Not canceling the bubble causes conflict with Google Tag Manager on some implementations.


gportela85:
fixed in latest

PR #3116, opened 2019-05-03
domLayout autoHeight viewport fix 

Redjaw:
I was having the following issue on my application that was returning a "100%" height instead of "auto".
image

Changing this property i managed to make it work in my project without changing the other viewport type behaviours.

Hope it helps.


gportela85:
Normally, when you want a PR to be merged or looked at, provide an example of the issue you are trying to fix.

Issue #1614, opened 2017-04-25
autoSizeColumns + maxWidth


kevinfairclough:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[ x] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/ceolter/ag-grid/blob/master/CONTRIBUTING.md#question
```

**Current behavior**

maxWidth doesn't allow the user to manually resize to a higher width, when used in conjunction with autoSizeColumns ideally we would still allow user make columns wider.

**Expected behavior**

A setting (or combination of) that allows autoSizeColumns to have a maxWidth, maxAutoSizeWidth?

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **ag-Grid version:** X.X.X

* **Browser:**

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

makinggoodsoftware:
Hi, 

Thank you for raising this issue for us, at this moment we are reviewing how we process Feature Requests/questions that come through from GitHub.

As I am sure you are aware, we offer two products:

ag-Grid - this is entirely free to use in line with our MIT license
ag-Grid Enterprise - our commercial version which funds the development of both versions.

We currently have a very full pipeline of work which can be viewed here: https://www.ag-grid.com/ag-grid-pipeline/

This will be our priority for the coming months and as a result, we don’t have capacity to take feature requests/questions from GitHub, but you might find that yours has already been raised in our pipeline.

We will do a periodic review of GitHub issues for any bugs raised and these can also be tracked on our pipeline page.

Thanks for your understanding.
ag-Grid Team.

kai-softr:
This seems to be possible since v34.

PaulSusset:
> This seems to be possible since v34.

Can you share how?

kai-softr:
> Can you share how?

Sure, I've successfully implemented this and got rid of the flicker that was caused with the previous workaround solution. I am just using the new `autoSizeStrategy` options and define my `defaultMaxWidth` and `columnLimits` with specific `minWidth`'s and `maxWidth`'s per column. Sometimes these values update dynamically then update the `autoSizeStrategy` grid option and also call `api.autoSizeColumns(...)` and pass the same parameters `defaultMaxWidth, columnLimits`.

PR #3138, opened 2019-05-17
Adding suppressAutoSizeAll property to ColDef

bamblack:
Adding in property to allow columns to be configured to not be auto-sized during `gridApi.columnApi.autoSizeAllColumns()`

### DIFF
--- a/dist/ag-grid-community.js
+++ b/dist/ag-grid-community.js
@@ -7243,7 +7243,11 @@ var ColumnController = /** @class */ (function () {
     ColumnController.prototype.autoSizeAllColumns = function (source) {
         if (source === void 0) { source = "api"; }
         var allDisplayedColumns = this.getAllDisplayedColumns();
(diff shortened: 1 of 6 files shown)


tagliala:
Hi,

I was interested in this feature because I wanted to resize all columns except one that I know can be wide. I've tried with `maxWidth`, but it doesn't allow users to manually resize the column in case of need (ag-grid/ag-grid#1614), so I ended up with this workaround:

```js
const onFirstDataRendered = function (params) {
  const autoResizableColumns = params.columnApi.columnModel.columnDefs
    .filter((column) => !column.hide && !column.suppressAutoSizeAll)
    .map((column) => column.field)

  params.columnApi.autoSizeColumns(autoResizableColumns, false)
}
```

LaysDragon:
yeah,I have the same situation need to resize all column except some column.

PR #3198, opened 2019-06-11
Update docs path to be correct

mwood23:


seanlandsman:
This has been incorporated - thanks for raising it

PR #3432, opened 2019-10-08
AG-3327 - move ChartBuilder into charts module

alowdon:
First step of moving the ChartBuilder into the charts module along with copies of relevant interfaces. Once this is done we can start to evolve the charts API and the grid options API separately, with the proxies translating into the options interfaces required when calling the ChartBuilder.


yay:
Looks good!

PR #3464, opened 2019-10-24
AG-3390 - improved function handling when generating React examples

alowdon:


alowdon:
@seanlandsman Have you had chance to look at this?

PR #3540, opened 2019-11-26
Use rowDataChangeDetectionStrategy / deltaRowDataMode for pinned rows

ajroe:
Currently if you set the `rowDataChangeDetectionStrategy` or `deltaRowDataMode` props they are only used for `rowData`

I believe this should also be applied to `pinnedTopRowData` and `pinnedBottomRowData`.

The issue I am facing is that I have a pinned top row, where the data in each cell is a complex object. As I cannot set the change detection strategy for this row, the default change detection strategy `DeepValueCheck` is applied, which is extremely slow (4-5 seconds to render). I would like to be able to override this to use an `IdentityCheck` instead.

### DIFF
--- a/packages/ag-grid-react/src/agGridReact.ts
+++ b/packages/ag-grid-react/src/agGridReact.ts
@@ -142,7 +142,7 @@ export class AgGridReact extends Component<AgGridReactProps, {}> {
     }
 
     private getStrategyTypeForProp(propKey: string) {
-        if (propKey === 'rowData') {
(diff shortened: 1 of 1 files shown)


travisje:
This is a bit of a blocker for us too as we have `pinnedTopRowData` that isn't picking up the `rowDataChangeDetectionStrategy` of `IdentityCheck`. Would be good to note this limitation in the documentation too.

ColinRTaylor:
@ajroe Since this was not merged,  did you ever find a way around this ? Having this problem as well

Issue #3160, opened 2019-05-29
row autoHeight doesn't work when using a Cell Renderer


areisle:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[] support request => see 'Requesting Community Support'
```

**Providing a Reproducible Scenario**
if you comment out  `cellRenderer` on the column definition, autoHeight works as excepted
https://embed.plnkr.co/d0dELGfRax8UPEJ8ujTu/

**Current behavior**
autoHeight seems to be ignored when the cell renderer is set

**Expected behavior**
autoHeight to be applied

**Please tell us about your environment:**

* Operating system: CentoOS Linux 7

* **ag-Grid version:** 20.2.0

* **Browser:** Chrome 72

* **Language:** all

CorreyL:
From @areisle's plunker:

https://embed.plnkr.co/d0dELGfRax8UPEJ8ujTu/

## When `cellRenderer` is defined in `columnDefs[0]`:

![image](https://user-images.githubusercontent.com/16601729/58583939-f8ae4d80-8209-11e9-8f01-d7f5d7b1bd65.png)

## When `cellRenderer` is not defined in `columnDefs[0]` (i.e. commented out, per the suggestion):
![image](https://user-images.githubusercontent.com/16601729/58584028-30b59080-820a-11e9-8760-6e9018876fb7.png)

Notice how in the second screenshot, the remainder of the text in the first row is no longer obscured, as it is in the first screenshot.

thielmal:
I'm having the same problem. The autoHeight calculation seems to never use the cellRenderer when creating the dummy elements to measure height.

heianqishizhisuiyuan:
you call use api.resetRowHeights() to reset the table of height

areisle:
@heianqishizhisuiyuan that only works if I call it every time I update the any of the data in the table which is not ideal. I've made a fork of the plnkr illustrating this https://next.plnkr.co/plunk/jqJ8pzU1rMrRl7vh

makinggoodsoftware:
Hi,

Thank you for your email. We are aware of this issue and have a feature request in our backlog:

You can follow the status of feature requests, bugs and releases using our pipeline: 
https://www.ag-grid.com/ag-grid-pipeline/

[...]

PR #3701, opened 2020-02-28
Fix AutoHeight for React Components (cellRendererFramework)

csit-intern-james:
This PR seems to be related to then new comments in the issue: #3160 

Minimum code to replicate bug (can't get `plunkr` to work) is in [this repo](https://github.com/csit-intern-james/ag-grid-react-autoheight-bug).

Bug seen on Firefox/73.0.1 on Ubuntu. The code changes in this PR attempts to fix the bug, although it is admittedly quite hacky; perhaps someone here has a better way to solve this :smile: 

From my understanding, the main cause of this issue is the race condition between React portals rendering the components onto the grid and `core/ts/rendering/autoHeightCalculator.ts` inferring the height from a dummy component.

I have tried a few ways of solving this:
1. Adding an interface to allow `IComponent`s to specify the height directly;
2. Invoke component rendering when laying out the components in the dummy container in `core/ts/rendering/autoHeightCalculator.ts`;
3. Rendering the component during `init` in `reactComponent` and removing the duplicated HTML after a while :sweat_smile:

Hopefully this will be enough information for somebody more experienced to find a better solution :smile:

### DIFF
--- a/community-modules/react/src/reactComponent.ts
+++ b/community-modules/react/src/reactComponent.ts
@@ -44,7 +44,7 @@ export class ReactComponent extends BaseReactComponent {
     public init(params: any): Promise<void> {
         this.eParentElement = this.createParentElement(params);
 
-        // we populate the innerHTML first, so that autoRowHeight can be calulated first
+        // we populate the innerHTML first, so that autoRowHeight can be calculated first
         if (params.colDef.autoHeight) {
             const reactComponent = React.createElement(this.reactComponent, params);
             this.eParentElement.innerHTML = renderToStaticMarkup(reactComponent);
@@ -85,7 +85,7 @@ export class ReactComponent extends BaseReactComponent {
         this.portal = portal;
         if (!params.colDef.autoHeight) {
             this.parentComponent.mountReactPortal(portal!, this, resolve);
-        } else { // remove the hack that allowed autoHeight to work (note, does not affect display, so it's an visually invisible change to the user)
+        } else { // remove the hack that allowed autoHeight to work (note: does not affect display, so it's a visually invisible change to the user)
             this.parentComponent.mountReactPortal(portal!, this, (value: any) => {
                 resolve(value);
                 setTimeout(() => {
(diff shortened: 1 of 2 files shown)


seanlandsman:
@csit-intern-james thanks very much for this - we have a fix for the autoheight issue but your PR has provided inspiration for a wider fix around this sort of issue

Thanks very much for the PR - we appreciate it!

Our next release is due out next week

DarkKnight1992:
@seanlandsman is the fix released ?

seanlandsman:
the release is due out next week

jwalkers:
> the release is due out next week

@seanlandsman Should autoheight now work with cellRenderers or is there a fix I need to implement? Currently not having any luck

joeferraro:
Still seeing this issue in `23.1.0`

taroy1:
Closed? I still see the issue, too, in 23.1.1. Any update on this very anticipated fix?

gportela85:
@taroy1 This issue has been fixed, if you are still seeing it, it might be a different problem. Can you provide a reproducible scenario?

taroy1:
Thanks for the prompt reply.

I am still seeing that autoHeight has no effect on my column definition. I
am using the server-side row model "infinite" to do paging of data in
combination with a custom cellRenderer on the column. The cellRenderer
generates a small chunk of HTML from a template. Is autoHeight on the
column definition supported with that combination of options?

Thanks very much!

gportela85:
@taroy1 Can we have that plunkr reproducing the issue? Also, since this seems to be a slightly different variation of the problem that was mentioned here initially, can you please file as a new issue?

taroy1:
I'll try to work on a plunkr example but that will be more involved because
it requires getting rows from a server. How shall I best go about filing a
new issue. We are customers using the enterprise edition.

Thanks,
Troy Osborne
4DV Analytics

gportela85:
@taroy1 If you are a customer using enterprise, please, file a support ticket through Zendesk.

seanlandsman:
@taroy1 If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

taroy1:
I have submitted issue #12535 which describes the bug with Plunker examples.

Thanks again,
Troy

PR #3750, opened 2020-03-27
AG-3312 - new histogram type in integrated charts

jimhigson:


charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+    frequency: number = 0;
+    domain: number[];
+
+    constructor ([domainMin, domainMax]: number[]) {
+        this.domain = [domainMin, domainMax];
+    };
+
+    addDatum (datum : any, yKey: string) {
+        this.data.push(datum);
+        this.total += yKey? datum[yKey] : 1;
```
yay: typo

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+        const [domainMin, domainMax] = this.domain;
+        return domainMax - domainMin;
+    }
+
+    get relativeHeight() : number {
+        return this.total / this.domainWidth;
+    };
+
+    getY( constantWidth: boolean ) {
+        return constantWidth? this.total: this.relativeHeight;
```
yay: typo?

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+
+    addDatum (datum : any, yKey: string) {
+        this.data.push(datum);
+        this.total += yKey? datum[yKey] : 1;
+        this.frequency ++;
+    };
+
+    get domainWidth() : number {
+        const [domainMin, domainMax] = this.domain;
+        return domainMax - domainMin;
```
yay: this is very minor, but I usually only create a variable if I access value more than once, I would just do `return this.domainMax - this.domainMin` like you did in the two methods below

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+    @reactive('change') formatter?: (bin: HistogramBin) => string;
+}
+
+export class HistogramBin {
+    data: any[] = [];
+    total: number = 0;
+    frequency: number = 0;
+    domain: number[];
+
+    constructor ([domainMin, domainMax]: number[]) {
```
yay: we can use `[number, number]` type if you only expect two numbers

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+        this.domain = [domainMin, domainMax];
+    };
+
+    addDatum (datum : any, yKey: string) {
+        this.data.push(datum);
+        this.total += yKey? datum[yKey] : 1;
+        this.frequency ++;
```
yay: typo

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+
+    highlightStyle: HighlightStyle = {
+        fill: 'yellow'
+    };
+
+    protected highlightedDatum?: HistogramSelectionDatum;
+
+    /*  during processData phase, used to unify different ways of the user specifying
+        the bins */
+    get derivedBins(): [number, number][] {
```
yay: It's good practice to use getters only when returning a precomputed value or when the computation is trivial.
If it isn't, it's better to use a traditional method like `getDerivedBins` or even `computeDerivedBins`.
This doesn't suggest that `derivedBins` are already there and you are just `getting` them.

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+
+        const {
+            xAxis : {scale : xScale},
+            yAxis : {scale : yScale},
+            fill, stroke, strokeWidth
+        } = this;
+
+        const selectionData: HistogramSelectionDatum[] = [];
+
+        const defaultLabelFormatter = (b:HistogramBin) => String(b.total);
```
yay: typo

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+
+        const selectionData = this.generateSelectionData();
+
+        this.updateRectSelection(selectionData);
+        this.updateTextSelection(selectionData);
+    }
+
+    private generateSelectionData(): HistogramSelectionDatum[] {
+
+        if( !this.seriesItemEnabled ) {
```
yay: Can you use `if (!this.seriesItemEnabled) {` please? I know this can be argued as a personal preference, but the existing code already follows a certain convention, which is quite standard too (built in VS Code and WebStorm code formatters use this style as default).

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+            yAxis : {scale : yScale},
+            fill, stroke, strokeWidth
+        } = this;
+
+        const selectionData: HistogramSelectionDatum[] = [];
+
+        const defaultLabelFormatter = (b:HistogramBin) => String(b.total);
+        const {
+            label: {
+                formatter : labelFormatter = defaultLabelFormatter,
```
yay: typo

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -0,0 +1,579 @@
...
+                fontSize: labelFontSize,
+                fontFamily: labelFontFamily,
+                fill: labelColor,
+                x: xMinPx + w/2,
+                y: yMaxPx + h/2
+            };
+
+            selectionData.push({
+                series: this,
+                seriesDatum: bin,  // required by SeriesNodeDatum, but might not make sense here
```
yay: Good point. I think it makes sense to make an exception for aggregated datums like you did here. The users will still want to access the `total` info for custom tooltips. This makes the `key` info useless though. Need to think about how to document this properly.

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
jimhigson: usually I would agree but I think array destructuring is a special case for clarity's sake. It wouldn't be `return this.domainMax - this.domainMin`, it'd be `return this.domain[1] - this.domain[1]`

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
jimhigson: +1

PR #3752, opened 2020-03-28
Provide access to GridApi and ColumnApi in getQuickFilterText params

rwdalpe:
Most column definition function params provide a reference to
GridApi and ColumnApi objects. Those references are easily accessible
when building the GetQuickFilterTextParams object, but were not included.
This adds them to the GetQuickFilterTextParams definition.

This also:

* Updates documentation for `getQuickFilterText` to reflect
  the new parameter properties.

* Formats documentation for the `getQuickFilterText` parameters to
  be more consistent with other column definition function
  documentation.

### DIFF
--- a/community-modules/core/src/ts/entities/colDef.ts
+++ b/community-modules/core/src/ts/entities/colDef.ts
@@ -378,15 +378,6 @@ export interface IsColumnFuncParams {
     columnApi: ColumnApi | null | undefined;
 }
 
-export interface GetQuickFilterTextParams {
-    value: any;
-    node: RowNode;
-    data: any;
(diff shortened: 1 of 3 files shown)


rwdalpe:
I opened this against the master branch but I'm not certain if that is the correct destination. It looks like other PRs have merged into `latest` or other branches. If this should target a different destination please let me know.

rwdalpe:
Are there changes I could make to this PR that will increase its chance of being accepted?

gportela85:
@rwdalpe the work to improve interfaces is due next major release. But thanks for pointing this one out.

PR #3806, opened 2020-04-27
fix: PreConstruct mistake

oppoffice:
fix: props.postConstructMethods to props.preConstructMethods


seanlandsman:
Thanks for the PR @oppoffice - I've applied this change to our development branch

Thanks for the PR - we appreciate it!

PR #3913, opened 2020-06-22
Mark lastRow in IServerSideGetRowsParams as optional

akoso:
This change matches what the docs describe here https://www.ag-grid.com/javascript-grid-server-side-model-datasource/#implementing-the-server-side-datasource and as well as my observations if I leave that parameter empty with ts-ignore

It's my first contribution so please let me know if there is anything else that I should update


akoso:
Actually playing with it more to make it work properly this parameter is not optional - would be great to update the docs to match this

PR #3928, opened 2020-06-28
Fix misspelling of community

Stargator:
Noticed when following the directions and copying commands that there was a misspelling. So instead of ignoring it, I would like to fix it.

### DIFF
--- a/community-modules/angular/README.md
+++ b/community-modules/angular/README.md
@@ -9,7 +9,7 @@ ag-Grid Angular Component
 ------
 
 ag-Grid is a fully-featured and highly customizable JavaScript data grid.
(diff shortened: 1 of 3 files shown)


Stargator:
I think this should be addressed. It's minor, but it affects anyone who's trying to duplicate three steps

WuglyakBolgoink:
up

Stargator:
Pinging @gportela85 and @seanlandsman just to see if there's anything I can do to make this easier

WuglyakBolgoink:
ping

gportela85:
@Stargator please, submit PR's against `latest` instead of `master`

Stargator:
Ok, will do

Issue #4022, opened 2020-08-14
Bug in mergeDeep when used by createFinalParams when "oldValue" is null and "newValue" is an object


dturcotte:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[] support request => see 'Requesting Community Support'
```

**Customers of ag-Grid**

If you are a customer you are entitled to use the ag-Grid's customer support system (powered by Zendesk). Please use that channel for guaranteed response from the ag-Grid team with regards bugs, feature requests and support.

**Requesting Community Support**

If you are not a customer of ag-Grid, ag-grid staff will label your issue as managed-by-the-community. This means that ag-Grid staff is not going to be actively looking into it and it will get closed if inactive for more than one month. The community is welcome to help with this question/support issue.

**Providing a Reproducible Scenario**
plunker: https://embed.plnkr.co/plunk/of8AD2ekbrHgpZ3C

**Current behavior**

Error `Uncaught TypeError: Cannot read property '<propery name>' of null` and the grid cells stop rendering after the cell that generates the error.

The problem is here in the `mergeDeep` code: https://github.com/ag-grid/ag-grid/blob/master/community-modules/core/src/ts/utils/object.ts#L73-L78

`typeof null` evaluates to `object`. If `oldValue` is `null` and `newValue` is an object, it attempts another `mergeDeep`.  Then, line 73 assumes `dest` is defined: https://github.com/ag-grid/ag-grid/blob/master/community-modules/core/src/ts/utils/object.ts#L73 and you see the error.

Note that if both `oldValue` and `newValue` are `null`, you're protected by line 75.

You get here from merging cell renderer params: https://github.com/ag-grid/ag-grid/blob/master/community-modules/core/src/ts/components/framework/userComponentFactory.ts#L485-L491

If a field such as `someValue.name` contains an empty string `''` then `oldValue` (and `value` as seen in the `cellRendererParams` function) comes out as `null`.  Then, if `userParams` (such as those from a `cellRendererParams` function) returns an object, you arrive in the above situation.

This seems roundabout but here's a very practical use case:

You have a grid and each row shows the user who created the record in a column `Created By`. Common in GraphQL APIs, your data might look like this:

```tsx
{
  // ...rowData
  createdBy: {
    name: 'John Smith',
    id: '1',
  }
}
```

[...]

PR #4023, opened 2020-08-14
Fix case in mergeDeep where oldValue is null and newValue is an object

dturcotte:
See this issue for a detailed description: https://github.com/ag-grid/ag-grid/issues/4022

TL;DR: `typeof null === 'object'` therefore if `oldValue` is `null` and `newValue` is an object then line 73 throws an error.


dturcotte:
https://github.com/ag-grid/ag-grid/issues/4037

PR #4040, opened 2020-08-24
Add managed tooltip bean if the cell can have a tooltip

2wheels4:


gportela85:
@2wheels4 this issue has been fixed in the latest branch

Issue #2911, opened 2019-01-24
[v20] Two different horizontal scrolls when I have custom horizontal scroll size


MaKCbIMKo:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[] support request => see 'Requesting Community Support'
```

**Current behavior**
I have custom scroll styles/sizes and I specify `scrollbarWidth` parameter as well. But in that case second horizontal scroll appears with default styles.
I have specified custom scroll style for `.ag-body-horizontal-scroll-viewport`, but second scroll appears at `.ag-center-cols-viewport`
(in case with v19 -> everything works just fine).

**Expected behavior**
There is no second horizontal scroll at all.

**Please tell us about your environment:**
Windows 10, VS Code, Angular, Angular-cli, Typescript version 3.2.2

* **ag-Grid version:** 20.0.0

* **Browser:** N/A

* **Language:** TypeScript 3.2.2

makinggoodsoftware:
Hi,

Can you provide a reproducible scenario using one of our examples or your source code in a plunker and we will investigate?

Accepted reproducible scenarios are
- A description of the detailed steps to reproduce your behaviour in one of our examples in the docs.
- A plunker

If you decide to send us a plunkr, from any example in our website use the plunkr button in there to fork your own code by following the steps below:

- Select the framework that is appropriate to you from the drop-down
- Open it in plunker. (Use the button plunker in our example)
- Add your changes so that the behaviour is reproduced
- Save and Freeze the plunker(On the top left corner)
- Send us the link to the plunker(You can copy the URL from the browser)

If reporting a bug make sure to state.

Actual behaviour.
Expected behaviour. If possible back this up with our docs/examples if possible
***

Note that this issue is getting closed, but you are welcomed to create a new issue if you can provide us with a reproducible scenario

If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

jiangyh1024:
@makinggoodsoftware I can reproduce this too. please see the plunker. scrolling to the end, there is a horizontal scroll bar even though 'suppressHorizontalScroll: true,' is set. I found that the 

```
.ag-center-cols-viewport {
    width: 100%;
    overflow-x: auto;
}
```
'overflow-x: auto;' caused this problem

[...]

PR #4078, opened 2020-09-19
Use more precise viewport width in sizeColumnsToFit.

pepperbc:
clientWidth rounded to the nearest integer, causing a horizontal scrollbar if that rounding increased the width. Some browsers handled this correctly, but was causing a horizontal scrollbar 50% of the time on Firefox. Might be related to #2911 , as the same temporary fix for that worked for me. (.ag-center-cols-viewport {overflow: hidden;})

There were several other areas in the code I found that similarly use clientWidth instead of getBoundingRect().width - it might be worth taking a look at those, in order to do things consistently throughout (and potentially fix other bugs). This should be the minimal change to fix the horizontal scrollbar after sizeColumnsToFit.

### DIFF
--- a/community-modules/core/src/ts/gridPanel/gridPanel.ts
+++ b/community-modules/core/src/ts/gridPanel/gridPanel.ts
@@ -1060,7 +1060,7 @@ export class GridPanel extends Component {
     // method will call itself if no available width. this covers if the grid
     // isn't visible, but is just about to be visible.
     public sizeColumnsToFit(nextTimeout?: number) {
-        const availableWidth = this.eBodyViewport.clientWidth;
+        const availableWidth = this.eBodyViewport.getBoundingClientRect().width;
 
         if (availableWidth > 0) {
             this.columnController.sizeColumnsToFit(availableWidth, "sizeColumnsToFit");


gportela85:
@pepperbc Looking at #2911 I can see the problem was there in v20, but it is no longer present in v24, see below.

v20 - https://plnkr.co/edit/2laIEPGS1RkkxJM6?preview
v24 - http://plnkr.co/edit/phCTwNYmFRq24IIX?preview

Can you provide a detailed scenario of the bug you are trying to fix? el.getBoundingClientRect().width has a different meaning than el.clientWidth and should not be replaced in all scenarios.

example
```
el = document.body.appendChild(document.createElement('div'));
el.style.position = 'absolute';
el.style.top = 0;
el.style.width = '100px';
el.style.height = '100px';
el.style.backgroundColor = 'green';
el.style.transform = 'scale(0.5)';
console.log(el.clientWidth); // 100
console.log(el.getBoundingClientRect().width); // 50
```

pepperbc:
[http://plnkr.co/edit/BV82eNoqvP4aE6yZ?preview] is doing it consistently for me - the fractional width is rounding up and causing a horizontal scrollbar.

Looked into the difference between the calls, and also noticed that they were different in the plnkr as well sometimes - thanks for the info. Unsure how to fix this otherwise, other than a fairly hacky + 1 to the width.  Any ideas?

pepperbc:
@gportela85 I realize you closed this before I sent the update, but could you take a look at that plinkr?

pepperbc:
@gportela85 Have you had a chance to look at this?

gportela85:
@pepperbc I've tried opening this on Chrome, Firefox for both Mac and Windows and did not see a horizontal scrollbar. I'm nut sure how to reproduce this.

PR #4080, opened 2020-09-20
Support custom debounce time and backspace in select editor search

2wheels4:
It is difficult to enter more than 2 or 3 characters before the search string is cleared, so it would be helpful to make this configurable


gportela85:
@2wheels4 can you submit PRs against `latest` instead of master?

PR #4126, opened 2020-10-08
Allow ranged deselection.

wjaspers:
A customer intended to select large numbers of rows, and also intended to allow
users to deselect parts of these ranges.

The existing functionality only supported adding rows to a range selection.
This feature allows the user to `click`, then `shift+click` in any
direction on the grid, allowing for ranged selection and deselection of nodes.

Because `RowNode.setSelected` may be triggered programatically (ie when updating
rows without a user interaction), we prevent the range selection from happening.

When the user toggles the selection of a row via checkbox, with the `shift` key
held down, we allow the range selection or deselection to occur.

### DIFF
--- a/community-modules/core/src/ts/entities/rowNode.ts
+++ b/community-modules/core/src/ts/entities/rowNode.ts
@@ -28,6 +28,8 @@ export interface SetSelectedParams {
     rangeSelect?: boolean;
     // used in group selection, if true, filtered out children will not be selected
     groupSelectsFiltered?: boolean;
+    // used when a range selection is permitted; ie. a user interaction sets the range
+    canBeRangeSelect?: boolean;
 }
 
 export interface RowNodeEvent extends AgEvent {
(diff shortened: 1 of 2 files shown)


makinggoodsoftware:
Hi,

gportela85:
We already have a patch for this being worked on at the moment, should be available in the next release.

kiril-matev:
Hello!

Thank you for the context.

There are two parts to your requirement - (1) a regression (clicking row checkboxes holding SHIFT doesn't select range in v24.1) and (2) feature request (support de-selection of row ranges inside the selected row range). I'll cover them below:

1) In v24.1 there's a regression with SHIFT+clicking checkboxes which is logged under:
**AG-4754 [Regression] Clicking row checkboxes while holding SHIFT key selects individual records instead of the range**

This regression is fixed in the upcoming ag-grid v25.

2) Support de-selection of row ranges inside the selected row range when holding SHIFT
We have added this requirement to our backlog and we are tracking it with the following reference and description:
**AG-4811 [Row Selection] Allow de-selecting row range from the selected rows range**

This feature request will be a part of our next release - v25 scheduled for early December.

Thanks again for bringing this up with us.

Issue #4154, opened 2020-10-20
Make RowNode a generic 


stevenhankin:
I'm submitting a ...
```
[] bug report => see 'Providing a Reproducible Scenario'
[X] feature request => do not use Github for feature requests, see 'Customers of ag-Grid'
[] support request => see 'Requesting Community Support'
```

Current behaviour.
Expected behaviour. If possible back this up with our docs/examples if possible

**Current behavior**
Currently RowNode.data type is "any"
This allows type bugs to occur, especially when creating renderers

**Expected behavior**
Should be able to declare the type of the data element
```ts
const deleteRowHandler = (node: RowNode<OrderItem>) => ...
```

* **ag-Grid version:** 22.0.0

* **Language:**  TypeScript 3.7

makinggoodsoftware:
Hi,

We have flagged this as a good_suggestion

This means that this is a candidate to be in a future release.

At this moment we don’t have a ticket number for this that you can use to track in our pipeline:https://www.ag-grid.com/ag-grid-pipeline/

But we should be updating this ticket with this information as soon as we process the good suggestions, as we do these in batches.

Thanks for your feedback!

kiril-matev:
Hello!

Thank you for this suggestion..

We agree this would be a valuable addition to the product and we have added a feature request for this.

We have added this requirement to our backlog and we are tracking it with the following reference and description:
**AG-4807 [API] Allow generics around type**

We have no timeline for this at the moment, but this can still be picked up by our team when they're making modifications to this functional area of ag-Grid. Please use the functionality currently provided by ag-Grid until this is delivered at a later point.

You can check whether this item is selected for implementation by taking a look at the NEXT RELEASE tab of the product pipeline:
https://www.ag-grid.com/ag-grid-pipeline/

I would recommend signing up for automatic release notifications for ag-grid. Then, following a release, check whether the item was implemented by looking for it in the changelog page.

Thanks again for bringing this up with us.

makinggoodsoftware:
Hi,

In order to help us keep our issues queue clear, we are going to close this issue

The official way to check the status for an issue raised here is to check its AG-xxxx ID against our pipeline and changelog

…the pipeline (if not yet developed)

https://www.ag-grid.com/ag-grid-pipeline/

... or the changelog, (if the issue has been resolved)

https://www.ag-grid.com/ag-grid-changelog/

[...]

PR #4160, opened 2020-10-26
Feat/4154 make rownode generic

stevenhankin:
For feature #4154 I was thinking about something along the lines of this draft PR

Screenshot 2020-10-27 at 08 10 34

It's probably a naive example and can be taken a lot further (e.g. determining the columns based on the data type)

Would also be nice to type the props for the Grid components


makinggoodsoftware:
Hi,

Thanks for this, note that we just recently have decided to go ahead and create a ticket in our pipeline to track the work to add generics for the rowNode.

We will follow up soon on the issue you raised.

In the meantime, I will close this, but feel free to follow up there!

Thanks!

PR #4210, opened 2020-11-23
Fix simple object check when value is null

soetji:
`deepCloneDefinition()` breaks when `value` is `null`. `typeof null` is `object` but `null` has no properties so `value.constructor` crashes.

This effects `api.getColumnDefs()`.


makinggoodsoftware:
Hi,

ceolter:
thank you, silly bug 👍 

i have put this fix into our development branch, which will be merged into master upon our next release. as such, am closing the pr.

thank you very much for taking the time to raise, made my job easier in fixing.

PR #4277, opened 2021-01-07
Add css "ag-row-level-N" for height calc

fvyshkov:


makinggoodsoftware:
Hi,

kiril-matev:
Hello!

Thank you for submitting this.

We agree this change would be a valuable addition to the product and we have added a feature request for this.

We have added this to our backlog and we are tracking it with the following reference and description:
**AG-5012 Feedback from Github (multiple items relating to product and docs)**

You can see whether this item is selected for implementation on the NEXT RELEASE tab of the product pipeline:

I would recommend signing up for ag-Grid new release notifications. This way you'll know as soon as a new version is out and you can check whether this was implemented on the changelog page.

Thanks again for bringing this up with us.

makinggoodsoftware:
Hi,

In order to help us keep our issues queue clear, we are going to close this issue

The official way to check the status for an issue raised here is to check its AG-xxxx ID against our pipeline and changelog

…the pipeline (if not yet developed)

... or the changelog, (if the issue has been resolved)

https://www.ag-grid.com/ag-grid-changelog/

We keep issues raised in the pipeline through github open for a while so that the community can add some feedback to them, and close them when there is no activity for a while.

If you want to add further feedback to a closed ticket with this comment, we would recommend opening a new ticket as we don’t monitor closed issues.

PR #4303, opened 2021-01-18
Solve performance bottleneck for tables with hundreds of columns

valentinstn:
We are using `ag-grid` with many (many hundreds) of (mostly hidden) columns. These lines caused a significant performance bottleneck during the initial rendering of the table.

The fix basically removes the need to run `indexOf()` calls in loops and uses an indexed map instead at performance-critical points in the code. In my example, the total rendering time could be reduced from roughly 10 to 1 second.


makinggoodsoftware:
Hi,

ceolter:
I have implemented this in latest, will be in the next release.
https://github.com/ag-grid/ag-grid/commit/579ef06bd01d71154fe1cf802f9426c7b674b676

Issue #4302, opened 2021-01-16
Usual Rows can be full-width


abadakhshan:
**I'm submitting a 
```
[] bug report 
[x] feature request 
[] support request 
```

**Current behavior**

the rows' width equals sum of columns' width and there is an empty space after last column.

![2222](https://user-images.githubusercontent.com/16729513/104814734-c78ef080-5825-11eb-902e-10cb17bc2965.png)

**Expected behavior**

It would be nice if there is a option that rows would be full-width. In this situation if sum of columns' width is less than available space, row should fill available space exactly like header row does currently.

![111](https://user-images.githubusercontent.com/16729513/104814738-ccec3b00-5825-11eb-9b29-8c68c315ac9c.png)

makinggoodsoftware:
Hi,

Thanks for your feedback, note that we don’t think this is a bug, as the system works as expected in this area.

Note that we don’t have the capacity to take feature requests as we have a full pipeline.
https://www.ag-grid.com/ag-grid-pipeline/

If you are a customer, you should be using ag-Grid Enterprise in which case you should have access to our customer support system (via zendesk), and that is the only channel where we accept feature requests.

If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

---

We are going to label this as managed-by-the-community as it is probably there are some ways to mitigate this and hopefully someone from the community can help you with a workaround/fix.

makinggoodsoftware:
Hi,

This ticket has been flagged as managed-by-community for a while now and there has not been any activity, to help us tidy up we are going to close it.

Thanks

PR #4334, opened 2021-02-05
Usual Rows can be full-width

abadakhshan:
Add **rowFillExtraWidth** option to grid to handle this issue

#4302


makinggoodsoftware:
Hi,

ceolter:
I don't see how this fixes anything?
All that was done is a new property put into GridOptions, but I don't see any code uses this property.

Also the PR has lots of formatting changes, making it difficult to see what the intended changes are.

ceolter:
closing until code doesn't include formatting changes and it's more clear what this is doing

abadakhshan:
Hello @ceolter 

This PR is responded to issue #4302.

It adds a property to GridOptions (rowFillExtraWidth) and based on this property do something to fill extra width.
Add [fillExtraWidth ](https://github.com/abadakhshan/ag-grid/blob/430575b6d314701d9e970bbd8bb585a979fdb47b/community-modules/core/src/ts/columnController/columnController.ts#L3508) method to `coloumnController `and call it when needed.

All formats are based on https://github.com/ag-grid/ag-grid/blob/master/.prettierrc.
If it is incorrect let me know where is the correct prettier config for formatting.

Thank you.

gportela85:
@abadakhshan The `prettier` plugin is not in use at the moment. As a general rule when submitting PRs, do not change code styling. Submit a PR containing changes that are relevant to the bugfix/feature only. It's also better for visibility of what is actually being changed.

abadakhshan:
Ok.
I will update this PR.

Thank you.

abadakhshan:
Hello @gportela85, @ceolter 

I created another [PR](https://github.com/ag-grid/ag-grid/pull/4382)

Thank you for your advice.

PR #4340, opened 2021-02-08
Use correct component selector in angular README

bamminger:
Fixed wrong replacement of ag-grid-angular in angular README.md


makinggoodsoftware:
Hi,

ceolter:
Thank you. This is already fixed. See:
https://github.com/ag-grid/ag-grid/blob/master/community-modules/angular/README.md

PR #4382, opened 2021-03-04
Add 'rowFillExtraWidth' option to grid to handle extra area on grid.

abadakhshan:
Add the `rowFillExtraWidth` option to the grid to handle this issue #4302.

If `rowFillExtraWidth` set to the `true` the grid's row will fill all width of the grid if the sum of the width of columns is less than the width of the grid. otherwise, there is an empty space in the extra area.

### DIFF
--- a/community-modules/angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
+++ b/community-modules/angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
@@ -438,6 +438,7 @@ export class AgGridAngular implements AfterViewInit {
     @Input() public showOpenedGroup : any = undefined;
     @Input() public suppressClipboardApi : any = undefined;
     @Input() public suppressModelUpdateAfterUpdateTransaction : any = undefined;
+    @Input() public rowFillExtraWidth: boolean = undefined;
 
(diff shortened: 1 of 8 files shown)


gportela85:
@abadakhshan It seems like the feature you want could be achieved with 1 line of CSS and doesn't need to be part of the core grid.
see: https://plnkr.co/edit/VSbpBG2fMpsk34GJ 

CSS: 

```css
.ag-center-cols-container { 
  min-width: 100%;
}
```

abadakhshan:
@gportela85  It works for LTR but there are some problems in RTL mode
see: https://plnkr.co/edit/cYAVGedjMMuUEG4I

gportela85:
@abadakhshan In that case, the fix is a little more complicated than initially presented. Instead of using `left` to set the position of the column, we should be using `start`. Meaning, setLeft in LTR and setRight in RTL. The `setLeftFeature` will probably need an update. I'll file this as a Feature Request. Thanks for pointing this out.

gportela85:
@abadakhshan I filed this as `AG-5136 - The center container should be allowed to cover 100% of the viewport` and added a detailed description of problems with RTL.

Thanks again for reporting this and for the effort with the Pull Request, but we will go about adding this improvement with a different approach.

PR #4419, opened 2021-03-26
Fix small typo

dangrussell:


makinggoodsoftware:
Hi,

makinggoodsoftware:
Hi,

Thanks for this! This has now been merged and will be part of our next release ETA 1/2 weeks 25.2

Please keep an eye on our pipeline to keep track of this

Thanks

PR #4429, opened 2021-03-31
fix(angular): zone aware

### LINKED ISSUE #4392: Angular AG-GRID without zone
```
[] bug report => see 'Providing a Reproducible Scenario'
[x] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
```

**Requesting Community Support**
I'm happy to create a PR myself to handle this but is this something you would accept or would it be better to manage it outside of AG-Grid?

**Providing a Reproducible Scenario**
- create an angular application without zone turned on.
- use custom components inside ag-grid

**Current behavior**
Uncaught ReferenceError: Zone is not defined

**Expected behavior**
Expect it to either run without zone or handle the error so we can trigger change detection manually

* **AG Grid version:** 25.1.0

* **Browser:**
All

* **Language:** Typescript

Jordan-Hall:
closes #4392

This change would require the parent wrapper component to fire change detection. This is a better solution than firing the change detection internally in case it's not required.

Repo: https://stackblitz.com/edit/ag-grid-angular-hello-world-3aca88?file=src%2Fapp%2Fapp.component.ts 

Issue:
AG grid was always forcing angular to run outside of Zone. But if the zone is turned off the minute you add an event the component will fire an error saying ngZone is undefined.

### DIFF
--- a/grid-packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
+++ b/grid-packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
@@ -1,4 +1,4 @@
-import {Injectable, NgZone} from "@angular/core";
+import { Injectable, NgZone } from "@angular/core";
 import {VanillaFrameworkOverrides} from "ag-grid-community";
 
 @Injectable()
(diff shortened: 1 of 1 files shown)


Jordan-Hall:
@davis4acca this should fix the issue

makinggoodsoftware:
Hi,

makinggoodsoftware:
Hi,

Thanks for this! This has now been merged and will be part of our next release ETA 1/2 weeks 25.2

Please keep an eye on our pipeline to keep track of this

Thanks!

PR #4499, opened 2021-05-31
CHG - add  `data : any` to `IFilterParams`

tarekbazine:
add `data : any` to the `IFilterParams` interface, so that we have a similar interface declaration as `ICellRendererParams`


makinggoodsoftware:
Hi,

ceolter:
This doesn't make sense. The init() method is called once when the filter is created. At this point the filter is not looking at a specific row, so there is no sensible ```data``` item to pass in. If I am missing something, please advise what the proposed ```data``` attribute would contain. Thanks.

PR #4512, opened 2021-06-09
[core] Introduce a default value parser that preserves the data type

spion:
Context:

When the user updates a cell value via the default editor, the resulting value is always a string. If the column participates in numeric operations e.g. as an aggregate value, the resulting aggregates will be invalid (due to use of the addition operator on strings) or the value may be refused.

Steps to reproduce:

* Open the ag-grid example page: https://www.ag-grid.com/example.php
* Add Country to row groups
* Add "Bank balance" to aggregate values
* Expand a country group and update the first value in its children
* Update the value for a second time.

Expectation: The second update works

Result: The second update is ignored (the editor produces a string which is rejected)

Various alternative options considered, including their pros and cons can be found in this gist: https://gist.github.com/spion/b095cf835d9703367280da356fbde487

### DIFF
--- a/community-modules/core/src/ts/rendering/cellComp.ts
+++ b/community-modules/core/src/ts/rendering/cellComp.ts
@@ -1386,6 +1386,23 @@ export class CellComp extends Component implements TooltipParentComp {
         }
     }
 
+    private defaultValueParser(params: NewValueParams) {
+        // The default value parser attempts to preserve the type of the field
+        // if the field is a Number or BigInt
+
+        if (typeof params.oldValue === 'number') {
+            let tryParseNumber = Number(params.newValue)
+            if (!Number.isNaN(tryParseNumber)) return tryParseNumber
+
+        } else if (typeof params.oldValue === 'bigint') {
+            try {
+                return BigInt(params.newValue)
+            } catch {
+                return params.newValue;
+            }
+        }
+        return params.newValue;
+    }
     private parseValue(newValue: any): any {
         const colDef = this.getComponentHolder();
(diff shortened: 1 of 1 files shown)


ceolter:
Hi Gorgi,

Thanks for the PR.

I agree with the issue, it is problem.

The proposed solution is a breaking change - there is the possibility that a user depends on the new value been a string. Eg they could be providing their own Cell Editor and converting to a string on purpose, eg to add another component to the value (eg old value was minutes, new value could be minutes with some formatting, or maybe some added meta-data to include an image / color / etc). If the developer was depending on the type changing, I would think the developer is been silly, however that wouldn't change the fact this could break things for some people.

Having said that, I like the solution, and if we were writing the grid for the first time, this would be the reasonable approach.

From the other options you present, I like the change to the aggregator. Having a check for string makes sense. That would also cover the case where the developer has provided strings in the data initially and wants to aggregate.

To push this forward, some more analysis would be needed from us.

Generally speaking the approach and proposed solutions all are fair. Should we speak again in person, that fact you have gone through our code and put this type of thought into it will be a good bases for some conversations :)

Also right now I'm in the middle of a massive CellComp refactor! (coincidence!!) We are reworking the UI to allow plugging React (and later other frameworks) onto the front end, to allow a pure React offering (not just a wrapper). This means we're breaking up all the components (CellComp is my current work) and putting all logic into a Controller class, with the intention that CellComp will use the Controller, and a future CellCompReact will use the same Controller. So making reusable logic, and the Component will only be concerned with setting things into the dom.

PR #4523, opened 2021-06-13
Fix link in console warning

nclavaud:
The console shows a helpful message, but the links are outdated.


makinggoodsoftware:
Hi,

ceolter:
Thank you. I have implemented this on my local dev machine. Nothing wrong with the proposed PR, just that we work of a different branch for the next release, so easier to implement it locally. Thanks ;)

Issue #4543, opened 2021-06-25
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.


dangrussell:
.scss files use "/" for division. Should be changed to use math.div()

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(ag-param(grid-size), 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

    ╷
246 │         margin: ag-param(grid-size) / 2 0;
    │                 ^^^^^^^^^^^^^^^^^^^^^^^
    ╵
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-base\sass\parts\_widgets.scss 246:17         ag-theme-base-part-widgets()    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-base\sass\_ag-theme-base-mixin.scss 39:5     ag-theme-base()
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-alpine\sass\_ag-theme-alpine-mixin.scss 8:5  ag-theme-alpine()
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-alpine\sass\ag-theme-alpine.scss 7:5         @import

DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(ag-param(grid-size), 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

    ╷
255 │         margin-bottom: ag-param(grid-size) / 2;
    │                        ^^^^^^^^^^^^^^^^^^^^^^^
    ╵
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-base\sass\parts\_widgets.scss 255:24         ag-theme-base-part-widgets()    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-base\sass\_ag-theme-base-mixin.scss 39:5     ag-theme-base()
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-alpine\sass\_ag-theme-alpine-mixin.scss 8:5  ag-theme-alpine()
    node_modules\@ag-grid-enterprise\all-modules\dist\styles\ag-theme-alpine\sass\ag-theme-alpine.scss 7:5         @import

[...]

PR #4551, opened 2021-06-29
Removing deprecated division symbol

lounsbrough:
This PR resolves https://github.com/ag-grid/ag-grid/issues/4543

This is the result of running `sass-migrator division **/*.scss`

https://www.npmjs.com/package/sass-migrator

### DIFF
--- a/community-modules/core/src/styles/ag-theme-alpine/sass/_ag-theme-alpine-mixin.scss
+++ b/community-modules/core/src/styles/ag-theme-alpine/sass/_ag-theme-alpine-mixin.scss
@@ -99,7 +99,7 @@
 
     .ag-charts-format-top-level-group-toolbar {
         margin-top: ag-param(grid-size);
-        @include ag-theme-rtl((padding-left: ag-param(icon-size) / 2 + ag-param(grid-size) * 2));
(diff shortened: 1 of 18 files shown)


TheSeg:
Switching to multiplication where possible is the right move. The only issue with the approach using `math.div` is that Libsass projects will break. While the best choice for everyone is to move away from Libsass, using `math.div` is more breaking than meets the eye.

The Bootstrap project solved this situation by creating a custom `divide()` function for these situations. That way you still get division while not requiring `@use "sass:math";`. Everyone wins!

* [Bootstrap `devide()` PR](https://github.com/twbs/bootstrap/pull/34245)
* Bootstrap's `divide()` function:
https://github.com/twbs/bootstrap/pull/34245/files#diff-de6111a37eb67d027961043ce738e80a3ac4ad61a6e7704d35cb3168369cba01R223

makinggoodsoftware:
Hi,

gportela85:
@lounsbrough Thanks for submitting this PR but merging this would actually break our build as is. This PR is also updating files within the `dist` folder.

We are already tracking a ticket for this issue and it will be fixed in the next release.

lounsbrough:
@gportela85 thanks for the update, much appreciated!

PR #4559, opened 2021-07-02
[API] Allow passing in source argument to columnApi.setColumnWidth(s)

namng5555:
Passing in the source argument allows more flexibility with simulating different types of resizing. For example, we may want to implement some UI for resizing in which the 'ui' sources may be more suitable. This was the case at my company where we've implemented a column resize indicator which replaces the native realtime resizing implementation:

indicator-demo-1

(Note: the data in the table is randomly generated, no real customer information is displayed)

However, we've run into a problem where we expect the event source of the resize drag to be `uiColumnDragged`, but since the indicator was built with `columnApi.setColumnWidth`, the event source becomes `api`. This is because under the hood `columnController.setColumnWidths` defaults the source argument to `api` and there's no way in the public API to change that. Although I do understand that logically the event source of `columnApi.setColumnWidth` should be `api` since... well... it is coming from an API, I believe that it would be more flexible to open up the `source` argument so we can differentiate between different types of column resizing using the public methods.

Would it be possible to add the `source` argument to `columnApi.setColumnWidth(s)`? We currently access the private `columnController` in order to get around this and this would help us get on the standard track.

Would appreciate any response to my proposed changes, thank you!


namng5555:
Not sure who to ping for review, but I see @makinggoodsoftware responding to a lot of PRs
Sorry if this is the wrong person to ping 😅

makinggoodsoftware:
Hi,

You did ping the right person! Let us have a look.

makinggoodsoftware:
Hi,

Apologies for the late reply! Note that this was made part of v26, you should see there these changes.

Hope this helps!

Issue #4561, opened 2021-07-04
DateFilter : UI don't update when change date if ag-grid is inside WebComponent


Bakino:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

**Providing a Reproducible Scenario**
When the ag-grid is included inside a WebComponent, the UI don't update when user choose a date in the picker so it does not display the second condition as usual.

Here is a plunkr that show the problem : https://plnkr.co/edit/fH9XoUODN6qAdf7o?preview

**Current behavior**
When choose a date, it does not display the second condition

**Expected behavior**
When choose a date, it should display the second condition

* **AG Grid version:** 25.3.0

* **Browser:**
Reproduced on Chrome and Firefox

* **Language:** 
Tested in ES6

makinggoodsoftware:
Hi,

Thanks for your feedback, note that we don’t support WebComponents .... So we think this would not be a bug, but if anything support or an FR

Note that we don’t have the capacity to take feature requests as we have a full pipeline.
https://www.ag-grid.com/ag-grid-pipeline/

If you are a customer, you should be using ag-Grid Enterprise in which case you should have access to our customer support system (via zendesk), and that is the only channel where we accept feature requests.

If you are an ag-Grid Enterprise Customer, please raise this in our Zendesk support system - contact info@ag-grid.com for access. This is our primary channel for Support.

---

We are going to label this as managed-by-the-community as it is probably there are some ways to mitigate this and hopefully someone from the community can help you with a workaround/fix.

makinggoodsoftware:
Hi,

 

This ticket has been flagged as managed-by-community/waiting-for-repro for a while now and there has not been any activity, to help us tidy up we are going to close it.

 

Thanks

PR #4562, opened 2021-07-04
Use getRootNode() instead of document to be compatible with WebComponents #4561

Bakino:
This modification fixes the problem described in ticket #4561


makinggoodsoftware:
Hi,

Webcomponents is not supported anynmore I am afraid, we are going to proceed to close this PR

Thanks

PR #4579, opened 2021-07-10
Replace Tslint with Eslint

gportela85:


.vscode/settings.json
```diff
@@ -15,12 +7,51 @@
...
+        { "directory": "./enterprise-modules/column-tool-panel" },
+        { "directory": "./enterprise-modules/excel-export" },
+        { "directory": "./enterprise-modules/filter-tool-panel" },
+        { "directory": "./enterprise-modules/master-detail" },
+        { "directory": "./enterprise-modules/rich-select" },
+        { "directory": "./enterprise-modules/server-side-row-model" },
+        { "directory": "./enterprise-modules/set-filter" },
+        { "directory": "./enterprise-modules/multi-filter" },
+        { "directory": "./enterprise-modules/viewport-row-model" },
+        { "directory": "./enterprise-modules/all-modules" }
```
seanlandsman, line 41: How about:

```
"eslint.workingDirectories" : [
    "./community-modules/"
    "./enterprise-modules/"
]
```

seanlandsman (approved):
Looks ok to me. Would need to ensure changes made to the react project work with "legacy' wrapper too, including umd bundle

.vscode/settings.json
gportela85: if we change this array to only include `./community-modules` and `./enterprise-modules` then `eslintrc.js` will assume that `tsconfig.eslint.json` is inside these folders and we will not be able to have different configs for each package. 
I'll try to make the eslintrc.js file a little cleaner for each package before merging

seanlandsman: fair enough

PR #4580, opened 2021-07-11
Brazilian Portuguese translations file

gubertcalixto:
Creating Brazilian Portuguese translations file.


makinggoodsoftware:
Hi,

Thanks for this, unfortunately, we are not looking to create a repo of localized strings.

Hope this makes sense, we will close this PR

PR #4595, opened 2021-07-21
Ag 3944 angular component types

StephenCooper:


community-modules/angular/gulpfile.js
```diff
@@ -32,38 +30,13 @@ const mainPostCompileRename = () => {
...
 
-const updateProperties = (callback) => {
-    gridToNg.updatePropertiesSrc(callback);
-};
-
 const compileSource = (callback) => {
     return ngc('./tsconfig-src.json', callback);
 };
 
-const copyFromModuleSource = () => {
```
StephenCooper, line 43: copying is done from within the grid-packages so removing from here.

community-modules/angular/package.json
```diff
@@ -53,7 +53,6 @@
     "ts-node": "~8.3.0",
     "tslint": "~5.18.0",
     "typescript": "~3.7.5",
-    "replace-in-file": "^4.1.0",
-    "typescript-simple": "8.0.6"
```
StephenCooper, line 57: No longer required after deleting watch task.

community-modules/core/src/ts/main.ts
```diff
@@ -323,7 +323,8 @@ export {
     DndSourceCallbackParams,
     EditableCallbackParams,
     SuppressPasteCallbackParams,
```
StephenCooper, line 325: Adding missing Interfaces as revealed by now typing angular component.

seanlandsman (approved):
Looks good to me, but you'll also need to change the Vue (not Vue 3) projects to use ComponentUtil.PUBLIC_EVENTS as they currently rely on ComponentUtil.EVENTS

Otherwise looks good

rmc-software (approved):

StephenCooper:
> Looks good to me, but you'll also need to change the Vue (not Vue 3) projects to use ComponentUtil.PUBLIC_EVENTS as they currently rely on ComponentUtil.EVENTS
> 
> Otherwise looks good

I only see EVENTS being used in the Vue 3 module.

PR #4598, opened 2021-07-22
Ag 3944 add existing but missing interfaces

StephenCooper:
Been through GridOptions and ColDef and replaced a number of `any` options with their existing type where this is obvious from the source code.


community-modules/core/src/ts/entities/colDef.ts
```diff
@@ -356,7 +356,7 @@ export interface ColDef extends AbstractColDef, IFilterDef {
     onCellContextMenu?: (event: CellContextMenuEvent) => void;
 
     /** Icons for this column. Leave blank to use default. */
-    icons?: { [key: string]: string; };
```
StephenCooper, line 359: This type is incorrect as the icons property accepts a function as stated in the docs.

rmc-software (approved):

PR #4604, opened 2021-07-28
Ag 5559 drive docs with ast

StephenCooper:
The aim here is to generate interfaces and type data from the code via AST parsing to provide accurate information to the documentation pages.


grid-packages/ag-grid-docs/documentation/doc-pages/grid-api/grid-options.json
```diff
@@ -0,0 +1,1533 @@
+{
+  "suppressBrowserResizeObserver": {
+    "description": "*****************************************************************************************************\nIf you change the properties on this interface, you must also update PropertyKeys to be consistent. *\n*****************************************************************************************************",
```
StephenCooper, line 3: I will need to correct this description parsing in the next stage. It is not used yet in the ApiDocumentation component.

grid-packages/ag-grid-docs/package.json
```diff
@@ -13,7 +13,8 @@
     "validate-es5": "npx eslint documentation/doc-pages",
     "test": "npx jest && cd documentation && npm run test",
     "hash": "sh ../../scripts/hashDirectory.sh > .hash",
-    "testtt": "node dev-server.js"
+    "testtt": "node dev-server.js",
+    "update-properties": "node -e \"require('./updateGridAndColumnProperties.js').updatePropertiesBuilt()\""
```
StephenCooper, line 17: Need to update the name of this file and task to be more accurate.

grid-packages/ag-grid-docs/documentation/src/components/ApiDocumentation.jsx
```diff
@@ -387,40 +434,106 @@ const FunctionCodeSample = ({ framework, name, type }) => {
...
+    })
 
-    return lines;
+    return allLines;
 };
 
 const getLinkedType = (type, framework) => {
     if (!Array.isArray(type)) {
-        type = [type];
+        type = type.split('|').map(t => t ? t.trim() : t);
```
StephenCooper, line 531: Plan to use a regex here to improve this and make it cleaner.

grid-packages/ag-grid-docs/documentation/src/components/ApiDocumentation.jsx
```diff
@@ -387,40 +434,106 @@ const FunctionCodeSample = ({ framework, name, type }) => {
...
+
+        } else {
+
+            // We have an interface
+            const lines = [`interface ${name} {`];
+
+            const properties = Object.entries(interfaceType.type);
+            properties.sort(([p1,], [p2,]) => {
+                // Sort alphabetically but with $scope at the end
+                if (p1 === '$scope') return 1;
```
StephenCooper, line 512: Any thoughts on hiding the $scope property from the Angular, React, Vue frameworks?
Or not worth removing now seeing as will soon be removed for real.

rmc-software (approved):

grid-packages/ag-grid-docs/documentation/src/components/ApiDocumentation.jsx
seanlandsman: I don't think it'll be soon - it'll still be there in 26.x and as that'll be our LTS version it's unlikely to be removed entirely (from docs) soon

seanlandsman: but that said, I can't see why we'd display this in any context other than Angular (even though angular isn't angularjs...)

seanlandsman (commented):
Looks fine to me - I've not really reviewed the actual AST changes in depth but the docs & resulting angular outputs look good as far as I can tell

grid-packages/ag-grid-docs/documentation/src/components/ApiDocumentation.jsx
StephenCooper: Yeah, wasn't sure if would need it in both Javascript and Angular to catch users of AngularJS. 

But sounds like hiding it from Vue and React would make sense.

Issue #4659, opened 2021-09-06
ag-charts-angular install failing


nikhil-kuyya:
**I'm submitting a ag-charts-angular install failing**  (check one with "online and offiline")
```
node 14.17.6
[] bug report => see 'installing the ag-charts-angular with npm install --save ag-charts-angular'
```

**Requesting Community Support**
Please provide here community inputs.

**Providing a Reproducible Scenario**
Got to [Npm RunKit](https://npm.runkit.com/ag-charts-angular) and press run.

![image](https://user-images.githubusercontent.com/46197824/132242090-a67a1c07-8c43-4c66-8a14-aff02fbff49f.png)

If reporting a bug make sure to state.

Current behaviour : failing install with angular : @12.1.0 and @ag-grid-enterprise/all-modules": "^26.0.0" and hope it will get reproduced in all environment.
Expected behaviour. It should install 

**Please tell us about your environment:**
Windows 10 Pro

* **AG Grid version:** 26.0.0

* **Browser:** Chrome

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

nikhil-kuyya:
I think issue is wrong tagname for @angular/core : https://github.com/ag-grid/ag-grid/pull/4660.
hope i made proper pr.
correct me if i am wrong any process and anything.

nikhil-kuyya:
also npm install works with --legacy-peer-deps i don't this is good to use this way or update the package in the pr above.

nikhil-kuyya:
@StephenCooper thank you merging,hope it is fine. and can i know will this package get released.

StephenCooper:
Will be released along with other fixes in the next minor version. No fixed
date yet.

DavidBradbury:
Hi there, is there a planned date for the next minor version at this point? We would use the older version but it is no longer compatible with our dependencies and installing with --legacy-peer-deps doesn't work for us.

StephenCooper:
Targeting the start of October for next minor release.

PR #4660, opened 2021-09-07
Update package.json | Invalid Tagname i guess | causing install issue

nikhil-kuyya:
I think this is a causing the issue: ! Invalid tag name "^>=8.0.0":
Can we please release this new change.
i opened the issue in https://github.com/ag-grid/ag-grid/issues/4659


StephenCooper (approved):
Thanks for the PR to correct this. 👍

PR #4748, opened 2021-10-28
Upgrade vue-property-decorator to 9.1.2

tlyau62:


makinggoodsoftware:
Hi,

seanlandsman:
Thanks for the suggestion and the PR!

Issue #4763, opened 2021-11-10
No way to specify desired "max-width" of Columns while exporting excel


hvsharma:
Hi,

My use case is this:

1) Auto-size the column width according to the column contents.
2) Provide a max-width parameter with Autosize.

I didn't find any way to do it for now. 

**Example:** 

A table has short-short cell contents (1-40 characters). But then there is a particular column for which the contents are like 100 characters+.

**Desired Functionality:**

I pass a max-width parameter:

"excelExportParams.maxColumnWidth=50" 

With this, all the small columns are auto-sized, but for the large columns, the width will not exceed 50 in the excel sheet.

makinggoodsoftware:
HI,

This seems like a support question and It sounds like you have an ag-grid license/evaluating. Note that github is not our main channel for support for customers.

You should contact us through zendesk, you should have access already at https://ag-grid.zendesk.com/

If that is not the case, or you are having trouble, please contact us at info@ag-grid.com where they will be able to assist you with your access.

Thanks

PR #4765, opened 2021-11-10
4763: Add maxColumnWidth parameter for excel export

hvsharma:
Added maxColumnWidth parameter to fix the following issue:

[Issue 4763](https://github.com/ag-grid/ag-grid/issues/4763)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .idea/codeStyles/Project.xml
 .idea/encodings.xml
 /.idea/atlassian-ide-plugin.xml
+.idea/ag-grid.iml
 .vscode
 
 grid-packages/ag-grid-docs/src/jira_reports/prod/credentials.json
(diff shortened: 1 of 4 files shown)


makinggoodsoftware:
Hi,

kiril-matev:
Hello!

Thank you for bringing this up with us. 

The functionality you described wasn't part of the requirements when we built this feature. This is why we have added a feature request in our backlog to track this requirement and implement it in the future.

You can track it with this reference:
**AG-6010 [Excel Export] Allow setting a maximum column width for the exported Excel columns**

You can follow the status of feature requests, bugs and releases using our pipeline:

**Now that this requirement is recorded in our backlog, we will soon close the GitHub issue because we will not update it when this feature is added in a future version of AG Grid. The best way to get an update on this item is to sign up for AG Grid new release notifications. Once a new AG Grid version is released, look for the item reference above (AG-XXXX) on the changelog page to see if it's resolved in that version:
https://www.ag-grid.com/ag-grid-changelog/**

PR #4789, opened 2021-11-22
fix endless timeout loop on autoHeight

yosefclinch:
Currently When setting autoHeight = true, its lead to endless loop of setTimeout requests.
Its happen because small logical mistake in the measureHeight function - while its trying to recall itself 5 times, it using `x++` to update index, but this statment return the old x value, not the updted one. Need to use ether `++x` (which update x and return updated value) or `x+1` (which not update x, just return updated value)


gportela85:
Setting `timesCalled + 1` as the param will never change the value of `timesCalled` and this should cause the infinite loop. Please, submit a Bug Report with an example of how to reproduce the issue.

PR #4792, opened 2021-11-23
Fix ignored return value from getPopupPosition

lbueker:
Since version 26 the return value from getPopupPosition gets ignored, resulting in the cell editor using the default value 'over'.

### DIFF
--- a/community-modules/core/src/ts/rendering/cell/cellComp.ts
+++ b/community-modules/core/src/ts/rendering/cell/cellComp.ts
@@ -513,7 +513,7 @@ export class CellComp extends Component implements TooltipParentComp {
             keepWithinBounds: true
         };
 
-        const positionCallback = position === 'under' ?
+        const positionCallback = positionToUse === 'under' ?
             popupService.positionPopupUnderComponent.bind(popupService, positionParams)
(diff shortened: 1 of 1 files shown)


gportela85:
Can you please submit a Bug Report with an example of the issue you are fixing?

lbueker:
@gportela85 Done. #4796

kiril-matev:
Hello!

Thank you for reporting this - much appreciated!

Based on your description we've now reproduced this issue as a regression.

We have added this bug to our backlog and we are tracking it with the following reference and description:
**AG-6064 [Regression] When cell editor popup is used it cannot be displayed under the edited cell (because getPopupPosition return value is ignored by editors)**

We try to fix bugs from one release to the next, so this should be fixed in the next release or the one after if it was raised too close to the next release date.

You can see the next release date on the product pipeline page:

**Now that this issue is recorded in our backlog, we will soon close the GitHub issue because we will not update it when this bug is fixed in a future version of AG Grid. The best way to get an update on this item is to sign up for AG Grid new release notifications. Once a new AG Grid version is released, look for the item reference above (AG-XXXX) on the changelog page to see if it's resolved in that version:
https://www.ag-grid.com/ag-grid-changelog/**

Issue #4798, opened 2021-11-25
Code to expand all nodes fails on v26 when using server-side row-model-type


jeffmath:
**Providing a Reproducible Scenario**

- Use the following grid-options:
`{ rowModelType: 'serverSide' }`

- Execute code like this in a button-click handler, to expand all nodes:
`grid.api.forEachNode((node: RowNode) => node.setExpanded(true));`

**Current behavior**
The expand-all code in the handler does not work.  An error appears in the console:
```
main-es2015.3a26d002c2580b38c7d9.js:1 TypeError: Cannot read properties of undefined (reading 'rowIndex')
    at t.binarySearchForDisplayIndex (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.getRowUsingDisplayIndex (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at main-es2015.3a26d002c2580b38c7d9.js:5322
    at main-es2015.3a26d002c2580b38c7d9.js:5322
    at Array.forEach (<anonymous>)
    at t.findBlockAndExecute (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.getRowUsingDisplayIndex (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.getRow (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.getRow (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.createOrUpdateRowCtrl (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at main-es2015.3a26d002c2580b38c7d9.js:5322
    at Array.forEach (<anonymous>)
    at t.redraw (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.redrawAfterModelUpdate (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.onPageLoaded (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at main-es2015.3a26d002c2580b38c7d9.js:5322
    at Set.forEach (<anonymous>)
    at main-es2015.3a26d002c2580b38c7d9.js:5322
    at e.dispatchToListeners (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at e.dispatchEvent (main-es2015.3a26d002c2580b38c7d9.js:5322)
    at t.onModelUpdated (main-es2015.3a26d002c2580b38c7d9.js:5322)
[...]

PR #4799, opened 2021-11-25
Add missing undefined checks to BlockUtils methods

jeffmath:
This is meant to fix Issue #4798.


gportela85:
We are in the process of upgrading Typescript, but so far we are stuck with v3.6 which does not support the `?`. I'll push a fix for this.

Thanks for reporting

gportela85:
see: https://github.com/ag-grid/ag-grid/commit/f3c4a2b4a6fb0636886910fb5032c6bd5fdc09d4

jeffmath:
@gportela85 Thank you, I appreciate that.  My boss, who is with Allied Security Trust, says he will be purchasing an ag-grid license for our product, soon.

PR #4872, opened 2022-01-14
Fix cellClassRules quadratic perf

StefanFehrenbach:
I was investigating some rendering perf issues in our app. Turns out we were spending a lot of time in `processClassRules`. (To be fair, we're doing silly things with ~25 cell class rules that aren't actually useful.) I came across this code.

Currently it is (accidentally?!) quadratic. With every iteration of the loop we remove and add all previous (unique) classes. The comment makes me think that's not necessary. We can collect classes and only apply them at the very end. This also matches the scope of `classesToApply` and `classesToRemove`.

I believe this is what @ceolter intended in 6eec566babc8e33d185ab8cd24ab858a96793856 all along, the `}` just got misplaced. Note though that I haven't actually executed a grid instance with this patch applied, so I'm opening this as a draft PR.

### DIFF
--- a/community-modules/core/src/ts/styling/stylingService.ts
+++ b/community-modules/core/src/ts/styling/stylingService.ts
@@ -37,15 +37,15 @@ export class StylingService extends BeanStub {
                 if (singleClass == null || singleClass.trim() == '') { return; }
                 resultOfRule ? classesToApply[singleClass] = true : classesToRemove[singleClass] = true;
             });
-
-            // we remove all classes first, then add all classes second,
-            // in case a class appears in more than one rule, this means it will be added
-            // if appears in at least one truthy rule
-            if (onNotApplicableClass) {
-                Object.keys(classesToRemove).forEach(onNotApplicableClass);
-            }
-            Object.keys(classesToApply).forEach(onApplicableClass);
         }
+
+        // we remove all classes first, then add all classes second,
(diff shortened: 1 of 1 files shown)


StefanFehrenbach:
I tried to run tests locally. I checked out my branch, did `npm install` and `npm run test`. The latter failed because it couldn't find `ng`, which I take to be some Angular thing. I don't have time to debug this, right now.
Marking this as ready for review anyway. Maybe somebody with a working setup could run the tests, or trigger CI on this patch, or something?

StefanFehrenbach:
https://plnkr.co/edit/8QUFR2H6QXV2rUJS

This uses a lot of cell class rules.

I've done this:
1. scroll down until all rendered rows disappear
2. start timer when the first row gets rendered
3. wait
4. stop timer when the values in the Year column show up (they get rendered after everything else, I guess because of the cell renderer)

At N = 1000 it takes ~15s for me
At N = 2000 it takes ~55s for me

55s is a lot more than 2 x 15s.

I know that thousands of cell class rules are pretty unrealistic. That's just to make the effect bigger so it's easier to measure. It would be much better to instrument the `on(Not)ApplicableClass` callbacks to see how often they get called. (And repeatedly with the same arguments.)

gportela85:
thanks for providing the test scenario

PR #4964, opened 2022-03-11
docs(testing): adds row click unit test for angular

mjunior:
## Overview

This pull request adds an example of how to test a function called from `rowClick` event. In the last few days, I searched a lot about that and I couldn't find a solution on google.

I talked with @StephenCooper and he helped me find this solution with `waitForAsync` and `whenStable`.

I'd like to put this example on documentation for helping more developers. Feel free to edit what will be necessary.

Ps: fakeAsync with tick() or flush() don't work in this context.


StephenCooper:
Thanks for your input and highlighting the gaps in some of our Angular testing docs. In the last release I have added a new page with a worked example of testing async code. 

https://ag-grid.com/angular-data-grid/testing-async/

PR #4976, opened 2022-03-15
AG-2168 - Filter pivoted columns

AndrewGlazier:


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -59,6 +59,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel
 
     // standard stages
     @Autowired('filterStage') private filterStage: IRowNodeStage;
+    @Autowired('filterAggregatesStage') private filterAggregatesStage: IRowNodeStage;
```
AndrewGlazier, line 62: I'd like this to be optional and living within the row grouping package, but as it has to deep clone due to the way the node pipeline has been set up anyway, it's already a required stage, the deep clone logic could be duplicated into csrm, but as it's also needed for the event where no aggregation filter is present, it would be another duplication of logic.

Would like to rework the pipeline to simplify the logic and lower the code complexity however it would be quite a large change

community-modules/core/src/ts/columns/columnModel.ts
```diff
@@ -3118,6 +3118,12 @@ export class ColumnModel extends BeanStub {
                     const colDef = abstractColDef as ColDef;
                     if (columnCallback) {
                         columnCallback(colDef);
+
+                        const supportedFilters = ['agTextColumnFilter', 'agNumberColumnFilter'];
+                        if (!supportedFilters.some(filter => filter === colDef.filter)) {
+                            const message = `AG Grid -  Only text and number floating filters are supported on secondary columns.`;
+                            doOnce(() => console.warn(message), 'secondaryColumns.unsupportedFilter');
+                        }
```
AndrewGlazier, line 3126: Not certain if this is the appropriate way to do this

community-modules/core/src/ts/filter/filterManager.ts
```diff
@@ -185,6 +191,10 @@ export class FilterManager extends BeanStub {
         this.allAdvancedFilters.forEach(filterWrapper => {
             if (filterWrapper.filterPromise!.resolveNow(false, isFilterActive)) {
                 const resolvedPromise = filterWrapper.filterPromise!.resolveNow(null, filter => filter);
+                if (!filterWrapper.column.isPrimary()) {
```
AndrewGlazier, line 194: Aggregate filtering only works with secondary columns, rather than all aggregate columns due to this line. Not certain if this is the desired functionality or not.

community-modules/core/src/ts/filter/filterManager.ts
```diff
@@ -327,12 +336,29 @@ export class FilterManager extends BeanStub {
         );
     }
 
+    public doesAggregatedRowPassFilter(params: {
+        rowNode: RowNode;
+        filterInstanceToSkip?: IFilterComp;
+    }): boolean {
+        // [TODO] pass in an external filter specific to aggregate nodes
+        if (this.gridOptionsWrapper.isExternalFilterPresent() && !this.gridOptionsWrapper.doesExternalFilterPass(params.rowNode)) {
```
AndrewGlazier, line 344: This probably needs changed, as it uses the external filter also used for advanced filters, but holding off until requirements discussed, as not sure if a user provided filter is a requirement yet.

community-modules/core/src/ts/rowNodes/selectableService.ts
```diff
@@ -30,6 +30,13 @@ export class SelectableService extends BeanStub {
         }
     }
 
+    public updateSelectableAfterAggregateFiltering(rowNode: RowNode): void {
+        if (this.isRowSelectableFunc) {
+            const nextChildrenFunc = (node: RowNode) => node.childrenAfterAggregateFilter;
+            this.recurseDown(rowNode.childrenAfterFilter, nextChildrenFunc);
```
AndrewGlazier, line 36: I find the inconsistency between this argument and the nextChildrenFunc odd, but this is reflective of the previous function (lines 26-31)

PR #4978, opened 2022-03-15
Ag 3353 callbacks standard

StephenCooper:
Main idea is that every grid callback that the user can hook into will have params that have at least the following properties. 

```ts
export interface AgGridCommon {
    /** The grid api. */
    api: GridApi;
    /** The column api. */
    columnApi: ColumnApi;
    /** Application context as set on `gridOptions.context`. */
    context: any;
}
```
This will result in a consistent interface for users to work with and hopefully less confusion or work arounds requiring them to save a reference to the grid apis from the `onGridReady` event.


community-modules/core/src/ts/entities/iGridCallbacks.ts
```diff
@@ -0,0 +1,33 @@
+import { ColumnApi } from "../columns/columnApi";
+import { GridApi } from "../gridApi";
+import { CellPosition } from "./cellPosition";
+
+export interface BaseCallback {
+    api: GridApi;
+    columnApi: ColumnApi;
+    context: any;
+}
+export interface ITabToNextCellParams {
```
StephenCooper, line 10: These become internal interfaces for when we are creating the params for the callbacks.

gportela85: I think we have a problem where we use the word params to mean the params we pass in a Grid event. Maybe these should be just events (keep params in the public API for backwards compatibility).

This way we could simply add context to`AgGridEvent` and use that instead of `BaseCallbackParams`.

Just an idea, as `BaseCallbackParams` seems to be essentially duplicating `AgGridEvent`.

StephenCooper: `AgGridEvent` also has the `type` which is very much event specific. I get what you mean about the overloading of params though, especially as these properties are not really something the user will specify.

community-modules/client-side-row-model/src/clientSideRowModel/immutableService.ts
```diff
@@ -1,28 +1,20 @@
...
-
 import { ClientSideRowModel } from "./clientSideRowModel";
 
+
 @Bean('immutableService')
 export class ImmutableService extends BeanStub implements IImmutableService {
 
     @Autowired('rowModel') private rowModel: IRowModel;
     @Autowired('rowRenderer') private rowRenderer: RowRenderer;
-    @Autowired('columnApi') private columnApi: ColumnApi;
```
StephenCooper, line 24: No longer required as the apis are handled in the grid options wrapper.

community-modules/core/src/ts/columns/columnModel.ts
```diff
@@ -2625,6 +2626,7 @@ export class ColumnModel extends BeanStub {
                 providedColumnGroup: providedColumnGroup,
                 location: location,
                 api: this.gridOptionsWrapper.getApi()!,
+                columnApi: this.gridOptionsWrapper.getColumnApi()!,
```
StephenCooper, line 2629: Add missing property to a ColDef callback. We have to apply these in location as there is no central place to apply the apis, like with grid options wrapper.

community-modules/core/src/ts/components/framework/userComponentFactory.ts
```diff
@@ -80,91 +81,91 @@ export class UserComponentFactory extends BeanStub {
     @Autowired('userComponentRegistry') private readonly userComponentRegistry: UserComponentRegistry;
     @Optional('frameworkComponentWrapper') private readonly frameworkComponentWrapper: FrameworkComponentWrapper;
 
-    public getHeaderCompDetails(colDef: ColDef, params: IHeaderParams): UserCompDetails | undefined {
+    public getHeaderCompDetails(colDef: ColDef, params: WithoutGridCommon<IHeaderParams>): UserCompDetails | undefined {
```
StephenCooper, line 84: We use the `WithoutGridCommon` type so that we do not have to cast to interfaces which can lead to typing errors being hidden. It omits the common properties that will be attached centrally so that we do not have to worry about them where the params are created.

community-modules/core/src/ts/components/framework/userComponentFactory.ts
```diff
@@ -330,7 +330,11 @@ export class UserComponentFactory extends BeanStub {
         paramsFromGrid: any,
         paramsFromSelector: any = null
     ): any {
-        const params = {} as any;
+        const params: AgGridCommon = {
```
StephenCooper, line 333: Apply the common params for all user components params.

community-modules/core/src/ts/entities/gridOptions.ts
```diff
@@ -832,8 +817,10 @@ export interface GridOptions {
     suppressKeyboardEvent?: (params: SuppressKeyboardEventParams) => boolean;
 
     // *** Localisation *** //
-    /** A callback for localising text within the grid. */
+    /** @deprecated - Use `getLocaleText` instead. */
     localeTextFunc?: (key: string, defaultValue: string, variableValues?: string[]) => string;
+    /** A callback for localising text within the grid. */
+    getLocaleText?: (params: GetLocaleTextParams) => string;
```
StephenCooper, line 823: Where callbacks were not using params and it is viable to rename the function we have created a parameterized version and deprecated the original.

PR #4984, opened 2022-03-16
AG-5837 - Area series missing values

manapeirov:
- fixed stacked area chart not showing a point for a series
- fixed stacked chart domain when undefined values present in data


charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -314,15 +314,17 @@ export class AreaSeries extends CartesianSeries {
             // find the sum of y values in the stack, used for normalization of stacked areas and determining yDomain of data
             const total = { sum: 0, absSum: 0 };
             for (let i = 0; i < stack.length; i++) {
-                const y = +stack[i]; // convert to number as the value could be a Date object
+                if (stack[i] !== undefined) {
```
alantreadway, line 317: Query: I notice we're using `isNaN()` for the other case fixed in this PR, should we be consistently using that?

alantreadway (approved):
LGTM other than the one query.

PR #4989, opened 2022-03-18
AG-5824 - Area Series skip invalid x values instead of rendering a gap

manapeirov:


alantreadway:
I'm still trying to analyse the change, but in the meantime are there any tests that could or should be added to exercise this change in behaviour?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -260,109 +266,118 @@ export class AreaSeries extends CartesianSeries {
...
+            // Y datum
+            yKeys.forEach((yKey, i) => {
+                if (!(yKey in datum)) {
+                    doOnce(() => console.warn(`The key '${yKey}' was not found in the data: `, datum), `${yKey} not found in data`);
+                    return;
+                }
+                const value = datum[yKey];
+                const seriesYs = yData[i] || (yData[i] = []);
+
+                if (!seriesItemEnabled.get(yKey)) {
```
alantreadway, line 303: In this case, am I right in thinking the resulting `yKeys` array has an element which is an array of all `0`s?

Is this skipped later in processing (maybe via a similar `!seriesItemEnabled.get(yKey)` guard)? If so, it might make sense to do this check as the first thing in this `forEach()` block to avoid needing to do the 2x lookups (existence of `yKey` and value of `datum[yKey]`) before this, and remove the creation of the all-`0` array?

**If this is a non-trivial, wherever `this.yKeys` is consumed, then I wouldn't worry about it as part of this PR.**

alantreadway: @manapeirov and I discussed this offline, essentially there are some assumptions in dependent code that the all-`0` array is present - I'm happy if we don't tackle this now.

alantreadway:
The fonts in the updated visual tests look wrong - let me have a play.

alantreadway (approved):
LGTM - thanks for updating the tests, I pushed a fix for the font rendering, I may need to see if I can make this more robust on MacOS.

PR #4996, opened 2022-03-21
fix(vue3): declarative columns fix

fromatlantis:
Fix declarative columns for vue3


seanlandsman (approved):
Looks great - thanks for providing this PR

PR #4998, opened 2022-03-21
Ag 5824/line series invalid x values

manapeirov:


manapeirov:
N.B. https://ag-grid.atlassian.net/browse/AG-5824?focusedCommentId=39859

charts-packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -20,11 +20,29 @@ export abstract class CartesianSeries extends Series {
...
+    protected checkDatum<T>(value: T, isContinuousScale: boolean): T | string | undefined {
+        if (isContinuousScale && isContinuous(value)) {
+            return value;
+        } else if (!isContinuousScale) {
+            if (!isDiscrete(value)) {
+                return String(value);
+            }
+            return value;
+        }
+    }
```
alantreadway, line 44: There is an implicit `return undefined` here - could you add it explicitly, or remove the case? (Otherwise it looks to my eye like the case was missed during update?)

alantreadway (commented):
Fonts are wrong on the new chart snapshots :-/ Can you delete them and regenerate via the NPM `test` script?

This should do it until I can come up with a more robust MacOS approach:
```
npx lerna --scope ag-charts-community run test -- -- -u
```

alantreadway (approved):
LGTM

PR #5035, opened 2022-03-30
AG-6483 - Minimise Charts non-theme defaults.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6483

Removes all but cartesian axis variations from the `defaults.ts` file, pushing any variations into the `chartTheme.ts` module.

Bonus:
- Adds types to `chartTheme.ts` to aid with maintainability.
- Fixes `darkTheme.ts` for hierarchical charts.
- Removed incorrect use of an instance of `Padding` in the base theme config.


manapeirov (commented):
Looks good!

Just was unsure why some of the properties have been cast to a type, was it erroring if not?

alantreadway:
> Looks good!
> 
> Just was unsure why some of the properties have been cast to a type, was it erroring if not?

Do you mean things like:
https://github.com/ag-grid/ag-grid/pull/5035/files#diff-c6e69780aba072688d22cf3519cdbf9bfe8abe24282b89e7ac36fb20f8b539dfR212

I'm now relying on inferred types on the `getXDefaults()` - the return type of these went from `any` to unspecified, meaning it's inferred from the `return` statements in the function. Unfortunately **any string literally is inferred as `string`**, meaning that without the cast (e.g. `'bold' as FontWeight`) that the return type would be incompatible with the `AgChartOptions` interface contract, and cause a compile error.

Having said that, I might be able to improve type-safety by extracting these cases as constants - I'll see if I can make that tweak now.

alantreadway:
@manapeirov pushed change to remove the `as` casts - thanks for calling this out!

manapeirov (approved):

PR #5036, opened 2022-03-30
AG-6329 - calculate bandwidth for bar/column sparklines based on the …

manapeirov:
Improved bandWidth calculation for bar/column series with continuous x axis

### DIFF
--- a/enterprise-modules/sparklines/src/sparkline/sparkline.ts
+++ b/enterprise-modules/sparklines/src/sparkline/sparkline.ts
@@ -337,6 +337,7 @@ export abstract class Sparkline extends Observable {
         this.tooltip.toggle(false);
     }
 
+    protected smallestInterval?: { x: number, y: number } = undefined;
     // Fetch required values from the data object and process them.
     private processData() {
         const { data, yData, xData } = this;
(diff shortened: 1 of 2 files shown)


enterprise-modules/sparklines/src/sparkline/sparkline.ts
```diff
@@ -356,6 +357,22 @@ export abstract class Sparkline extends Observable {
         const { type: xValueType } = this.axis;
         const xType = xValueType !== 'number' && xValueType !== 'time' ? 'category' : xValueType;
 
+        const isContinuousX = xType === 'number' || xType === 'time';
```
alantreadway, line 360: Nit: Should we have a common definition for `isContinuousX` somewhere - I've seen variables with this name elsewhere, are they calculated the same way?

No action required, but I think it's worth thinking about if there's common domain logic that we should have a single implementation for?

alantreadway (approved):
LGTM, just one nit for us to think about on a wider scale than this PR potentially?

enterprise-modules/sparklines/src/sparkline/sparkline.ts
manapeirov: Typically in standalone charts, isContinuousX and isContinuousY will be the value of `xAxis.scale instanceof ContinuousScale` and `yAxis.scale instanceof ContinuousScale` respectively. That's a great point Alan, we can definitely add something to have less duplication. 

For sparklines it's a little different, we process the data first then create the axes.

PR #5042, opened 2022-04-01
AG-6586 - Prototype AJV-based Chart Options Validation

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6586

**WIP DO NOT MERGE**

Investigation into the feasibility of using `ajv` to validation Charts `AgChartOptions` structures.

Implementation detail:
- Introduces a build step (see `generate-ajv.js`) which takes `agChartOptions.ts` as the source of truth and:
  - Uses `typescript-json-schema` to create a matching JSONSchema of the options interface contract.
  - Uses `ajv` to create a standalone validator of that JSONSchema.
- During `AgChartsV2` execution, if we detect that we're in a development environment then perform the validation.

Performance:
- Anecdotally validation takes 1-2ms from a few spot checks on examples.

TODO:
- [x] Work out kinks with `webpack` build - can't find generated `.js` modules at present.
- [ ] Determine if we can lazy-load the AJV validator code, as after optimisation it's still ~400KB (before gzip). Maybe consider a distinct module?
- [ ] Get all tests to pass (some are failing validation!).
- [ ] Consider moving general code (e.g. `generate-ajv.js` and the error reduction code in `AgChartV2.ts`) to somewhere it could be re-used?

### DIFF
--- a/charts-packages/ag-charts-community/generate-ajv.js
+++ b/charts-packages/ag-charts-community/generate-ajv.js
@@ -1,9 +1,10 @@
 const fs = require('fs');
 const { join, resolve } = require('path');
 const Ajv = require('ajv');
+const ajvKeywords = require('ajv-keywords');
 const standaloneCode = require('ajv/dist/standalone').default;
 const tjs = require('typescript-json-schema');
-const uglifyjs = require('uglify-js');
+const terser = require('terser');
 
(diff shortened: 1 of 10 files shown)


seanlandsman:
Looks fine to me - changes look self contained

PR #5049, opened 2022-04-04
AG-6565 retain column width when toggling pivot mode

eddking:
This follows from the previous PR for AG-6565, since column widths were not preserved on flex columns if manually changed


gportela85:
This is not a bug, and it does not require this PR. The problem is that `defaultColDef` is setting `flex` instead of `initialFlex`. It is the same issue as https://ag-grid.atlassian.net/browse/AG-6525 that was rejected.

PR #5054, opened 2022-04-05
AG-1228 - Aggregation filtering docs

AndrewGlazier:


grid-packages/ag-grid-docs/documentation/doc-pages/aggregation-filtering/examples/agg-filtering-group/main.ts
```diff
@@ -28,7 +20,5 @@ document.addEventListener('DOMContentLoaded', () => {
   const gridDiv = document.querySelector<HTMLElement>('#myGrid')!
   new Grid(gridDiv, gridOptions)
 
-  fetch('https://www.ag-grid.com/example-assets/olympic-winners.json')
```
seanlandsman, line 31: Is the data in getData() sufficient for this example? I assume olympic-winners.json is larger so perhaps more useful?

grid-packages/ag-grid-docs/documentation/doc-pages/aggregation-filtering/examples/suppress-filtered-only/main.ts
```diff
@@ -1,34 +1,24 @@
...
     flex: 1,
     resizable: true,
-    floatingFilter: true,
   },
+  groupDefaultExpanded: -1,
   suppressAggFilteredOnly: true,
-  autoGroupColumnDef: {
-    headerName: 'Country',
-    field: 'athlete',
-  },
```
seanlandsman, line 23: Without this auto group col is this example substantially different in behaviour? This might be fine and the intent though...

grid-packages/ag-grid-docs/documentation/doc-pages/aggregation-filtering/examples/agg-filtering-group/main.ts
AndrewGlazier: It's helpful here to have a smaller data set as I want the group rows expanded by default, but not to be bogged down in the entries of just one group (I'm trying to make both the leaf and group nodes quite visible so the users can see the side effects of filtering on both)

grid-packages/ag-grid-docs/documentation/doc-pages/aggregation-filtering/examples/filters/main.ts
```diff
@@ -1,33 +1,23 @@
...
+  columnDefs: [
+    { field: 'country', rowGroup: true, hide: true },
+    { field: 'athlete' },
+    { field: 'year', filter: 'agNumberColumnFilter', floatingFilter: true },
+    { field: 'total', aggFunc: 'sum' },
+  ],
   defaultColDef: {
     flex: 1,
     resizable: true,
-    floatingFilter: true,
```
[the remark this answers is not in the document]
AndrewGlazier: A little, but my thoughts are to direct users specifically to where the feature is applied and hopefully help them recognise that the feature does not exist on every column, just specifically the aggregations

grid-packages/ag-grid-docs/documentation/doc-pages/aggregation-filtering/examples/suppress-filtered-only/main.ts
AndrewGlazier: I wouldn't say it's substantially different, as I've moved the athlete field to a column of it's own. It does behave slightly differently to the original example but as the focus is filtering rather than the specific grouping behaviour I think it still fills the requirement

PR #5076, opened 2022-04-08
AG-6647 - Fix Pie Chart initial resize.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6647

Attempts to fix the initial rendering of pie charts for a couple of frames at the wrong size.

Root causes:
- `ChartTheme` was setting `width` and `height` properties, despite `autoSize: true` being the intrinsic default - this triggered multiple re-evaluations during option application due to effectively flip-flopping on whether we're rendering a fixed size canvas or one tied to the containing element dimensions.
- There was a race between two triggers for `Scene.render()` calls:
  - Assignment of `scene.root = root` in the `Chart` constructor was triggering a rendering cycle before the size of the container was established as a side-effect of the `this.autoSize = true` assignment, which was racing with...
  - First callback by the `SizeMonitor.observe()` which was being called asynchrnoulsy, and triggering another rendering cycle.

Fixes:
- Remove hardcoded `width` and `height` from the base theme to prevent spurious `autoSize` flip-flopping.
- Re-orders the call to `this.autoSize = true` vs. `scene.root = root` so canvas size can be established before the first render - this is assisted by....
- Forcing the first `SizeMonitor.observe()` callback to happen synchronously.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
+++ b/charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
@@ -189,9 +189,6 @@ export class ChartTheme {
 
     private static getChartDefaults() {
         return {
-            width: 600,
-            height: 300,
-            autoSize: true,
             background: {
                 visible: true,
                 fill: 'white'
--- a/charts-packages/ag-charts-community/src/chart/chart.ts
+++ b/charts-packages/ag-charts-community/src/chart/chart.ts
@@ -407,9 +407,9 @@ export abstract class Chart extends Observable {
 
         const scene = new Scene(document);
         this.scene = scene;
+        this.autoSize = true; // Triggers width/height calc - needs to happen before root group assignment.
         scene.root = root;
         scene.container = element;
-        this.autoSize = true;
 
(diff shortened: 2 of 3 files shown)


charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -189,9 +189,6 @@ export class ChartTheme {
 
     private static getChartDefaults() {
         return {
-            width: 600,
-            height: 300,
-            autoSize: true,
```
manapeirov, line 194: What would happen if the user sets autoSize to false but then doesn't provide width/height, do we use the defaults on the chart instance? 

I think I remember wanting to remove the width/height properties from the theme before but there were some restrictions on the integrated charts side, not sure what the issue was exactly

alantreadway: I think you are right to flag this as a risk - but in terms of doing the right thing, IMO if you are overriding the default `autoSize: true` then you must be explicitly setting `width` and `height`, I don't really see (from an API-contract perspective) much wiggle room on this?

PR #5080, opened 2022-04-11
AG-6119 - Highlight Charts legend items are clickable

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6119

Fix pointer rendering when hovering over a legend item - we now use the the `pointer` cursor type (aka link-hover pointer type).

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/chart.ts
+++ b/charts-packages/ag-charts-community/src/chart/chart.ts
@@ -1199,6 +1199,7 @@ export abstract class Chart extends Observable {
     }
 
     private pointerInsideLegend = false;
+    private pointerOverLegendDatum = false;
     private handleLegendMouseMove(event: MouseEvent) {
         if (!this.legend.enabled) {
             return;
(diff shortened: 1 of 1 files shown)


charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1208,20 +1209,26 @@ export abstract class Chart extends Observable {
         const datum = this.legend.getDatumForPoint(offsetX, offsetY);
 
         const pointerInsideLegend = this.legendBBox.containsPoint(offsetX, offsetY);
-        if (pointerInsideLegend) {
-            if (!this.pointerInsideLegend) {
-                this.pointerInsideLegend = true;
-            }
-        } else if (this.pointerInsideLegend) {
+        const pointerOverLegendDatum = pointerInsideLegend && datum != null;
```
manapeirov, line 1212: nit: in charts, `!== undefined` is commonly used for almost all of these checks, might be better to have the same for consistency? or could just have `&& datum`

manapeirov (approved):

charts-packages/ag-charts-community/src/chart/chart.ts
alantreadway: Fixed

PR #5088, opened 2022-04-12
Fix errors due to this.cellComp being undefined in refreshCell

### LINKED ISSUE #5085: this.cellComp is undefined in CellCtrl.createCellRendererParams with AgGridReact
```
```

I've been working on upgrading the version of ag-grid used by the open source mlflow project from 25.3.0 to 27.1.0, and trying to improve performance when reloading data or loading more data by implementing the `getRowId` method.

However, I'm getting errors when updating the row data within ag-grid in `CellCtrl.createCellRendererParams` because `this.cellComp` is undefined. The full stack trace is:

```
TypeError: Cannot read properties of undefined (reading 'getParentOfValue')
	at CellCtrl.createCellRendererParams (cellCtrl.js:470:1)
	at CellCtrl.showValue (cellCtrl.js:215:1)
	at CellCtrl.refreshCell (cellCtrl.js:564:1)
	at rowCtrl.js:502:1
	at Array.forEach (<anonymous>)
	at RowCtrl.onRowNodeDataChanged (rowCtrl.js:501:1)
	at eventService.js:104:1
	at Set.forEach (<anonymous>)
	at processEventListeners (eventService.js:99:1)
	at EventService.dispatchToListeners (eventService.js:109:1) "Uncaught TypeError: Cannot read properties of undefined (reading 'getParentOfValue')"
```

I've tried to reproduce this with a minimal example but without any luck. And the error isn't consistently reproducible but will happen randomly, so possibly there's a race condition somewhere. The full code I'm using can be seen at https://github.com/adamreeve/mlflow/blob/d9adb6a9c7574c97b194dc193ebb8bc1a2f2fb4b/mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js

Does this look like I'm doing something wrong or is this a bug within ag-grid?

I can see comments elsewhere in `CellCtrl` stating that cellComp may not be set yet due to async in React (https://github.com/ag-grid/ag-grid/blob/6647b6a84307834a9d72f11baa5e988c03b6f162/community-modules/core/src/ts/rendering/cell/cellCtrl.ts#L648-L649), so I think the correct  fix might be to just exit out of `CellCtrl.refreshCell` if `this.cellComp` is not set. (Exiting early in `createCellRendererParams` or `showValue` isn't enough as then the same error is raised in `applyUserStyles`). This isn't an area I'm very familiar with but I'm happy to open a PR with that change if you'd like.

**Current behavior** `Cannot read properties of undefined` error when updating row data.

**Expected behavior** Not crashing.

Fedora Linux 35, using a React app based on craco.

* **AG Grid version:** 27.1.0

* **Browser:** Chrome 100 (`Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.88 Safari/537.36`)

* **Language:** ES5

adamreeve:
Fixes #5085.

I'm not familiar with the ag-grid codebase, but this fixes the crashes and seems to work correctly as far as I can tell (although I tested this by making the corresponding change in the generated JavaScript, I haven't got the build working).

### DIFF
--- a/community-modules/core/src/ts/rendering/cell/cellCtrl.ts
+++ b/community-modules/core/src/ts/rendering/cell/cellCtrl.ts
@@ -645,10 +645,6 @@ export class CellCtrl extends BeanStub {
     }
 
     public onCellChanged(event: CellChangedEvent): void {
-        // because of async in React, the cellComp may not be set yet, if no cellComp then we are
-        // yet to initialise the cell, so no need to refresh.
-        if (!this.cellComp) { return; }
-
         const eventImpactsThisCell = event.column === this.column;
 
         if (eventImpactsThisCell) {
@@ -663,6 +659,10 @@ export class CellCtrl extends BeanStub {
     // + rowCtrl: api refreshCells() {animate: true/false}
     // + rowRenderer: api softRefreshView() {}
     public refreshCell(params?: { suppressFlash?: boolean, newData?: boolean, forceRefresh?: boolean; }) {
+        // because of async in React, the cellComp may not be set yet, if no cellComp then we are
+        // yet to initialise the cell, so no need to refresh.
+        if (!this.cellComp) { return; }
+
         // if we are in the middle of 'stopEditing', then we don't refresh here, as refresh gets called explicitly
         if (this.suppressRefreshCell || this.editing) { return; }
 


adamreeve:
Closing as not needed, this was already fixed by 516e5007a2aa09f2f1f26ca75cf9e7ab01745346

PR #5089, opened 2022-04-13
AG-6674 - Fix sparklines label `fontSize` type

manapeirov:
https://ag-grid.atlassian.net/browse/AG-6674

Correct `fontSize` type to be consistent with the implementation AND standalone charts.


alantreadway (approved):
LGTM - but don't merge until code-freeze is off.

PR #5102, opened 2022-04-19
fix peerDependencies

shadoworion:


StephenCooper:
Thanks for reporting. We will be releasing a patch for this later today.

PR #5104, opened 2022-04-19
Fix ag-grid-react peer dependencies for 27.2.0

ghmeier:
At the moment, ag-grid-react@27.2.0 cannot be installed with @ag-grid-community/core@27.2.0 because of the minor version required in peer depencencies. It should be updated to match the dev dependencies and allow all ag-grid packages to be installed at the latest version.


StephenCooper:
Thanks for reporting this issue. We will release a corrected patch later today.

PR #5112, opened 2022-04-22
Ag 6207/continuous axis labels

manapeirov:

PR #5119, opened 2022-04-26
AG-5243 - Allow setting axis label auto rotate to avoid category axis…

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-5243

The intent is to allow the user to specify an `autoRotate` axis label property, which is `true` by default and applies to horizontal category axes (in the X direction).

When `autoRotate` is `true`, rotate the labels to prevent overlap when the chart width/height is too small, or when labels are too large. This applies to `category` axis type and is only relevant for horizontal axes as this is the case were the labels would otherwise be placed parallel to the axis line.

The approach is to modify the `parallelLabels` and `regularFlipFlag` to alter the orientation of the lables. If the total length of all labels in the new orientation still exceeds the available axis range, half the labels are removed iteratively until the length of visible labels is not larger than the available space. This change was made as part of https://github.com/ag-grid/ag-grid/pull/5112.

Note that this also only applies if the user has not explicitly specified a label rotation angle in the options.

### DIFF
--- a/charts-packages/ag-charts-community/src/axis.ts
+++ b/charts-packages/ag-charts-community/src/axis.ts
@@ -92,9 +92,14 @@ export class AxisLabel {
     rotation: number = 0;
 
     /**
+     * By default, if the category axis labels collide, they are rotated so that they are positioned perpendicular to the horizontal axis line to reduce overlap.
+     */
+    autoRotate: boolean = true;
+
+    /**
      * By default labels and ticks are positioned to the left of the axis line.
      * `true` positions the labels to the right of the axis line.
-     * However, if the axis is rotated, its easier to think in terms
+     * However, if the axis is rotated, it's easier to think in terms
      * of this side or the opposite side, rather than left and right.
      * We use the term `mirror` for conciseness, although it's not
      * true mirroring - for example, when a label is rotated, so that
@@ -387,8 +392,8 @@ export class Axis<S extends Scale<D, number>, D = any> {
         const requestedRangeMin = Math.min(requestedRange[0], requestedRange[1]);
         const requestedRangeMax = Math.max(requestedRange[0], requestedRange[1]);
         const rotation = toRadians(this.rotation);
-        const parallelLabels = label.parallel;
         const labelRotation = normalizeAngle360(toRadians(label.rotation));
+        let parallelLabels = label.parallel;
 
         group.translationX = this.translation.x;
         group.translationY = this.translation.y;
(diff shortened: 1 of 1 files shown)


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -91,10 +91,15 @@ export class AxisLabel {
      */
     rotation: number = 0;
 
+    /**
+     * By default, if the category axis labels collide, they are rotated so that they are positioned perpendicular to the horizontal axis line to reduce overlap.
+     */
+    autoRotate: boolean = true;
```
alantreadway, line 97: Do we need to document this in `agChartOptions.ts` (and `iAgChartOptions.ts`)? Should the default be specified here on in `chartTheme.ts`, where many of the defaults are defined?

alantreadway (commented):
Maybe we could hold-off merging these changes until we have a solid foundation for testing them in #5120 ?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -499,52 +504,81 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
+            ? parallelFlipFlag * Math.PI / 2
+            : (regularFlipFlag === -1 ? Math.PI : 0);
+
+        labelSelection.each(label => {
+            label.textBaseline = parallelLabels && !labelRotation
+                ? (sideFlag * parallelFlipFlag === -1 ? 'hanging' : 'bottom')
+                : 'middle';
+            label.textAlign = parallelLabels
+                ? labelRotation ? (sideFlag * alignFlag === -1 ? 'end' : 'start') : 'center'
+                : sideFlag * regularFlipFlag === -1 ? 'end' : 'start';
```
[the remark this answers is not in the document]
manapeirov: done

charts-packages/ag-charts-community/src/axis.ts
manapeirov: Done

charts-packages/ag-charts-community/src/axis.ts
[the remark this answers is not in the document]
manapeirov: Done

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -91,10 +91,15 @@ export class AxisLabel {
      */
     rotation: number = 0;
 
+    /**
+     * By default, if the category axis labels collide, they are rotated so that they are positioned perpendicular to the horizontal axis line to reduce overlap.
+     */
+    autoRotate: boolean = false;
```
alantreadway, line 97: Just debating if we should include the `iAgChartOptions.ts` and `agChartOptions.ts` addition of this field in this PR.

If the Integrated Charts save includes this field with no further changes, then irrespective of whether it's documented or not, it's public - could you confirm if this is the case or not?

alantreadway: Discussed offline, the Integrated Charts save functionality WILL include `autoRotate` as it's part of the theme defaults - with that in mind, I think sticking with the current approach in the PR is fine, but let's flag it with Rob.

alantreadway (approved):
LGTM

charts-packages/ag-charts-community/src/axis.ts
manapeirov: The `autoRotate` property would be present in the `chartModel` (for integrated charts save and restore) if we have it present in the chartTheme default options (which we currently do). We could remove it from `chartTheme.ts` and specify the default just here and in the axis`update` method. 

Discussed offline and decided to leave it for now.

PR #5120, opened 2022-04-26
AG-6719 - Add Chart Axis test suite

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6719

As part of the Chart Axis Label Improvements epic, we want to add some visual test coverage for the problematic cases we aim to address. This will allow us to track progress, and later ensure we don't regress.

Cases covered:
- Category axis
  - Variable length text
  - Uniform length text
- Grouped category axis.
- Number axis.
- Time axis.
- Manually rotated permutations of all above cases.

Bonus:
- Refactors out common test `Canvas` image extraction code, and adds ability to specify a bounding box for comparisons.


charts-packages/ag-charts-community/src/chart/axis/axisExamples.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    let parent: typeof axis.group.parent = axis.group;
+    while (parent != null) {
+        bbox = parent.matrix.transformBBox(bbox);
+
+        parent = parent.parent;
+    }
+
+    const { x, y, width, height } = bbox;
+    return { x, y, width, height };
+}
```
alantreadway, line 55: @manapeirov your input on the approach here would be great - after some experimentation, this seemed the most reliable way to work out what the bounding-box on the `Canvas` an axis occupied, in the coordinate-space of the `Canvas` itself?

manapeirov: Looks good!

manapeirov:
Really enjoyed looking at this PR, learned so much from it! Might be good to add some tests for axes in the `top` and `right` positions because the labels will be placed on the opposite side of the axis line (`mirrored`).

manapeirov (approved):

alantreadway:
> Might be good to add some tests for axes in the `top` and `right` positions because the labels will be placed on the opposite side of the axis line (`mirrored`).

I'll take a look at adding these as a follow-up perhaps, we can probably automate generation of these cases too.

PR #5121, opened 2022-04-27
AG-6719 - Add flipped axes test cases

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6719

As per feedback on #5120, adds flipped axis snapshots and automated generation of these derived cases.

Bonus:
- Adds NPM script to force snapshot update.


manapeirov (approved):
Looks great, not sure if grouped category axis case makes sense being at the top, seems like it doesn't render correctly. charts-packages/ag-charts-community/src/chart/axis/__image_snapshots__/axis-examples-test-ts-axis-examples-for-grouped-category-axis-flip-it-should-render-to-canvas-as-expected-1-snap.png

alantreadway:
> not sure if grouped category axis case makes sense being at the top, seems like it doesn't render correctly

Agreed it looks quite borked - it's technically a possible configuration, so having it for completeness and visibility makes sense to me.

I'll raise a specific follow-up JIRA for us to look into this case.

PR #5127, opened 2022-04-28
AG-6677 - Better null/undefined handling for scatter series charts.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6677

Completely removes scatter series datum from the plotted chart which are missing either an xKey or yKey value.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
+++ b/charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
@@ -16,7 +16,7 @@ import { Text } from "../../../scene/shape/text";
 import { HdpiCanvas } from "../../../canvas/hdpiCanvas";
 import { Marker } from "../../marker/marker";
 import { MeasuredLabel, PlacedLabel, PointLabelDatum } from "../../../util/labelPlacement";
-import { isContinuous } from "../../../util/value";
+import { isContinuous, isNumber } from "../../../util/value";
 
 interface ScatterNodeDatum extends SeriesNodeDatum {
     readonly point: {
@@ -58,6 +58,7 @@ export class ScatterSeries extends CartesianSeries {
     private yDomain: number[] = [];
     private xData: any[] = [];
     private yData: any[] = [];
+    private datum: any[] = [];
     private sizeData: number[] = [];
     private sizeScale = new LinearScale();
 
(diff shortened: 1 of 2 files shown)


charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -16,7 +16,7 @@ import { Text } from "../../../scene/shape/text";
 import { HdpiCanvas } from "../../../canvas/hdpiCanvas";
 import { Marker } from "../../marker/marker";
 import { MeasuredLabel, PlacedLabel, PointLabelDatum } from "../../../util/labelPlacement";
-import { isContinuous } from "../../../util/value";
+import { isContinuous, isNumber } from "../../../util/value";
```
manapeirov, line 19: Is this being used anywhere?

charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -199,17 +200,25 @@ export class ScatterSeries extends CartesianSeries {
...
-        this.xData = data.map(d => d[xKey]);
-        this.yData = data.map(d => d[yKey]);
+        const xyData = data.map(d => ({
+            x: d[xKey],
+            y: d[yKey],
+            datum: d,
+            size: sizeKey ? d[sizeKey] : null,
+            label: String(labelKey ? d[labelKey] : null),
+        }))
+            .filter(({x, y}) => x != null && x !== NaN && y != null && y !== NaN);
```
manapeirov, line 210: We have some utility functions which are used to validate data, for continuous data, there's the `isContinuous` function in value.ts. Could that be used here?

manapeirov: We also have `checkDomainXY` in cartesianSeries.ts which is for data validation

alantreadway: > We have some utility functions which are used to validate data, for continuous data, there's the `isContinuous` function in value.ts. Could that be used here?

Based upon the implementation, it won't handle non-numeric data:
```
export function isContinuous(value: any): boolean {
    return isNumeric(value) || isDate(value);
}
```

This is an example that fails to render if we don't allow non-numeric values:
https://ag-grid.com/javascript-charts/gallery/bubble-with-categories/

alantreadway: I'll give `checkDomainXY` a try, that looks like it will handle the above example.

manapeirov: I think in the other series types for example Line Series, we check the type of the axis, if it's continuous we use the `isContinuous` function, if it's a category we use the `isDiscrete` helper function. I added my own function `checkDatum` in cartesianSeries.ts because I needed the return values to be in a different format to to return type of `checkDomainXY` in the case of lineSeries data validation

charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
alantreadway: Good spot, removed.

charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
alantreadway: Ok - I went with `checkDatum` in the end as it doesn't construct redundant arrays vs. `checkDomainXY`. I also reworked the code to minimise the number of extra array creations (the code is now much closer to the original code, except the introduction of the `this.datum` intermediate array.

alantreadway:
PTAL @manapeirov

charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -198,14 +199,19 @@ export class ScatterSeries extends CartesianSeries {
...
 
         const data = xKey && yKey && this.data ? this.data : [];
+        const xScale = xAxis.scale;
+        const yScale = yAxis.scale;
+        const isContinuousX = xScale instanceof ContinuousScale;
+        const isContinuousY = yScale instanceof ContinuousScale;
 
-        this.xData = data.map(d => d[xKey]);
-        this.yData = data.map(d => d[yKey]);
+        this.datum = data.filter(d => this.checkDatum(d[xKey], isContinuousX) != null && this.checkDatum(d[yKey], isContinuousY) != null);
```
manapeirov, line 207: nit: for consistency with the rest of the checks in this file could we change it to be `!== undefined`

manapeirov (approved):
Looks good thanks Alan!

charts-packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
alantreadway: Ok - fixed

manapeirov (approved):

PR #5131, opened 2022-04-29
Ag 5243/category axis auto rotate

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-6633.

- Handle navigator zoom correctly.
- Add ‘padding’ in calculation, to make spacing a higher priority.
- Make Colliding labels detection more intelligent - auto rotate should be applied to category axes with variable width labels ‘if any of the labels width > the average tick gap’.
- Switch `autoRotate: true` to 45 degree rotation.


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -597,6 +561,56 @@ export class Axis<S extends Scale<D, number>, D = any> {
             }
         }
 
+        groupSelection.selectByTag<Line>(Tags.Tick)
+        .each((line, _, index) => {
```
alantreadway, line 565: Nit: Indentation is wrong here - apply `prettier`?

alantreadway (approved):
LGTM other than one nit.

charts-packages/ag-charts-community/src/axis.ts
manapeirov: done

PR #5141, opened 2022-05-05
Ag 6693/tooltips hidden series

manapeirov:
This PR is for: https://ag-grid.atlassian.net/browse/AG-6693

The intent is to not display tooltips for series which have been hidden via the legend. This applies to both area and pie series.

The pie series uses `pickNode` which recursively checks if the given point is inside any of the pie sectors (it is not using `tooltipTracking` to pick the closest node, the pointer has to be inside the node). If all the sectors have been hidden via the legend, the pie series `group.visible` will be set to `false`. Hit testing will return the last node even if it is not visible as `group.render()` is not invoked if the group visibility is false, which means `updatePath()` is not invoked and hence the node data is not updated, which is why a false hit is returned. A check has been added to see if the `series.group.visible` is `false`, if so, we prevent hit testing all together, rather than hit testing and getting a false hit.

This is not sufficient for the area series as the area series group visibility is not controlled based on the `seriesItemEnabled` map (due to certain requirements specific to area series). For the area series, the approach is to check if the item that has been hit is enabled via the `seriesItemEnabled` map, and return an empty html string if it's not enabled.


alantreadway (approved):
LGTM - nice sleuthing!

PR #5146, opened 2022-05-09
Ag 5243/category axis auto rotate

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-5243 and https://ag-grid.atlassian.net/browse/AG-6207.

The intent is to apply axis label rotation automatically and remove labels in order to avoid collisions when the chart width/height is too small, or when labels are too large. This applies to all axis types.

The approach is to have an `autoRotate` property which can be set to `true`, `false` or any rotation angle which can be used to automatically rotate the labels when they collide. This rotation is applied first and then the labels are removed if the rotation is not sufficient to prevent collisions. `autoRotate` can be applied on both Y and X axes.

The default autoRotate value has been set to 335 for consistency with previous integrated charts behaviour.
AutoRotate behaves exactly as the `rotation` property does, but will only take effect if `rotation` has not been specified. 

Changes to the chart axis range via the navigator will trigger a layout if the orientation of the labels has changed.

When removing labels, the first and last labels are not removed.


charts-packages/ag-charts-community/src/chart/navigator/navigator.ts
```diff
@@ -98,7 +98,11 @@ export class Navigator {
                     clipSeries = true;
                 }
                 axis.visibleRange = [min, max];
+                const oldLabelAutoRotated = axis.labelAutoRotated;
                 axis.update();
+                if (axis.labelAutoRotated !== oldLabelAutoRotated) {
```
alantreadway, line 103: Is it possible to embed this logic in the `set labelAutoRotated()` method, for consistency with how other properties ripple up to `Chart.layoutPending`?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -498,67 +501,63 @@ export class Axis<S extends Scale<D, number>, D = any> {
 
         let useWidth = parallelLabels; // When the labels are parallel to the axis line, use the width of the text to calculate the total length of all labels
 
-        if (labelRotation) {
-            // If the specified label rotation angle results in a non-parallel orientation, use the height of the texts to calculate the total length of all labels
+        let totalLabelLength = calculateLabelsLength(labelBboxes, useWidth);
```
alantreadway, line 504: I'm not seeing where we check for any label being larger than the average tick width?

manapeirov: It happens when we invoke `calculateLabelsLength` (https://github.com/ag-grid/ag-grid/pull/5146/files#diff-fef1ca1efb17bd1d7f66e6004300cccb5162017f16da14a7f56a5941404304e6R495)

charts-packages/ag-charts-community/src/chart/navigator/navigator.ts
manapeirov: I think we would need to pass an instance of the chart to axis which means we'd have to add a lot more changes, if you prefer this though I could do that

alantreadway: In light of the reactive property rework, if we don't have a `Chart` reference in `Axis` then this is probably ok since this code will just be removed or replaced anyway.

alantreadway (approved):

PR #5149, opened 2022-05-10
Ag 3559/grouped category axis

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-3559.

The intent is to remove axis labels automatically in order to avoid collisions when the chart width/height is too small, or when labels are too large. This applies to the grouped category axis type.

The approach is to check the bbox of the label text and if it exceeds the available space set the visibility to `false`.


charts-packages/ag-charts-community/src/chart/axis/groupedCategoryAxis.ts
```diff
@@ -309,12 +312,24 @@ export class GroupedCategoryAxis extends ChartAxis<BandScale<string | number>> {
...
-                if (isHorizontal) {
-                    label.rotation = autoRotation;
+                const availableRange = datum.leafCount * bandwidth;
+                const bbox = labelBBoxes.get(label.id);
+
+                if (bbox && bbox.width > availableRange) {
+                    label.visible = false;
                 } else {
-                    label.rotation = -Math.PI / 2;
+                    if (isHorizontal) {
```
alantreadway, line 328: Nit: Merge this with the enclosing `else` to remove a level of nesting.

alantreadway (approved):
LGTM

PR #5157, opened 2022-05-12
Fix example documentation for cell editor stopping propagation.

sorohan:
"keyCode" variable was renamed to "key" in a previous commit, but a few instances of "keyCode" remained.


sorohan:
Did you need an issue created for this? It's just a small documentation error I noticed.

StephenCooper:
Thanks!

PR #5161, opened 2022-05-12
Ag 6207/continuous axis labels

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-6207 and https://ag-grid.atlassian.net/browse/AG-5243

The approach for handling overlapping labels due to reduced chart size or large labels has been altered slightly for the continuous axis.

We dynamically set the tick count to a number between 2 and 5 depending on the available axis range, which means we get a dynamic "nice" domain and the labels displayed will initially be determined by the tick count. This should avoid most overlaps for the continuous axis type. If we determine that label overlapping is still likely, then label hiding logic will be triggered.

The other changes in this PR include minor bug fixes and improvements.
- Fix autoRotation being applied when labelRotation has been set to `0` by the user, though the user could have bypassed this by setting `autoRotate` to `false `.
- Consider labels will protrude the axis line (first and last)
- Keep first and last labels for continuous axes only if the tick count has not been specified, if it has been specified, keeping the last label is likely to cause a collision

N.B. even with the nice domain, it looks like with the time axis the tick intervals are not always even and it is not guaranteed to have a grid line for the last tick, you can see this in the snapshot images


charts-packages/ag-charts-community/src/chart/chartAxis.ts
```diff
@@ -48,6 +49,16 @@ export class ChartAxis<S extends Scale<any, number> = Scale<any, number>> extend
...
     }
 
+    /**
+     * For continuous axes, if tick count has not been specified, set the number of ticks based on the available range
+     */
+    calculateTickCount(availableRange: number): void {
+        if (!(this.scale instanceof ContinuousScale)) { return; } // Discrete axes do not require a tick count, the tick count will be the number of categories
+
+        const tickInterval = 70; // Approximate number of pixels to allocate for each tick
+        this._calculatedTickCount = this.tick.count || Math.min(Math.max(2, Math.floor(availableRange / tickInterval)), 5); // We don't want less than 2 ticks or more than 5 ticks (this will be a hint, the actual tick count after `nice` is applied might be different)
```
alantreadway, line 59: I feel like the limit of `5` should be a bit more dynamic perhaps?

E.g. perhaps we allow a tick per 100px of availableRange?

alantreadway (commented):
One comment, and we seem to have lost the top-line on three examples - which seems counter to the intent of the change?

charts-packages/ag-charts-community/src/chart/chartAxis.ts
alantreadway: I pushed a change to remove the upper limit - since we're aiming for a specific pixels / tick ration, having a hard-coded cap of 5 doesn't make sense for me. If for larger charts things don't look right, tuning the `70` target might be the way forward instead?

alantreadway:
Merging as the change is looking good, we can follow-up if the removal of the calculated tick count limit of `5` is a problem?

PR #5165, opened 2022-05-13
AG-6207 - Disable new tick algorithm for time axes

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6207

We prefer the previous time-axis tick calculation behaviour, so we're disabling the new tick algorithm for them.


manapeirov (approved):
Looks good

PR #5198, opened 2022-05-18
AG-6786 - Remove redundant boilerplate change-detection get/set.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6786

Replace repeated change-detection `set`/`get` pattern with a simple field declarataion and an annotation to implement generic (but functionally equivalent) change detection processing.

### DIFF
--- a/charts-packages/ag-charts-community/src/scene/node.ts
+++ b/charts-packages/ag-charts-community/src/scene/node.ts
@@ -21,6 +21,47 @@ export enum RedrawType {
     MAJOR, // Significant change in rendering.
 }
 
+export function SceneChangeDetection(opts?: {
+    redraw?: RedrawType,
+    type?: 'normal' | 'transform' | 'path' | 'font',
+    transform?: (o: any) => any,
+    changeCb?: (o: any) => any,
+}) {
+    const { redraw = RedrawType.TRIVIAL, type = 'normal', changeCb, transform } = opts || {};
+
+    return function (target: any, key: string) {
+        // `target` is either a constructor (static member) or prototype (instance member)
+        const privateKey = `__${key}`;
+
+        if (!target[key]) {
+            Object.defineProperty(target, key, {
+                set: function (value: any) {
+                    const oldValue = this[privateKey];
+                    if (transform) {
+                        value = transform(value);
+                    }
+                    if (value !== oldValue) {
+                        this[privateKey] = value;
+                        if (type === 'normal') {
+                            this.markDirty(redraw);
+                        } else if (type === 'transform') {
+                            this.markDirtyTransform(redraw);
+                        }
+                        if (changeCb) {
+                            changeCb(this);
+                        }
+                    }
+                },
+                get: function (): any {
+                    return this[privateKey];
+                },
+                enumerable: true,
+                configurable: true
+            });
+        }
+    }
+}
+
 /**
  * Abstract scene graph node.
  * Each node can have zero or one parent and belong to zero or one scene.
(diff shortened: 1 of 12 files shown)


charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -565,32 +531,20 @@ export abstract class Node { // Don't confuse with `window.Node`.
...
-    set visible(value: boolean) {
-        if (this._visible !== value) {
-            this._visible = value;
-            this.markDirty(RedrawType.MINOR);
-        }
-    }
-    get visible(): boolean {
-        return this._visible;
-    }
+    @SceneChangeDetection({ redraw: RedrawType.MAJOR })
```
manapeirov, line 534: Is the redraw type for this major or minor?

charts-packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -90,38 +90,14 @@ export abstract class Shape extends Node {
...
-    set fillOpacity(value: number) {
-        if (this._fillOpacity !== value) {
-            this._fillOpacity = value;
-            this.markDirty(RedrawType.MINOR);
-        }
-    }
-    get fillOpacity(): number {
-        return this._fillOpacity;
-    }
+    @SceneChangeDetection()
```
manapeirov, line 93: Do these need redrawType?

charts-packages/ag-charts-community/src/scene/shape/path.ts
```diff
@@ -1,6 +1,25 @@
...
+export function ScenePathChangeDetection(opts?: {
+    redraw?: RedrawType,
+    changeCb?: (t: any) => any,
+}) {
+    const { redraw = RedrawType.MAJOR, changeCb: optChangeCb } = opts || {};
+
+    const changeCb = (o: any) => {
+        if (!o._dirtyPath) {
+            o._dirtyPath = true;
+            o.markDirty(redraw);
```
manapeirov, line 14: how come we need to do markDirty, does this not already happen in SceneChangeDetecion?

charts-packages/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -20,159 +38,59 @@ export class Text extends Shape {
...
-    set x(value: number) {
-        if (this._x !== value) {
-            this._x = value;
-            this.markDirty(RedrawType.MAJOR);
-        }
-    }
-    get x(): number {
-        return this._x;
-    }
+    @SceneChangeDetection()
```
manapeirov, line 41: how come this isn't  @SceneChangeDetection({ redraw: RedrawType.MAJOR })

manapeirov (commented):
Thanks for explaining this, changes look great. Just some questions for my understanding

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: I'm thinking that visibility on/off is a major event, since items behind it may need to be redrawn?

charts-packages/ag-charts-community/src/scene/shape/path.ts
alantreadway: No only for type `'normal'` or `'transform'` cases - the intent being that this is a more specialised case where we want to detect changes more granularly.

charts-packages/ag-charts-community/src/scene/shape/shape.ts
alantreadway: Good spot - fixed.

charts-packages/ag-charts-community/src/scene/shape/text.ts
alantreadway: Good spot - fixed.

alantreadway:
@manapeirov PTAL - applied rename that we discussed to improve readability.

manapeirov (approved):
Looks good :) thank you!

PR #5213, opened 2022-05-23
Ag 3838/bar series continuous axes

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-3838.

Previously for the Column and Bar series, one of the axes had to be a category axis with a band scale. The intent of this is to allow the Column/Bar series to be rendered using a continuous scale; time and number.

The approach is to add data validation and find the smallest data interval in `processData()`. As the data is represented by rectangles which have a width, a step is calculated using the smallest data interval, the domain length and the available range. The step must be a minimum of `1px` for the rects to be visible, which means there is a max number of steps, if this is exceeded, overlap may occur on the chart.

The scale range is modified by reducing one step length, this is to prevent the last item from being clipped.

Snapshot tests have been added.


charts-packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -330,22 +330,49 @@ export class BarSeries extends CartesianSeries {
...
-        if (!xAxis) {
+        const xAxis = this.getCategoryAxis();
+        const yAxis = this.getValueAxis();
+
+        if (!(xAxis && yAxis)) {
             return false;
         }
 
+        const setSmallestXInterval = (curr: number, prev: number) => {
+            if (this.smallestDataInterval == undefined) {
```
alantreadway, line 346: Should this be `===`?

manapeirov: fixed

alantreadway (approved):

PR #5216, opened 2022-05-24
AG-6791 - Add Charts LineSeries Marker distinct rendering layer

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6791

Attempts to generally optimise:
- Rendering of `opacity !== 1` cases, where all scene-graph objects have the same opacity in a layer. (Helps with series highlighting cases where the only change is opacity)

Specifically optimises `LineSeries` as a spike into the approach we could apply to other series types:
- Moves highlighted marker rendering to its own layer `Group`, which contains all marker nodes in the highlighted state, with only the highlighted marker visible.
- Pushes series opacity changes up to the `Group` to benefit from layer opacity optimisation, when possible.

Bonus:
- More debugging options including rendering stats.

PR #5218, opened 2022-05-24
AG-6791 -Refactor Charts series opacity calculation.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6791

Refactor per-datum contract for `Series.getOpacity()` to allow simplification/optimisation of where opacity is applied.

This is a follow-up to #5216.


manapeirov (approved):
Looks good :)

PR #5227, opened 2022-05-26
AG-6820 - Optimise AreaSeries chart rendering.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6820

Introduces layered rendering for `AreaSeries`. This was more complex than for `LineSeries` since `AreaSeries` actually manages multiple data series.

Principle differences in approach:
- Adds a distinct layer for each data series within the `AreaSeries` to allow opacity/highlight optimisations.
- Splits all previous groups into sub-selections of new layer groups.
- Adds an abstraction for `Series.pickNode` so that `Series` sub-classes can override the default behaviour of all interactable nodes being in one group.
- Adds a highlight layer with all marker nodes in highlighted state (same pattern as `LineSeries`.
  - Since only one node is ever `visible === true` splitting into different groups is unnecessary.

Bonus:
- Adds `agChartsSceneDirtyTree` debug property that adds dirty-tree debugging output.
- Fixes scaling in rendering mode `composite` when device pixel ratio != 1.
- Fixed `getOpacity()` for multi-data-series `Series` instances.


charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -688,27 +764,39 @@ export class AreaSeries extends CartesianSeries {
...
+                    marker.tag = AreaSeriesTag.Marker;
+                });
             return updateMarkers.merge(enterMarkers);
         }
 
-        this.markerSelection = update(markerSelection);
-        this.highlightMarkerSelection = update(highlightMarkerSelection);
+        seriesGroup.markerSelection = update(markerSelection, data);
+        if (idx === 0) {
+            this.highlightMarkerSelection = update(highlightMarkerSelection, highlightData);
```
manapeirov, line 790: is it worth adding a separate function for updating `highlightMarkerSelection`?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -634,33 +706,37 @@ export class AreaSeries extends CartesianSeries {
         });
     }
 
-    private updateStrokeSelection(): void {
-        const updateStrokes = this.strokeSelection.setData(this.strokeSelectionData);
+    private updateStrokeSelection(idx: number): void {
+        const seriesGroup = this.seriesGroups[idx];
+        const { strokeSelection } = seriesGroup;
+        const updateStrokes = strokeSelection.setData(this.strokeSelectionData.slice(idx, idx + 1));
```
manapeirov, line 712: How come we can't do `this.strokeSelectionData[idx]`

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -589,26 +625,62 @@ export class AreaSeries extends CartesianSeries {
...
+                labelSelection: Selection.select(group).selectAllByTag<Text>(AreaSeriesTag.Label),
+                markerSelection: Selection.select(group).selectAllByTag<Marker>(AreaSeriesTag.Marker),
+            });
+        }
+    }
+
+    private updateFillSelection(idx: number): void {
+        const seriesGroup = this.seriesGroups[idx];
+        const { fillSelection } = seriesGroup;
+        const updateFills = fillSelection.setData(this.fillSelectionData.slice(idx, idx + 1));
```
manapeirov, line 663: does `this.fillSelectionData[idx]` not work here?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: My reading was that `setData()` wanted an array - so the choices are either:
- `setData([ this.strokeSelectionData[idx] ])`
- `setData(this.strokeSelectionData.slice(idx, idx + 1))`

Neither of which is particularly satisfying. The later could at least have a lower memory overhead potentially (it's just a view on an existing array - not sure if that optimisation is actually a thing though).

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: Ditto

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
manapeirov: oh so there's only one thing in the selection, in this case we don't need a selection for each series? as long as each series has its own group it will not re-render right? or does it require to have it's own selections too

manapeirov: As in, is it possible to still have just one fill selection, one stroke selection etc and select the seriesGroup and then select all fill Paths in all subgroups? (let me know if what I'm saying is gibberish)

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -399,7 +401,13 @@ export class AreaSeries extends CartesianSeries {
     }
 
     updateNodes() {
-        this.group.visible = this.visible && this.xData.length > 0 && this.yData.length > 0;
+        const visible = this.visible && this.xData.length > 0 && this.yData.length > 0;;
```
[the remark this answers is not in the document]
alantreadway: Fixed

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: Done

alantreadway:
@manapeirov PTAL?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: Pushed a change to remove the `Selection` in this case - there is now just a singular `Path` in both `fill` and `stroke` cases.

manapeirov (approved):

PR #5237, opened 2022-05-30
Update floatingFilter.js

ABUdhay:
This change will restrict exposing of onFloatingFilterChanged every first time we press Tab key from any of Text filter.


ABUdhay (commented):
Approved

PR #5249, opened 2022-06-06
AG-6820 - Further Chart rendering optimisations for larger data-sets

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6820

Attempts to address scaling issues with the Charts scene rendering which were identified whilst testing in the O(50K->150K) range. Focuses on line, area and scatter series as the types most likely to see use with very large data-sets:
- Switches to use of `OffscreenCanvas` + `ImageBitmap` APIs and bitmap copying of layers to main canvas for better rendering performance when a large number of points are rendered.
- Adds support to `Group` for an optimised dirty-check path, for groups where the number of visible/dirty elements is typically a low % of the total children. This is important for highlight series, where only one node will be visible / dirty at a time.
- Inlines more `@SceneChangeDetection` cases for better execution performance.
- Introduces path-level dirty detection, enabled in specific cases (Line and area series plots, which are complex single-paths that don't scale well from a change-detection/render perspective).

### DIFF
--- a/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
+++ b/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
@@ -8,6 +8,7 @@ export class HdpiCanvas {
     readonly document: Document;
     readonly element: HTMLCanvasElement;
     readonly context: CanvasRenderingContext2D;
+    readonly imageSource: HTMLCanvasElement;
 
     // The width/height attributes of the Canvas element default to
     // 300/150 according to w3.org.
@@ -21,6 +22,7 @@ export class HdpiCanvas {
     }) {
         this.document = document;
         this.element = document.createElement('canvas');
+        this.imageSource = this.element;
         this.context = this.element.getContext('2d')!;
 
         const { style } = this.element;
--- a/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
+++ b/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
@@ -22,8 +22,8 @@ export class HdpiCanvas {
     }) {
         this.document = document;
         this.element = document.createElement('canvas');
-        this.imageSource = this.element;
         this.context = this.element.getContext('2d')!;
+        this.imageSource = this.context.canvas;
 
         const { style } = this.element;
 
--- a/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
+++ b/charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
@@ -94,6 +94,10 @@ export class HdpiCanvas {
         Object.freeze(this);
     }
 
+    snapshot() {
+        // No-op for compatibility with HdpiOffscreenCanvas.
+    }
+
     clear() {
         this.context.save();
         this.context.resetTransform();
@@ -327,7 +331,7 @@ export class HdpiCanvas {
         return size;
     }
 
-    static overrideScale(ctx: CanvasRenderingContext2D, scale: number) {
+    static overrideScale(ctx: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D, scale: number) {
         let depth = 0;
         const overrides = {
             save() {
--- a/charts-packages/ag-charts-community/src/chart/chart.ts
+++ b/charts-packages/ag-charts-community/src/chart/chart.ts
@@ -1186,21 +1186,34 @@ export abstract class Chart extends Observable {
 
     private checkLegendClick(event: MouseEvent): boolean {
         const datum = this.legend.getDatumForPoint(event.offsetX, event.offsetY);
-
-        if (datum) {
-            const { id, itemId, enabled } = datum;
-            const series = find(this.series, (s) => s.id === id);
-
-            if (series) {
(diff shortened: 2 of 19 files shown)


charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -553,15 +555,17 @@ export abstract class Node { // Don't confuse with `window.Node`.
         return this._dirty;
     }
 
-    markClean(force = false) {
+    markClean({ force = false, recursive = true } = {}) {
```
alantreadway, line 558: Check if defaults are being picked up

charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
```diff
@@ -327,7 +331,7 @@ export class HdpiCanvas {
         return size;
     }
 
-    static overrideScale(ctx: CanvasRenderingContext2D, scale: number) {
+    static overrideScale(ctx: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D, scale: number) {
```
manapeirov, line 334: Can we use `CanvasTransform` here as well or is that not specific enough

charts-packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -297,18 +297,18 @@ export class LineSeries extends CartesianSeries {
...
-        this.nodeSelection = update(nodeSelection, nodeData);
-        this.highlightSelection = update(highlightSelection, nodeData);
+        const updateSelection = nodeSelection.setData(nodeData);
+        updateSelection.exit.remove();
+        const enterSelection = updateSelection.enter.append(Group);
+        enterSelection.append(MarkerShape);
+        enterSelection.append(Text);
+        this.nodeSelection = updateSelection.merge(enterSelection);
+    
+        const updateSelection2 = highlightSelection.setData(nodeData);
```
manapeirov, line 310: shall we call this `updateHighlightSelection`?

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1274,22 +1283,27 @@ export abstract class Chart extends Observable {
...
+            this.element.style.cursor = newSeries.cursor;
         }
 
         this.lastPick = newPick;
         this.highlightedDatum = datum;
 
-        seriesToUpdate.add(newPick?.datum?.series);
-
-        this.update(ChartUpdateType.SERIES_UPDATE);
+        let updateAll = newSeries == null || lastSeries == null;
```
manapeirov, line 1315: in this case do we need to do any update at all?

charts-packages/ag-charts-community/src/scene/shape/path.ts
```diff
@@ -48,6 +47,14 @@ export class Path extends Shape {
         return this._dirtyPath;
     }
 
+    checkPathDirty() {
```
manapeirov, line 50: is this `setPathDirty`?

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: Done

charts-packages/ag-charts-community/src/canvas/hdpiCanvas.ts
alantreadway: We could from a typing perspective, but I think the intent is that the thing we're overriding must be one of these context types, so I don't think the changes makes sense.

charts-packages/ag-charts-community/src/scene/scene.ts
```diff
@@ -93,27 +101,35 @@ export class Scene {
 
     private _nextZIndex = 0;
     private _nextLayerId = 0;
-    addLayer(opts?: { zIndex?: number, name?: string }): HdpiCanvas | undefined {
+    addLayer(opts?: { zIndex?: number, name?: string }): HdpiCanvas | HdpiOffscreenCanvas | undefined {
         const { mode } = this.opts;
-        if (mode !== 'composite' && mode !== 'dom-composite') {
+        const layeredModes = ['composite', 'dom-composite', 'adv-composite'];
```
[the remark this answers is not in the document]
alantreadway: No - `'simple'` isn't a layered mode :-)

charts-packages/ag-charts-community/src/chart/chart.ts
alantreadway: Yes, because the opacity needs to potentially change for the series layers. (On mouse-move out of the series area, we need to update all series to be opaque)

charts-packages/ag-charts-community/src/scene/shape/path.ts
alantreadway: `setPathDirty()` isn't right either - the result could be no change in the `dirtyPath` value of `false`?

charts-packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
alantreadway: Fixed

manapeirov (approved):

PR #5250, opened 2022-06-06
AG-5105 - Chart Axis CrossLines Support

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-5105 and https://ag-grid.atlassian.net/browse/AG-6289.

The aim is to allow the users to render shaded regions/ranges or lines along the x and y axes to highlight specific parts of the chart.

The approach is to create a line path and an area path which can be configured via the user options and placed at the coordinates corresponding to the specified data values.

These ranges/lines are initialised in `AgChartsV2` factory. `Axis.ts` contains the functions which will append and remove them from the crossLines group.  `Axis.ts`updates the crossLines in its `update` function which will provide the required properties such as `scale`, `gridLength` and `visible` to the ranges/lines.

A label can also be added and positioned with respect to the line/region.

Snapshot tests have been added.


charts-packages/ag-charts-community/src/chart/shapes/Annotation.ts
```diff
@@ -0,0 +1,176 @@
+import { Path } from "../../scene/shape/path";
```
alantreadway, line 1: Just wondering if `shapes/` is the right spot for this? Maybe it deserves it's own top-level folder?

charts-packages/ag-charts-community/src/chart/shapes/Annotation.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    private updateRangePath() {
+        const { annotationRange, pathData = { points: [] } } = this;
+        const points = pathData.points;
+        const { path } = annotationRange;
+
+        path.clear();
+        points.forEach((point, i) => {
+            const { x, y } = point;
+            path[i > 0 ? 'lineTo' : 'moveTo'](x, y);
```
alantreadway, line 172: Nit: I'd just use a direct method call with an `if` block for the two cases - it took me extra time to parse what was going on here, so the readability could be improved in this respect.

alantreadway (commented):
Generally looking good :-)

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -235,6 +256,14 @@ export class Axis<S extends Scale<D, number>, D = any> {
         return this._calculatedTickCount ?? this.tick.count;
     }
 
+    attachCrossLine(crossLine: CrossLine) {
```
alantreadway, line 259: Nit: Could this and `detachCrossLine()` be `private`?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -192,6 +197,8 @@ export class Axis<S extends Scale<D, number>, D = any> {
     ticks: any[];
 
     readonly axisGroup = new Group({ name: `${this.id}-axis`, layer: true, zIndex: 50 });
+    readonly crossLineGroup: Group = new Group({ name: `${this.id}-CrossLines`, zIndex: 150 });
```
alantreadway, line 200: Should the `zIndex` here be below the series? `Series.SERIES_LAYER_ZINDEX` should probably be used as a reference point too, if so?

charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
```diff
@@ -0,0 +1,248 @@
...
+
+        const autoRotation = parallel
+            ? parallelFlipFlag * Math.PI / 2
+            : (regularFlipFlag === -1 ? Math.PI : 0);
+
+        crossLineLabel.rotation = autoRotation + labelRotation;
+
+        const bbox = crossLineLabel.computeTransformedBBox();
+
+        crossLineLabel.textBaseline = 'middle';
```
alantreadway, line 237: Nit: Maybe twiddle all of the `crossLineLabels` before calling `computeTransformedBBox()`, in case they impact the result?

alantreadway (commented):
Generally LGTM - some things that you may want to check prior to merge:
- Do a boarder check that the `computeTransformedBBox()` hasn't introduced any unexpected gremlins (maybe check the auto-label-rotation example at least)?
- Verify that the API Explorer is doing something sane with the new configuration options?

alantreadway:
@manapeirov heads-up, there are some merge conflict with `latest` to resolve?

charts-packages/ag-charts-community/src/axis.ts
manapeirov: Yes, done!

charts-packages/ag-charts-community/src/axis.ts
manapeirov: done

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -200,6 +207,20 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
     private gridlineGroupSelection = Selection.select(this.gridlineGroup).selectAll<Group>();
 
+    private _crossLines?: CrossLine[] = [];
+    set crossLines(value: CrossLine[] | undefined) {
+        this._crossLines?.forEach(crossLine => this.detachCrossLine(crossLine));
+
+        this._crossLines = value;
+
+        this._crossLines?.forEach(crossLine => {
+            this.attachCrossLine(crossLine);
```
alantreadway, line 217: Should we also assign `scale` + `gridLength` here too, in case of a different order of execution?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -703,6 +743,17 @@ export class Axis<S extends Scale<D, number>, D = any> {
         lineNode.visible = ticks.length > 0;
 
         this.positionTitle();
+
+        this.crossLines?.forEach((crossLine) => {
+            crossLine.scale = scale;
```
alantreadway, line 748: If we're assigning `scale` and `gridLength` here, maybe we don't need to do it elsewhere?

charts-packages/ag-charts-community/src/chart/agChartV2.test.ts
```diff
@@ -184,6 +184,30 @@ const EXAMPLES: Record<string, TestCase> = {
         options: examples.TRUNCATED_LEGEND_ITEMS,
         assertions: cartesianChartAssertions({ axisTypes: ['number', 'category'], seriesTypes: ['bar'] }),
     },
+    SCATTER_CROSSLINES: {
```
alantreadway, line 187: I'm wondering if these deserve their own test suite in the `crosslines/` folder?

charts-packages/ag-charts-community/src/chart/test/examples.ts
```diff
@@ -1078,6 +1078,442 @@ export const TRUNCATED_LEGEND_ITEMS: AgCartesianChartOptions = {
...
+}
+
+const yAxisCrossLineStyle = {
+    fill: 'pink',
+    fillOpacity: 0.2,
+    stroke: 'red',
+    strokeWidth: 1,
+}
+
+export const SCATTER_CROSSLINES: AgCartesianChartOptions = {
```
alantreadway, line 1095: Just wondering if we should create actual examples for these cases added (we'll need examples for the documentation?), and load them the same way we do for other examples?

alantreadway (approved):
LGTM

PR #5253, opened 2022-06-07
AG-1555 Auto header height & multi line header text

eddking:


community-modules/core/src/ts/headerRendering/cells/column/headerCellComp.ts
```diff
@@ -12,9 +12,11 @@ export class HeaderCellComp extends AbstractHeaderCellComp<HeaderCellCtrl> {
     private static TEMPLATE = /* html */
         `<div class="ag-header-cell" role="columnheader" tabindex="-1">
             <div ref="eResize" class="ag-header-cell-resize" role="presentation"></div>
+            <div ref="eHeaderCompWrapper" class="ag-header-cell-comp-wrapper"></div>
```
gportela85, line 15: This element needs `role="presentation"`

PR #5265, opened 2022-06-10
Fix broken typings for agGridColumn

omerman:
Hello, current AgGridColumn types are broken.

image

I removed redundant code in order to make the typings work again.


gportela85:
@omerman Sorry, this can't be merged because the package needs to be updated as well.

I have made this change here: https://github.com/ag-grid/ag-grid/commit/fdec9743b48d0e66697a5be4819414de47fe3992

Thanks for submitting it.

PR #5280, opened 2022-06-15
AG-6815 - Rework chart axis layout calculations

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6815

Reworks the Chart Axis layout calculations to:
- Be more explicit about taking an iterative approach to finding a solution.
- Use more than a fixed number of iterations (2 previously, possibly), especially in light of recent additions such as auto-hiding of labels/ticks depending on available space.
- Push more responsibility back into the `Axis` class hierarchy to help simplify `CartesianChart.performLayout()`.
- Detangle the mess of dependencies in layout calculations (previously 4-5x loops over axis, now just one).
- Remove redundant code duplication (e.g. `Axis.computeBBox()` rather than delegating to `Group.computeBBox()`).

See also:
https://ag-grid.atlassian.net/browse/AG-6878
https://ag-grid.atlassian.net/browse/AG-6783

### DIFF
--- a/charts-packages/ag-charts-community/src/axis.ts
+++ b/charts-packages/ag-charts-community/src/axis.ts
@@ -6,7 +6,6 @@ import { Text, FontStyle, FontWeight } from "./scene/shape/text";
 import { Arc } from "./scene/shape/arc";
 import { Shape } from "./scene/shape/shape";
 import { BBox } from "./scene/bbox";
-import { Matrix } from "./scene/matrix";
 import { Caption } from "./caption";
 import { createId } from "./util/id";
 import { normalizeAngle360, normalizeAngle360Inclusive, toRadians } from "./util/angle";
@@ -176,11 +175,13 @@ export class Axis<S extends Scale<D, number>, D = any> {
     ticks: any[];
 
     readonly axisGroup = new Group({ name: `${this.id}-axis`, layer: true, zIndex: 50 });
-    private axisGroupSelection: Selection<Group, Group, D, D>;
-    private lineNode = new Line();
+    private readonly tickLineGroup = this.axisGroup.appendChild(new Group());
+    private readonly titleGroup = this.axisGroup.appendChild(new Group());
+    private tickLineGroupSelection = Selection.select(this.tickLineGroup).selectAll<Group>();
+    private lineNode = this.tickLineGroup.appendChild(new Line());
 
     readonly gridlineGroup = new Group({ name: `${this.id}-gridline`, layer: true, zIndex: 0 });
-    private gridlineGroupSelection: Selection<Group, Group, D, D>;
+    private gridlineGroupSelection = Selection.select(this.gridlineGroup).selectAll<Group>();
 
     readonly line: {
         /**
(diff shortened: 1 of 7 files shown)


alantreadway:
Rendering is now mostly consistent with the pre-refactor output - the navigator is mis-positioned but I will have a fix for that shortly.

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -707,17 +703,16 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
-            const bbox = this.computeBBox({ excludeTitle: true });
+            const bbox = this.tickLineGroup.computeBBox();
             const titleRotationFlag = sideFlag === -1 && parallelFlipRotation > Math.PI && parallelFlipRotation < Math.PI * 2 ? -1 : 1;
 
             titleNode.rotation = titleRotationFlag * sideFlag * Math.PI / 2;
-            // titleNode.x = titleRotationFlag * sideFlag * (lineNode.y1 + lineNode.y2) / 2; // TODO: remove?
-            titleNode.x = titleRotationFlag * sideFlag * (requestedRange[0] + requestedRange[1]) / 2;
+            titleNode.x = Math.floor(titleRotationFlag * sideFlag * (requestedRange[0] + requestedRange[1]) / 2);
 
             if (sideFlag === -1) {
```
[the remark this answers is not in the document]
alantreadway: Fixed

alantreadway:
> Rendering is now mostly consistent with the pre-refactor output - the navigator is mis-positioned but I will have a fix for that shortly.

Navigator positioning now fixed too.

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -288,50 +195,161 @@ export class CartesianChart extends Chart {
...
+                    const safeY = Math.max(shrinkRect.y, shrinkRect.y + shrinkRect.height + 1);
+                    axis.translation.x = Math.floor(shrinkRect.x + (axisWidths.left ?? 0));
+                    axis.translation.y = Math.floor(safeY - axisThickness - axisOffset);
+                    break;
+                case ChartAxisPosition.Left:
+                    axis.translation.x = Math.floor(shrinkRect.x + axisOffset + axisThickness);
+                    axis.translation.y = Math.floor(shrinkRect.y + (axisWidths.top ?? 0));
+                    break;
+            }
             axis.update();
```
manapeirov, line 348: we still have to call update multiple times or two times per pass?

alantreadway: I was just looking at this code and was having the same thought. I'm going to try removing, since the only axis mutation after the previous call is to the `translation` property, which should only impact the main axis group position.

alantreadway: Hmmm... yes it's needed, the axis renders offset if I don't do this :-P

I'm not going to dig into it here, but I think half of the (over-)complexity of this axis code is due to the fact that we're not using the containing `Group` to transform the axis - all nested children are being individually rotated and positioned, which means we need different code-paths for vertical vs. horizontal rendering, and that we can't reposition the axis trivially, as we should be able to do here :-P

alantreadway:
Merging - will follow-up separately on any further feedback on Mana's return.

PR #5282, opened 2022-06-15
AG-6884 - groupHideOpenParents fix to allow cell copying

AndrewGlazier:


enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -691,7 +691,8 @@ export class ClipboardService extends BeanStub implements IClipboardService {
             processCellCallback: this.gridOptionsWrapper.getProcessCellForClipboardFunc(),
             processRowGroupCallback: (params) => params.node.key!,
             processHeaderCallback: this.gridOptionsWrapper.getProcessHeaderForClipboardFunc(),
-            processGroupHeaderCallback: this.gridOptionsWrapper.getProcessGroupHeaderForClipboardFunc()
+            processGroupHeaderCallback: this.gridOptionsWrapper.getProcessGroupHeaderForClipboardFunc(),
+            hideOpenParents: false,
```
gportela85, line 695: Instead of creating a local param that overrides the value of `hideOpenParents` I would make it `exportGroupHideOpenParents: true` and tweak the conditional in the GridSerializer.

It took me a while to understand why this works, it's better to make it explicit.

dpoliyivets:
Hi guys! We're your enterprise customers and this is currently blocking us from releasing an upgrade to ag-grid v27.

We've monkey-patched the current fix and it does fix the issue with copying row data, however, there's still a bug with footer data when you have the `groupIncludeTotalFooter` flag set to true. The previous version included the total row when selecting all grid data (CTRL + A) and hitting the "Copy with headers" option from the grid context menu.

Just highlighting the issue to you so that you're aware of this. Thanks!

ezgif com-gif-maker (4)

kiril-matev:
Hi @dpoliyivets!

Thank you for bringing this up.

As an AG Grid Enterprise customer please use our support system to report issues. This is the quickest way to get a response to any query.

The groupHideOpenParents property doesn't have to be used together with the groupIncludeTotalFooter property. This is why this should be reported as a separate issue as it happens even without groupHideOpenParents being set. 

Please open a new Github issue for this, so I can respond there and so it's more transparent to all users.

Thank you for your time.

kiril-matev:
Hi @dpoliyivets!

AG Grid footer row group column value cannot be copied to clipboard and isn't visible in the export. These limitations are documented here:
https://www.ag-grid.com/javascript-data-grid/grouping-footers/#group-footer-limitations 

This is why we have added a feature request in our backlog to track this requirement and implement it in the future.

You can track it with this reference:
**AG-6920 [Row Grouping] Allow copying the group column cell value for the group and total footer rows**

You can follow the status of feature requests, bugs and releases using our pipeline:

Now that this requirement is recorded in our backlog, we will soon close the GitHub issue because we will not update it when this feature is added in a future version of AG Grid. The best way to get an update on this item is to sign up for AG Grid new release notifications. Once a new AG Grid version is released, look for the item reference above (AG-XXXX) on the changelog page to see if it's resolved in that version:
https://www.ag-grid.com/ag-grid-changelog/

PR #5289, opened 2022-06-18
Invalid aria prop `aria-description` on <div> tag.

RiteshParyali:
Invalid aria prop `aria-description` on <div> tag.

It's not ignored by **eslint**, and it's giving error,

image


seanlandsman:
This is a react issue - please see https://github.com/facebook/react/issues/21035

PR #5292, opened 2022-06-20
AG-6902 - Fix Chart legend hover not updating series opacity.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6902

Fixes edge-case of changing hover between legend items. The root cause was that the `Group.dirty` flag was not being cleared in the `Group.dirty === RedrawType.TRIVIAL` case, which meant subsequent dirty processing didn't percolate `TRIVIAL` updates up the tree.


manapeirov (approved):
LGTM :)

PR #5297, opened 2022-06-22
AG-6907 - Improve number axis tick calculations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6907

After #5280 some gallery + other examples had a reduced tick count. The root cause of this seems to be that the `availableRange` taken into account in `ChartAxis.calculateTickCount()` is now more accurate and smaller as it takes other axes into account.

I've taken another swing at improving the `ChartAxis.calculateTickCount()` calculation, specifically for number axes - we now try and estimate the size of the largest label, and use this as the basis (with some padding) for calculating the optimal number of ticks. For `y` axes we assume a single line of text for the label, and apply more generous padding.

Bonus:
- I updated a gallery example to remove labels with decimals using a formatter.
- I then found there was a bug in the axis label-skipping code which meant that we ended up with 9 labels, each with 2 ticks between them, and a gap where a 3rd tick should've been. The reason for this seems to be that the empty label case still has a `BBox` with a non-zero height, triggering the skipping process unnecessarily. This is now fixed.
- I then found that our label formatter typings don't allow for an `undefined` result - this is also fixed.

### DIFF
--- a/charts-packages/ag-charts-community/src/axis.ts
+++ b/charts-packages/ag-charts-community/src/axis.ts
@@ -513,7 +513,7 @@ export class Axis<S extends Scale<D, number>, D = any> {
         this.fractionDigits = (ticks as any).fractionDigits >= 0 ? (ticks as any).fractionDigits : 0;
 
         // Update properties that affect the size of the axis labels and measure the labels
-        const labelBboxes: Map<number, BBox> = new Map();
+        const labelBboxes: Map<number, BBox | null> = new Map();
         let labelCount = 0;
 
         let halfFirstLabelLength = false;
@@ -531,9 +531,10 @@ export class Axis<S extends Scale<D, number>, D = any> {
                 node.visible = node.parent!.visible;
                 if (node.visible !== true) { return; }
 
-                labelBboxes.set(index, node.computeBBox());
+                const userHidden = node.text === '' || node.text == undefined;
+                labelBboxes.set(index, userHidden ? null : node.computeBBox());
 
-                if (node.text === '' || node.text == undefined) { return; }
+                if (userHidden) { return; }
                 labelCount++;
 
                 if (index === 0 && (node.translationY === scale.range[0])) {
@@ -547,12 +548,14 @@ export class Axis<S extends Scale<D, number>, D = any> {
 
         const step = availableRange / labelCount;
 
-        const calculateLabelsLength = (bboxes: Map<number, BBox>, useWidth: boolean) => {
+        const calculateLabelsLength = (bboxes: Map<number, BBox | null>, useWidth: boolean) => {
             let totalLength = 0;
             let rotate = false;
             const lastIdx = bboxes.size - 1;
             const padding = 12;
             for (let [i, bbox] of bboxes.entries()) {
+                if (bbox == null) { continue; }
+
                 const divideBy = (i === 0 && halfFirstLabelLength) || (i === lastIdx && halfLastLabelLength) ? 2 : 1;
                 const length = useWidth ? bbox.width / divideBy : bbox.height / divideBy;
                 const lengthWithPadding = length <= 0 ? 0 : length + padding;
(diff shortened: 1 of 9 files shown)


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -531,9 +531,10 @@ export class Axis<S extends Scale<D, number>, D = any> {
                 node.visible = node.parent!.visible;
                 if (node.visible !== true) { return; }
 
-                labelBboxes.set(index, node.computeBBox());
+                const userHidden = node.text === '' || node.text == undefined;
+                labelBboxes.set(index, userHidden ? null : node.computeBBox());
```
manapeirov, line 535: Can we not just move this to set after the return on line 537? instead of setting it to null

charts-packages/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -467,7 +467,7 @@ export interface AgAxisLabelOptions {
     /** Format string used when rendering labels for time axes. */
     format?: string;
     /** Function used to render axis labels. If `value` is a number, `fractionDigits` will also be provided, which indicates the number of fractional digits used in the step between ticks; for example, a tick step of `0.0005` would have `fractionDigits` set to `4` */
-    formatter?: (params: AgAxisLabelFormatterParams) => string;
+    formatter?: (params: AgAxisLabelFormatterParams) => string | undefined;
```
manapeirov, line 470: Should it be able to return undefined?

charts-packages/ag-charts-community/src/chart/chartAxis.ts
```diff
@@ -62,15 +62,39 @@ export class ChartAxis<S extends Scale<any, number> = Scale<any, number>> extend
...
         const tickInterval = optimalTickInteralPx * tickIntervalRatio;
-        this._calculatedTickCount = this.tick.count || Math.max(2, Math.floor(availableRange / tickInterval));
+        this._calculatedTickCount = Math.max(2, Math.floor(availableRange / tickInterval));
+    }
+
+    protected calculateTickInternalEstimate() {
+        const { domain, label: { fontSize }, direction } = this;
+
+        const padding = fontSize * 1.5;
+        if (direction === 'y') {
```
manapeirov, line 86: can we use `ChartAxisDirection.Y` here?

charts-packages/ag-charts-community/src/chart/chartAxis.ts
```diff
@@ -62,15 +62,39 @@ export class ChartAxis<S extends Scale<any, number> = Scale<any, number>> extend
...
-        const optimalTickInteralPx = 70;
-        const tickIntervalRatio = Math.pow(Math.log(availableRange) / Math.log(optimalRangePx), 2);
+        const minTickIntervalRatio = 0.75;
+        const tickIntervalRatio = Math.max(Math.pow(Math.log(availableRange) / Math.log(optimalRangePx), 2), minTickIntervalRatio);
         const tickInterval = optimalTickInteralPx * tickIntervalRatio;
-        this._calculatedTickCount = this.tick.count || Math.max(2, Math.floor(availableRange / tickInterval));
+        this._calculatedTickCount = Math.max(2, Math.floor(availableRange / tickInterval));
+    }
+
+    protected calculateTickInternalEstimate() {
```
manapeirov, line 82: Is this meant to be called `calculateTick**Interval**Estimate()`?

charts-packages/ag-charts-community/src/axis.ts
alantreadway: No, we need the `Map` entry to be present, otherwise we remove ticks/gridlines later in the calculations.

alantreadway: Specifically, we need an entry for ticks and gridlines to still render:
- https://github.com/ag-grid/ag-grid/blob/3e2265f87bf34b0d177b4d262d5e1d97620f6dd9/charts-packages/ag-charts-community/src/axis.ts#L646
- https://github.com/ag-grid/ag-grid/blob/3e2265f87bf34b0d177b4d262d5e1d97620f6dd9/charts-packages/ag-charts-community/src/axis.ts#L669

charts-packages/ag-charts-community/src/chart/chartAxis.ts
alantreadway: Good call, done

alantreadway:
PTAL @manapeirov

manapeirov (approved):
LGTM

PR #5302, opened 2022-06-23
AG-6887 - Series highlight z-order improvements

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6887

Whilst reviewing the z-order changes made for `28.0.0` it became clear there were some easily achievable improvements, which are captured as part of this PR:
- Highlighted series is now temporarily bumped in `zIndex`, so hidden series can be more easily seen when hovering over the series itself or it's legend item.
- Highlighted datum rendering is at a higher `zIndex` than all series, so highlighted datum can be easily seen. (No change in this PR - just stating new `28.0.0` behaviour which has been approved)

Bonus:
- Cleanup Group and Scene `zIndex` processing.


manapeirov (approved):
Looks good :)

PR #5303, opened 2022-06-23
AG-5740 - Add more consistent series.visible handling.

alantreadway:
Attempts to make `visible: false` behave more consistently across series. After this change, setting `visible: false` means that series will appear as if they have been toggled disabled from the legend.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
+++ b/charts-packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
@@ -43,31 +43,72 @@ export function groupSeriesByType(seriesOptions: SeriesOptions[]) {
 export function reduceSeries(series: any[], enableBarSeriesSpecialCases: boolean) {
     let options: any = {};
 
+    const NA = Symbol();
+    const SKIP = Symbol();
     const arrayValueProperties = ['yKeys', 'fills', 'strokes', 'yNames', 'hideInChart', 'hideInLegend'];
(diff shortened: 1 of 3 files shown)


alantreadway:
Merging to allow @rmc-software to progress the Integrated Charts rework - will follow-up with cleanup.

charts-packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
```diff
@@ -43,31 +43,72 @@ export function groupSeriesByType(seriesOptions: SeriesOptions[]) {
...
-                    options.hideInLegend = [...(options.hideInLegend || []), ...(s.yKey ? [s.yKey] : s.yKeys)];
-                }
-            } else if (enableBarSeriesSpecialCases && property === 'grouped') {
-                if (s[property] === true) {
-                    options[property] = s[property];
-                }
-            } else {
-                options[property] = s[property];
-            }
+    const stringValueProperties = ['yKey', 'fill', 'stroke', 'yName', 'visible'];
```
manapeirov, line 49: How come visible is a stringValueProperty

alantreadway: Good spot - I already fixed this misalignment in the follow-up cleanup PR here (these become the `DEFAULTING_ARRAY_REDUCER` case):
https://github.com/ag-grid/ag-grid/pull/5305/files

PR #5315, opened 2022-06-28
Improved documentation for gridOptions and framework bindings

ngehlert:
Currently it's not very clear what happens if you specify an option on the gridOptions and with a framework binding
https://www.ag-grid.com/angular-data-grid/grid-interface/#grid-options-1

I took a look at the source code and it looks like (and I can also confirm this from my testing) that the values from the framework bindings will override the settings made in gridOptions
https://github.com/ag-grid/ag-grid/blob/latest/grid-packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts#L204


StephenCooper (approved):
Thanks for this addition to our docs.

PR #5321, opened 2022-06-29
AG-6931 - Fix api explorer update cases

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6931

Attempts to fix several breaks found in testing WRT API Explorer interactive Chart updates. Breaks-down into several cases of break:
- These options were both unsupported (and unimplemented) options accidentally introduced into the options interface contract in 27.1.0 - they have been removed so the options contract aligned with the implementation:
  - `title.padding`
  - `subtitle.padding`
  - `series[].title.padding`
  - `series[].marker.fillOpacity`
  - `series[].marker.strokeOpacity`
- These options were due to editor breakage in the API Explorer itself:
  - `series[].lineDash`
  - `series[].fills`
  - `series[].strokes`
-  There was one actual bug in `axis[].tick.count` due to residual state when a Falsy count of `0` is presented - this is now fixed.


charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -266,7 +267,7 @@ export class CartesianChart extends Chart {
             visited[position] = (visited[position] ?? 0) + 1;
 
             // for multiple axes in the same direction and position, apply padding at the top of each inner axis (i.e. between axes).
-            const axisPadding = axis.title && axis.title.padding.top || 15;
+            const axisPadding = axis.title ? Caption.PADDING : 15;
```
manapeirov, line 270: I guess we can just make this `Caption.PADDING` or 15... don't need to check if there is a title anymore

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -861,37 +861,37 @@ export abstract class Chart extends Observable {
...
+        if (!title) {
+            return;
+        }
+        title.node.visible = title.enabled;
+
+        if (title.enabled) {
             title.node.x = this.width / 2;
             title.node.y = paddingTop;
-            titleVisible = true;
+            title.node.visible = true;
```
manapeirov, line 875: This is already happening on line 870, do we still need it to be set to true here

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -861,37 +861,37 @@ export abstract class Chart extends Observable {
...
+            return;
         }
-        if (subtitle) {
-            subtitle.node.visible = subtitleVisible;
+        subtitle.node.visible = title.enabled && subtitle.enabled;
+        
+        if (title.enabled && subtitle.enabled) {
+            subtitle.node.x = this.width / 2;
+            subtitle.node.y = paddingTop + spacing;
+            subtitle.node.visible = true;
```
manapeirov, line 890: same with subtitle

charts-packages/ag-charts-community/src/chart/chart.ts
alantreadway: Good spot - fixed.

alantreadway:
@manapeirov PTAL

manapeirov (approved):

PR #5326, opened 2022-07-01
AG-6939 - Fix chart deprecations warnings.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6939

Re-enabled @Deprecated() checks in Charts, and fixes many examples using deprecated properties.

Additionally reinstates `Marker.{fill,stroke}Opacity` which is needed by `ScatterSeries`, and now also works correctly for `LineSeries` and `AreaSeries`.

### DIFF
--- a/grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
+++ b/grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
@@ -20,6 +20,7 @@ type AliasTypeProps<T> = {
     breakIndex?: number,
     min?: T,
     max?: T,
+    step?: T,
     unit?: string,
 };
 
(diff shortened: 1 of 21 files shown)


grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
```diff
@@ -190,13 +192,13 @@ export const getPrimitiveEditor = ({ meta, desc }: JsonModelProperty, key: strin
     return { editor, editorProps: { ...meta, ...editorProps } };
 };
 
-const setStepEditorProp = (editorProps: Record<string, any>, { min, max }: JsonModelProperty['meta']) => {
-    if (min == null || max == null) {
+const setStepEditorProp = (editorProps: Record<string, any>, { min, max, step }: JsonModelProperty['meta']) => {
+    if (min == null || max == null || step != null) {
```
manapeirov, line 196: is this supposed to be `step == null` ?

manapeirov (approved):
LGTM

grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
alantreadway: No - if `step` is explicitly set, then the intent is no to try and execute the method (which tries to guess a step).

PR #5348, opened 2022-07-12
Ag 6825 fix suppress sorting for set filter

AndrewGlazier:


enterprise-modules/set-filter/src/setFilter/setValueModel.ts
```diff
@@ -403,6 +401,25 @@ export class SetValueModel implements IEventEmitter {
...
     }
 
+    private uniqueStringArray(values: any[]): (string | null)[] {
+        const stringifiedResults = _.toStrings(values);
+        if (!stringifiedResults) {
+            return [];
+        }
+
+        const uniqueValues = this.uniqueValues(stringifiedResults);
+        return stringifiedResults.filter(item => {
```
alantreadway, line 411: Just wondering if `Object.values(uniqueValues)` would give the same result as the filter applied here?

alantreadway (commented):
Looks like a reasonable refactor, just one suggestion for a potential code reduction, but needs testing.

enterprise-modules/set-filter/src/setFilter/setValueModel.ts
AndrewGlazier: The issue here is that uniqueValues itself is applying the sort, in the event of numeric items (or even numeric strings) we're generating keys as such:

'30',
'2,
'1',

javascript maps sort numeric strings/numbers when inserted into an object.

Pop this into console to prove the hypothesis:
new Array(10).fill(undefined).reduce((acc, _, idx, arr) => ({ ...acc, [`${arr.length - idx}`]: arr.length - idx }), {});

Note how despite the fact the order should be inverted, the output in the console is actually linear 1-10.

alantreadway: @AndrewGlazier ok sounds good - could you add a comment to highlight this for future readers, as my natural instinct would have been to refactor this? :-)

PR #5351, opened 2022-07-14
Ag 5105/annotations

manapeirov:
Documentation for the new cross lines feature:  https://ag-grid.atlassian.net/browse/AG-5105 + https://ag-grid.atlassian.net/browse/AG-6289


alantreadway (approved):
LGTM - the example is a bit verbose, but I wonder if we looked at crossLine theming support that might reduce some of the options down a little?

manapeirov:
@alantreadway PTAL

PR #5357, opened 2022-07-14
Ag 5105/annotations

manapeirov:
Default fill and stroke for crossLines via chart theme


charts-packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -243,10 +243,17 @@ function calculateSeriesPalette<T extends SeriesOptionsTypes>(context: Preparati
     return paletteOptions as T;
 }
 
-function prepareAxis<T extends AxesOptionsTypes>(input: T, ...defaults: T[]): T {
+function prepareAxis<T extends AxesOptionsTypes>(axis: T, defaults: Omit<T, "crossLines"> & { crossLines: AgCrossLineOptions }): T {
```
alantreadway, line 246: Nit: Maybe rename `defaults` to `axisTheme` to reflect the changed expectations?

charts-packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -243,10 +243,17 @@ function calculateSeriesPalette<T extends SeriesOptionsTypes>(context: Preparati
...
     const removeOptions = { top: DELETE, bottom: DELETE, left: DELETE, right: DELETE } as any;
-    return jsonMerge(...defaults, input, removeOptions);
+
+    // Special cross lines case where we have an arrays of cross line elements which need their own defaults.
+    if (axis.crossLines) {
+        const { crossLines: crossLinesTheme } = defaults;
+        axis.crossLines = axis.crossLines.map((crossLine) => jsonMerge(crossLinesTheme, crossLine));
+    }
+
+    return jsonMerge(defaults, axis, removeOptions);
```
alantreadway, line 256: We may need to remove `crossLines` from `default` before using it here, otherwise you may be seeing some console warnings about mismatching types?

manapeirov: Do you mean by using the removeOptions?

alantreadway:
FYI: https://github.com/ag-grid/ag-grid/pull/5354 merged

charts-packages/ag-charts-community/src/chart/mapping/prepare.ts
alantreadway: I think either way is fine (`axesTheme` is a resuilt of a `jsonMerge()` in  `prepareOptions()` - so directly mutating is practically fine) - but it might be cleaner to copy & remove in this method.

You can't use `removeOptions` directly since it is applied last (and would remove any user-supplied `crossLines` property - but you could add an extra input to `jsonMerge` immediate after `axesTheme`/`defaults` to clean-up before applying user-supplied options?

manapeirov:
@alantreadway PTAL

alantreadway (approved):
LGTM

PR #5364, opened 2022-07-15
AG-5105 - Add enabled property to `AgCrossLineOptions` and `AgCrossLi…

manapeirov:
- Add `enabled` flag to crossLine options for consistency
- Change rendering order depending on type of crossLine

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
+++ b/charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
@@ -11,6 +11,7 @@ import { ChartAxisDirection } from "../chartAxis";
 import { CrossLineLabelPosition, Point, labeldDirectionHandling, POSITION_TOP_COORDINATES, calculateLabelTranslation } from "./crossLineLabelPosition";
 
 export class CrossLineLabel {
+    enabled?: boolean = undefined;
     text?: string = undefined;
     fontStyle?: FontStyle = undefined;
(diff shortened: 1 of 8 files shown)


charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
```diff
@@ -77,7 +79,7 @@ export class CrossLine {
     }
 
     update(visible: boolean) {
-        if (!this.type) { return; }
+        if (!this.enabled || !this.type) { return; }
```
alantreadway, line 82: Does the update case work correctly? I.e. for an existing chart with crossLines enabled, does disabling them actually remove them?

charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
```diff
@@ -96,8 +98,10 @@ export class CrossLine {
             this.updateRangeNode();
         }
 
-        this.updateLabel();
-        this.positionLabel();
+        if (this.label.enabled) {
```
alantreadway, line 101: Ditto

charts-packages/ag-charts-community/src/chart/crossline/crossLine.ts
manapeirov: yes

manapeirov:
@alantreadway PTAL

PR #5368, opened 2022-07-18
AG-6797 - Fix Rect border rendering.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6797

Rendering updates around `Rect` and `crisp: true` so:
- `Rect` stroke aka border is renderer WITHIN the `Rect` bounds, rather than partially overlapping. This addresses the issues raised in AG-6797.
- Reworks `crisp: true` / alignment to take device-pixel-ratio into account properly.

Bonus:
- Removes redundant `strokeShadow` handling.
- Improves usage of alignment calculations (makes alignment a proper method, and removes the need to pass the same constant to each invocation).
- Fixes `Group` optimised rendering mode, to correctly remove dangling references to removed children.

Screenshot 2022-07-18 at 11 37 56


manapeirov (approved):

charts-packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -528,8 +538,11 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
 
         groupScale.range = [0, xBandWidth!];
 
-        const grouped = true;
-        const barWidth = grouped ? groupScale.bandwidth : xBandWidth!;
+        const barWidth = groupScale.bandwidth >= 1 ?
```
manapeirov, line 541: Does this still work for continuous axes? We're not using the `xBandWidth` calculated above anymore?

alantreadway: `xBandwidth` was never being used before for `barWidth`, since `grouped = true` meant the value assigned as always `groupScale.bandwidth` :-(

PR #5379, opened 2022-07-19
Ag 5105/annotations

manapeirov:
Automatically account for cross line labels outside of the series rect during layout calculations to avoid clipping of the cross line labels


charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -210,6 +210,24 @@ export class CartesianChart extends Chart {
             clipSeries = result.clipSeries;
             seriesRect = result.seriesRect;
 
+            if (count === 0) {
```
alantreadway, line 213: The logic looks right below, but this seems like an odd spot for it? Could this be done without the guard before the `do`?

alantreadway (approved):
LGTM!

PR #5390, opened 2022-07-21
AG-6671 - Fix sparklines item highlighting

manapeirov:
Don't break out of the loop early when finding closest node during highlighting as this approach breaks in the case where the provided sparkline data is not ordered


alantreadway (approved):
LGTM

PR #5395, opened 2022-07-22
Ag 5105/cross lines

manapeirov:
- Validate the values supplied to cross lines 
- Only clamp values for range cross lines
- Do not display line for clamped range boundaries
- Do not display line cross lines outside of the axis domain
- Make crossLine invisible if update results in no path data
- Update crossLine docs
- Set the default enabled values for cross lines and cross line label to false in chartTheme

### DIFF
--- a/charts-packages/ag-charts-community/src/util/value.ts
+++ b/charts-packages/ag-charts-community/src/util/value.ts
@@ -34,3 +34,15 @@ export function isDiscrete(value: any): boolean {
 export function isContinuous(value: any): boolean {
     return isNumeric(value) || isDate(value);
 }
+
+export function checkDatum<T>(value: T, isContinuousScale: boolean): T | string | undefined {
+    if (isContinuousScale && isContinuous(value)) {
+        return value;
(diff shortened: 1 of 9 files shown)


charts-packages/ag-charts-community/src/util/value.ts
```diff
@@ -34,3 +34,15 @@ export function isDiscrete(value: any): boolean {
 export function isContinuous(value: any): boolean {
     return isNumeric(value) || isDate(value);
 }
+
+export function checkDatum<T>(value: T, isContinuousScale: boolean): T | string | undefined {
```
alantreadway, line 38: Looks good - could you follow-up with a change to use this from the other places we have this same code?

alantreadway (approved):

charts-packages/ag-charts-community/src/util/value.ts
manapeirov: Sure will do

manapeirov: done

PR #5402, opened 2022-07-26
AG-5320 - Fix bins calculation when `binCount` is specified

manapeirov:
PR for: https://ag-grid.atlassian.net/browse/AG-5320

Try to stick to the configured `binCount`.


alantreadway (approved):
LGTM

Issue #5406, opened 2022-07-26
Incorrect type for "rowPinned" property of "rowNode"


JohannCooper:
**I'm submitting a ...**
```
[X] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

**Current behavior**
The current type for the `rowPinned` property of the `rowNode` type is defined as follows in `community-modules/core/src/ts/entities/rowNode.ts`:
```ts
/** Either 'top' or 'bottom' if row pinned, otherwise `undefined` or `null`. */
public rowPinned: string;
```
Not only is this type incorrect (excludes the `undefined` and `null` cases), it is also not as narrow as it could be.

**Expected behavior**
I would suggest changing the type to the following:
```ts
/** Either 'top' or 'bottom' if row pinned, otherwise `undefined` or `null`. */
public rowPinned: "top" | "bottom" | undefined | null;
```

**Please tell us about your environment:**

* **AG Grid version:** 28.0.0

* **Browser:** Any

* **Language:** TypeScript

gportela85:
this should be addressed by this PR - https://github.com/ag-grid/ag-grid/pull/5407

Thanks for reporting

PR #5407, opened 2022-07-27
AG-7048 - Improved Model/Pinned types

gportela85:
Fixes the following:

https://github.com/ag-grid/ag-grid/issues/5406


StephenCooper (commented):
Nice work!

PR #5410, opened 2022-07-27
AG-7049 - Drag and drop hovering

AndrewGlazier:


gportela85 (changes requested):
I think the `getNewInsertIndex` should be something short and simple, and because we have access to the `this.horizontal` property, there is no need to pass a flag.

I haven't tested the code below, but I don't any reason why it shouldn't work

```
private getNewInsertIndex(draggingEvent: DraggingEvent): number {
    const { clientX, clientY } = draggingEvent.event;
    const midPoints = this.childColumnComponents.map(childComp => {
        const rect = childComp.getGui().getBoundingClientRect();
        if (this.horizontal) {
            return rect.x + rect.width / 2;
        }
        return rect.y + rect.height / 2
    });

    const pointerPosition = this.horizontal ? clientX : clientY;

    midPoints.push(pointerPosition);

    midPoints.sort((a,b) => a - b);

    return midPoints.findIndex(value => value === pointerPosition);
}
```

enterprise-modules/row-grouping/src/rowGrouping/columnDropZones/baseDropZonePanel.ts
```diff
@@ -208,50 +208,60 @@ export abstract class BaseDropZonePanel extends Component {
...
+        // not hovering a non-ghost component
+        if (index === -1) {
+            // if mouse is below or right of all components then new index should be placed last
+            const isLast = this.childColumnComponents.every(child => {
+                const rect = child.getGui().getBoundingClientRect();
+                if (isVertical) {
+                    return mouseLocation > rect.bottom;
+                }
+                return mouseLocation > rect.right;
+            });            const rects = this.childColumnComponents.map(child => child.getGui().getBoundingClientRect());
```
gportela85, line 233: line break here

AndrewGlazier:
Line break / horizontal param are fair points.

The implementation you've suggested would be simpler, however it comes with an undesirable change in behaviour. Because you're comparing to find where the index of the pointer lays within that list, hovering a component would now have a different behaviour, hovering on the right of the component would insert after, hovering on the left would insert before.

The required behaviour 
Hovering the component will insert the ghost before
Hovering the component while the ghost is left of the hovered component will insert after
Hovering the LHS of all components will insert first
Hovering the RHS of all components will insert last


I don't think I can simplify it much further more than what I've done unfortunately

gportela85 (approved):

gportela85 (changes requested):
This code is no longer considering when `enableRtl: true`, looking at the logic, this will probably have a flickering effect when rtl is true. Can you you confirm it works?

gportela85 (approved):

PR #5414, opened 2022-07-28
AG-7052 - Reverse stacked bar and area series legend items.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7052

Reverses the ordering of vertically stacked series legend items, so they are visually consistent in order with the series display order. This is consistent with Excel's stacked series renderings.

Also adds a `legend.reverseOrder` option, since all of our competitors also offer this option, which should allow users to workaround any issues for non-combination charts due to the change.


charts-packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -168,6 +168,9 @@ export class Legend {
         return this._position;
     }
 
+    /** Reverse the display order of legend items if `true`. */
+    reverseOrder = false;
```
manapeirov, line 172: Shall we have this default in the theme?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -893,23 +893,28 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
...
-                        strokeOpacity: marker.strokeOpacity ?? strokeOpacity,
-                    },
-                });
+        if (!data || !data.length || !xKey || !yKeys.length) {
+            return;
+        }
+
+        // Area stacks should be listed in the legend in reverse order, for symmetry with the
+        // vertical stack display order.
+        for (let index = yKeys.length - 1; index >= 0; index--) {
```
manapeirov, line 902: So setting `reverseOrder` to `true` will be previous behaviour?

alantreadway: No, we debated this a bit last night - `reverseOrder` is distinct, and isn't meant to give the previous behaviour - it just happens that in many cases (single series cases) it would allow the previous behaviour.

charts-packages/ag-charts-community/src/chart/legend.ts
alantreadway: Fix pushed.

alantreadway:
@manapeirov PTAL - thanks!

manapeirov:
LGTM thanks for adding it to the theme

manapeirov (approved):

PR #5416, opened 2022-07-28
AG-6797 - Fix Rect border rendering edge cases & add tests.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6797

Re-attempt to fix rendering edge cases for `Rect` with borders. Also add visual regression tests to highlight any rendering anomalies (we should roll this strategy out to other parts of the scene graph over time too).

Whilst testing, it seemed that `0` width rectangles were not being rendered consistently (e.g. stroke would not be rendered beyond the end of what was essentially a line - as soon as `width >= 1` then the stroke extends by an additional `strokeWidth / 2` pixels 🤦


manapeirov (approved):
LGTM

PR #5419, opened 2022-07-29
AG-7009 - Add inRangeFloatingFilterDateFormat option

AndrewGlazier:


community-modules/core/src/ts/utils/date.ts
```diff
@@ -19,6 +19,63 @@ export function serialiseDate(date: Date | null, includeTime = true, separator =
...
+
+    const fullYear = padStartWidthZeros(date.getUTCFullYear(), 4);
+    const months = [
+        'January', 'February', 'March', 'April', 'May', 'June',
+        'July', 'August', 'September', 'October', 'November', 'December',
+    ];
+    const days = [
+        'Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday',
+    ]
+    const replace: { [key:string]: number | string } = {
```
alantreadway, line 51: Nit: We're doing a lot of `string` building and calcs here - would it make sense to make the values lambdas that execute the expression as needed?

(Not sure if there's a significant impact either way, but just seems wrong to calculate the full set of all possible substitutions on every use?)

community-modules/core/src/ts/utils/date.ts
```diff
@@ -19,6 +19,63 @@ export function serialiseDate(date: Date | null, includeTime = true, separator =
...
+        YYYY: fullYear.slice(fullYear.length - 4, fullYear.length),
```
alantreadway, line 52: We probably need to document these formats somewhere (or reference somewhere that does)?

community-modules/core/src/ts/utils/date.ts
```diff
@@ -19,6 +19,63 @@ export function serialiseDate(date: Date | null, includeTime = true, separator =
...
+        dddd: days[date.getUTCDay()],
+        ddd: days[date.getUTCDay()].slice(0, 3),
+        dd: days[date.getUTCDay()].slice(0, 2),
+        do: `${date.getUTCDay()}${calculateOrdinal(date.getUTCDay())}`,
+        d: date.getUTCDay(),
+    };
+    const regexp = new RegExp(Object.keys(replace).join('|'), 'g');
+    return format.replace(regexp, (match) => {
+        if (match in replace) {
+            return `${replace[match]}`;
```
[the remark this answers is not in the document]
AndrewGlazier: This is just a type thing, as my objects above have a mix of numeric or string types, and the replacer func wants explicitly a string

community-modules/core/src/ts/utils/date.ts
AndrewGlazier: I agree, I did look for a comprehensive ISO doc, but it was quite hard to find one. As moment is a widely used library that I knew had decent docs I used their docs here https://momentjs.com/docs/#/displaying/format/ as a guideline, though I doubt we want to reference moments docs in our docs

community-modules/core/src/ts/utils/date.ts
AndrewGlazier: I did consider doing this, but decided the impact of this use case was very low, we only use this function when updating the second value of the in range filter when showing floating filters once when the user makes an input, as such I wasn't really concerned at all about performance.

This was laziness on my side though, and as you've brought it up and I've placed it in a utils function I should do it, so I'll make this change now

community-modules/core/src/ts/utils/date.ts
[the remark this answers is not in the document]
AndrewGlazier: I've moved this into the lambdas where required and tightened the typing now also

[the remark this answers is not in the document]
alantreadway: Ok - I have a personal preference for `String(value)` in this case, but we seem to be evenly split in AG-Grid mono-repo on the style you've used vs. that - so LGTM

community-modules/core/src/ts/utils/date.ts
alantreadway: Looks good - thanks!

community-modules/core/src/ts/utils/date.ts
alantreadway: @rmc-software one for you?

alantreadway:
@AndrewGlazier maybe worth checking read-only mode works as expected too?

community-modules/core/src/ts/filter/provided/date/dateFloatingFilter.ts
```diff
@@ -45,14 +46,13 @@ export class DateFloatingFilter extends SimpleFloatingFilter {
...
-        const operatingSystemLocale = Intl.DateTimeFormat()?.resolvedOptions()?.locale;
+        const format = this.filterParams.inRangeFloatingFilterDateFormat;
         if (isRange) {
-            return `${dateFrom?.toLocaleDateString(operatingSystemLocale) || null}-${dateTo?.toLocaleDateString(operatingSystemLocale) || null}`;
+            return `${dateToFormattedString(dateFrom, format)}-${dateToFormattedString(dateTo, format)}`;
         }
 
         if (dateFrom != null) {
-            return dateFrom.toLocaleDateString(operatingSystemLocale);
+            return `${dateToFormattedString(dateFrom, format)}`
```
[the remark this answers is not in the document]
AndrewGlazier: It was a side effect of the null handling, but I've changed the behaviour around that now.

community-modules/core/src/ts/utils/date.ts
rmc-software: @AndrewGlazier best if we take this one offline

AndrewGlazier:
> @AndrewGlazier maybe worth checking read-only mode works as expected too? https://ag-grid.com/javascript-data-grid/filter-api/#example-filter-api-readonly
Verified works as I would expect at least

alantreadway (approved):

community-modules/core/src/ts/utils/date.ts
rmc-software: @alantreadway @AndrewGlazier marking as resolved. We'll follow up with the doc changes later.

PR #5422, opened 2022-07-29
AG-7062 - Add series highlight label handling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7062

Attempts to fix series label obstruction by highlighted datum.

Adds a special group for highlighted item labels (in the same layer as the existing highlight, but higher z-index) to render an unobscured label for the highlighted datum. We then follow a similar processing pattern as for highlighted items - i.e. identifying the 0 or 1 highlighted label datum, and then updating the special highlighted item label selection with this.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/series/series.ts
+++ b/charts-packages/ag-charts-community/src/chart/series/series.ts
@@ -136,6 +136,8 @@ export abstract class Series<C extends SeriesNodeDataContext = SeriesNodeDataCon
     // for large-scale data-sets, where the only thing that routinely varies is the currently
     // highlighted node.
     readonly highlightGroup: Group;
+    readonly highlightNodes: Group;
+    readonly highlightLabels: Group;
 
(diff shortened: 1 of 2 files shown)


charts-packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -136,6 +136,8 @@ export abstract class Series<C extends SeriesNodeDataContext = SeriesNodeDataCon
     // for large-scale data-sets, where the only thing that routinely varies is the currently
     // highlighted node.
     readonly highlightGroup: Group;
+    readonly highlightNodes: Group;
+    readonly highlightLabels: Group;
```
manapeirov, line 140: This is a really minor nit but I feel like these should be called `highlightNode` and `hghlightLabel`

alantreadway: Done!

manapeirov (approved):

Issue #5068, opened 2022-04-07
Grid encounters "this.rowIndex is null" when serverSideDataSource changes


harish-prakash:
**I'm submitting a ...**
```
[x] bug report
[] feature request
[] support request => see 'Requesting Community Support'
```

**Providing a Reproducible Scenario**
I build the server-side datasource with a `useCallback` and change it when external filters are updated. I need to use this method because, the grid's native external filter solution (reference: [React Data Grid: External Filter](https://www.ag-grid.com/react-data-grid/filter-external/)) does not work for server-side row model, plus I cannot use the request's filterModel because the my external filter is not a column in the grid.

The worst part is, it is not always reproducible thus I cannot provide a plunkr

**Current behavior**
Line 312 in rowNode.ts causes the error because a potential `null` rowIndex is ignored with the use of `!`
https://github.com/ag-grid/ag-grid/blob/c44bb64f1939515f905e7dca6fd68321b55f8ce7/community-modules/core/src/ts/entities/rowNode.ts#L303-L313

**Expected behavior**
* **BugFix**: Handle the `null` rowIndex properly **(or)**
* Allow filterModels to work even when the column is not in the grid **(or)**
* Extend "external filtering" to server-side row model **(or)**
* Allow passing additional payloads to the server-side getRows request (something similar to passing cellRenderParams)

**AG Grid version:** 27.0.1

**Browser:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:100.0) Gecko/20100101 Firefox/100.0

**Language:** TypeScript 4.5.5

This was already reported once before but it was improperly closed off as a support request: https://github.com/ag-grid/ag-grid/issues/4767

kiril-matev:
Hi @harish-prakash!

Thank you for reporting this. 

Apologies for closing this issue earlier, this was done in error. You've submitted a bug report and we'd like to look into this further.

We're keen to start our investigation of this specific issue. We need a live sample to reproduce the issue as per our bug report template. It's fine if the plunker only reproduces the issue on every 5th attempt, we need an example which reproduces the issue. Once we have the example that reproduces the issue, we can then work on the reproduction steps.

Please provide a live example reproducing this issue.

We're looking forward to your response so we can progress this.

StephenCooper:
I am going to close this issue as without a reproduction we are unable to resolve this. 

If this is still happening in v29 then feel free to re-open.

StephenCooper:
[...]

Issue #4767, opened 2021-11-10
this.rowIndex is null in relation to cleanup in empty grids that use reactUi


fgblomqvist:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

**Current behavior**
I don't have time to create a repro atm, but this is the scenario:
1. I have a grid with `reactUi={true}` and that uses the partial serverside model, along with masterDetail
2. The grid gets initialized and fetches data, 0 rows are returned (expected) so the grid is empty
3. I update the datasource (since one of the params in getRows gets updated), which should trigger another data fetch.
4. Instead, the grid crashes with the following stack:

![Screen Shot 2021-11-10 at 6 50 40 PM](https://user-images.githubusercontent.com/649260/141166320-34021774-3715-4bb1-833a-5f169086418e.png)

Setting `reactUi` to `false` fixes it.

**Expected behavior**
It should not crash.

**Please tell us about your environment:**
React 17.0.2

* **AG Grid version:** Enterprise 26.1.0

* **Browser:**
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:93.0) Gecko/20100101 Firefox/93.0"

* **Language:** TypeScript 4.4

makinggoodsoftware:
HI,

This seems like a support question and It sounds like you have an ag-grid license/evaluating. Note that github is not our main channel for support for customers.

You should contact us through zendesk, you should have access already at https://ag-grid.zendesk.com/

If that is not the case, or you are having trouble, please contact us at info@ag-grid.com where they will be able to assist you with your access.

Thanks

fgblomqvist:
It is not a support question, it is a bug report, and while we have a license, it is more useful to post issues here where everyone can see them and find them rather than on a closed platform where it will only benefit the enterprise users.

harokevin:
Should be reopend

margarytaviter:
The error is pretty much similar to this DnD issue: https://github.com/ag-grid/ag-grid/issues/5156

kiril-matev:
Hi @fgblomqvist!

Thank you for reporting this. 

You are correct - this is a bug report and we'd like to look into this further. 

[...]

PR #5423, opened 2022-07-29
Fix stringification of RowIndex to prevent grid hard crashing

caitlinwolters-RL:
There have been several GitHub issues referencing this problem:

https://github.com/ag-grid/ag-grid/issues/5068
https://github.com/ag-grid/ag-grid/issues/4767

It is difficult to reproduce in a sandbox environment, but in real-world applications we are seeing this problem manifest throughout the grid.

Screen Shot 2022-07-29 at 12 15 40 PM

Essentially, In certain circumstances, `rowIndex` will briefly be null. If the row attempts to re-render while `rowIndex` is null, the grid hard crashes.

This often seems linked to filtering the grid in rapid succession, or updating rows when a filter is applied.

This simple fix prevents a hard crash and filtering works as intended.

### DIFF
--- a/community-modules/core/src/ts/entities/rowNode.ts
+++ b/community-modules/core/src/ts/entities/rowNode.ts
@@ -325,7 +325,7 @@ export class RowNode<TData = any> implements IEventEmitter {
             return 'b-' + this.rowIndex;
         }
 
-        return this.rowIndex!.toString();
+        return String(this.rowIndex);
     }
 
     private createDaemonNode(): RowNode {


StephenCooper:
Thanks for this PR, I am going to investigate this and the linked issues that you provided.

gportela85:
@caitlinwolters-RL @StephenCooper We never created short-circuit here on purpose, if the grid is attempting to do any operations on a null rowIndex, it's a sign that something else is wrong in the process. Preventing the grid from hard crashing here will just end up with a broken viewport anyway.

caitlinwolters-RL:
@gportela85 when we monkeypatch AG Grid to not hard crash here, our grid works as intended. Without the fix, our grid crashes when filtering multiple times in succession. 

This error appears to be the result of a race condition, which, in enterprise applications, is difficult to avoid given multiple layers of implementation.

gportela85:
@caitlinwolters-RL This is exactly the reason why we don't patch this, now it looks like everything is fine but you probably have a memory leak somewhere because functions are being called on rows that are no longer rendered and this will make it very hard to find out where. It would be better if you submitted a reproducible scenario of what you are fixing using `Plunkr`. We need to know exactly what race conditions are creating this scenario so we can patch this correctly. Simply hiding the error is not the way to go.

kiril-matev:
Hi @caitlinwolters-RL!

Thank you for looking into other issues where this was reported. 

Unfortunately none of these reports contained a live example so we didn't investigate. I've now reopened the issues you listed and asked the users for reproduction examples.

As @gportela85 mentioned we'd like to fix the underlying issue instead of apply a patch to deal with one of its symptoms. This is why we need to be able to reproduce the issue in a live example.

I understand it's difficult to build an example demonstrating an intermittent issue that only manifests itself occasionally. Even so, without a live example we're not sure what is the grid state at the time this issue occurs. This is why we can start our investigation with a live example from you that reproduces this issue even if it takes a long time to use it to reproduce it. The live scenario would answer important questions about how the grid is configured.

We're looking forward to your response so we can progress this.

genghongkai:
How to accurately add the added data to the correct position through applyTransaction when expanded  is false

genghongkai:
How to accurately add the added data to the correct position through applyTransaction when expanded is false ?

wjaspers:
Wouldn't this be easier?

```
return `${this.rowIndex}`;
```

PR #5439, opened 2022-08-04
AG-7097 - Improve AreaSeries stoke rendering to avoid overlapping with other series.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7097

Moves the `AreaSeries` stroke to be clipped by the area path, so that:
- There is no overlap between stacked series at all.
- The rendering is more accurate, in the case that the `strokeWidth` is significant and distorts the proportions of the stack.
- Strokes are clipped at the start/end and top/bottom of the series area correctly, as a side-effect.

Before:
Screenshot 2022-08-04 at 10 12 31

After:
Screenshot 2022-08-04 at 10 13 28


charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -627,10 +627,14 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
             let moveTo = true;
 
             stroke.stroke = strokes[seriesIdx % strokes.length];
-            stroke.strokeWidth = this.getStrokeWidth(this.strokeWidth, { itemId });
+            stroke.strokeWidth = this.getStrokeWidth(this.strokeWidth, { itemId }) * 2;
             stroke.strokeOpacity = strokeOpacity;
             stroke.lineDash = this.lineDash;
             stroke.lineDashOffset = this.lineDashOffset;
+            stroke.lineCap = 'round';
```
manapeirov, line 634: Should these (`lineCap` and `lineJoin`) be in `updatePaths`?

charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -627,10 +627,14 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
             let moveTo = true;
 
             stroke.stroke = strokes[seriesIdx % strokes.length];
-            stroke.strokeWidth = this.getStrokeWidth(this.strokeWidth, { itemId });
+            stroke.strokeWidth = this.getStrokeWidth(this.strokeWidth, { itemId }) * 2;
```
manapeirov, line 630: how come we * 2?

alantreadway: The intent is so that the rendered stroke width (after clipping half-way) is `strokeWidth` pixels.

alantreadway:
#5446 supersedes this PR.

PR #5440, opened 2022-08-04
AG-6966 - charts fixes for SonarCloud rule: Assignments should not be…

manapeirov:
… made from within sub-expressions

PR for: AG-6966](https://ag-grid.atlassian.net/browse/AG-6966)


alantreadway (approved):
LGTM, thanks for these!

PR #5449, opened 2022-08-05
Ag 7092/pie series radius

manapeirov:
This PR is for https://ag-grid.atlassian.net/browse/AG-7092

- Clamp `radiusKey` values which exceed the maximum available radius
- Set minimum radius to `0` if `radiusMin` has not been specified so that the smallest radius slice will be visible
- Add snapshot tests for polar series


alantreadway:
LGTM, but could we break the tests + snapshots into a distinct PR from the fixes here, so we can compare before and after (and have history we can look at in the future)?

alantreadway (approved):
LGTM - thanks for breaking the original PR up.

PR #5468, opened 2022-08-10
AG-7053 - Fix histogram scatter overlap series highlight

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7053

Specifically: Fixes scatter series highlight when used in combination with histogram series.

Generally:
- Reverses series pick order, so 'nearest' series have the opportunity to be picked first.
- Takes datum size into account, so 'dead-zone' of `0` distance extends to its extent, for point-based datums.


charts-packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -28,6 +28,7 @@ export interface SeriesNodeDatum {
         // in local (series) coordinates
         readonly x: number;
         readonly y: number;
+        readonly size: number;
```
manapeirov, line 31: Shall we extract this into its own type Point? because we have something called Coordinate in area series, it's not consistent

manapeirov (commented):
LGTM, this isn't related to this bug but we could add a task to change the implementation of `getLabelData` for `cartesianSeries` so that we can have similar behaviour to bubble series across all cartesian series for preventing label overlaps now that we have the `size` property available on the series node data (except for the bar series case)

alantreadway:
> LGTM, this isn't related to this bug but we could add a task to change the implementation of `getLabelData` for `cartesianSeries` so that we can have similar behaviour to bubble series across all cartesian series for preventing label overlaps now that we have the `size` property available on the series node data (except for the bar series case)

I've started an epic for more general label collision detection and prioritisation: https://ag-grid.atlassian.net/browse/AG-7153

I've also added a task to address this particular concern.

charts-packages/ag-charts-community/src/chart/series/series.ts
alantreadway: Good call, I've created a `Point` and `SizedPoint` type, and retrofitted it in a bunch of places.

I stopped at the Scene graph, since I was a little concerned that we might impact GC / memory allocation by switching from raw `x` and `y` parameters to an object, which may need to be continuously reconstructed as we traverse the scene graph.

alantreadway:
@manapeirov PTAL

manapeirov (approved):
Looks great thanks for making that change

PR #5472, opened 2022-08-11
Ag 2299 add context to all events

StephenCooper:
Add `gridOptions.context` to all events.


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -52,9 +50,6 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel
     @Autowired('selectionService') private selectionService: SelectionService;
     @Autowired('filterManager') private filterManager: FilterManager;
     @Autowired('valueCache') private valueCache: ValueCache;
-    @Autowired('columnApi') private columnApi: ColumnApi;
```
StephenCooper, line 55: Removed as no longer used in the class now that these are attached centrally in the EventService.

PR #5486, opened 2022-08-17
AG-7178 - Fix axis gap

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7178

The gap observed was the result of not rounding in `axisThickness` calculations, leading to later calculations `floor()`/`ceil()`ing in opposite directions for the non-pixel aligned value.


manapeirov (approved):
LGTM

PR #5493, opened 2022-08-18
Ag 7070/cross lines

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7070.

- created a new scene graph node called `range` which will draw the path for the fill of the range and also the the start and end lines of the range 
- crossLine will calculate valid canvas coordinate values for the `range` node to render correctly
- crossLine also handles creation and positioning of the label

Todo: add snapshot tests for the various label positions as I found a bug during this refactoring


charts-packages/ag-charts-community/src/scene/shape/range.ts
```diff
@@ -0,0 +1,128 @@
...
+
+        if (this.dirty === RedrawType.NONE && !forceRender) {
+            if (stats) stats.nodesSkipped += this.nodeCount.count;
+            return;
+        }
+
+        let x1 = this.x1;
+        let y1 = this.y1;
+        let x2 = this.x2;
+        let y2 = this.y2;
```
alantreadway, line 60: Nit: Use a deconstruction here.

```suggestion
        let { x1, y1, x2, y2 } = this;
```

charts-packages/ag-charts-community/src/scene/shape/range.ts
```diff
@@ -0,0 +1,128 @@
...
+        ...Shape.defaultStyles,
+        strokeWidth: 1,
+    };
+
+    constructor() {
+        super();
+        this.restoreOwnStyles();
+    }
+
+    private effectiveStrokeWidth: number = Range.defaultStyles.strokeWidth;
```
alantreadway, line 18: Why do you need this?

alantreadway (commented):
Generally LGTM aside from a few points - could we add specific test coverage for `Range`, similar to `Rect`?

manapeirov:
@alantreadway PTAL

manapeirov:
@alantreadway let me know what you think

alantreadway (approved):
LGTM - thanks for making those updates!

PR #5521, opened 2022-08-30
AG-7202 - Fix secondary axis ticks when series are toggled via legend

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7202.

Fixes ticks being set to `NaN` values when the domain is undefined.


alantreadway (commented):
LGTM

PR #5536, opened 2022-09-07
Ag 6586/options validation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-6586

Add validation to the chart options

### DIFF
--- a/charts-packages/ag-charts-community/src/caption.ts
+++ b/charts-packages/ag-charts-community/src/caption.ts
@@ -1,12 +1,14 @@
 import { Text, FontStyle, FontWeight } from './scene/shape/text';
 import { PointerEvents } from './scene/node';
 import { Observable } from './util/observable';
+import { BOOLEAN, Validate } from './util/validation';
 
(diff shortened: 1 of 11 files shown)


alantreadway:
So I have concerns with the exact approach around the `Node`, `Shape` and `Text` objects - I feel there is a strong risk that we will add significant performance issues with introducing a wrapping call around the setter.

There is also a slight disconnect here: We shouldn't be performing validations during scene-graph updates - so this is perhaps highlighting an architectural defect (i.e. that users can directly set properties on the scene graph).

My prefered approach is to separate these concerns - which may mean introducing 'model' objects which receive user/theme options (and have @Validate() on their fields), and then during scene update we apply these to the (unvalidated) scene-graph?

alantreadway:
> So I have concerns with the exact approach around the `Node`, `Shape` and `Text` objects - I feel there is a strong risk that we will add significant performance issues with introducing a wrapping call around the setter.

Could we add @Validate() decorators to these (wrapping) setters perhaps?

PR #5543, opened 2022-09-09
AG-5220 - Treemap formatter

alexanderby:
https://ag-grid.atlassian.net/browse/AG-5220

Ability to customise TreeMap tiles' fill and stroke styles via a formatter function.


alantreadway (approved):
Nice concise change - LGTM

PR #5546, opened 2022-09-12
Ag 7234/histogram series

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7234. 

If both `binCount` and `bins` specified in the histogram series options, `binCount` will now take precedence. Integrated charts now reads from `bin` and writes to `binCount` to get an accurate count for the formatting panel.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
+++ b/charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
@@ -32,6 +32,8 @@ class HistogramSeriesLabel extends Label {
     formatter?: (params: { value: number }) => string = undefined;
 }
 
+const defaultBinCount = 10;
+
 export { HistogramTooltipRendererParams };
 
 interface HistogramNodeDatum extends SeriesNodeDatum {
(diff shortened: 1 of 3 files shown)


charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -262,6 +262,7 @@ export class HistogramSeries extends CartesianSeries<SeriesNodeDataContext<Histo
     private placeDataInBins(data: any[]): HistogramBin[] {
         const { xKey } = this;
         const derivedBins = this.deriveBins();
+        this.bins = derivedBins;
```
alantreadway, line 265: I feel like we should avoid having mixed-source properties (options vs. calculated) as it could lead to race conditions or unclear data flows - can we avoid the reassignment to `this.bins` somehow?

manapeirov:
@alantreadway Let me know if I should revert my last commit

manapeirov:
@alantreadway PTAL

charts-packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
alantreadway: Discussed offline, although avoiding mixed-source is desireable, we don't have a solution on how to feedback defaults to Integrated Charts right now, so this is the simplest solution. @manapeirov has verified that in all update scenarios for Standalone Charts that this 'caching' of `bins` doesn't lead to unexpected update behaviour (e.g. when switching between using `binCount` and `bins`, and vice-versa).

alantreadway (approved):

PR #5547, opened 2022-09-12
AG-4460 - Move wait for chart render into chart class

taktran:
Will be useful outside of the test function


alantreadway (approved):
LGTM - thanks for doing this!

manapeirov (approved):

PR #5560, opened 2022-09-16
Ag 6586/charts options validation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-6586

- Allow change detection as well as validation to be applied to certain properties.
- Create a wrapper for background to validate charts background options rather than applying properties directly on rect node


alantreadway (approved):
LGTM

PR #5561, opened 2022-09-16
AG-7200 - Allow setting padding of bars and columns to 0

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7200. 

- Added `groupPaddingInner` property to control padding between bars/columns within a group
- Remove rounding of groupScale coordinates if the `groupScale.padding` is set to `0px` as rounding prevents exactly zero padding
- Render all charts as before unless `groupPaddingInner` is supplied


alantreadway (approved):
LGTM

PR #5570, opened 2022-09-21
AG-5198 Doughnut centre colour

alexanderby:
https://ag-grid.atlassian.net/browse/AG-5198

Ability to set the Doughnut's centre colour with `innerCircle.fill` property.


charts-packages/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -1198,11 +1198,18 @@ export interface AgDoughnutInnerLabel {
     /** The colour to use for the inner label. */
     color?: CssColor;
     /** The margin in pixels before and after the inner label. */
-    margin?: number;
+    margin?: PixelSize;
```
alantreadway, line 1201: Good spot - thanks!

charts-packages/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -1254,6 +1261,8 @@ export interface AgPieSeriesOptions<DatumType = any> extends AgBaseSeriesOptions
     tooltip?: AgPieSeriesTooltip;
     /** Text lines to display inside the doughnut chart */
     innerLabels?: AgDoughnutInnerLabel[];
+    /** The centre of the doughnut chart */
```
alantreadway, line 1264: I would maybe expand the description a little here, as this is within the context of a `AgPieSeriesOptions` not a doughnut-specific options object?

E.g. `Configuration for the area inside the series, typically only visible when rendering a doughnut-type chart by using innerRadiusOffset`

Maybe we should revise the description for `innerLabels` similarly?

grid-packages/ag-grid-docs/documentation/doc-pages/charts-pie-series/index.md
```diff
@@ -134,6 +134,17 @@ series: [{
...
+
+```js
+series: [{
+    ...
+    innerCircle: {
+        fill: 'aliceblue',
+    }
+}]
+```
+
```
alantreadway, line 147: I would merge these with the previous example perhaps? Also, could we update the Plunker example to include a fill too?

alexanderby: From the design point of view, the colouring of the center should be avoided or done professionally with some precise colours :) I would prefer to not include it into the sample, so that people, who copy/paste the configuration from the docs, would stick to a better example :)

alantreadway (approved):
LGTM - thanks for making those updates :-)

PR #5574, opened 2022-09-21
AG-5117 Pie labels inside segments

alexanderby:
https://ag-grid.atlassian.net/browse/AG-5117

Ability to put the labels inside the pie segments using ~~`series.label.position`~~ `series.sectorLabel` property.


alexanderby:
Sorry if already renamed some descriptions, that should have been changed in another PR.

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -717,9 +817,53 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+                    const corners = [
+                        [bbox.x, bbox.y],
+                        [bbox.x + bbox.width, bbox.y],
+                        [bbox.x + bbox.width, bbox.y + bbox.height],
+                        [bbox.x, bbox.y + bbox.height],
+                    ];
+                    const { startAngle, endAngle } = datum;
+                    const sectorBounds = { startAngle, endAngle, innerRadius, outerRadius };
+                    if (corners.every(([x, y]) => isPointInArc(x, y, sectorBounds))) {
+                        text.fill = color;
```
alantreadway, line 860: It would be better to toggle `Text.visible` to turn the node off/on, as the `Group.render()` will not call `Text.render()` if the child node is `visible === false`, the current approach will still do most of the work (in our code) to render the `Text`.

charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -463,6 +465,15 @@ export class ChartTheme {
                     offset: 3,
                     minAngle: 20,
                 },
+                sectorLabel: {
+                    enabled: true,
```
alantreadway, line 469: Should this be `enabled: false` so there are no visual changes unless a user opts-in to using `sectorLabels`?

alexanderby: I see the other options like `label` and `title` have `enabled: true`. The sector label will remain invisible unless there will be a `sectorLabelKey`. Same way it works for the usual `label`.

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
alexanderby: Thanks for the hint, I see the usual `label` manages this with `fill` too. I will fix it there in another PR.

charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
alantreadway: Ok, makes sense - thanks for clarifying!

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -717,9 +817,54 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+        const { radiusScale } = this;
+        const innerRadius = radiusScale.convert(0);
+        const { fontSize, fontStyle, fontWeight, fontFamily, offset, color } = this.sectorLabel;
+
+        this.sectorLabelSelection.each((text, datum) => {
+            const sectorLabel = datum.sectorLabel;
+            const radius = radiusScale.convert(datum.radius, clamper);
+            const outerRadius = Math.max(0, radius);
+
+            text.visible = false;
```
alantreadway, line 834: Thanks for adding use of `visible` here :+1:

As discussed offline, we should avoid flipping between values, as every set could invalidate change-detection. Would you mind reworking the code here to avoid doing that?

alantreadway (approved):
LGTM - thanks for all the tweaks & updates!

PR #5583, opened 2022-09-22
Ag 6879/export chart as image

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-6879

Allow downloading chart as image in a specified size without flashing/ re-rendering of the visible chart

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartV2.ts
@@ -82,7 +82,13 @@ type SeriesOptionType<T extends Series> = T extends LineSeries
     ? AgTreemapSeriesOptions
     : never;
 
-type DownloadOptions = { width?: number; height?: number; fileName?: string; fileFormat?: string };
+type DownloadOptions = {
+    width?: number;
+    height?: number;
+    fileName?: string;
+    /** The standard MIME type for the image format to return. If you do not specify this parameter, the default value is a PNG format image. See `Canvas.toDataURL()` */
+    fileFormat?: string;
+};
 
 function chartType<T extends ChartType>(options: ChartOptionType<T>): 'cartesian' | 'polar' | 'hierarchy' {
     if (isAgCartesianChartOptions(options)) {
@@ -211,6 +217,10 @@ export abstract class AgChartV2 {
         });
     }
 
+    /**
+     * Returns the content of the current canvas as an image.
+     * @param opts The download options including `width` and `height` of the image as well as `fileName` and `fileFormat`.
+     */
     static download(chart: Chart, opts?: DownloadOptions) {
         let { width, height, fileName, fileFormat } = opts || {};
         const currentWidth = chart.width;
--- a/charts-packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartV2.ts
@@ -82,6 +82,8 @@ type SeriesOptionType<T extends Series> = T extends LineSeries
     ? AgTreemapSeriesOptions
     : never;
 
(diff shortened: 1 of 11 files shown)


charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -82,6 +82,8 @@ type SeriesOptionType<T extends Series> = T extends LineSeries
     ? AgTreemapSeriesOptions
     : never;
 
+type DownloadOptions = { width?: number; height?: number; fileName?: string; fileFormat?: string };
```
alantreadway, line 85: Nit: Should we be more explicit with the typing of `fileFormat`? If we're ultimately passing this to a `Canvas` API, can we re-use the typing for this parameter from there (assuming it's stricted that `string`)?

charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -202,6 +209,35 @@ export abstract class AgChartV2 {
...
 
+    static download(chart: Chart, opts?: DownloadOptions) {
+        let { width, height, fileName, fileFormat } = opts || {};
+        const currentWidth = chart.width;
+        const currentHeight = chart.height;
+
+        width = width ?? currentWidth;
+        height = height ?? currentHeight;
+
+        if (currentWidth !== width || currentHeight !== height) {
```
alantreadway, line 220: We probably need to check `devicePixelRatio` is `1` too here, otherwise we'll get inconsistent behaviour?

charts-packages/ag-charts-community/src/chart/agChartV2.ts
manapeirov: Canvas API: "The desired file format and image quality may be specified. If the file format is not specified, or if the given format is not supported, then the data will be exported as image/png. In other words, if the returned value starts with data:image/png for any other requested type, then that format is not supported."

`    /**
     * Returns the content of the current canvas as an image that you can use as a source for another canvas or an HTML element.
     * @param type The standard MIME type for the image format to return. If you do not specify this parameter, the default value is a PNG format image.
     */
    toDataURL(type?: string, quality?: any): string;`

manapeirov:
@alantreadway PTAL

charts-packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway: Ah, web 'API's :-P

Ok, maybe just document and point to `Canvas.toDataURL()` in a JSDoc?

charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -202,6 +211,37 @@ export abstract class AgChartV2 {
...
+            };
+
+            const clonedChart = AgChartV2.create(options as any);
+
+            clonedChart.waitForUpdate().then(() => {
+                clonedChart.scene.download(fileName, fileFormat);
+                clonedChart.destroy();
+            });
+        }
+    }
```
alantreadway, line 243:
```suggestion
        if (unchanged) {
            chart.scene.download(fileName, fileFormat);
            return;

        const options = {
            ...chart.userOptions,
            container: document.createElement('div'),
            width,
            height,
            autoSize: false,
            overrideDevicePixelRatio: 1,
        };

        const clonedChart = AgChartV2.create(options as any);

        clonedChart.waitForUpdate().then(() => {
            clonedChart.scene.download(fileName, fileFormat);
            clonedChart.destroy();
        });
    }
```

alantreadway: Nit: prefer less nesting when possible :-)

rmc-software (approved):

Issue #5586, opened 2022-09-22
AgGrid inside of WebComponents - Reorder Columns not possible


jogibear9988:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

**Customers of AG Grid**

If you are a customer you are entitled to use the AG Grid's customer support system (powered by Zendesk). Please use that channel for guaranteed response from the AG Grid team with regards bugs, feature requests and support.

**Requesting Community Support**

If you are not a customer of AG Grid, ag-grid staff will label your issue as managed-by-the-community. This means that AG Grid staff is not going to be actively looking into it and it will get closed if inactive for more than one month. The community is welcome to help with this question/support issue.

**Providing a Reproducible Scenario**
Accepted reproducible scenarios are
- A description of the detailed steps to reproduce your behaviour in one of our examples in the docs.
- A plunker

If you decide to send us a plunkr, from any example in our website use the plunkr button in there to fork your own code by following the steps below:

- Select the framework that is appropriate to you from the drop-down
- Open it in plunker. (Use the button plunker in our example)
- Add your changes so that the behaviour is reproduced
- Save and Freeze the plunker(On the top left corner)
- Send us the link to the plunker(You can copy the URL from the browser)

If reporting a bug make sure to state.

Current behaviour.
Expected behaviour. If possible back this up with our docs/examples if possible

**Current behavior**
Reordering of Columns is not possible cause the call to eDocument.elementsFromPoint in the DragAndDropService.prototype.findCurrentDropTarget does not return the elements from a shadowRoot

**Expected behavior**
It should be possibe to use the shadowRoot of the grid div:
var elementStack = this.gridOptionsWrapper.eGridDiv.getRootNode().elementsFromPoint(mouseEvent.clientX, mouseEvent.clientY);

**Please tell us about your environment:**
Windows120
* **AG Grid version:** X.X.X
28.1.1

jogibear9988:
change was introduced here: https://github.com/ag-grid/ag-grid/commit/fea16a6ed97662fbd9ee5050dd9db238c3e93932

VitaliiK91:
@jogibear9988 I think we are discussing same issues and ways to fix it here https://github.com/ag-grid/ag-grid/issues/5358

jogibear9988:
[...]

PR #5587, opened 2022-09-22
Update dragAndDropService.ts

jogibear9988:
fix for #5586 

Here is a fiddle showing the issue: https://jsfiddle.net/gxL15fum/
And if you remove the comment from line 7 to 67 the issue is fixed.
the code is mostly the same then yours in the patched method
fix is in this line: https://github.com/ag-grid/ag-grid/blob/521cb1c4f453c7cbe3bd4cc551a2164a6c1f7690/community-modules/core/src/ts/dragAndDrop/dragAndDropService.ts#L341


gportela85:
Hi @jogibear9988 thanks for pushing this, however, the change required for this fix is a bit bigger. The DragAndDrop service should not be adding elements outside of the ShadowDOM. I have pushed a fix for it that should be going into AG Grid v29.

PR #5598, opened 2022-09-26
AG-7298 - Add TimeAxis min+max properties

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7298

Improves `time` axis consistency with other axes types by adding `min` and `max` properties to allow override of the calculated domain of a `time` axis.


charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
```diff
@@ -43,23 +40,44 @@ export class TimeAxis extends ChartAxis<TimeScale> {
...
 
+    @Validate(OPT_DATE)
+    private _min?: Date = undefined;
+    set min(value: Date | undefined) {
+        if (this._min === value) {
+            return;
+        }
+
+        this._min = value;
+        this.scale.domain = [value!, this.scale.domain[1]];
```
manapeirov, line 59: does this mean we could be setting the domain to undefined?

charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
```diff
@@ -43,23 +40,44 @@ export class TimeAxis extends ChartAxis<TimeScale> {
...
-        this._domain = domain;
         this.setDomain(domain);
     }
+
     get domain(): Date[] {
         return this.scale.domain;
     }
 
+    @Validate(OPT_DATE)
+    private _min?: Date = undefined;
```
manapeirov, line 52: I think min and max should be initialised as `NaN` like in `numberAxis.ts`

charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
alantreadway: Yes - what should we do in this scenario if that isn't the right approach?

charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
alantreadway: Ok, I can make that change :+1:

charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
manapeirov: I think same as in numberAxis.ts would be good:
```@Validate(NUMBER_OR_NAN())
    protected _min: number = NaN;
    set min(value: number) {
        if (this._min !== value) {
            this._min = value;
            if (!isNaN(value)) {
                this.scale.domain = [value, this.scale.domain[1]];
            }
        }
    }
    get min(): number {
        return this._min;
    }```

manapeirov: We could even set it to `this._min` instead of `value` to make sure it's a Date

charts-packages/ag-charts-community/src/chart/axis/timeAxis.ts
alantreadway: On review, not sure that makes sense for `Date` - we end up with a type of `Date | NaN`, which seems a bit wierd.

alantreadway:
@manapeirov PTAL - can we discuss the `domain.scale` update offline if we aren't in alignment?

alantreadway:
@manapeirov PTAL, I've fixed the discrepancy between `TimeAxis` and `NumberAxis` domain handling, they are now working more consistently, and in the `AgChart.update()` case :tada:

manapeirov (approved):
LGTM!

PR #5608, opened 2022-09-27
AG-7290 Rename Pie series callout and label

alexanderby:
- Use series.calloutLine and series.calloutLabel instead.
- ~~Use Proxy to display warnings when assigning deprecated properties.~~


charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -252,9 +265,83 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+            get: (target, prop: keyof PieSeriesCalloutLabel) => {
+                return target[prop];
+            },
+            set: (target: any, prop: keyof PieSeriesCalloutLabel, value: any) => {
+                this._showLabelDeprecationWarning();
+                target[prop] = value;
+                return true;
+            },
+        });
+    }
```
alantreadway, line 289: I like the approach :-) This looks like it could be a generically applicable pattern for these types of deprecated property object - is there any chance we could refactor this into a reusable `@DeprecatedObject()` (or similarly named) decorator? This would also declutter the code and make the intent much clearer, if possible?

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -252,9 +265,83 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+        doOnce(
+            () =>
+                console.warn(
+                    'AG Charts - the use of series.callout in the pie chart series is deprecated. Please use series.calloutLine instead.'
+                ),
+            'deprecated use of series.callout property in pie chart series'
+        );
+    }
+    private _calloutLine = new PieSeriesCalloutLine();
+    private _createCalloutDeprecationProxy() {
```
alantreadway, line 318: Ditto

alantreadway (commented):
Nice - something to note, we need to add details to the JIRA 'Deprecation Notes' field about all of these changes, as that is what is used as part of the release-note creation process.

charts-packages/ag-charts-community/src/util/deprecation.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        const warnRenamed = () => warnDeprecated(key, `Use [${newPropName}] instead.`);
+        Object.defineProperty(target, key, {
+            get: function () {
+                warnRenamed();
+                return this[newPropName];
+            },
+            set: function (value: any) {
+                if (value !== this[newPropName]) {
```
alantreadway, line 56: Is this guard necessary?

charts-packages/ag-charts-community/src/util/deprecation.ts
```diff
@@ -0,0 +1,65 @@
...
+            return;
+        }
+
+        const symbol = Symbol(`__${key}__`);
+        Object.defineProperty(target, key, {
+            get: function () {
+                return this[symbol];
+            },
+            set: function (value: any) {
+                if (value !== def) {
```
alantreadway, line 29: Is this guard necessary?

alantreadway (commented):
Generally LGTM! Can you check if Integrated Charts needs changes too?

charts-packages/ag-charts-community/src/util/deprecation.ts
alexanderby: I don't know, it used to be there before. I guess if prevents the warning, when you set the default value.

charts-packages/ag-charts-community/src/util/deprecation.ts
alexanderby: It should prevent the warning, when setting the default value. But maybe the class properties can be unassigned. Not sure what it transpiles into. For some reason we have `someProperty = undefined` everywhere.

charts-packages/ag-charts-community/src/util/deprecation.ts
alantreadway: Ok, I can see this was the pattern we had before, so it's not new.

grid-packages/ag-grid-docs/documentation/doc-pages/charts-pie-series/index.md
```diff
@@ -40,18 +40,18 @@ Now we get labels, a legend, and the tooltips will also show labels along with t
...
 Each individual sector can be toggled on and off via the legend.
 
-You might notice that not all of the sectors in the chart above have a label. The reason for this is that certain sectors can be small, and if there's a cluster of small sectors their labels will overlap, resulting in a messy chart. To prevent this from happening the series will only show labels for sectors with an angle greater than a certain value, which by default is set to be `20` degrees. This value is adjustable via the `label.minAngle` config:
+You might notice that not all of the sectors in the chart above have a label. The reason for this is that certain sectors can be small, and if there's a cluster of small sectors their labels will overlap, resulting in a messy chart. To prevent this from happening the series will only show labels for sectors with an angle greater than a certain value, which by default is set to be `20` degrees. This value is adjustable via the `calloutLabel.minAngle` config:
 
 ```js
 series: [{
     ...
-    label: {
+    lacalloutLabelbel: {
```
alantreadway, line 48: Nit: Typo?

alantreadway (approved):
LGTM - we should get Rob to review the Integrated Charts changes too.

alexanderby:
There is a particular commit, that touches the integrated charts https://github.com/ag-grid/ag-grid/pull/5608/commits/4fa38edd9c2bf91c55b84d5b90e9a8d279979def
My guess was that setFont can be executed when getSelectedSeries: () => ChartSeriesType returns a different chart type (other than). So this `getSelectedSeries()` is executed inside of `setFont(...)` every time.

PR #5609, opened 2022-09-27
Enable support for `TValue` typing of `ValueFormatterParams` when using `ValueFormatterFunc`

astorije:
At the moment, we use something like this:

```tsx
const numericFormatter = ({
  value,
}: ValueFormatterParams<TData, number | undefined>) =>
  value === undefined ? '' : round(value, 2).toFixed(2);
```

This is great, but type safety can be improved a little further with this change:

```tsx
const numericFormatter = ({
  value
}): ValueFormatterFunc<TData, number | undefined> =>
  value === undefined ? '' : round(value, 2).toFixed(2);
```


StephenCooper:
TValue has been included.

PR #5620, opened 2022-09-29
Ag 7293/chart image export docs

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7293.

Not merging for now as it doesn't work in for the different frameworks


alantreadway:
We can close this as it's now redundant.

PR #5627, opened 2022-09-30
AG-7289 Rename Pie Slice or Segment to Sector

alexanderby:
Renamed pie slices and segments in docs, comments, file names.

According to Google Trends, Pie Slice is the most popular naming. Pie Sector is a bit less popular, but is a correct trigonometry term.


alantreadway (approved):
LGTM

alexanderby:
Screenshots were also renamed. Please be aware, that `npm run test-update` doesn't remove the old screenshots, so there could be some old unused.

alantreadway (approved):

PR #5636, opened 2022-10-03
RTI-1072 Hide Doughnut inner circle when no data

alexanderby:
The circle was still visible when all data was hidden through legend.


alantreadway:
LGTM - we should add a regression test for this case, but maybe after code-freeze?

alantreadway (approved):

PR #5637, opened 2022-10-03
RTI-1074 Hide doughnut inner circle when radius <0

alexanderby:
When the chart's radius appears to be less than the `innerRadiusOffset`, the doughnut inner circle should be hidden.

In future we should warn when outer radius is smaller than inner radius or less than 0.


alantreadway (approved):
LGTM

PR #5640, opened 2022-10-03
RTI-1070 Put single pie sector label in the center

alexanderby:


alantreadway (approved):
LGTM - we should add a regression test for this case post-release too.

PR #5645, opened 2022-10-05
AG-7323 - New gallery thumbnail generator

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7323

Moves gallery thumbnail generation into `ag-charts-community` to leverage `node-canvas`-based rendering we already have there.

Generation time is now <10s.

Bonus:
- Adds missing test coverage for the `cross-lines` gallery example.


manapeirov (approved):
Looks good, it seems like the thumbnails have less padding around them though, I think it looked better previously so it doesn't seem so squished. What do you think

alexanderby:
Simple Pie labels didn't fit in the screenshot

alantreadway:
I'm going to merge to unblock gallery updates as part of AG-7292 and follow-up on tweaks in a follow-up PR.

PR #5646, opened 2022-10-05
AG-7179 - SonarQube warning fixes.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7179

Fixes the remaining low-hanging fruit in terms of SonarQube lint warnings.


charts-packages/ag-charts-community/src/util/labelPlacement.ts
```diff
@@ -91,22 +69,19 @@ export function placeLabels(
             const y = d.point.y - r - l.height - padding;
             const { width, height } = l;
 
-            const withinBounds =
-                !bounds || rectContainsRect(bounds.x, bounds.y, bounds.width, bounds.height, x, y, width, height);
+            const withinBounds = !bounds || rectRectOverlap(bounds, x, y, width, height);
```
manapeirov, line 72: this was `rectContainsRect` before and now it's `rectRectOverlap`, could it be not overlapping but also not contained?

alantreadway: Reasonable question - given the nature of the calculation (placing labels), this seemed to be quite artificially restricting the displayed labels to be completely within the series area (even a 1px over would remove the label). Given that labels is an area we want to complete redesign, as a simplification I don't think it's a major issue?

manapeirov: I think series labels should be rendered only inside the series rect and shouldn't be allowed to overflow onto anything else including the axes

alantreadway: Ok, as the intent of this change is to fix SonarQube warnings and not change behaviour, I've reverted this aspect of the change - thanks for flagging!

alantreadway:
@manapeirov PTAL

manapeirov (approved):
Perfect thanks Alan

PR #5657, opened 2022-10-07
AG-7356 - Make Chart colour validations case-insensitive.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7356

Makes web-colour validation and use case-insensitive.


manapeirov (approved):
LGTM

PR #5662, opened 2022-10-10
Ag 7056/axis labels placement

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7056

There are two problems with the current axis label placement solution. The first is that we assume all labels are roughly the same size, so irregular sized labels may still overlap. The second is that (for continuous axes) we keep the first and last labels regardless of overlap, meaning the second to last and last labels may overlap. For numeric/linear axes we have a heuristic to calculate an approximate tick count which aims to avoid overlap, but this is a rough calculation and is not applied for any of the other axes.

The new approach involves re-calculating the ticks with a new tick count and nice domain rather than hiding them after they have been generated. The ticks are generated, formatted, positioned and rotated based on the configuration and tested for overlaps until a set of ticks in a particular orientation is produced with no overlap.

For category axis this isn’t possible (ticks will always be the domain) so we generate the ticks once, format them, if they overlap, first attempt to rotate them, and then if they still overlap, remove ticks based on the available space. The ticks are dropped evenly to allow for regularly spaced labels and grid lines.

The label bbox calculations have been improved to give more accurate measurements, removing the assumption that all labels are the same size, and fixing the issue with overlap of larger labels.

If tick count is provided as a time interval or a number, we generate the ticks once with this tick count, but drop the ticks just like we would with category axis if there is overlap. As the chart is resized and there is space, the labels will appear again. So the tick count will be honored only if there is enough space. If the tickCount is ridiculously large (more than 1000 ticks) there will probably be some overlap as we only loop a maximum of 10 times, each time halving the number of displayed ticks. 

There is still an issue with the first and last axis ticks sometimes bleeding out of the chart rect and getting clipped, this requires separate analysis.

Profiling using our `window.agChartsSceneStats = true` showed the performance is similar to before changes in this PR. No noticeable difference is seen in layout calculations for axes that have larger labels with more characters. We were computing the text bounding boxes in `updateLabels` before so no significant difference in the number of canvas calls to calculate text size as the loop iterates a maximum of 10 times but each time with less labels/text to compute. The difference now is the added matrix transformation to get the the bbox in the desired orientation for label overlap calculations and also the repeated calls to the tick interval calculation algorithm `scale.ticks()`.


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -382,22 +380,14 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
-    protected onLabelFormatChange(format?: string) {
+    protected onLabelFormatChange(format?: string, ticks?: any[]) {
         const { scale } = this;
         if (format && scale && scale.tickFormat) {
             try {
                 this.labelFormatter = scale.tickFormat({
-                    ticks: this.getTicks(),
-                    count: this.calculatedTickCount,
+                    ticks: ticks ?? this.getTicks(),
+                    count: ticks?.length ?? this.tick.count,
```
alantreadway, line 390: In the RHS of the `??`, should we be using `this.getTicks()` rather than `this.tick` as the basis as per the previous line?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -546,29 +536,76 @@ export class Axis<S extends Scale<D, number>, D = any> {
         const regularFlipRotation = normalizeAngle360(rotation - Math.PI / 2);
         const halfBandwidth = (scale.bandwidth || 0) / 2;
 
-        const ticks = this.getTicks();
+        let count = 0;
+        let labelOverlap = true;
+        let ticks: any[] = [];
```
alantreadway, line 541: Can we tighten this `any`?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -546,29 +536,76 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
-            ticks,
-        });
-        const gridlineGroupSelection = this.updateGridLines({
+        scale.domain = this.dataDomain;
+
+        if (this.nice && scale.nice) {
+            scale.nice!(this.tick.count);
+        }
+
+        while (labelOverlap && count < defaultTickCount) {
```
alantreadway, line 554: `count < defaultTickCount` seems a bit suspect? `count` is the iteration count, `defaultTickCount` is the default number of ticks?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -546,29 +536,76 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
+            let unchanged = true;
+            while (unchanged && count < defaultTickCount) {
```
alantreadway, line 556: Ditto

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -546,29 +536,76 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
+                const prevTicks = ticks;
+
+                const filteredTicks =
+                    (continuous && !tickCount) || count === 0 ? undefined : ticks.filter((_, i) => i % 2 === 0);
+
+                if (calculatePrimaryDomain) {
+                    scale.domain = this.dataDomain;
```
alantreadway, line 563: This is a repeat of line 548 - is it needed? (Does `scale.nice()` mutate `scale.domain`?

charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -657,30 +686,69 @@ export class Axis<S extends Scale<D, number>, D = any> {
...
+    }) {
+        const { scale } = this;
+        ticks = ticks ?? this.getTicks(tickCountOffset);
+        const data = ticks.map((t) => ({ tick: t, translationY: Math.round(scale.convert(t) + halfBandwidth) }));
+        const gridlineGroupSelection = this.updateGridLineGroupSelection({ gridLength, data });
+        const tickGroupSelection = this.updateTickGroupSelection({ data });
+
+        const translationYFn = (_: unknown, datum: any) => datum.translationY;
+        gridlineGroupSelection.attrFn('translationY', translationYFn);
+        tickGroupSelection.attrFn('translationY', translationYFn);
```
alantreadway, line 711: It might be simpler to inline this, as you can then lose the irrelevant types?
```suggestion
        gridlineGroupSelection.attrFn('translationY', (_, datum) => datum.translationY);
        tickGroupSelection.attrFn('translationY', (_, datum) => datum.translationY);
```

charts-packages/ag-charts-community/src/scale/timeScale.ts
```diff
@@ -158,18 +158,26 @@ export class TimeScale extends ContinuousScale {
...
+        start: number;
+        stop: number;
+        step?: number;
+        offset?: number;
+    }): CountableTimeInterval | TimeInterval | undefined {
         if (typeof interval === 'number') {
-            const tickCount = interval;
+            const tickCount = Math.max(0, interval - (offset ?? 0));
             const tickIntervals = this.tickIntervals;
+            const n = tickIntervals.length;
```
alantreadway, line 177: Is `n` strictly needed?

alantreadway:
Could you perform some profiling on the new approach to determine how much impact there is on layout time?

charts-packages/ag-charts-community/src/axis.ts
manapeirov: The ticks here are the ones passed into the function rather than `this.ticks`

charts-packages/ag-charts-community/src/axis.ts
manapeirov: Type of ticks has always been any[], wasn't thinking about changing that as part of this PR unless you think I should?

charts-packages/ag-charts-community/src/axis.ts
manapeirov: It's needed to start back at the original data array rather than modifying the modified domain based on the previous iteration/tick count calculations

charts-packages/ag-charts-community/src/axis.ts
manapeirov: sure

charts-packages/ag-charts-community/src/scale/timeScale.ts
manapeirov: not really

charts-packages/ag-charts-community/src/axis.ts
manapeirov: yes scale.nice() changes the scale.domain

charts-packages/ag-charts-community/src/axis.ts
alantreadway: Understood, but it looks like there is a scenario where we're passing a value for `ticks` which is based on `this.getTicks()` and `count` is based on `this.tick.count` - this seems inconsistent to me?

charts-packages/ag-charts-community/src/axis.ts
alantreadway: I was more concerned about `defaultTickCount` :D It doesn't seem to conceptually match here as the limit of iterations?

manapeirov: Switched it up a bit and added comments for more context

manapeirov:
@alantreadway PTAL

alantreadway (approved):
Looks great, nice work! LGTM.

PR #5679, opened 2022-10-12
AG-7233 Fix Legend/Tooltip formatter color mismatch

alexanderby:
- Run formatter during the data processing stage.
- Fix tooltip formatted color mismatch.
- Add `fillOpacity` to formatter return type.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/agChartOptions.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartOptions.ts
@@ -1194,6 +1194,7 @@ export interface AgPieSeriesFormatterParams<DatumType> {
 
 export interface AgPieSeriesFormat {
     fill?: CssColor;
+    fillOpacity?: Opacity;
     stroke?: CssColor;
     strokeWidth?: PixelSize;
 }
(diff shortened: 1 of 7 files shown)


alantreadway:
LGTM - I think the approach of pre-calculating the colours in `processData()` works for `PieSeries`, although we might want to be more careful adopting this approach elsewhere (`PieSeries` is inherently only going to have O(10) entries / callbacks by virtue of the amount of data that can be sensibly displayed).

alantreadway (approved):

alexanderby:
> LGTM - I think the approach of pre-calculating the colours in `processData()` works for `PieSeries`, although we might want to be more careful adopting this approach elsewhere (`PieSeries` is inherently only going to have O(10) entries / callbacks by virtue of the amount of data that can be sensibly displayed).

The issue seems to be relevant to Pies only, because the legend represents a single slice on the chart. While in other chart types it represents a whole category, and formatters edit a particular marker or a bar.

To avoid calling this formatter twice in `processData` and `update` methods, we will have to make the `renderLegend` process harder, split it into calculating the layout and rendering the legend items.

PR #5686, opened 2022-10-14
AG-7224 Datum ref in Line tooltip & click handler

alexanderby:
Provide a correct reference to a datum object for node click handler and tooltip renderer.


charts-packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -267,11 +277,11 @@ export class LineSeries extends CartesianSeries<LineContext> {
                             : '';
                 }
 
-                const seriesDatum = { [xKey]: xDatum, [yKey]: yDatum };
+                const datum = point.datum;
```
alantreadway, line 280: Nit: Maybe just unwrap this on line 242 with the `xDatum` and `yDatum`?

PR #5694, opened 2022-10-17
AG-7297 Recalculate layout when navigating X-axis

alexanderby:
Compare visible labels and rotation angles when dragging the navigator.

### DIFF
--- a/charts-packages/ag-charts-community/src/axis.ts
+++ b/charts-packages/ag-charts-community/src/axis.ts
@@ -378,6 +378,24 @@ export class Axis<S extends Scale<D, number>, D = any> {
         return this._visibleRange.slice();
     }
 
+    getVisibleLabels(): string[] {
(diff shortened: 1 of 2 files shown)


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -378,6 +378,24 @@ export class Axis<S extends Scale<D, number>, D = any> {
         return this._visibleRange.slice();
     }
 
+    getVisibleLabels(): string[] {
+        const labels: string[] = [];
```
alantreadway, line 382: I don't think we should be querying the state of the scene graph - normally data flows from opts -> models -> scene-graph, so reading model state back from scene-graph is going against the established flow.

Perhaps we can cache this information at the end of `Axis.updateLabels()` instead, and just allow it to be read here?

PR #5701, opened 2022-10-18
AG-7185 - Localise blank and not blank in filter text box

taktran:
## Changes

* Localise blank and not blank in filter text box

## Screenshot

Docs: https://ag-grid.com/javascript-data-grid/localisation/#example-localisation

Before (not localised)

Screenshot 2022-10-18 at 5 14 37 pm

Screenshot 2022-10-18 at 5 14 44 pm

After (localised)

Screenshot 2022-10-18 at 5 13 00 pm

Screenshot 2022-10-18 at 5 13 06 pm

### DIFF
--- a/community-modules/core/src/ts/filter/floating/provided/simpleFloatingFilter.ts
+++ b/community-modules/core/src/ts/filter/floating/provided/simpleFloatingFilter.ts
@@ -37,7 +37,7 @@ export abstract class SimpleFloatingFilter extends Component implements IFloatin
     // used by:
     // 1) NumberFloatingFilter & TextFloatingFilter: Always, for both when editable and read only.
     // 2) DateFloatingFilter: Only when read only (as we show text rather than a date picker when read only)
-    protected getTextFromModel(model: ProvidedFilterModel): string | null {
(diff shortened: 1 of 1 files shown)


community-modules/core/src/ts/filter/floating/provided/simpleFloatingFilter.ts
```diff
@@ -52,6 +52,12 @@ export abstract class SimpleFloatingFilter extends Component implements IFloatin
                 combinedModel.operator,
                 customOption2,
             ].join(' ');
+        } else if (
+            model.filterType === 'text'
+            && (model.type === 'blank' || model.type === 'notBlank')
```
alantreadway, line 57: Nit: Maybe consider using these constants so if there is a change to these symbols it'll be easier to spot in the future?
https://github.com/ag-grid/ag-grid/blob/774853e988807770385555d747158aed5aba33fa/community-modules/core/src/ts/filter/provided/simpleFilter.ts#L91-L92

taktran: Good shout, didn't see that. Updated.

PR #5706, opened 2022-10-19
Ag 7056/axis labels placement

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7056. 

Fixes label jittering at some pixels when chart is resized.

This was occuring due to the rounding of y translation values during axes.update. This rounding isn't required, the axis grid pixel alignment should occur at the scene node level when rendering the individual line, shape and text components.


alantreadway (approved):
LGTM

PR #5707, opened 2022-10-19
AG-7398 + AG-7399 Fix primary/secondary axis ticks

alexanderby:
- Start the primary axis grid lines at the same place as the secondary.
- Use the proper ticks count for primary/secondary axes.
- Merge the primary/secondary ticks together if nice=true.
- Properly use axes min/max when nice=false, don't try to align the primary/secondary ticks.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/cartesianChart.ts
+++ b/charts-packages/ag-charts-community/src/chart/cartesianChart.ts
@@ -287,6 +287,8 @@ export class CartesianChart extends Chart {
             return fn(value, compareTo);
         };
 
+        const axesByPosition: Map<ChartAxisPosition, ChartAxis[]> = new Map();
+
         // Set the number of ticks for continuous axes based on the available range
         // before updating the axis domain via `this.updateAxes()` as the tick count has an effect on the calculated `nice` domain extent
         axes.forEach((axis) => {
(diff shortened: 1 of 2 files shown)


charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -381,8 +383,26 @@ export class CartesianChart extends Chart {
...
+                axesByPosition.set(position, samePosAxes);
+            }
+            samePosAxes.forEach((otherAxis) => (otherAxis.gridPadding += axisThickness));
+            samePosAxes.push(axis);
         });
 
+        axesByPosition.forEach((axes) =>
+            axes.forEach((axis) => {
+                if (axis.gridPadding !== 0) {
+                    axis.update(axis.nice ? primaryTickCounts[axis.direction] : undefined);
```
alantreadway, line 401: This extra `Axis.update()` call makes me a little nervous, although it looks to be quite exceptional.

I'm wondering if we should prioritise trying to refactor the `Axis.update()` into smaller steps we can be more selective about invoking?

(Maybe in a follow-up PR?)

alantreadway:
LGTM, let's discuss the refactor in the office?

alantreadway (approved):

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
alexanderby: Yes, it would be great to have it refactored. Do you think double axes need any test? I haven't found one.

PR #5711, opened 2022-10-20
AG-7387 Improve data handling performance

alexanderby:
- Avoid deep data cloning.
- Prevent unnecessary re-renderings.

### DIFF
--- a/charts-packages/ag-charts-community/src/util/json.ts
+++ b/charts-packages/ag-charts-community/src/util/json.ts
@@ -142,11 +142,12 @@ export const DELETE = Symbol('<delete-property>') as any;
 
 const NOT_SPECIFIED = Symbol('<unspecified-property>') as any;
 
-/**
- * A collection of objects or arrays for which the
- * deep copying during the merge should be avoided.
- */
-export const avoidJsonMergeCopy = new WeakSet();
+export interface JsonMergeOptions {
+    /**
+     * Contains a list of properties where deep clones should be avoided
+     */
+    avoidDeepClone: string[];
+}
 
 /**
  * Merge together the provide JSON object structures, with the precedence of application running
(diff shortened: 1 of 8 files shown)


charts-packages/ag-charts-community/src/util/json.ts
```diff
@@ -148,8 +139,15 @@ export function jsonDiff<T extends any>(source: T, target: T, opts?: { stringify
...
  */
 export const DELETE = Symbol('<delete-property>') as any;
+
 const NOT_SPECIFIED = Symbol('<unspecified-property>') as any;
 
+/**
+ * A collection of objects or arrays for which the
+ * deep copying during the merge should be avoided.
+ */
+export const avoidJsonMergeCopy = new WeakSet();
```
alantreadway, line 149: Could we pass a context object with this sort of per-call context, rather than using a global?

charts-packages/ag-charts-community/src/util/json.ts
```diff
@@ -134,7 +125,7 @@ export function jsonDiff<T extends any>(source: T, target: T, opts?: { stringify
             continue;
         }
 
-        const diff = jsonDiff(lhs[prop], rhs[prop], { stringify });
+        const diff = jsonDiff(lhs[prop], rhs[prop]);
```
alantreadway, line 128: I was going to comment that we should just skip the `data` array during the diff, but I see that we merge BEFORE this, so I can see why your change to the merge processing makes sense now :-)

alantreadway (commented):
Update queue changes look good :+1: If we could avoid the global variable for the merge processing I think this is good to go :-)

alantreadway (approved):
Excellent, thanks for taking the feedback onboard! LGTM.

PR #5719, opened 2022-10-25
AG-7089 Fixed click and highlight when tooltip off

alexanderby:
Fixed node click events and highlighting when tooltip is disabled.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/chart.test.ts
+++ b/charts-packages/ag-charts-community/src/chart/chart.test.ts
@@ -1,9 +1,13 @@
 import { describe, expect, it, beforeEach, afterEach, jest } from '@jest/globals';
 import { toMatchImageSnapshot } from 'jest-image-snapshot';
-import { AgCartesianChartOptions } from './agChartOptions';
+import {
+    AgBarSeriesOptions,
+    AgCartesianChartOptions,
+    AgLineSeriesOptions,
+    AgPolarChartOptions,
+} from './agChartOptions';
 import { AgChartV2 } from './agChartV2';
-import { CartesianChart } from './cartesianChart';
-import { SeriesNodeDataContext } from './series/series';
+import { Chart } from './chart';
 import {
     waitForChartStability,
     setupMockCanvas,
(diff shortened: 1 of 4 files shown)


alantreadway (commented):
Code changes looks good, what testing strategy should we consider to avoid future regression?

charts-packages/ag-charts-community/src/chart/chart.test.ts
```diff
@@ -0,0 +1,225 @@
...
+                xKey: dataSamples.randomAnnual.categoryKey,
+                yKey: dataSamples.randomAnnual.valueKey,
+            },
+            getNodeData: (series) => series.contextNodeData[0].nodeData,
+            getNodePoint: (item) => [item.point.x, item.point.y],
+            getDatumValues: (datum, series) => {
+                const xValue = datum[series['xKey']];
+                const yValue = datum[series['yKey']];
+                return [xValue, yValue];
+            },
```
alantreadway, line 204: 🤔 Side-note: the fact we need these adaptor functions is probably pointing to the fact we should better standardise the `nodeData` format.

charts-packages/ag-charts-community/src/util/sanitize.ts
```diff
@@ -5,6 +5,6 @@ export function sanitizeHtml(text?: string): string {
     if (!text) {
         return '';
     }
-    element.innerText = text;
+    element.textContent = text;
```
alantreadway, line 8: Does this still play well with HTML support for `series[].tooltip.renderer`?
https://ag-grid.com/javascript-charts/tooltips/#using-custom-tooltips

alantreadway: Tested locally, still looks good in Standalone Charts and Integrated Charts examples.

alantreadway (approved):
LGTM

charts-packages/ag-charts-community/src/chart/chart.test.ts
alexanderby: I agree, it's totally different everywhere

charts-packages/ag-charts-community/src/util/sanitize.ts
alexanderby: Yes, but the custom HTML remains unsanitized. We only sanitise the fields and titles.

alexanderby:
The tests for Pies and Histograms will be added later.

PR #5725, opened 2022-10-27
AG-7445 - Fix incorrect positioning of series items when using a bar/…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7445

Adjust data domain instead of scale range when using bar/column series with continuous axis


alantreadway (approved):
LGTM

PR #5735, opened 2022-10-31
AG-7374 - Add unit tests for calculating `defaultTickFormat` in `time…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7374.

Added unit tests for the timeScale `defaultTickFormat` calculation

### DIFF
--- a/charts-packages/ag-charts-community/src/scale/timeScale.test.ts
+++ b/charts-packages/ag-charts-community/src/scale/timeScale.test.ts
@@ -3,8 +3,6 @@ import { TimeScale } from './timeScale';
 
 describe('TimeScale', () => {
     describe('#calculateDefaultTickFormat', () => {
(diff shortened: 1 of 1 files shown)


charts-packages/ag-charts-community/src/scale/timeScale.test.ts
```diff
@@ -3,22 +3,115 @@ import { TimeScale } from './timeScale';
 
 describe('TimeScale', () => {
     describe('#calculateDefaultTickFormat', () => {
+        const scale = new TimeScale();
```
alantreadway, line 6: Nit: This is bad practice for tests - you should initialise a fresh instance per test execution to ensure test isolation. It's also a memory leak outside a test or `beforeEach()` block :-/

alantreadway (approved):
LGTM

PR #5740, opened 2022-11-01
RTI-1118 Fixed highlight opacity applied twice

alexanderby:


charts-packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -316,7 +316,6 @@ export abstract class CartesianSeries<
         this.group.visible = visible;
         this.seriesGroup.visible = visible;
         this.highlightGroup.visible = visible && !!seriesHighlighted;
-        this.seriesGroup.opacity = this.getOpacity();
```
alantreadway, line 319: Opacity needs to be applied at the group(s) that map to layers so that opacity is applied during layer composition rather than during canvas rendering - is the opacity being applied elsewhere?

alexanderby: https://github.com/ag-grid/ag-grid/blob/5376db6f0a163ace3843cb9d49df2047d2c969ae/charts-packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts#L346

alantreadway (changes requested):
Let's discuss offline.

alantreadway (approved):
LGTM, this may still have performance issues, but they are pre-existing.

PR #5747, opened 2022-11-03
Rti 1110/axis label rotation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/RTI-1122

Using the same `axisLabelsOverlap` check for determining whether labels should be rotated before checking if they should be removed. 

The approach before was using a different check for rotation which assumed labels are similar sizes, so in some cases no auto rotation would be applied but then it would fail the overlap test later on (as this is a more accurate check) and lead to removal of labels.


grid-packages/ag-grid-enterprise/src/main.ts
```diff
@@ -42,7 +42,7 @@ export { getMultipleSheetsAsExcel, exportMultipleSheetsAsExcel } from '@ag-grid-
 
 /* COMMUNITY_EXPORTS_START_DO_NOT_DELETE */
 export { ColumnFactory } from "ag-grid-community";
-export { ColumnModel, ColumnState, ColumnStateParams, ApplyColumnStateParams, ISizeColumnsToFitParams, IColumnLimit } from "ag-grid-community";
```
alantreadway, line 45: Drop this change from the PR?

alantreadway (approved):
LGTM - nice work!

grid-packages/ag-grid-enterprise/src/main.ts
manapeirov: Oops dunno how that got in there. Removed now!

PR #5767, opened 2022-11-10
AG-6840 - Reduce ag-charts-community exports

alantreadway:
https://ag-grid.atlassian.net/browse/AG-6840

Attempts to minimise the exports from `main.ts` for `ag-charts-community`, as many undocumented and implementation specific types were being exported by default.

Main changes:
- Minimise exports from `main.ts`.
  - For undocumented Integrated Charts type dependencies, these have been moved to `integrated-charts-scene.ts` + `integrated-charts-theme.ts` and exported with the name `_Scene` + `_Theme` respectively from the `main.ts` exports.
  - For Integrated Charts use of implementation-specific fields/methods, I've introduced an `@ag-grid-enterprise/charts`-specific extension file `agChartsCommunityExtension.ts` to list the dependencies (we should aim to reduce this over time).
- Where possible, Integrated Charts dependencies have been eliminated (e.g. replacing `instanceof` checks with `type` field checks should simplify dependencies).


enterprise-modules/charts/src/charts/chartComp/services/chartOptionsService.ts
```diff
@@ -181,13 +168,13 @@ export class ChartOptionsService extends BeanStub {
...
 
     private static isMatchingSeries(seriesType: ChartSeriesType, series: SupportedSeries): boolean {
-        return seriesType === 'area' && series instanceof AreaSeries ? true :
-               seriesType === 'bar' && series instanceof BarSeries ? true :
-               seriesType === 'column' && series instanceof BarSeries ? true :
-               seriesType === 'histogram' && series instanceof HistogramSeries ? true :
-               seriesType === 'line' && series instanceof LineSeries ? true :
-               seriesType === 'pie' && series instanceof PieSeries ? true :
-               seriesType === 'scatter' && series instanceof ScatterSeries;
+        return seriesType === 'area' && series.type === seriesType ? true :
```
alexanderby, line 171: My eye got caught by it. Is it an equivalent of:
```typescript
series.type === seriesType && ['area', 'bar', ...etc].includes(seriesType)
```
?

PR #5770, opened 2022-11-10
AG-7506 - Fix demo page and integrated charts normalized area series …

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7506. 

Fixed broken normalized area series label formatting for the demo page and missing % formatting for integrated charts


alantreadway (approved):
LGTM - this was a much smaller gap than I expected!

Should we consider making the use of a percentage formatter the default in Standalone Charts for normalised series?

manapeirov:
I was thinking this but we agreed with Rob that % only makes sense for certain cases, normalized value is not always necessarily a percentage?

PR #5773, opened 2022-11-11
Update Jest to be compatible with TypeScript 4.0

alexanderby:
- Jest 25 is compatible with TS v3
- Jest 28 and 29 are compatible with TS >=4.3
- jest-serial-runner doesn't start, probably needs another release.


alantreadway (approved):
LGTM - I'll look into the `jest-serial-runner` issue as we'll need to do something to keep the e2e tests stable.

PR #5778, opened 2022-11-14
AG-7469 - Fix custom marker selection not updating

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7530.

There is an issue with the markerSelection not being cleared before updating. Even with the navigator disabled, hovering the chart shows incorrect highlight marker node size.

Before we only cleared the `markerSelection` if `this.marker.isDirty()` returned true, which would only occur if the `seriesMarker` properties change like marker `size` but here the marker size is changing via the custom marker `updatePath` method which is on a per marker basis.

### DIFF
--- a/enterprise-modules/charts/src/charts/chartComp/chartProxies/chartProxy.ts
+++ b/enterprise-modules/charts/src/charts/chartComp/chartProxies/chartProxy.ts
@@ -46,7 +46,7 @@ export abstract class ChartProxy {
     /**
      * Integrated Charts specific chart theme overrides
      */
-    private readonly integratedThemeOverrides = {
+    private readonly integratedThemeOverrides: AgChartThemeOverrides = {
         common: {
             padding: {
                 top: 25,


charts-packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -579,7 +579,8 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
     }
 
     protected isPathOrSelectionDirty(): boolean {
-        return this.marker.isDirty();
+        const isCustomMarker = typeof this.marker.shape === 'function';
```
alantreadway, line 582: The won't scale for large data-sets, as it forces re-render of custom marker always - could we make use of the `Path2D.clear({ trackChanges: true })` and `Path2D.isDirty()` methods to detect actual marker changes?

PR #5781, opened 2022-11-15
AG-7469 - Fix custom marker shape path not updating

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7530.

In the case of a custom marker shape, the shape path commands could be dependent on the node data and we may need to invoke updatePath() to update the path when the selections are updated (data for the marker node has changed).


alantreadway (approved):
LGTM - this seems a lot more straightforward than I was expecting :D

PR #5784, opened 2022-11-15
AG-7395 - Add id placeholder for empty keys

AndrewGlazier:


enterprise-modules/server-side-row-model/src/serverSideRowModel/blocks/blockUtils.ts
```diff
@@ -338,10 +338,13 @@ export class BlockUtils extends BeanStub {
     public createNodeIdPrefix(parentRowNode: RowNode): string | undefined {
         const parts: string[] = [];
         let rowNode: RowNode | null = parentRowNode;
-
         // pull keys from all parent nodes, but do not include the root node
         while (rowNode && rowNode.level >= 0) {
-            parts.push(rowNode.key!);
+            if (rowNode.key === '') {
+                parts.push('AgGrid_EmptyKey')
```
gportela85, line 344: This should probably go in the `Constants` file to keep consistent with GROUP_AUTO_COLUMN_ID.

AndrewGlazier: Seems the Constants file has been removed since I last visited this, can you check the change I've made to see if you agree that it's sufficient?

PR #5789, opened 2022-11-16
AG-7500 - Minimise repeated IAgChartOptions typings

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7500

Enables a single-source-of-truth for `AgChartOptions` and nested types. Achieves this by using stub types in `ag-grid-community`, and having `@ag-grid-enterprise/charts` extend these when imported.

Any project which import anything from `@ag-grid-enterprise/charts` will intrinsically have the extended typings.

For package builds (`ag-grid-enterprise`) @seanlandsman has kindly added support for ensure the overrides are loaded correctly, and so any import of `ag-grid-enterprise` also correctly has the extended typings.

Tested:
- Docs TS example verification is passing.
- `npm pack`ed `ag-grid-enterprise` has correct typings in a simple TS project.


grid-packages/ag-grid-docs/documentation/doc-pages/integrated-charts-customisation/examples/histogram-overrides/main.ts
```diff
@@ -1,7 +1,7 @@
 import { CreateRangeChartParams, FirstDataRenderedEvent, Grid, GridOptions } from '@ag-grid-community/core';
+import '@ag-grid-enterprise/charts';
```
StephenCooper, line 2: Should we add this to the example generator so that you don't have to remember to add it?

seanlandsman: I'd like to discuss this actually - maybe one for next week, but rather than importing this we could make it part of the registeration process.

ie having a user both register and import a module seems like duplication

alantreadway: I've raised https://ag-grid.atlassian.net/browse/AG-7597 to follow-up on this aspect.

PR #5797, opened 2022-11-17
AG-7546 - Review fixes for AgChart.updateDelta() changes.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7546

Review fixes for the new `AgChart.updateDelta()` factory update method.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartV2.ts
@@ -101,10 +101,11 @@ function chartType<T extends ChartType>(options: ChartOptionType<T>): 'cartesian
     throw new Error('AG Chart - unknown type of chart for options with type: ' + options.type);
 }
 
-type DeepPartial<T> = T extends object
-    ? {
-          [P in keyof T]?: DeepPartial<T[P]>;
-      }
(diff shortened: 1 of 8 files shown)


charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -123,13 +124,21 @@ export abstract class AgChart {
     /**
      * Update an existing `AgChartInstance`. Options provided should be complete and not
      * partial.
+     *
+     * NOTE: each call could trigger a redraw; multiple calls in quick succession could result in
```
taktran, line 128: I'm not sure the redraw issue will be as much of a problem with `update`, as the `options` are passed in complete. Not 100% sure though 😅

taktran (approved):
Looks good 👍🏼 

I've checked and the new `DeepPartial` works for `ChartOptionsService` too.

charts-packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway: Agree the case you hit is much less likely, but I added it as if a user spam `update()` then some form of 'flickering' aka unexpected renderings could happen (I've left the definition of that deliberately vague) irrespective of whether the configurations are correct/complete.

PR #5801, opened 2022-11-18
AG-7502 - Get chart options from Chart instead of storing in `chartProxy.chartOptions`

taktran:
Follow up to https://github.com/ag-grid/ag-grid/pull/5765

## Changes

* Get chart options from Chart `getOptions()` and delete `chartOptions` and `chartPalette` in `chartProxy`.  `chartProxy.agChartTheme` stores theme on creation, but during updates, changes are made directly to the chart using `AgChart.updateDelta`. Hence the Chart stores the state of the options and to retrieve it `getOptions()` can be used.
  * Special handling for `scatter.paired` to be persisted in `chartProxy` to make it explicit that it's only for Integrated Charts
* Remove chart options merges within child chart proxies since these were for styles, and this is now managed by the Chart

### DIFF
--- a/enterprise-modules/charts/src/charts/chartComp/chartController.ts
+++ b/enterprise-modules/charts/src/charts/chartComp/chartController.ts
@@ -1,7 +1,6 @@
 import {
     _,
     AgChartThemePalette,
-    AgEvent,
     Autowired,
     BeanStub,
     CellRange,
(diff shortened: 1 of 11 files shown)


alantreadway (approved):
LGTM - looks like a good step forward with the removal of a lot of `deepMerge()` calls and simplification of state!

PR #5807, opened 2022-11-21
AG-7492 Pie label auto padding

alexanderby:
- Calculate padding for Pie callout labels.
- Prevent chart from becoming too small.
- Shrink long text.


charts-packages/ag-charts-community/src/chart/polarChart.ts
```diff
@@ -61,17 +66,79 @@ export class PolarChart extends Chart {
...
+                    if (Math.min(padLeft, padRight) * 2 > padWidth) {
+                        padLeft = padWidth / 2;
+                        padRight = padWidth / 2;
+                    } else if (padLeft > padRight) {
+                        padLeft = padWidth - padRight;
+                    } else {
+                        padRight = padWidth - padLeft;
+                    }
+                }
+            }
```
alantreadway, line 129: Nit: Maybe break this into a distinct function/method given the level of nesting required here (the control flow can also be more easily broken with a `return` then too, which will reduce the nesting level and improve readability?)

charts-packages/ag-charts-community/src/chart/polarChart.ts
```diff
@@ -61,17 +66,79 @@ export class PolarChart extends Chart {
...
+        let radius = Math.max(0, Math.min(seriesBox.width, seriesBox.height) / 2);
+        polarSeries.forEach((series) => {
+            series.centerX = centerX;
+            series.centerY = centerY;
+            series.radius = radius;
+        });
+
+        const radiusRatioThreshold = 0.5;
+        const labelRepositionAttempts = 2;
+        for (let i = 0; i < labelRepositionAttempts; i++) {
```
alantreadway, line 88: Do we always iterate twice if there are callout labels?

It might be better to separate the calculation logic from the iteration/state logic (it's not clear in the current structure what the inputs and outputs are from the iteration here? (Looks like maybe `padLeft` and `padRight`?)

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -843,36 +846,87 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+                } else if (box.x + box.width - errPx > seriesRight) {
+                    t = (seriesRight - box.x) / box.width;
+                }
+                if (t < 1) {
+                    const length = Math.floor(label.text.length * t) - 1;
+                    text.text = label.text.substring(0, length) + '…';
+                }
+                text.visible = true;
+            } else {
+                text.visible = false;
```
alantreadway, line 893: Nit: Maybe check for this case first and just `return`, so we can un-nest the previous block?

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -843,36 +846,87 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+            const outerRadius = Math.max(0, radius);
+
+            if (label && outerRadius !== 0) {
+                const labelRadius = outerRadius + calloutLength + offset;
+
+                text.fontStyle = fontStyle;
+                text.fontWeight = fontWeight;
+                text.fontSize = fontSize;
+                text.fontFamily = fontFamily;
+                text.text = label.text;
```
alantreadway, line 871: Avoid double assignment to the scene graph - it yield false-positive change-detection in scenarios like this where we are performing calculations at this late stage:

e.g. we could end up doing this series of steps:
```
text.text = 'Abc 123 xyz'
...
<perform calculations>
...
text.text = 'Abc 123 ...'
```

If the label was previously the same clipped value at the end, the scene graph will still detect a change due to the first assignement.

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -843,36 +846,87 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+                text.x = datum.midCos * labelRadius;
+                text.y = datum.midSin * labelRadius;
+                text.fill = color;
+                text.textAlign = label.textAlign;
+                text.textBaseline = label.textBaseline;
+
+                // Clip text if there is overflow
+                const box = text.computeBBox();
```
alantreadway, line 879: Could we use a lower-level operation rather than having the state of the `Label` depend on itself?

charts-packages/ag-charts-community/src/chart/polarChart.ts
alexanderby: Iterating twice gives a precise result. But I added a "break" in case if no overflow detected.

alexanderby: Each iteration updates radius and center coordinates.

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
alexanderby: Done

charts-packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
alexanderby: So I create a virtual `tempTextNode`, assign props into it, check the bounding box, and then change the text of the text node from the scene graph if necessary. Maybe this can be optimised, but I will know for sure when working on labels collisions.

alantreadway (approved):
LGTM!

PR #5809, opened 2022-11-21
AG-7431 - Mark path dirty transofrm when seriesRoot transformed durin…

manapeirov:
Path render was not being invoked during re-render to apply the new transform


alantreadway (approved):
LGTM - could you just verify there are no major side-effects (performance-wise) of this change? E.g. high-volume partial redraw cases?

PR #5810, opened 2022-11-21
AG-7457 Fix callout line color mismatch

alexanderby:
- Use sectors stroke colors for callout line if unset.
- Remove unused Series.setColors method.


alantreadway (approved):
LGTM

PR #5813, opened 2022-11-22
AG-7550 Fix CrossFilter Pie tooltip override

alexanderby:
Note: chart tooltip options are different from series tooltip options.


alantreadway (approved):
LGTM

PR #5814, opened 2022-11-22
Ag 7431/line area series path update

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7431.

The reason why the line and area paths were not being updated was due to path/path2D not having a `computeBBox()` implementation, resulting in `forceRender` being set to false.

Now we only set `forceRender` to false if the current bbox is the same as the previous bbox and it is not infinity.


alantreadway (approved):
LGTM

PR #5819, opened 2022-11-25
Ag 7579 change detection via grid options service

StephenCooper:
With the new GridOptionsService we have an opportunity to update how ComponentUtils works to lay a solid foundation for making as many of the Grid properties reactive as we can. As such our aims include:

1. Ensure that a `PropertyChangedEvent` is fired for every changed grid option
2. Avoid bugs due to events not firing due to updated order breaking change detection
3. Remove as much custom handling from ComponentUtils as possible

I believe the following changes meet these requirements. There could be some followup items such as:

- removing GridApi set methods that are solely for component utils to be able to update grid options. If they are not triggering reactive behaviour then maybe JS users can just update their GridOptions config option and have the grid use the new values. 
- Make the quickFilter listen to propertyChanged event instead of having the special case handling because of a GridApi misnaming.

### DIFF
--- a/community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
+++ b/community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
@@ -98,8 +98,8 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel 
             animate
         });
 
-        this.addManagedListener(this.gridOptionsWrapper, GridOptionsWrapper.PROP_GROUP_REMOVE_SINGLE_CHILDREN, refreshMapListener);
-        this.addManagedListener(this.gridOptionsWrapper, GridOptionsWrapper.PROP_GROUP_REMOVE_LOWEST_SINGLE_CHILDREN, refreshMapListener);
+        this.addManagedPropertyListener('groupRemoveSingleChildren', refreshMapListener);
+        this.addManagedPropertyListener('groupRemoveLowestSingleChildren', refreshMapListener);
 
         this.rootNode = new RowNode(this.beans);
         this.nodeManager = new ClientSideNodeManager(this.rootNode,
--- a/community-modules/core/src/ts/components/componentUtil.ts
+++ b/community-modules/core/src/ts/components/componentUtil.ts
@@ -41,6 +41,7 @@ export class ComponentUtil {
             ...ComponentUtil.ARRAY_PROPERTIES,
             ...ComponentUtil.OBJECT_PROPERTIES,
             ...ComponentUtil.STRING_PROPERTIES,
+            ...ComponentUtil.FUNCTION_PROPERTIES,
             ...ComponentUtil.getEventCallbacks(),
         ]
             .forEach((key: keyof GridOptions) => this.coercionLookup[key] = 'none');
--- a/community-modules/core/src/ts/components/componentUtil.ts
+++ b/community-modules/core/src/ts/components/componentUtil.ts
@@ -30,12 +30,54 @@ export class ComponentUtil {
     public static FUNCTION_PROPERTIES = PropertyKeys.FUNCTION_PROPERTIES;
     public static ALL_PROPERTIES = PropertyKeys.ALL_PROPERTIES;
 
+    private static coercionLookup: Record<keyof GridOptions, 'number' | 'boolean' | 'none'>;
+    private static getCoercionLookup() {
+        if (this.coercionLookup) {
+            return this.coercionLookup;
+        }
+
+        this.coercionLookup = {} as any
+        [
+            ...ComponentUtil.ARRAY_PROPERTIES,
+            ...ComponentUtil.OBJECT_PROPERTIES,
+            ...ComponentUtil.STRING_PROPERTIES,
+            ...ComponentUtil.getEventCallbacks(),
+        ]
+            .forEach((key: keyof GridOptions) => this.coercionLookup[key] = 'none');
+        ComponentUtil.BOOLEAN_PROPERTIES
+            .forEach(key => this.coercionLookup[key] = 'boolean');
+        ComponentUtil.NUMBER_PROPERTIES
+            .forEach(key => this.coercionLookup[key] = 'number');
+        return this.coercionLookup;
+    }
+
     public static getEventCallbacks(): string[] {
         if (!ComponentUtil.EVENT_CALLBACKS) {
             ComponentUtil.EVENT_CALLBACKS = ComponentUtil.EVENTS.map(event => ComponentUtil.getCallbackForEvent(event));
         }
-
         return ComponentUtil.EVENT_CALLBACKS;
+    }
+
+    private static getValue(key: string, rawValue: any) {
+        const coercionStep = ComponentUtil.getCoercionLookup()[key as keyof GridOptions];
+
+        if (coercionStep) {
+            let newValue = rawValue;
+            switch (coercionStep) {
+                case 'number':
+                    newValue = ComponentUtil.toNumber(rawValue);
+                case 'boolean':
+                    newValue = ComponentUtil.toBoolean(rawValue);
+                case 'none': {
+                    // if groupAggFiltering exists and isn't a function, handle as a boolean.
+                    if (key === 'groupAggFiltering' && typeof rawValue !== 'function') {
+                        newValue = ComponentUtil.toBoolean(rawValue);
+                    }
+                }
+            }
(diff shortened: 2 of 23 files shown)


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -98,8 +98,8 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel
             animate
         });
 
-        this.addManagedListener(this.gridOptionsWrapper, GridOptionsWrapper.PROP_GROUP_REMOVE_SINGLE_CHILDREN, refreshMapListener);
```
StephenCooper, line 101: The addManagedPropertyListener is now typesafe and so is being used in place of all the existing GridOptionWrapper events.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -30,50 +30,73 @@ export class ComponentUtil {
     public static FUNCTION_PROPERTIES = PropertyKeys.FUNCTION_PROPERTIES;
     public static ALL_PROPERTIES = PropertyKeys.ALL_PROPERTIES;
 
+    private static coercionLookup: Record<keyof GridOptions, 'number' | 'boolean' | 'none'>;
+    private static getCoercionLookup() {
```
StephenCooper, line 34: This coercionLookup enables us to apply coercions one by one as required for the changed properties. As opposed to looping through every grid option and checking if the component has that key.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -30,50 +30,73 @@ export class ComponentUtil {
...
+
     public static getEventCallbacks(): string[] {
         if (!ComponentUtil.EVENT_CALLBACKS) {
             ComponentUtil.EVENT_CALLBACKS = ComponentUtil.EVENTS.map(event => ComponentUtil.getCallbackForEvent(event));
         }
-
         return ComponentUtil.EVENT_CALLBACKS;
     }
 
+    private static getValue(key: string, rawValue: any) {
```
StephenCooper, line 61: For each key / value on our component and also as part of onChange this method checks if it is a valid gridOption. (by inclusion in the coercion map)
Then if required will apply the coercion before returning the new value.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -30,50 +30,73 @@ export class ComponentUtil {
...
+
+        if (coercionStep) {
+            let newValue = rawValue;
+            switch (coercionStep) {
+                case 'number':
+                    newValue = ComponentUtil.toNumber(rawValue);
+                case 'boolean':
+                    newValue = ComponentUtil.toBoolean(rawValue);
+                case 'none': {
+                    // if groupAggFiltering exists and isn't a function, handle as a boolean.
```
StephenCooper, line 72: Special handling for this property with can be a boolean or a function

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
-        }
-        // *********  CODE ORDER TO AVOID BUGS *************** //
-        // If you want to call an update method that just calls through to gridOptionsWrapper.setProperty then it needs to be
-        // called before the values get copied across otherwise the change will not fire an event because the method
-        // gridOptionsWrapper.setProperty does a diff check first.
-
-        // All these manual calls are required in the current setup as changes to these properties are being listened to in the 
-        // rest of the code base which can be found by searching for: "addManagedListener(this.gridOptionsWrapper"
-
-        if (changesToApply.domLayout) {
```
StephenCooper, line 123: We no longer need these special cases as we now ensure that events are fired for every property change and also the coercion is handle for every property as part of the update. 

Previously these properties had to be added to work around a bug in ComponentUtils which is why they match properties that have already been setup as reactive.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
-
-        // check if any change for the simple types, and if so, then just copy in the new value
-        [
-            ...ComponentUtil.ARRAY_PROPERTIES,
-            ...ComponentUtil.OBJECT_PROPERTIES,
-            ...ComponentUtil.STRING_PROPERTIES,
-            ...ComponentUtil.getEventCallbacks(),
-        ]
-            .filter(keyExists)
-            .forEach(key => pGridOptions[key] = changesToApply[key].currentValue);
```
StephenCooper, line 197: We no longer loop through all GridOptions but instead loop through the changes which should be more efficient. Also we do not directly update gridOptions as part of ComponentUtils as that was the cause of the bug above. 

Instead all updates are passed off to the GridOptionService which is how we can now fire ProptertyChangedEvents consistently.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
         }
-        if (changesToApply.columnDefs) {
-            api.setColumnDefs(changesToApply.columnDefs.currentValue, "gridOptionsChanged");
-            delete changesToApply.columnDefs;
-        }
         if (changesToApply.defaultColDef) {
             api.setDefaultColDef(changesToApply.defaultColDef.currentValue, "gridOptionsChanged");
             delete changesToApply.defaultColDef;
         }
-        if (changesToApply.pivotMode) {
```
StephenCooper, line 231: pivot mode handled by common approach now as I added the dynamic api call to GridApi which I believe was missing in error.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
-            columnApi.setPivotMode(ComponentUtil.toBoolean(changesToApply.pivotMode.currentValue));
-            delete changesToApply.pivotMode;
+        if (changesToApply.columnDefs) {
+            api.setColumnDefs(changesToApply.columnDefs.currentValue, "gridOptionsChanged");
```
StephenCooper, line 129: For the three columnDef props we still require special handling so that they can apply the correct event source.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
-                const setterName = `set${property.charAt(0).toUpperCase()}${property.substring(1)}`;
+        Object.keys(changesToApply).forEach(key => {
+            const gridKey = key as keyof GridOptions;
+
+            const coercedValue = ComponentUtil.getValue(gridKey, changesToApply[gridKey].currentValue);
+            // Ensure the GridOptions property gets updated and fires the change event as we
+            // cannot assume that the dynamic Api call updated GridOptions or fired the event.
+            // If the dynamic api did already update GridOptions then the change detection in the 
+            // setter should prevent the event being fired twice.
+            api.__setProperty(gridKey, coercedValue);
```
StephenCooper, line 142: Now every change goes through GridOptionsService to ensure events are fired.

community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -92,158 +115,44 @@ export class ComponentUtil {
...
+
+            const setterName = `set${key.charAt(0).toUpperCase()}${key.substring(1)}`;
+            if (dynamicApi[setterName]) {
+                dynamicApi[setterName](coercedValue);
```
StephenCooper, line 146: Still call dymanicApis as they may contain more logic then just updating gridOptionsService. 
If the dynamic api does call gridOptionsService too then as it will be passed the exact same value the change detection will prevent two events being fired.

So this should be a safe approach even if methods come and go off GridApi.

community-modules/core/src/ts/gridOptionsService.ts
```diff
@@ -60,20 +60,11 @@ export class GridOptionsService {
 
     @Autowired('gridOptions') private readonly gridOptions: GridOptions;
     private propertyEventService: EventService = new EventService();
-    private coercionLookup: Record<keyof GridOptions, 'number' | 'boolean' | 'none'>;
```
StephenCooper, line 63: Remove setting coercion as that is really the job of the ComponentUtils and really should only apply to frameworks. Anyone calling gridApi methods directly should call them with the correct type.

rmc-software (approved):

PR #5821, opened 2022-11-25
AG-7407 - Hide axes, crossLines, series and legend based on bounds ca…

manapeirov:
PR for: https://ag-grid.atlassian.net/browse/AG-7407

When the crossLinePadding exceeds available boundary, prioritise axes, remove crossLines.
If axisWidths exceed available chart space, remove axes, series and crossLines.

This is done by toggling the relevant group visibility.

Layers should honour group visibility based on parent nodes' visible flag.

TODO: should  navigator also be hidden?
Look into why legend recreates groups on performLayout

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/cartesianChart.ts
+++ b/charts-packages/ag-charts-community/src/chart/cartesianChart.ts
@@ -76,7 +76,10 @@ export class CartesianChart extends Chart {
             navigator.width = shrinkRect.width;
         }
 
-        const { seriesRect } = this.updateAxes(shrinkRect);
+        const { seriesRect, hideSeries } = this.updateAxes(shrinkRect);
+
+        this.seriesRoot.visible = !hideSeries;
+        legend.group.visible = !hideSeries;
 
         this.seriesRect = seriesRect;
         this.series.forEach((series) => {
@@ -137,6 +140,8 @@ export class CartesianChart extends Chart {
         let clipSeries = false;
         let seriesRect = this.seriesRect?.clone();
         let count = 0;
+        let hideAxes = false;
+        let hideCrossLines = false;
         do {
             Object.assign(axisWidths, lastPass);
 
@@ -144,17 +149,27 @@ export class CartesianChart extends Chart {
             lastPass = ceilValues(result.axisWidths);
             clipSeries = result.clipSeries;
             seriesRect = result.seriesRect;
+            hideAxes = result.hideAxes;
+            hideCrossLines = result.hideCrossLines;
 
             if (count++ > 10) {
                 console.warn('AG Charts - unable to find stable axis layout.');
                 break;
             }
-        } while (!stableWidths(lastPass));
+        } while (!stableWidths(lastPass) && !hideAxes);
 
         this.seriesRoot.enabled = clipSeries;
+
+        // update visibility of axes and crosslines
+        this.axes.forEach((axis) => {
+            axis.axisGroup.visible = !hideAxes;
+            axis.gridlineGroup.visible = !hideAxes;
+            axis.crossLineGroup.visible = !hideCrossLines;
+        });
+
         this._lastAxisWidths = axisWidths;
 
-        return { seriesRect };
+        return { seriesRect, hideSeries: hideAxes };
     }
 
     private updateAxesPass(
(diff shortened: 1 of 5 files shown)


charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -71,12 +71,19 @@ export class CartesianChart extends Chart {
...
+        const { seriesRect, hideSeries } = this.updateAxes(shrinkRect);
+
+        if (navigator.enabled && !hideSeries) {
+            navigator.x = seriesRect.x;
+            navigator.width = seriesRect.width;
+        }
+
+        this.seriesRoot.visible = !hideSeries;
+        legend.group.visible = legend.enabled && !hideSeries;
+        navigator.rs.visible = navigator.enabled && !hideSeries;
```
alantreadway, line 86: Nit: I'm not really a fan of delving into an objects implementation (i.e. twiddling nested object references) as it breaks encapsulation - is there a better way to manage the visibility in the scene graph nodes?

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -137,24 +144,39 @@ export class CartesianChart extends Chart {
...
+        let hideCrossLines = false;
         do {
             Object.assign(axisWidths, lastPass);
 
             const result = this.updateAxesPass(axisWidths, inputShrinkRect.clone(), seriesRect);
             lastPass = ceilValues(result.axisWidths);
             clipSeries = result.clipSeries;
             seriesRect = result.seriesRect;
+            hideAxes = result.hideAxes;
+            hideCrossLines = result.hideCrossLines;
```
alantreadway, line 157: We might want to consider using an object with these fields rather than multiple local variables to clean things up?

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -137,24 +144,39 @@ export class CartesianChart extends Chart {
...
 
             if (count++ > 10) {
                 console.warn('AG Charts - unable to find stable axis layout.');
                 break;
             }
-        } while (!stableWidths(lastPass));
+        } while (!stableWidths(lastPass) && !hideAxes);
```
alantreadway, line 163: I think the `hideAxes` should probably be part of the `stableWidths()` calculation, or handled similarly, since the value could toggle on a boundary condition similarly to widths?

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -137,24 +144,39 @@ export class CartesianChart extends Chart {
...
+
+        const hideSeries = seriesRect.width < 1 || seriesRect.height < 1;
+        hideCrossLines = hideCrossLines || hideSeries;
+
+        // update visibility of axes and crosslines
+        this.axes.forEach((axis) => {
+            axis.axisGroup.visible = !hideAxes;
+            axis.gridlineGroup.visible = !hideAxes;
+            axis.crossLineGroup.visible = !hideCrossLines;
+        });
```
alantreadway, line 175: Encapsulation? (It seems like if we disable axes, then everything else should also be disabled, so encapsulating this behaviour in Axis makes sense and should cleanup the code here?)

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -167,12 +189,23 @@ export class CartesianChart extends Chart {
...
+        let hideAxes = false;
         let primaryTickCounts: Partial<Record<ChartAxisDirection, number>> = {};
 
         const crossLinePadding = lastPassSeriesRect ? this.buildCrossLinePadding(lastPassSeriesRect, axisWidths) : {};
-        const axisBound = this.buildAxisBound(bounds, crossLinePadding);
+        let { axisBound, hideCrossLines } = this.buildAxisBound(bounds, crossLinePadding);
+
         const seriesRect = this.buildSeriesRect(axisBound, axisWidths);
 
+        if (seriesRect.width === 0 || seriesRect.height === 0) {
```
alantreadway, line 200: Nit: Are these numbers rounded, and if not should they be?

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -233,11 +266,17 @@ export class CartesianChart extends Chart {
     private buildAxisBound(bounds: BBox, crossLinePadding: Partial<Record<AgCartesianAxisPosition, number>>) {
         const result = bounds.clone();
         const { top = 0, right = 0, bottom = 0, left = 0 } = crossLinePadding;
+        if (result.width <= left + right || result.height <= top + bottom) {
```
alantreadway, line 269: Nit: Could you add some parenthesis here to improve readability of the logic here?

alantreadway (commented):
Generally looks good, but I think we need to improve the encapsulation between model elements to reduce complexity.

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
manapeirov: done

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
manapeirov: fixed

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -111,14 +126,20 @@ export class CartesianChart extends Chart {
...
 
-        const stableWidths = <T extends typeof axisWidths>(other: T) => {
-            return Object.entries(axisWidths).every(([p, w]) => {
-                const otherW = (other as any)[p];
-                if (w != null || otherW != null) {
-                    return w === otherW;
-                }
-                return true;
-            });
+        const stableWidths = <T extends typeof axisWidths>(
```
alantreadway, line 129: Nit: Maybe rename this to `stableOutputs`?

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -111,14 +126,20 @@ export class CartesianChart extends Chart {
...
+            otherAxisWidths: T,
+            otherVisibility: Partial<VisibilityMap>
+        ) => {
+            return (
+                visibility.axes === otherVisibility.axes &&
```
alantreadway, line 134: Should we also check `visibility.crossLines`?

manapeirov: fair point

charts-packages/ag-charts-community/src/chart/cartesianChart.ts
manapeirov: done

manapeirov:
PTAL @alantreadway

alantreadway (approved):
LGTM

PR #5827, opened 2022-11-28
AG-7588 Treemap value label formatter

alexanderby:
- Add value label formatter (or key).
- Stop getting value text from colorKey.
- Remove series.labels.color label config.


alantreadway (approved):
LGTM

PR #5849, opened 2022-12-02
AG-7585 Treemap API improvements & docs

alexanderby:
- Untie tile and tooltip label text from colorKey.
- Ability to get depth and parent in formatter and tooltip renderer.
- Separate tile and text shadows.
- Highlight text colour (can be used in all types).
- Ability to prevent groups highlighting.
- Fixed not displaying formatted fill in the tooltip.
- API docs update and complex colouring example.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/series/series.ts
+++ b/charts-packages/ag-charts-community/src/chart/series/series.ts
@@ -95,9 +95,15 @@ class SeriesHighlightStyle {
     enabled?: boolean = undefined;
 }
 
+class TextHighlightStyle {
+    @Validate(OPT_COLOR_STRING)
+    color?: string = 'black';
+}
+
 export class HighlightStyle {
     readonly item = new SeriesItemHighlightStyle();
     readonly series = new SeriesHighlightStyle();
+    readonly text = new TextHighlightStyle();
 }
 
 export class SeriesTooltip {
--- a/grid-packages/ag-grid-docs/documentation/doc-pages/charts-treemap-series/examples/custom-colors/main.ts
+++ b/grid-packages/ag-grid-docs/documentation/doc-pages/charts-treemap-series/examples/custom-colors/main.ts
@@ -37,17 +37,13 @@ const options: agCharts.AgChartOptions = {
                     color: undefined,
                 },
             },
-            formatter: ({ datum, depth, parent, highlighted }) => {
+            formatter: ({ depth, parent, highlighted }) => {
                 if (depth < 2) {
                     return {};
                 }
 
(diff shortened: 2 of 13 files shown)


charts-packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -95,9 +95,15 @@ class SeriesHighlightStyle {
...
 
+class TextHighlightStyle {
+    @Validate(OPT_COLOR_STRING)
+    color?: string = 'black';
+}
+
 export class HighlightStyle {
     readonly item = new SeriesItemHighlightStyle();
     readonly series = new SeriesHighlightStyle();
+    readonly text = new TextHighlightStyle();
```
alantreadway, line 106: I don't see any update for cartesian series types related to this addition? (Importantly, the `AgSeriesHighlightStyle` change means that docs for cartesian series types will include this options - so we should either rectify that or add support in those other series types?)

grid-packages/ag-grid-docs/documentation/doc-pages/charts-treemap-series/examples/custom-colors/main.ts
```diff
@@ -0,0 +1,63 @@
...
+                },
+            },
+            formatter: ({ datum, depth, parent, highlighted }) => {
+                if (depth < 2) {
+                    return {};
+                }
+
+                const stroke = highlighted ? 'black' : 'white';
+                const mainColor = parent.name === 'Foodstuffs' ? [224, 64, 32] : [32, 92, 224];
+                const siblings = parent.children.sort((a, b) => b.exports - a.exports);
```
alantreadway, line 47: `Array.sort()` mutates the array - this raises an interesting point regarding mutability of the callback parameters here?

alantreadway (commented):
Doc updates look great, just a couple of things to follow-up on.

grid-packages/ag-grid-docs/documentation/doc-pages/charts-treemap-series/examples/custom-colors/main.ts
alexanderby: Yes, probably the coloring is unnecessary complex here, I will make just 2 plain colors for different branches.

charts-packages/ag-charts-community/src/chart/series/series.ts
alexanderby: Will do just for Treemap. It could also be useful for Pies.

alantreadway (approved):
LGTM - thanks for the updates!

PR #5853, opened 2022-12-05
Ag 7539 move widgets

StephenCooper:
Move chart widgets into enterprise charts package. 

Reference Color.ts from `ag-charts-community` via _Utils.


alantreadway (approved):
LGTM from a Charts perspective - thanks for this!

PR #5863, opened 2022-12-08
Ag 4835/legend improvements

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-4835 & https://ag-grid.atlassian.net/browse/AG-7584

- Arrange legend items in a grid layout using a `maxHeight` constraint.
- Paginate grid using a `maxWidth` constraint.
- Add pagination component to handle page changes.


charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -755,19 +757,24 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
 
+        const legendSpacing = legend.spacing;
         const width = this.width;
         const height = this.height - captionAutoPadding;
-        const legendSpacing = legend.spacing;
+
+        const horizontalLegendWidth = legend.maxWidth ?? width - legendSpacing * 2;
+        const horizontalLegendHeight = legend.maxHeight ?? height * 0.1;
+
+        const verticalLegendWidth = legend.maxWidth ?? width * 0.1;
```
alantreadway, line 767: So we're by default scaling to 10% of the canvas width/height? Does that look right in common scenarios, like `position: 'right'` where we might want to take up more like 90% of the height?

alantreadway (commented):
Nice! Just one query, and could we add some test coverage (I believe I added a `legend.test.ts` recently that you could extend perhaps?)

What are your plans for documenting the new options?

charts-packages/ag-charts-community/src/chart/chart.ts
manapeirov: So if the legend is vertical (position `right` or `left`) we use the full chart height. If horizontal, we use 10 %

alantreadway: On line 765 we're calculating the 10% height value too? Does this seem like a good default?

charts-packages/ag-charts-community/src/chart/gridLayout.ts
```diff
@@ -0,0 +1,245 @@
...
+                }
+
+                guess = pageIndices < guess && pageIndices > minGuess ? pageIndices : guess;
+                continue;
+            }
+
+            processedBBoxCount += pageIndices.length * pageIndices[0].length; // this is an estimate, not all rows/columns will have the same length
+            rawPages.push(pageIndices);
+            break;
+        }
```
alantreadway, line 84: Nit: Maybe we should move this into a distinct function, to reduce the complexity of the enclosing function?

(`gridLayout()` is then orchestrating things, and delegating to other functions to perform calculations?)

charts-packages/ag-charts-community/src/chart/gridLayout.ts
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    // rawPages[pageNumber][secondaryDim][primaryDim]
```
alantreadway, line 87: Nit: remove?

alantreadway (commented):
Very close :-)

alantreadway (approved):
LGTM!

PR #5864, opened 2022-12-08
AG-7624 - Improve IOS/iPadOS Safari 12+ memory use

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7624

Attempts to mitigate and/or workaround differences in iOS/iPadOS Safari 12+ canvas memory allocation vs. every other browser.

Related WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=195325

Changes in this PR:
- For mobile devices, assume DPR of 1 to avoid allocating excessive amounts of memory for chart rendering (2^2->2^3 memory footprint reduction).
- Cleanup of canvas resources on page unload to avoid Safari not disposing of resources on even full-page reload.
- Collapse axis canvas count to 1 optimally (more if `crosslines` are present though) - reduces memory usage per chart by a few MB.

Observed improvements:
- I can now load the Stacked Area examples repeatedly by hitting reload on iOS. Before this change, after 8 reloads the chart would fail to load.
- Chrome is reporting a reduction in memory footprint of ~20MB for the multi-chart Plunker: https://plnkr.co/edit/h521Vb2i5Quincno


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -221,7 +221,7 @@ export class Axis<S extends Scale<D, number>, D = any> {
         return this._scale;
     }
 
-    protected readonly axisGroup = new Group({ name: `${this.id}-axis`, layer: true, zIndex: Layers.AXIS_ZINDEX });
+    protected readonly axisGroup = new Group({ name: `${this.id}-axis`, zIndex: Layers.AXIS_ZINDEX });
```
alexanderby, line 224: Need some explanation what could be the consequences of axes not being a layer.

alexanderby (approved):

PR #5881, opened 2022-12-14
AG-7655 Scales improvements

alexanderby:
- Simplify scales.
- Use single way flow for updating scales.
- Remove unused code.


alantreadway:
Looking at this snapshot diff:
https://github.com/ag-grid/ag-grid/pull/5881/files#diff-47289c474dafb2cb4dc4bf657be090e683660ec8661dda22061e0df4fd899401

...it looks like axis label collision detection is no longer working? The main principle behind the detection is to dynamically adjust the ticks/count and check for collisions iteratively - so that might be a clue as to where to look?

alantreadway (approved):
LGTM - thanks for fixing the collision issue!

PR #5921, opened 2022-12-21
AG-6682 - split filter params interfaces into internal/external

peterjrreynolds:


grid-packages/ag-grid-docs/documentation/doc-pages/filter-number/index.md
```diff
@@ -8,13 +8,9 @@ The [Provided Filters](/filter-provided/) and [Simple Filters](/filter-provided-
...
+Number Filters are configured though the `filterParams` attribute of the column definition:
 
-<interface-documentation interfaceName='IProvidedFilterParams' names='["buttons","closeOnApply","debounceMs","readOnly"]' config='{"description":""}' overrideSrc="filter-provided/resources/provided-filters.json"></interface-documentation>
-
-In addition, the following parameters are also available for number filters:
-
-<interface-documentation interfaceName='INumberFilterParams' names='["alwaysShowBothConditions","filterOptions","defaultOption","defaultJoinOperator","suppressAndOrCondition","caseSensitive","textFormatter","trimInput","inRangeInclusive","includeBlanksInEquals","includeBlanksInLessThan","includeBlanksInGreaterThan","includeBlanksInRange","allowedCharPattern","numberParser","comparator","browserDatePicker"]' config='{"description":""}' overrideSrc="filter-number/resources/number-filter-params.json"></interface-documentation>
+<interface-documentation interfaceName='INumberFilterParams' config='{"description":""}' overrideSrc="filter-number/resources/number-filter-params.json"></interface-documentation>
```
StephenCooper, line 13: Guessing this now displays the same list as before in one section?

community-modules/core/src/ts/filter/provided/date/dateFilter.ts
```diff
@@ -24,6 +25,10 @@ export interface DateFilterModel extends ISimpleFilterModel {
     dateTo: string | null;
 }
 
+// internal type
```
StephenCooper, line 28: I think for all of these interfaces that are now split we need to provide good JsDoc annotation. These should help the user know which interface they should be using.

i.e for the IDateFilterParams this should talk about being the properties that can be used to configure the provided date filter. 
While the DateFilterParams is what the grid will provide to you if you are creating a custom date filter.

StephenCooper:
Have you looked if there are any examples that can now use the simpler interfaces to configure the provided filters. This would be a good way to show how to use them. 

We could not do this previously as the interface would complain about all the grid provided properties.

grid-packages/ag-grid-docs/documentation/doc-pages/filter-number/index.md
peterjrreynolds: Yes. They all had to specify overrides before because they had to exclude the external parameters

peterjrreynolds:
> Have you looked if there are any examples that can now use the simpler interfaces to configure the provided filters. This would be a good way to show how to use them.
> 
> We could not do this previously as the interface would complain about all the grid provided properties.

The examples which already had types are using the correct ones as I maintained the names for the external types (it's the internal ones with the new names). Can definitely add to some more though

grid-packages/ag-grid-docs/documentation/doc-pages/filter-api/examples/filter-api-readonly/main.ts
```diff
@@ -37,11 +37,11 @@ const columnDefs: ColDef[] = [
     filter: 'agMultiColumnFilter',
     filterParams: {
       filters: [
-        { filter: 'agTextColumnFilter', filterParams: { readOnly: true } },
-        { filter: 'agSetColumnFilter', filterParams: { readOnly: true } },
+        { filter: 'agTextColumnFilter', filterParams: { readOnly: true } as IProvidedFilterParams },
```
StephenCooper, line 40: Could these not be the more specific interfaces for text and set filters?

StephenCooper (approved):
LGTM apart from one example which have commented on. 

I like this pattern and I hope we can do the same for other provided components.

PR #5922, opened 2022-12-21
AG-1896 - add rowSelected and selectionChanged source property

peterjrreynolds:


community-modules/core/src/ts/events.ts
```diff
@@ -81,7 +81,36 @@ export interface RowDataUpdatedEvent<TData = any> extends AgGridEvent<TData> { }
...
+ * - `uiSelectAll` - select all in header clicked
+ * - `rowDataChanged` - row data updated which triggered selection updates
+ * - `parentSelected`- parent selected when groupSelectsChildren enabled
+ * - `rowGroupChanged` - grouping changed which updated the selection
+ * - `selectAll` - API selectAll called
+ * - `deselectAll` - API deselectAll called
+ * - `selectAllFiltered` - API selectAllFiltered called
+ * - `deselectAllFiltered` - API deselectAllFiltered called
+ */
+export type SelectionEventType =
```
peterjrreynolds, line 98: Open to better names for these

StephenCooper: Maybe `SelectionSourceType`

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -140,19 +140,20 @@ export class ClientSideNodeManager {
         const selectionChanged = nodesToUnselect.length > 0;
         if (selectionChanged) {
             nodesToUnselect.forEach(rowNode => {
-                rowNode.setSelected(false, false, true);
+                rowNode.setSelected(false, false, true, 'rowDataChanged');
```
StephenCooper, line 143: I think I would be tempted to pass the source into this method to prevent duplication and ensure that if it is used in other places that we have the flexibility to use a different source.

community-modules/core/src/ts/events.ts
```diff
@@ -81,7 +81,36 @@ export interface RowDataUpdatedEvent<TData = any> extends AgGridEvent<TData> { }
...
+    'rowClicked' |
+    'checkboxSelected' |
+    'uiSelectAll' |
+    'rowDataChanged' |
+    'parentSelected' |
+    'rowGroupChanged' |
+    'selectAll' |
+    'deselectAll' |
+    'selectAllFiltered' |
+    'deselectAllFiltered';
```
StephenCooper, line 110: I wonder if these should start with `api` seeing as that is where they are called from?

community-modules/core/src/ts/headerRendering/cells/column/selectAllFeature.ts
```diff
@@ -187,9 +187,9 @@ export class SelectAllFeature extends BeanStub {
         const value = this.cbSelectAll.getValue();
 
         if (value) {
-            this.selectionService.selectAllRowNodes(this.filteredOnly);
+            this.selectionService.selectAllRowNodes('uiSelectAll', this.filteredOnly);
         } else {
-            this.selectionService.deselectAllRowNodes(this.filteredOnly);
+            this.selectionService.deselectAllRowNodes('uiSelectAll', this.filteredOnly);
```
StephenCooper, line 192: should this be `uiDeselectAll` to be consistent with other sources

community-modules/core/src/ts/interfaces/iRowNode.ts
```diff
@@ -167,8 +168,9 @@ export interface IRowNode<TData = any> extends BaseRowNode<TData>, GroupRowNode<
      * @param newValue -`true` for selection, `false` for deselection.
      * @param clearSelection - If selecting, then passing `true` will select the node exclusively (i.e. NOT do multi select). If doing deselection, `clearSelection` has no impact. Default: `false`
      * @param suppressFinishActions - Pass `true` to prevent the `selectionChanged` from being fired. Note that the `rowSelected` event will still be fired. Default: `false`
+     * @param source - Source property that will appear in the `selectionChanged` event. Default: `'api'`
      */
-    setSelected(newValue: boolean, clearSelection?: boolean, suppressFinishActions?: boolean): void;
+    setSelected(newValue: boolean, clearSelection?: boolean, suppressFinishActions?: boolean, source?: SelectionEventType): void;
```
StephenCooper, line 173: We should probably expose the params method as well as this one. I will do that as part of the IRowNode introduction ticket because this method is getting a few too many args.

community-modules/core/src/ts/events.ts
peterjrreynolds: I was referring to the values themselves, but good point. Although I originally named this to be in sync with `ColumnEventType`. Probably worth renaming that too as it refers to sources and not event types (I can keep the old name and make it deprecated).

I think it's worth keeping `event` in the name (`ColumnSourceType` sounds a bit misleading), so `SelectionEventSourceType` and `ColumnEventSourceType`

community-modules/core/src/ts/interfaces/iRowNode.ts
StephenCooper: Actually this looks like the params method is just for internal convenience and was not intended for external use

community-modules/core/src/ts/events.ts
peterjrreynolds: Actually, the `ColumnEventType` is used as both a source and an event type, so will leave it alone.

community-modules/core/src/ts/entities/rowNode.ts
```diff
@@ -1015,19 +1019,20 @@ export class RowNode<TData = any> implements IEventEmitter, IRowNode<TData> {
         this.selected = newValue;
 
         if (this.eventService) {
-            this.dispatchLocalEvent(this.createLocalRowEvent(RowNode.EVENT_ROW_SELECTED));
+            this.dispatchLocalEvent(Object.assign({}, this.createLocalRowEvent(RowNode.EVENT_ROW_SELECTED), {source}));
```
StephenCooper, line 1022: Nit: I personally prefer the spead syntax of { ...ob1, ...ob2}

community-modules/core/src/ts/events.ts
```diff
@@ -81,7 +81,38 @@ export interface RowDataUpdatedEvent<TData = any> extends AgGridEvent<TData> { }
...
+ * - `uiDeselectAll` - select all in header clicked (deselect)
+ * - `rowDataChanged` - row data updated which triggered selection updates
+ * - `parentSelected`- parent selected when groupSelectsChildren enabled
+ * - `rowGroupChanged` - grouping changed which updated the selection
+ * - `apiSelectAll` - API selectAll called
+ * - `apiDeselectAll` - API deselectAll called
+ * - `apiSelectAllFiltered` - API selectAllFiltered called
+ * - `apiDeselectAllFiltered` - API deselectAllFiltered called
+ */
+export type SelectionEventSourceType =
```
StephenCooper, line 99: I would order these and the JsDoc above alphabetically to give a nicer grouping and make it easier to skim.

community-modules/core/src/ts/gridApi.ts
```diff
@@ -681,22 +681,22 @@ export class GridApi<TData = any> {
 
     /** Select all rows, regardless of filtering and rows that are not visible due to grouping being enabled and their groups not expanded. */
     public selectAll() {
```
StephenCooper, line 683: I thought we were going to set the source as a default arguement to then enable the user to override them if they wanted?

community-modules/core/src/ts/entities/rowNode.ts
```diff
@@ -859,12 +859,13 @@ export class RowNode<TData = any> implements IEventEmitter, IRowNode<TData> {
      * @param clearSelection - If selecting, then passing `true` will select the node exclusively (i.e. NOT do multi select). If doing deselection, `clearSelection` has no impact.
      * @param suppressFinishActions - Pass `true` to prevent the `selectionChanged` from being fired. Note that the `rowSelected` event will still be fired.
      */
-    public setSelected(newValue: boolean, clearSelection: boolean = false, suppressFinishActions: boolean = false) {
+    public setSelected(newValue: boolean, clearSelection: boolean = false, suppressFinishActions: boolean = false, source: SelectionEventSourceType = 'api') {
```
StephenCooper, line 862: You need to update the IRowNode.ts file to match this.

community-modules/core/src/ts/events.ts
StephenCooper: Also maybe the sources should not include the deselect versions. As we are not consistent with having this split. So maybe its best to be consistent in not providing this.

For example if a user clicks the row to trigger an event they will have to check the selected status to know which way it is going. So they should be able to do the same with selectAll / deselectAll?

community-modules/core/src/ts/entities/rowNode.ts
peterjrreynolds: Should have already been updated

StephenCooper (approved):
LGTM

PR #5932, opened 2022-12-23
AG-6834 - Add `axis.label.minGap` option to prevent eager axis label …

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-6834

Added `axis.label.minGap` property to allow configuration of gap between axis labels when deciding whether tick count should be reduced


alexanderby (approved):
Awesome!

PR #5951, opened 2023-01-04
Tweak naming of Chart options for better consistency.

alantreadway:
Renames `avoidCollision` to `autoCollisionRemoval` for better consistency with other automatic label-layout behaviour options.

Moves `seriesToggleEnabled` as a nested property of `item`, which better maps to its conceptual relationship.


manapeirov (approved):
LGTM thank you!

PR #5961, opened 2023-01-06
AG-7762 - Force legend grid layout if maxWidth or maxHeight is config…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7762

If maxWidth/maxHeight is larger than chart width/height, the legend still gets displayed
For consistency, we are forcing legend grid layout to display at least one row or column of legend items if the maxWidth/maxHeight is smaller than chart width.height


alantreadway (approved):
LGTM

PR #5964, opened 2023-01-06
AG-7754 Color string values for Treemap example

alexanderby:
- Access for the Treemap node datum has been removed.
- Use Hex colors for Treemap overview example.


alantreadway (approved):
LGTM

PR #5987, opened 2023-01-10
AG-7792 - Fix inconsistent legend placement when paged

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7792

Ensures the that `BBox` used for legend layout takes legend multiple pages and their sizes into account.


manapeirov (approved):
LGTM!

PR #6006, opened 2023-01-13
Update dom.ts fixes function "addStylesToElement"

leonelvsc:
Currently if an object containing some of it's properties with ie: null | undefined the function throws an Error when calling ".replace()"

### DIFF
--- a/community-modules/core/src/ts/utils/dom.ts
+++ b/community-modules/core/src/ts/utils/dom.ts
@@ -367,8 +367,10 @@ export function addStylesToElement(eElement: any, styles: any) {
         if (!key || !key.length) { return; }
         // changes the key from camelCase into a hyphenated-string
         const parsedKey = key.replace(/[A-Z]/g, s => `-${s.toLocaleLowerCase()}`);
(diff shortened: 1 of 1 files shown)


leonelvsc:
Could someone review this ?

StephenCooper:
Hi @leonelvsc, 

Can you please provide a reproduction of how you run into this error? That will help with the validation process of this PR.

leonelvsc:
Hi, yes with a function that returns an object in the row or cell style ie { 'background-color': 'red' } ... Then if I want to delete that style I set that property value to null. It was working on previous versions of ag-grid, but with the last update it expects a string value

StephenCooper:
Thanks I can reproduce now and can see where this is coming from. 

We have some other changes in this area to make but I have raised AG-7917 to resolve this issue. 

I would recommend reporting these type of issues initially as a Github Issue as opposed to directly raising a PR. We do not monitor our PRs in the same way as Issues which is why it took a while to notice this. 

I am going to close this now as we will be fixing this along with other changes in this area which you can see have caused a merge conflict. 

Thanks for reporting this!

PR #6008, opened 2023-01-13
AG-7775 Simplify Array utilities

alexanderby:
- Remove unnecessary Array utilities.
- Accept numbers and dates for `extent` utility.
- Make `extent` always return numbers.


alantreadway (approved):
LGTM, nice work!

alantreadway (approved):
Thanks for the updates to Sparklines.

PR #6015, opened 2023-01-16
AG-7783 - Switch Legend to use HTML tooltips

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7783
https://ag-grid.atlassian.net/browse/AG-7790

Switches `Legend` to use the HTML-based `Tooltip` class, removing the (ab)use of `Canvas.title` to manage legend tooltips.

Additionally introduces `TooltipManager` to handle conflicts from multiple tooltip sources correctly.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
+++ b/charts-packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
@@ -1,0 +1,63 @@
+import { Tooltip, TooltipMeta } from '../tooltip/tooltip';
+
+interface TooltipState {
+    content: string;
+    meta: TooltipMeta;
+}
+
+/**
+ * Manages the tooltip HTML an element. Tracks the requested HTML from distinct dependents and
+ * handles conflicting tooltip requests.
+ */
+export class TooltipManager {
+    private readonly states: Record<string, TooltipState> = {};
+    private readonly tooltip: Tooltip;
+    private appliedState?: TooltipState;
+
+    public constructor(tooltip: Tooltip) {
+        this.tooltip = tooltip;
+    }
+
+    public updateTooltip(callerId: string, meta?: TooltipMeta, content?: string) {
+        if (content == null) {
+            content = this.states[callerId]?.content;
+        }
+
+        delete this.states[callerId];
+
+        if (meta != null && content != null) {
+            this.states[callerId] = { content, meta };
+        }
+
+        this.applyStates();
+    }
+
+    private applyStates() {
+        let contentToApply: string | undefined = undefined;
+        let metaToApply: TooltipMeta | undefined = undefined;
+
+        // Last added entry wins.
+        Object.entries(this.states)
+            .reverse()
+            .slice(0, 1)
+            .forEach(([_, { content, meta }]) => {
+                contentToApply = content;
+                metaToApply = meta;
+            });
+
+        if (metaToApply === undefined || contentToApply === undefined) {
+            this.appliedState = undefined;
+            this.tooltip.toggle(false);
+            return;
+        }
+
+        if (this.appliedState?.content === contentToApply) {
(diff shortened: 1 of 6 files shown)


charts-packages/ag-charts-community/package.json
```diff
@@ -1,105 +1,105 @@
...
-  "publishConfig": {
-    "access": "public"
-  },
-  "lint-staged": {
-    "*.{ts,js,md,json}": "charts-packages/ag-charts-community/.lint-staged/run.sh prettier",
-    "*.ts": "charts-packages/ag-charts-community/.lint-staged/run.sh autodocs"
-  }
-}
\ No newline at end of file
+    "name": "ag-charts-community",
```
alexanderby, line 2: Maybe we should exclude this file from Prettier, or make it have 2 space tabulation?

charts-packages/ag-charts-community/src/chart/interaction/baseManager.ts
```diff
@@ -6,7 +6,9 @@ export type Listener<E> = {
 export abstract class BaseManager<EventTypes extends string = never, EventType = never> {
     protected readonly registeredListeners: { [I in EventTypes]?: Listener<EventType>[] } = {};
 
-    protected constructor() {}
+    protected constructor() {
```
alexanderby, line 9: If it's abstract class, TypeScript will probably not allow instantiating it?

charts-packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -0,0 +1,63 @@
...
+        delete this.states[callerId];
+
+        if (meta != null && content != null) {
+            this.states[callerId] = { content, meta };
+        }
+
+        this.applyStates();
+    }
+
+    private applyStates() {
```
alexanderby, line 35: 🇺🇸

alexanderby (approved):

charts-packages/ag-charts-community/src/chart/interaction/baseManager.ts
alantreadway: Fair point - fixed distinctly in: https://github.com/ag-grid/ag-grid/commit/917db69eeeeb2488096e505724c03980be03325b

charts-packages/ag-charts-community/package.json
alantreadway: Undone for now, I don't want to drop `prettier` for `package.json`, but equally I don't think we can configure file-specific rules (I'll need to investigate).

alantreadway:
Backed out the contentious points as they weren't really related to the core change.

charts-packages/ag-charts-community/package.json
alantreadway: Fixed as part of: https://github.com/ag-grid/ag-grid/commit/9d3de58f20ae2502b036e1a08f72571953369708

alantreadway:
@alexanderby PTAL

charts-packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
alexanderby: Why there are multiple states?

alexanderby (approved):

charts-packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
alantreadway: There are currently two sources of tooltip state, the the chart (although maybe that should be the series themselves) and the legend - practically for any given (x,y) there should only be one tooltip source, but managing the state by source like this does mean that if there is a bug in the state updates for one source, it doesn't block the other source(s).

This is following the same pattern as other exclusive resources, such as the `cursor` and `highlight` state.

PR #6025, opened 2023-01-17
AG-7703 Simplify time utilities

alexanderby:


alantreadway (approved):
Nice work - it looks like this reimplementation might be more efficient, do you know if that is the case?

alexanderby:
I think we don't use those intervals extensively :) For 10 ticks it shouldn't matter.

PR #6027, opened 2023-01-18
Add Charts agChartsSceneDebug option.

alantreadway:
Adds `window.agChartsSceneDebug` to render bounding boxes of scene nodes for debugging.

Supports specification of scene graph ids, group names or aliases:
- `layout` - alias for all layout-related groups.

Screenshot 2023-01-18 at 11 54 59

### DIFF
--- a/charts-packages/ag-charts-community/src/scene/group.ts
+++ b/charts-packages/ag-charts-community/src/scene/group.ts
@@ -104,16 +104,6 @@ export class Group extends Node {
         super.markDirty(source, type, parentType);
     }
 
-    findNode(id: string | RegExp): Node[] | undefined {
-        if (typeof id === 'string' && this.name === id) {
-            return [this];
-        } else if (typeof id !== 'string' && this.name != null && id.test(this.name ?? '')) {
-            return [this];
-        }
-
-        return super.findNode(id);
-    }
-
     // We consider a group to be boundless, thus any point belongs to it.
     containsPoint(_x: number, _y: number): boolean {
         return true;
--- a/charts-packages/ag-charts-community/src/scene/group.ts
+++ b/charts-packages/ag-charts-community/src/scene/group.ts
@@ -104,6 +104,16 @@ export class Group extends Node {
         super.markDirty(source, type, parentType);
     }
 
+    findNode(id: string | RegExp): Node[] | undefined {
+        if (typeof id === 'string' && this.name === id) {
+            return [this];
+        } else if (typeof id !== 'string' && this.name != null && id.test(this.name ?? '')) {
+            return [this];
+        }
+
+        return super.findNode(id);
+    }
+
     // We consider a group to be boundless, thus any point belongs to it.
     containsPoint(_x: number, _y: number): boolean {
         return true;
--- a/charts-packages/ag-charts-community/src/scene/node.ts
+++ b/charts-packages/ag-charts-community/src/scene/node.ts
@@ -335,22 +335,14 @@ export abstract class Node extends ChangeDetectable {
         }
     }
 
-    findNode(id: string | RegExp): Node[] | undefined {
-        if (typeof id === 'string') {
-            if (this.id === id) {
-                return [this];
-            }
+    findNodes(predicate: (node: Node) => boolean): Node[] | undefined {
+        const result: Node[] = predicate(this) ? [this] : [];
 
-            if (this.childSet[id]) {
-                return [this.children.find((n) => n.id === id)!];
-            }
-        } else if (id.test(this.id)) {
-            return [this];
-        }
-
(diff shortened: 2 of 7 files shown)


charts-packages/ag-charts-community/src/scene/group.ts
```diff
@@ -104,6 +104,16 @@ export class Group extends Node {
         super.markDirty(source, type, parentType);
     }
 
+    findNode(id: string | RegExp): Node[] | undefined {
```
alexanderby, line 107: Did you mean `findNode**s**`? Then the result can be legally an array, whether it is empty or not.

alexanderby: Why it has to be overriden? Why can't the RegExp be used for other nodes?

charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -334,6 +335,27 @@ export abstract class Node extends ChangeDetectable {
...
+    findNode(id: string | RegExp): Node[] | undefined {
+        if (typeof id === 'string') {
+            if (this.id === id) {
+                return [this];
+            }
+
+            if (this.childSet[id]) {
+                return [this.children.find((n) => n.id === id)!];
+            }
+        } else if (id.test(this.id)) {
```
alexanderby, line 347: Children may also be tested positive for the same RegExp.

charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -334,6 +335,27 @@ export abstract class Node extends ChangeDetectable {
         }
     }
 
+    findNode(id: string | RegExp): Node[] | undefined {
```
alexanderby, line 338: Maybe it could be any predicate function?

charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -334,6 +335,27 @@ export abstract class Node extends ChangeDetectable {
...
+            if (this.childSet[id]) {
+                return [this.children.find((n) => n.id === id)!];
+            }
+        } else if (id.test(this.id)) {
+            return [this];
+        }
+
+        const result: Node[] = [];
+        for (const child of this.children) {
+            result.push(...(child.findNode(id) ?? []));
```
alexanderby, line 353: children.filter?

charts-packages/ag-charts-community/src/scene/group.ts
alantreadway: Good point, I'll rename (I was originally not considering `RegExp` matches, so the result was 0...1)

`Group` is a special case as it has a `name` property, which is not currently supported by other `Node` types.

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: Hmmmm.. that might clean this code up a bit, good idea

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: This is now cleaned-up - I couldn't use `Array.filter()` due to the need to recurse.

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: Fixed

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: Reworked

alantreadway:
@alexanderby PTAL

charts-packages/ag-charts-community/src/scene/node.ts
```diff
@@ -334,6 +335,19 @@ export abstract class Node extends ChangeDetectable {
         }
     }
 
+    findNodes(predicate: (node: Node) => boolean): Node[] | undefined {
```
alexanderby, line 338: Just curious: why it returns `undefined` instead of `[]`? I think it's okay to return an empty array, e.g. like in `querySelectorAll`.

alexanderby (approved):

charts-packages/ag-charts-community/src/scene/node.ts
alantreadway: Maybe a premature optimisation, we can probably go either way here

alexanderby: Ah, so it's for not creating an empty array for each node?

alantreadway: Fixed

PR #6033, opened 2023-01-18
AG-7775 Simplify Observable superclass

alexanderby:


alantreadway (approved):
LGTM, let's do a smoke test once this lands just to be sure.

PR #6036, opened 2023-01-19
Ag 7761/legend item truncation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7761

- Modified legend constraints docs example so it doesn't jump when the user interacts with the `item.maxWidth` slider
- Split legend item text label truncation into its own function


alantreadway (approved):
LGTM

PR #6040, opened 2023-01-19
AG-7848 Update browser utils

StephenCooper:
The browser utils need to be updated to match the current state of our supported browsers. There is some dead code relating to older versions of Edge and IE that can be removed.

Also the Firefox detection is soon to fail as the property used has been deprecated. 

I have also reduced the code require to get the browser version to just what is currently require, i.e the Safari version


community-modules/core/src/ts/dragAndDrop/dragAndDropService.ts
```diff
@@ -475,6 +474,15 @@ export class DragAndDropService extends BeanStub {
         ghost.style.top = `${top}px`;
     }
 
+
+    private getBodyWidth(): number {
+        return document.body?.clientWidth ?? (window.innerHeight || document.documentElement?.clientWidth || -1);
+    }
+
+    private getBodyHeight(): number {
```
gportela85, line 482: same as `getBodyWidth`

PR #6042, opened 2023-01-19
AG-7775 Rework object utilities

alexanderby:


alexanderby:
Maybe it will be better to review the 3 commits separately, because I'm not sure if any of these utilities has any use outside.

alantreadway (approved):
LGTM - discussed risky use-cases in Integrated Charts offline, they are unaffected from preliminary testing.

PR #6050, opened 2023-01-20
AG-3390 - Add website only docs runner

taktran:
## Changes

* Add a `npm run docs-serve-website-only` command to skip example generation, and only serve the gatsby server. For website development


seanlandsman (approved):
Looks fine to me

PR #6051, opened 2023-01-20
AG-7775 - Remove redundant Path2d.cubicArc().

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7775

Remove redundant Path2D.cubicArc().


alexanderby (approved):
Since we use arcs only for very basic shapes, and probably will not need to compute any shapes intersections, we should be fine.

PR #6052, opened 2023-01-20
AG-7746 - Fix log axis scale min value

manapeirov:

Validation for log axis min and max values, both cannot be less than 1


alantreadway (approved):
LGTM

PR #6054, opened 2023-01-20
AG-7871 - Fix main navigation current page bug

markdurrant:
- Removed 'Documentation' as the default current page for main nav selected item. 
- Stoped using `useState`/`setActive(link.name)` so the nav changes selected item when the user clicks non-nav links. E.g. links in the footer.


seanlandsman (approved):
looks good to me

PR #6055, opened 2023-01-20
AG-7462 - Check year change using first and last tick values instead …

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7462
- Use tick values to determine changes in year across axis ticks


alantreadway (approved):
LGTM

PR #6061, opened 2023-01-23
AG-7806 - handle filter destroyed in tool panel

peterjrreynolds:


enterprise-modules/filter-tool-panel/src/filterToolPanel/toolPanelFilterComp.ts
```diff
@@ -142,19 +152,24 @@ export class ToolPanelFilterComp extends Component {
...
         _.setAriaExpanded(this.eFilterToolPanelHeader, false);
-        this.agFilterToolPanelBody.removeChild(this.agFilterToolPanelBody.children[0]);
+        this.removeFilterComponent();
 
         _.setDisplayed(this.eExpandChecked, false);
         _.setDisplayed(this.eExpandUnchecked, true);
     }
 
+    private removeFilterComponent(): void {
+        this.agFilterToolPanelBody.removeChild(this.agFilterToolPanelBody.children[0]);
```
gportela85, line 170: This seems a bit obscure, why is it always the first child? Should we user `clearElement` from the utils here? 

Also `removeFilterComponent` reads like we are removing `Component` instance. This should probably be called something like `removeFilterElement`.

enterprise-modules/filter-tool-panel/src/filterToolPanel/toolPanelFilterComp.ts
```diff
@@ -119,6 +125,10 @@ export class ToolPanelFilterComp extends Component {
         _.setDisplayed(this.eExpandChecked, true);
         _.setDisplayed(this.eExpandUnchecked, false);
 
+        this.addFilterComponent();
```
gportela85, line 128: Same thing as the comment below, this should probably be called `addFilterElement`

gportela85:
looks good

PR #6064, opened 2023-01-23
Use package warning instead of module

StephenCooper:
Aim is to make the missing enterprise module slightly more accurate when working with the umd package in javascript. 

So instead of 

```
AG Grid: unable to use colDef.rowGroup as the RowGroupingModule is not registered. Check if you have registered the module:
           
    import { ModuleRegistry } from '@ag-grid-community/core';
    import { RowGroupingModule } from '@ag-grid-enterprise/row-grouping';
    
    ModuleRegistry.registerModules([ RowGroupingModule ]);

For more info see: https://www.ag-grid.com/javascript-grid/modules/
```

We would now get

```
AG Grid: unable to use colDef.rowGroup as package 'ag-grid-enterprise' has not been imported. Check that you have imported the package:
            
    import 'ag-grid-enterprise';
            
For more info see: https://www.ag-grid.com/javascript-grid/packages/
```


seanlandsman (approved):
Looks good to me

PR #6065, opened 2023-01-23
AG-7293 - Add basic Charts FW wrapper export of chart reference.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7293

Adds support for exposing the Chart instance managed by FW wrapper tags, to allow use of `AgChart.download()` and other API features.

- Angular ([example](https://plnkr.co/edit/p58uM0Z9NtjceJ1Y?open=app%2Fapp.component.ts)):
  - Exposes `AgChartsAgnular.chart` for parent elements to be able to access via a `@ViewChild()` decorated property.
  - Adds a `onChartReady` event which is triggered after the initial rendering of the chart completes.
- React ([example](https://plnkr.co/edit/HXnjbkzjXtKXtcMR?open=index.jsx&preview)):
  - Exposes `AgChartsReact.chart` for parent elements to be able to access via a `ref`.
  - Adds a `onChartReady` event which is triggered after the initial rendering of the chart completes.
- Vue ([example](https://plnkr.co/edit/WHUxB6dxBRoZaeEx?open=main.js)):
  - Adds a `onChartReady` callback which is triggered after the initial rendering of the chart completes.
  - Exposes `AgChartsVue.chart` and AgChartsVue3.chart` for parent elements to be able to access (I'm not familiar with whether this is actually necessary).

### DIFF
--- a/charts-packages/ag-charts-react/src/agChartsReact.ts
+++ b/charts-packages/ag-charts-react/src/agChartsReact.ts
@@ -4,6 +4,8 @@ import { AgChartInstance, AgChart, AgChartOptions } from "ag-charts-community";
 
 export interface AgChartProps {
     options: AgChartOptions;
+    onChartReady?: (chart: AgChartInstance) => void;
+    containerStyle: {};
 }
 
 interface AgChartState {
@@ -12,12 +14,12 @@ interface AgChartState {
 export class AgChartsReact extends Component<AgChartProps, AgChartState> {
     static propTypes: any;
 
-    private chart!: AgChartInstance;
+    public chart?: AgChartInstance;
 
     protected chartRef: RefObject<HTMLElement>;
 
-    constructor(public props: any, public state: any) {
-        super(props, state);
+    constructor(public props: AgChartProps) {
+        super(props);
         this.chartRef = createRef();
     }
 
@@ -37,7 +39,12 @@ export class AgChartsReact extends Component<AgChartProps, AgChartState> {
 
     componentDidMount() {
         const options = this.applyContainerIfNotSet(this.props.options);
-        this.chart = AgChart.create(options);
+
+        const chart = AgChart.create(options);
+        this.chart = chart;
+
+        (chart as any).chart.waitForUpdate()
+            .then(() => this.props.onChartReady?.(chart));
     }
 
     private applyContainerIfNotSet(propsOptions: any) {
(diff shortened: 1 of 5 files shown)


charts-packages/ag-charts-react/src/agChartsReact.ts
```diff
@@ -12,12 +14,12 @@ interface AgChartState {
...
     static propTypes: any;
 
-    private chart!: AgChartInstance;
+    public chart?: AgChartInstance;
 
     protected chartRef: RefObject<HTMLElement>;
 
-    constructor(public props: any, public state: any) {
-        super(props, state);
+    constructor(public props: AgChartProps) {
```
StephenCooper, line 21: What the reason behind moving the `state`?

charts-packages/ag-charts-react/src/agChartsReact.ts
```diff
@@ -4,6 +4,8 @@ import { AgChartInstance, AgChart, AgChartOptions } from "ag-charts-community";
 
 export interface AgChartProps {
     options: AgChartOptions;
+    onChartReady?: (chart: AgChartInstance) => void;
+    containerStyle: {};
```
StephenCooper, line 8: Is `containerStyle` meant to be included here?

seanlandsman: it allows users to override the `div` we need to create to host the chart

seanlandsman (approved):
looks like a good first stab at this

charts-packages/ag-charts-react/src/agChartsReact.ts
alantreadway: There was previously a deprecation warning about using the two-arg constructor at this call-site; I looked at the AgGridReact* tag implementations and it looked like this could be removed safely - but LMK if I'm missing something?

charts-packages/ag-charts-react/src/agChartsReact.ts
alantreadway: It was referenced on line 36, because I changed the constructor from taking `any` to `AgChartProps` the interface here needs to be extended to match with the code. I'm not really sure of the history/reasoning for this specific item though.

alantreadway:
Talked offline with @StephenCooper about the outstanding points, no further action required - merging shortly.

PR #6066, opened 2023-01-23
Ag 7746/log axis domain

manapeirov:


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -1032,7 +1033,9 @@ export class Axis<S extends Scale<D, number>, D = any> {
 
     // For formatting (nice rounded) tick values.
     formatTickDatum(datum: any, index: number): string {
-        const { label, labelFormatter, fractionDigits } = this;
+        const { label, labelFormatter, fractionDigits, scale } = this;
+
+        const logScale = scale instanceof LogScale;
 
         return label.formatter
```
alantreadway, line 1040: The logic here is quite unreadable, would you mind splitting these nested ternary statements this into a simple `if ... else if ... else ...` block?

charts-packages/ag-charts-community/src/util/array.ts
```diff
@@ -33,17 +33,25 @@ export function extent(values: Array<number | Date>): [number, number] | undefin
  * finds the min and max using a process appropriate for stacked values. Ie,
  * summing up the positive and negative numbers, and returning the totals of each
  */
-export function findMinMax(values: number[]): { min?: number; max?: number } {
+export function findMinMax(values: number[]): { min?: number; max?: number; extent?: [number, number] } {
```
alantreadway, line 36: We should really rename this to `findMinMaxExtent`?

alantreadway: Should we be combining the responsibilities for `min`, `max` and `extent` though? Seems like we end up with two distinct (and later divergent `extent` calculations?

manapeirov:
@alantreadway PTAL

PR #6070, opened 2023-01-24
AG-5744 - reset set filter excel mode on filter close

peterjrreynolds:


community-modules/core/src/ts/events.ts
```diff
@@ -48,6 +48,10 @@ export interface AgGridEvent<TData = any> extends AgGridCommon<TData>, AgEvent {
 
 export interface ToolPanelVisibleChangedEvent<TData = any> extends AgGridEvent<TData> {
     source: string | undefined;
```
peterjrreynolds, line 50: This is named a bit misleadingly (as is the event itself). `source` contains the active tool panel (or undefined if hidden). This is not inline with other `source` properties, which is why I've added `key`. There was also no way to track when a specific panel is hidden, which is why I've added `previousKey`

enterprise-modules/menu/src/menu/enterpriseMenu.ts
```diff
@@ -304,6 +304,8 @@ export class EnterpriseMenu extends BeanStub {
         if (this.mainMenuList) {
             this.mainMenuList.setParentComponent(this.tabbedLayout);
         }
+
+        this.addDestroyFunc(() => this.destroyBean(this.tabbedLayout));
```
peterjrreynolds, line 308: `tabbedLayout` bean is created, but not destroyed. Closing the menu destroys it, so we need to pass on.

community-modules/core/src/ts/events.ts
```diff
@@ -48,6 +48,10 @@ export interface AgGridEvent<TData = any> extends AgGridCommon<TData>, AgEvent {
...
+    /** Key for tool panel that is now visible, or `undefined` if hidden  */
+    key: string | undefined;
```
gportela85, line 52: it might be less confusing to have `key` always present with the name of the panel, and have another property like `visible: boolean`.

enterprise-modules/menu/src/menu/enterpriseMenu.ts
gportela85: good find :)

community-modules/core/src/ts/events.ts
peterjrreynolds: After wider discussions, following this approach but with a temporary new internal event. In next major release will swap out with existing event.

community-modules/core/src/ts/events.ts
peterjrreynolds: Fixed the source property in the new event.

PR #6072, opened 2023-01-25
AG-7293 - Rework Chart API docs ready for FW cases.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7293

Update Chart API docs so we can demonstrate how to use `AgChart.download()` in a way that makes sense for FW cases.


alantreadway:
Updated after feedback from @rmc-software - here's where we've landed:
localhost_8000_javascript-charts_api-themes_ (2)
localhost_8000_javascript-charts_api-themes_ (3)
localhost_8000_javascript-charts_api-themes_ (4)
localhost_8000_javascript-charts_api-themes_ (5)
localhost_8000_javascript-charts_api-themes_ (6)

As Rob has reviewed offline, I'm going for merge now.

PR #6076, opened 2023-01-25
AG-7906 Use standard import for AgChart

StephenCooper:
The following could be made cleaner by just directly importing AgChart in our typescript examples. So instead of 

```
  import * as agCharts from "ag-charts-community";
  agCharts.AgChart.update(chart, options);

 agCharts.time.month;
```
We would have the more standard

```
import { AgChart, time } from "ag-charts-community";
AgChart.update(chart, options);

time.month
```

### DIFF
--- a/grid-packages/ag-grid-docs/documentation/doc-pages/charts-overview/examples/custom-marker-shapes/main.ts
+++ b/grid-packages/ag-grid-docs/documentation/doc-pages/charts-overview/examples/custom-marker-shapes/main.ts
@@ -1,5 +1,4 @@
-import * as agCharts from "ag-charts-community";
-import { AgChartOptions } from "ag-charts-community";
+import { AgChart, AgChartOptions, Marker } from "ag-charts-community";
 import { getData } from "./data";
 
 var markerSize = 10
@@ -147,10 +146,10 @@ const options: AgChartOptions = {
   ],
 }
 
-var chart = agCharts.AgChart.create(options)
+var chart = AgChart.create(options)
 
 function heartFactory() {
-  class Heart extends agCharts.Marker {
+  class Heart extends Marker {
     rad(degree: number) {
       return (degree / 180) * Math.PI
     }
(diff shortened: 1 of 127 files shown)


grid-packages/ag-grid-docs/documentation/doc-pages/charts-overview/examples/custom-marker-shapes/main.ts
```diff
@@ -147,10 +146,10 @@ const options: AgChartOptions = {
   ],
 }
 
-var chart = agCharts.AgChart.create(options)
+var chart = AgChart.create(options)
 
 function heartFactory() {
-  class Heart extends agCharts.Marker {
+  class Heart extends Marker {
```
StephenCooper, line 152: Need to check this one is working.

StephenCooper: Now working as expected.

grid-packages/ag-grid-docs/src/example-generation/chart-vanilla-to-vue3.ts
```diff
@@ -7,13 +7,21 @@ function processFunction(code: string): string {
...
 
-function getImports(componentFileNames: string[]): string[] {
+function getImports(componentFileNames: string[], bindingImports: BindingImport[]): string[] {
     const imports = [
         "import { createApp } from 'vue';",
-        "import * as agCharts from 'ag-charts-community';",
         "import { AgChartsVue } from 'ag-charts-vue3';",
     ];
 
+    const chartsImport = bindingImports.find(i => i.module === "'ag-charts-community'");
```
StephenCooper, line 16: I know that this code is duplicated across generators but I would like to tackle that in a separate work item.

alantreadway (approved):
This is excellent, thank you for doing this!

PR #6085, opened 2023-01-26
AG-7461 Fix detecting arc intersections

alexanderby:
This PR solves the regression after replacing fake cubic arcs with actual arcs. It is the first step in solving the actual point click issue.


alantreadway (approved):
LGTM - thanks!

PR #6092, opened 2023-01-27
AG-7867 Fix tick SI format

alexanderby:


alantreadway (commented):
Is there a test case we can easily add to avoid future regressions?

alantreadway (approved):
Thanks for the updates @alexanderby - let's catch-up face-to-face tomorrow on the FP issue you've highlighted?

PR #6093, opened 2023-01-27
AG-6834 - Add `tick.interval` option

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7908


charts-packages/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -572,6 +572,9 @@ export interface AgAxisNumberTickOptions extends AgAxisBaseTickOptions {
      * The axis is not guaranteed to use exactly this number of ticks, but will try to use a number of ticks that is close to the number given.
      */
     count?: number;
+    /** The step value between ticks specified as a number. If the configured interval results in dense ticks given the data domain, the ticks will be removed.
+     */
+    // interval?: number;
```
alexanderby, line 577: I wish it could be uncommented.

charts-packages/ag-charts-community/src/scale/timeScale.ts
```diff
@@ -225,18 +225,18 @@ export class TimeScale extends ContinuousScale {
...
-            return interval;
-        }
+        const { tickIntervals } = this;
 
+        let countableTimeInterval;
         let step;
-        const tickCount = this.tickCount ?? 10;
-        const tickIntervals = this.tickIntervals;
+
+        const tickCount = count ?? 10;
```
alantreadway, line 239: Is this the right place for the fallback default of `10` to be declared? (Just thinking that we will shortly want to change this default, it would be good to know that there aren't any hidden default values lurking around)

alantreadway (commented):
Nice, liking the test test coverage! Are there really no visual snapshot changes?

charts-packages/ag-charts-community/src/scale/timeScale.ts
```diff
@@ -285,10 +285,50 @@ export class TimeScale extends ContinuousScale {
...
+            return interval.range(new Date(start - 1), new Date(stop + 1));
+        }
+
+        const step = Math.abs(interval);
+
+        if (this.isDenseInterval({ start, stop, interval: step })) {
+            return [];
+        }
+
+        const timeInterval = tickIntervals.find((tickInterval) => tickInterval[2] === step && tickInterval[1] === 1);
```
alexanderby, line 314:
```ts
tickInterval[2] * tickInterval[1] === step
```

alexanderby:
```ts
// reverse() can be replaced with a backward "for" loop
tickIntervals.reverse().find((tickInterval) =>  step % tickInterval[2] === 0);
```
This way you will be able to get the interval and use the `.every(step / foundInterval)`

charts-packages/ag-charts-community/src/scale/timeScale.ts
alexanderby: This is what I was working on today, trying to find a single source of truth for all those `10`s

manapeirov:
> Nice, liking the test test coverage! Are there really no visual snapshot changes?

Apparently all unchanged, maybe because my changes only apply to the case where tick.interval is specified which never used to be the case, so previous behaviour should be unchanged

charts-packages/ag-charts-community/src/chart/agChartOptions.ts
manapeirov: Patience my child

manapeirov:
@alantreadway PTAL

alantreadway (approved):
LGTM

PR #6106, opened 2023-01-31
AG-7905 - Fix listener handling on chart type changes

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7905

Fixes setting of `listeners` and `legend.listeners` using `AgChart.updateDelta()` for Integrated Charts use-cases.

Additionally adds a new example to demonstrate this.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartV2.ts
@@ -395,7 +395,7 @@ function applyChartOptions(chart: Chart, processedOptions: Partial<AgChartOption
         registerListeners(chart, processedOptions.listeners);
     }
     if (processedOptions.legend?.listeners) {
-        Object.assign(chart.legend.listeners, processedOptions.legend.listeners);
+        Object.assign(chart.legend.listeners, processedOptions.legend.listeners ?? {});
     }
 
     chart.processedOptions = jsonMerge([chart.processedOptions ?? {}, processedOptions], noDataCloneMergeOptions);
--- a/charts-packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/charts-packages/ag-charts-community/src/chart/agChartV2.ts
@@ -353,19 +353,19 @@ function debug(message?: any, ...optionalParams: any[]): void {
 }
 
 function applyChartOptions(chart: Chart, processedOptions: Partial<AgChartOptions>, userOptions: AgChartOptions): void {
+    let skip = ['type', 'data', 'series', 'autoSize', 'listeners', 'theme', 'legend.listeners'];
     if (isAgCartesianChartOptions(processedOptions)) {
-        applyOptionValues(chart, processedOptions, {
-            skip: ['type', 'data', 'series', 'axes', 'autoSize', 'listeners', 'theme'],
-        });
+        // Append axes to defaults.
+        skip.push('axes');
     } else if (isAgPolarChartOptions(processedOptions) || isAgHierarchyChartOptions(processedOptions)) {
-        applyOptionValues(chart, processedOptions, {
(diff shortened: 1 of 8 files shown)


charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -353,20 +353,20 @@ function debug(message?: any, ...optionalParams: any[]): void {
 }
 
 function applyChartOptions(chart: Chart, processedOptions: Partial<AgChartOptions>, userOptions: AgChartOptions): void {
+    let skip = ['type', 'data', 'series', 'autoSize', 'listeners', 'theme', 'legend.listeners'];
```
alexanderby, line 356: That's it: `legend.listeners`

charts-packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -395,7 +395,7 @@ function applyChartOptions(chart: Chart, processedOptions: Partial<AgChartOption
         registerListeners(chart, processedOptions.listeners);
     }
     if (processedOptions.legend?.listeners) {
-        Object.assign(chart.legend.listeners, processedOptions.legend.listeners);
+        Object.assign(chart.legend.listeners, processedOptions.legend.listeners ?? {});
```
alexanderby, line 398: Curious how is `undefined` treated here

charts-packages/ag-charts-community/src/util/observable.ts
```diff
@@ -8,6 +8,10 @@ export class Observable {
     private allEventListeners = new Map<string, Set<TypedEventListener>>();
 
     addEventListener(type: string, listener: TypedEventListener): void {
+        if (typeof listener !== 'function') {
+            throw new Error('AG Charts - listener must be a Function');
```
alexanderby, line 12: A warning probably?

alexanderby (approved):

charts-packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway: LHS is never `undefined` (as it's out implementation model), but if it were then `Object.assign()` fails with:
```
> Object.assign(undefined, {})
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.assign (<anonymous>)
```

alexanderby: I mean, what happened when the second argument was undefined?

charts-packages/ag-charts-community/src/util/observable.ts
alantreadway: So I added a warning in the `AgChartsV2.ts` changes, and this method never gets called - this is just belt & braces guard against the internal state becoming corrupted if there are any other use-cases.

charts-packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway: Yes, the root problem before was that the `jsonDiff` output was resulting in `legend: { listeners: undefined }` (which is correct behaviour), and then this code was applying that `undefined` value, even though these was distinct handling later in this method :-(

Adding `legend: { listeners: {} }` to the base theme removes the `undefined` from the mix, and this addition means we're not trying to process the update in two different ways 🤞

charts-packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway:
```
> Object.assign({}, undefined)
{}
```

So its strictly unnecessary, but I feel that the addition of `?? {}` here at least clarifies the intent and doesn't rely on knowledge of the detail of `Object.assign()` to be readable.

PR #6114, opened 2023-02-01
AG-7746 - Allow logAxis min/max to be any non-zero number

manapeirov:


alantreadway (approved):
LGTM!

PR #6126, opened 2023-02-06
AG-7850 - Replace TimeInterval WeakMap with constructor arg.

alantreadway:


alexanderby (commented):
Singular form can be used for the callbacks names.

PR #6143, opened 2023-02-09
AG-7991 - Add support for chained decorators.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7991

Adds a `@Default()` decorator to allow properties that don't default to `undefined` to declare what should be their alternative default.

This is important, as `AgChartV2` performs a JSON-diff, and properties that are removed end up with `undefined` values to allow reset of model properties that are no longer explicitly configured (on subsequent updates there is nothing in the diff for these properties, only on the transition from having a value to not having a value). `undefined` is not a suitable value to set for some options.

To enable this, I've implemented some utility code to allow decorators to be chained and moved the existing decorators to this pattern.

NOTE: I've deliberately not moved @SceneChangeDetection to the new pattern as it is performance sensitive.


manapeirov (approved):
LGTM!

PR #6148, opened 2023-02-10
Ag 7908/axis tick interval

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7908

Addresses Davids comments:
https://ag-grid.atlassian.net/browse/AG-7908?focusedCommentId=47309

1/

Requesting too many ticks via a small interval, warns the user as expected. 

However, no ticks at all are rendered. If we are ignoring the provided interval, we should display the default ‘nice’ tick number.

2/

Invalid options of providing a string is handled correctly. 
When provided with a negative interval, it is ignored but the message does not mention why.

Property [interval] cannot be set to [-1]; expecting an optional tick count Number value or, for a time axis, a Time Interval such as 'agCharts.time.month', ignoring.

3/
When providing an interval of 0, it is ignored (I think), and no warning message is displayed.

4/
Axis intervals don’t work properly at two decimal places.
https://plnkr.co/edit/dXcBmGSOa89vAaCn?open=main.js
The axis interval should be 0.75, but alternates between 0.7 and 0.8

### DIFF
--- a/charts-packages/ag-charts-community/src/scale/continuousScale.ts
+++ b/charts-packages/ag-charts-community/src/scale/continuousScale.ts
@@ -110,7 +110,17 @@ export abstract class ContinuousScale implements Scale<any, any> {
         }
     }
 
-    protected isDenseInterval({ start, stop, interval }: { start: number; stop: number; interval: number }): boolean {
+    protected isDenseInterval({
+        start,
+        stop,
+        interval,
+        count,
+    }: {
+        start: number;
+        stop: number;
+        interval: number | TimeInterval;
+        count?: number;
+    }): boolean {
         const { range } = this;
         const domain = stop - start;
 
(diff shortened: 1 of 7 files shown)


charts-packages/ag-charts-community/src/scale/continuousScale.ts
```diff
@@ -110,18 +110,31 @@ export abstract class ContinuousScale implements Scale<any, any> {
...
         const domain = stop - start;
 
         const min = Math.min(range[0], range[1]);
         const max = Math.max(range[0], range[1]);
 
         const availableRange = max - min;
-        const stepCount = domain / interval;
-        if (stepCount >= availableRange) {
+        const step = typeof interval === 'number' ? interval : 1;
+        count ??= domain / step;
```
alexanderby, line 132: I haven't seen `??=` before

charts-packages/ag-charts-community/src/util/ticks.ts
```diff
@@ -47,8 +47,8 @@ export class NumericTicks extends Array<number> {
 }
 
 export function range(start: number, stop: number, step: number): NumericTicks {
-    const fractionalRemainder = step % 1;
-    const fractionDigits = fractionalRemainder === 0 ? 0 : -Math.floor(Math.log10(fractionalRemainder));
+    const fractionalRemainder = step.toString().split('.')[1];
```
alexanderby, line 50: I'm afraid it can return `1.25e-20` or something. Will it have an impact on this calculation?

alantreadway: Maybe `step % 1` would work? E.g. `120 % 1 === 0` and `1.2 % 1 !== 0`

But you'd still need to do something like this to cover all cases:
```
const countDigits = (expNo) => {
  const parts = expNo.split('e');
  return Math.max((parts[0].split('.')[1]?.length ?? 0) - Number(parts[1]), 0);
};

fractionalDigits = countDigits((step % 1).toExponential());
```

manapeirov:
@alantreadway PTAL

alantreadway (approved):
LGTM - thanks Mana!

PR #6156, opened 2023-02-13
AG-7850 - Make snapTo: start the default time interval alignment.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7850

Change the default time interval behaviour to `start` alignment.


alantreadway:
DIscussed offline with @manapeirov - I've reviewed all examples and generally things are no worse than before, and in some cases better, so let's move forward with this change.

PR #6160, opened 2023-02-14
AG-7963 Performance Improvements for Quick Filter

StephenCooper:


community-modules/core/src/ts/rendering/cell/cellCtrl.ts
```diff
@@ -539,7 +539,13 @@ export class CellCtrl extends BeanStub {
 
         const valueParser = colDef.valueParser;
 
-        return exists(valueParser) ? this.beans.expressionService.evaluate(valueParser, params) : newValue;
+        if (exists(valueParser)) {
```
StephenCooper, line 542: Calling callback directly provides a more consistent performance, (especially in Chrome due to v8 optimisations). 

Also most users will not be using expressions so lets not add extra function depth to their hot path.

community-modules/core/src/ts/valueService/expressionService.ts
```diff
@@ -13,17 +13,12 @@ export class ExpressionService extends BeanStub {
         this.logger = loggerFactory.create('ExpressionService');
     }
 
-    public evaluate(expressionOrFunc: Function | string | undefined, params: any): any {
-        if (typeof expressionOrFunc === 'function') {
-            // valueGetter is a function, so just call it
-            const func = expressionOrFunc;
-            return func(params);
-        } else if (typeof expressionOrFunc === 'string') {
+    public evaluate(expression: string | undefined, params: any): any {
```
StephenCooper, line 16: Only accept a string so that this is only used when the user has actually provided an expression statement.

PR #6191, opened 2023-02-20
AG-7986 - handle same value in select cell editor

peterjrreynolds:


community-modules/core/src/ts/rendering/cellEditors/selectCellEditor.ts
```diff
@@ -56,7 +56,7 @@ export class SelectCellEditor extends PopupComponent implements ICellEditorComp
         // we don't want to add this if full row editing, otherwise selecting will stop the
         // full row editing.
         if (this.gridOptionsService.get('editType') !== 'fullRow') {
-            this.eSelect.onValueChange(() => params.stopEditing());
+            this.eSelect.onItemSelected(() => params.stopEditing());
```
gportela85, line 59: call `this.eventService.dispatchEvent(event)` instead

gportela85: Somewhere in the init method add

`this.addManagedEventListener(this.eventService, event, () => params.stopEditing())`

community-modules/core/src/ts/widgets/agSelect.ts
```diff
@@ -8,6 +8,7 @@ import { IAgLabel } from './agAbstractLabel';
 import { setAriaExpanded } from "../utils/aria";
 
 export class AgSelect extends AgPickerField<HTMLSelectElement, string> {
+    public static EVENT_ITEM_SELECTED = 'selectedItem';
```
gportela85, line 11: This should probably go in the Events class as `EVENT_SELECT_ITEM_SELECTED`

community-modules/core/src/ts/widgets/agSelect.ts
```diff
@@ -128,6 +132,12 @@ export class AgSelect extends AgPickerField<HTMLSelectElement, string> {
         return super.setValue(value, silent);
     }
 
+    public onItemSelected(callbackFn: () => void): this {
```
gportela85, line 135: this should be removed

PR #6193, opened 2023-02-20
AG-7522 Support Typescript Generic context

StephenCooper:


StephenCooper:
The changes here are too far reaching for the value that they bring. Will apply a much more targeted approach to common use cases as a first step and see if requests come from users for adding generic support to other areas of the grid.

Issue #5687, opened 2022-10-16
The `autoHeight` not working with Column Spanning


donnguyen:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

We have 2 columns, both with `autoHeight=true`.

**Current behavior**
- Seems like the auto height is not working with a row that spanning on both columns. Like the following image:
<img width="361" alt="image" src="https://user-images.githubusercontent.com/3904511/196028705-031ba16f-ea89-435b-9817-a8e212b398ad.png">

**Expected behavior**
- The autoHeight should be working as normal, even the row is spanning on multiple columns (all with `autoHeight=true)

**Please tell us about your environment:**

* **AG Grid version:** 28.2.0

* **Browser:**

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36 Edg/106.0.1370.42

* **Language:** [all]

jpike88:
maybe related to #5688

StephenCooper:
This seems to be working in this example. Note how Row 1 changes the row height if you change the width of Column B.

https://plnkr.co/edit/QZT1InENPIB34vgp?open=main.ts

Can you provided / update this to show how it does not work?

donnguyen:
hi @StephenCooper,

Thanks for your response and sorry for the reply.

We just need to make the 3rd column in your example to `autoHeight` (that is what I meant by `We have 2 columns, both with autoHeight=true.`),  the issue will happen like following:
<img width="1381" alt="Screenshot 2022-10-26 at 16 50 40" src="https://user-images.githubusercontent.com/3904511/197995497-73b3c1a7-92d1-495b-9dfa-476abd181a9e.png">

Hope this helps

kiril-matev:
Hello @donnguyen,

Thank you for reporting this - much appreciated!

Based on your description we've now reproduced this issue as a bug.

We have added this bug to our backlog and we are tracking it with the following reference and description:
**AG-7523 When a column with autoHeight & wrapText=true spans into another column with autoHeight & wrapText=true, the row has default instead of auto-height and the cell value appears on a single row instead of wrapped**

[...]

PR #6196, opened 2023-02-20
AG-7523 fix autoheight with column spanning

Kreash:
Fix autoheight issue for [Column Spanning](https://www.ag-grid.com/javascript-data-grid/column-spanning/)

The problem was identified in the following [issue](https://github.com/ag-grid/ag-grid/issues/5687)

issue number in the pipeline: **AG-7523**

### DIFF
--- a/grid-community-modules/core/src/ts/entities/rowNode.ts
+++ b/grid-community-modules/core/src/ts/entities/rowNode.ts
@@ -569,7 +569,6 @@ export class RowNode<TData = any> implements IEventEmitter, IRowNode<TData> {
     }
 
     public checkAutoHeights(): void {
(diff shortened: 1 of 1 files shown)


StephenCooper:
Thank you for contributing this. I have linked this fix to our internal jira so that when that ticket is prioritised this can be tested and validated. 

I am going to close this now as the changes will need to flow through our internal pipeline. 

Thank you again though for this.

Kreash:
In that case, is it possible to get a fix for the next release? 
This error makes us write a lot of crutches

StephenCooper:
@Kreash Just to let you know we should have this fixed as part of 29.3 which is targeted for the end of the month.

PR #6197, opened 2023-02-21
AG-8028 Respect tick minSpacing

alexanderby:
https://ag-grid.atlassian.net/browse/AG-8028


charts-packages/ag-charts-community/src/axis.ts
```diff
@@ -473,14 +473,16 @@ export class Axis<S extends Scale<D, number, TickInterval<S>>, D = any> {
...
-    private setTickCount(count?: TickCount<S> | number) {
+    private setTickCount(count?: TickCount<S> | number, minTickCount?: number, maxTickCount?: number) {
         const { scale } = this;
         if (!(count && scale instanceof ContinuousScale)) {
             return;
         }
 
         if (typeof count === 'number') {
             scale.tickCount = count;
+            scale.maxTickCount = minTickCount ?? 0;
```
manapeirov, line 484: This needs to be `scale.minTickCount`

alantreadway (commented):
Code changes look fine, but could we look into why the scale on some of the examples look off - also can we run these past Mana and David?

charts-packages/ag-charts-community/src/axis.ts
alexanderby: Oops

PR #6204, opened 2023-02-21
AG-7461 Fix series node click to check distance is zero

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-7461

This takes the simplistic approach of checking the pick distance is zero, and doesn't address separating highlighting and clicking behaviour / configuration.

### DIFF
--- a/charts-packages/ag-charts-community/src/chart/chart.ts
+++ b/charts-packages/ag-charts-community/src/chart/chart.ts
@@ -891,6 +891,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         | {
               series: Series<any>;
               datum: SeriesNodeDatum;
+              distance: number;
           }
         | undefined {
         const {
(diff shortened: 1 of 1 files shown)


charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1006,19 +1007,11 @@ export abstract class Chart extends Observable implements AgChartInstance {
     }
 
     private checkSeriesNodeClick(event: InteractionEvent<'click'>): boolean {
-        const { lastPick } = this;
+        const pick = this.pickSeriesNode({ x: event.offsetX, y: event.offsetY });
 
-        if (lastPick?.datum) {
-            const { datum } = lastPick;
-            datum.series.fireNodeClickEvent(event.sourceEvent, datum);
+        if (pick?.distance === 0) {
```
alantreadway, line 1012: I'm just wondering if it would be better to remove the dependency on `tracking.tooltip` here: https://github.com/ag-grid/ag-grid/pull/6204/files#diff-8ff81f59743760dfbaa596f9e1cf165391ebb1a38b88b65a26801613c15272efR904

Perhaps just add a boolean parameter to `pickSeriesNode()` for 'exactMatchOnly', and then the caller can decide if we're using default strategies or exact-match strategy? (So for the other use-cases, we'd supply `tooltip.tracking` for now, except for the call-site just above here)?

lsjroberts: Yep, that makes a lot of sense

alantreadway (approved):
Excellent, thanks for the updates!

PR #6205, opened 2023-02-21
AG-7522 Provide TContext generic to events and callbacks

StephenCooper:


community-modules/core/src/ts/entities/colDef.ts
```diff
@@ -74,7 +74,7 @@ export interface IAggFunc<TData = any, TValue = any> {
     (params: IAggFuncParams<TData, TValue>): any;
 }
 
-export interface IAggFuncParams<TData = any, TValue = any> extends AgGridCommon<TData> {
+export interface IAggFuncParams<TData = any, TValue = any> extends AgGridCommon<TData, any> {
```
StephenCooper, line 77: Leave TContext off ColDef interfaces for now to avoid potential future breaking changes if we want to do more with TValue

PR #6214, opened 2023-02-23
AG-8058 Add interaction range for nearest, exact or pixel distance to tooltip and node clicks

lsjroberts:
For https://ag-grid.atlassian.net/browse/AG-8058

This adds the `tooltip.range` and `series[].nodeClickRange` options which can take the values `'nearest'`, `'exact'` or a distance in canvas pixels.

It replaces and deprecates `tooltip.tracking`.

This needs some examples and testing, but thought I'd get feedback on the approach first.


charts-packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -126,6 +127,7 @@ export class ChartTheme {
                     color: 'black',
                 },
             },
+            nodeClickRange: 'exact' as AgChartInteractionRange,
```
lsjroberts, line 130: I'm inclined to default this to `5` pixels or similar, as small points on a line chart can be tricky to hit.

alantreadway: Lets run this choice for default past David?

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -915,7 +916,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
             if (!match || distance == null) {
                 continue;
             }
-            if (!result || result.distance > distance) {
+            if ((!result || result.distance > distance) && distance <= (maxDistance || Infinity)) {
```
alantreadway, line 919: Just wondering if Falsey evaluation might be a problem, should we use the `??` operator instead here:
```suggestion
            if ((!result || result.distance > distance) && distance <= (maxDistance ?? Infinity)) {
```

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -967,23 +969,38 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
         }
 
-        const pick = this.pickSeriesNode({ x: offsetX, y: offsetY }, !this.tooltip.tracking);
+        const isPixelRange = typeof range === 'number' && Number.isFinite(range);
+        const pick = this.pickSeriesNode(
+            { x: offsetX, y: offsetY },
+            range === 'exact',
+            // @ts-ignore Typescript doesn't recognise `range` must be a number here
+            isPixelRange ? range : undefined
+        );
```
alantreadway, line 978: I'd avoid // @ts-ignore where possible - maybe try:
```suggestion
        let pixelRange = undefined;
        if (typeof range === 'number' && Number.isFinite(range)) {
          pixelRange = range;
        }
        const pick = this.pickSeriesNode(
            { x: offsetX, y: offsetY },
            range === 'exact',
            pixelRange,
        );
```

charts-packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1006,9 +1023,31 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
+            return true;
+        }
+
+        // Then check for an exact match or within the given range
+        const isPixelRange = typeof nodeClickRange === 'number' && Number.isFinite(nodeClickRange);
+
+        const pick = this.pickSeriesNode(
+            { x: event.offsetX, y: event.offsetY },
+            nodeClickRange === 'exact',
+            // @ts-ignore Typescript doesn't recognise `nodeClickRange` must be a number here
```
alantreadway, line 1041: Likewise, lets avoid // @ts-ignore where possible

alantreadway (commented):
Looking good - have you considered doc updates?

Also we'll need to make sure the that JIRA `Deprecation Notes` field is populated to capture the deprecation.

charts-packages/ag-charts-community/src/chart/chart.ts
lsjroberts: Yeah, ts-ignore is not great. Only issue is I use `isPixelRange` again further down. But it does seem the only way to make ts happy is to directly wrap it in that `if` statement. We are a little behind on ts version but let's maybe save that for another day!

Will update as per your suggestion.

lsjroberts:
Fixed merge conflicts, I'll pop doc updates and tests on a separate PR.

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -967,23 +969,38 @@ export abstract class Chart extends Observable implements AgChartInstance {
             return;
         }
 
-        const pick = this.pickSeriesNode({ x: offsetX, y: offsetY }, !this.tooltip.tracking);
+        const isPixelRange = typeof range === 'number' && Number.isFinite(range);
```
alantreadway, line 972: Nit: Maybe move this to after the `if` block, and change to `const isPixelRange = pixelRange != null;` so that the logic isn't duplicated?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -1006,9 +1023,35 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
+        const nodeClickRange = datum?.series.nodeClickRange;
+
+        // First check if we should fire the event based on nearest node
+        if (datum && nodeClickRange === 'nearest') {
+            datum.series.fireNodeClickEvent(event.sourceEvent, datum);
+            return true;
+        }
+
+        // Then check for an exact match or within the given range
+        const isPixelRange = typeof nodeClickRange === 'number' && Number.isFinite(nodeClickRange);
```
alantreadway, line 1036: Ditto

alantreadway (approved):
Couple of nits but otherwise looks great! LGTM

PR #6219, opened 2023-02-24
AG-8048 Multiline Text node

sryazanov:
Then text splites with /\r?\n/ regex and renders in multiple lines.

All Text nodes now handle "new line" symbols. Captions may have `lineHeight` parameter, which is used to compute offset of a next line. If it's undefined (default), than it's derived from font metrics (not actual text size, so captions with the same font/size would have the same line height regardless of contents).


alantreadway (commented):
Generally looking good - just some specific points:

- Multi-line axis labels don't seem to be centred (especially on Y-axis), is it possible to address this?
- Were you able to verify that the Integrated Charts title-edit functionality is working ok after this change?

alantreadway (approved):
Excellent, thanks for the updates :-) LGTM

alantreadway (approved):
Thanks for the title edit updates, I can't visualise what it'll look like but I'll review it once merged. LGTM

PR #6220, opened 2023-02-24
AG 8058 Interaction range docs for tooltips and node clicks

lsjroberts:

Adds docs for `tooltip.range` and `series[].nodeClickRange` with switchable examples.

### DIFF
--- a/grid-packages/ag-grid-docs/documentation/doc-pages/charts-tooltips/index.md
+++ b/grid-packages/ag-grid-docs/documentation/doc-pages/charts-tooltips/index.md
@@ -15,8 +15,8 @@ The default chart tooltip has the following template:
 
 ```html
 <div class="ag-chart-tooltip">
-    <div class="ag-chart-tooltip-title"></div>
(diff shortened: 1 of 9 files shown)


grid-packages/ag-grid-docs/documentation/doc-pages/charts-tooltips/index.md
```diff
@@ -106,35 +101,34 @@ The actual type of the `params` object passed into the tooltip renderer will dep
 
 ```ts
 interface AgTooltipRendererParams {
```
alantreadway, line 103: Hmmm, hadn't noticed this before... I'd actually query if this should even be in the docs (inline) like this, would a link to an API doc table section be sufficient here?

alantreadway (commented):
Looks great, have you verified all FW permutations of the new examples work as expected (this is a common area to trip up on with new examples)?

lsjroberts:
@alantreadway Checked each FW and sub-FW, confirmed all work as expected.

PR #6222, opened 2023-02-24
AG-8052 Simplify Selection

alexanderby:
https://ag-grid.atlassian.net/browse/AG-8052


charts-community-modules/ag-charts-community/src/axis.ts
```diff
@@ -715,12 +716,12 @@ export class Axis<S extends Scale<D, number, TickInterval<S>>, D = any> {
             if (visible) {
                 anyTickVisible = true;
             }
-            return visible;
+            node.visible = visible;
```
alantreadway, line 719: Looks slightly suspicious since removing `return` alters the control flow?

alantreadway (commented):
Generally LGTM - I'll check out the branch and kick the wheels before merging a bit later today if everything checks out (I'd rather get this merged earlier in the release cycle to give us some chance of finding unforeseen side-effects before release).

charts-community-modules/ag-charts-community/src/axis.ts
alexanderby: It is the end of the function, so the return is unnecessary.

alexanderby: So, previously this was a callback for the `attr()` method, but now it is a callback for the `each()` method.

alantreadway:
I made some fixes to Sparklines (it was also using `Selection` heavily), but this is working without too much effort as it was following the exact same patterns as Standalone Charts.

PR #6225, opened 2023-02-27
AG-8058 Add tests for node click range 'nearest' and pixel distance

lsjroberts:
Final PR for https://ag-grid.atlassian.net/browse/AG-8058


alantreadway (approved):
LGTM

PR #6229, opened 2023-02-28
AG-8058 Add coerced editor to api explorer for mixed union types

lsjroberts:

This adds a simple string editor that coerces the type so the API editor can handle mixed union types like `type Foo = number | 'bar' | 'baz'`.

### DIFF
--- a/grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
+++ b/grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
@@ -3,25 +3,23 @@
  * type of the particular option.
  */
 
-import React, { useState } from 'react';
-import { HuePicker, AlphaPicker } from 'react-color';
+import { FontFamily, FontSize, FontStyle, FontWeight, Opacity, Ratio } from 'ag-charts-community';
 import classnames from 'classnames';
-
-import { FontFamily, FontWeight, FontStyle, FontSize, Opacity, Ratio } from 'ag-charts-community';
-
+import React, { useState } from 'react';
+import { AlphaPicker, HuePicker } from 'react-color';
+import { JsonModelProperty, JsonProperty } from '../expandable-snippet/model';
 import { doOnEnter } from '../key-handlers';
 import styles from './Editors.module.scss';
-import { JsonProperty, JsonModelProperty } from '../expandable-snippet/model';
 
 type AliasTypeProps<T> = {
-    default?: T,
-    options?: T[],
-    suggestions?: T[],
-    breakIndex?: number,
-    min?: T,
-    max?: T,
-    step?: T,
-    unit?: string,
+    default?: T;
+    options?: T[];
+    suggestions?: T[];
+    breakIndex?: number;
+    min?: T;
+    max?: T;
+    step?: T;
+    unit?: string;
 };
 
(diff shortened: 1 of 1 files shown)


grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
```diff
@@ -181,9 +175,28 @@ export const getPrimitiveEditor = ({ meta, desc }: JsonModelProperty, key: strin
...
+            'string',
+            'boolean',
+        ];
+
+        if (options.some((o) => validUnionTypes.includes(o))) {
+            options = options.filter((o) => validUnionTypes.includes(o));
+            return { editor: CoercedEditor, editorProps: meta };
+        }
+
+        options = options.filter((v) => v.startsWith("'") && v.endsWith("'")).map((v) => v.substring(1, v.length - 1));
```
lsjroberts, line 199: Actual relevant section of code above, outside the prettier changes

grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
```diff
@@ -299,26 +324,47 @@ export const JsonEditor = (props) => (
...
+            fromStringValue={(value) => {
+                if (value.toLowerCase() === 'false') return false;
+                if (value.toLowerCase() === 'true') return true;
+                if (!isNaN(value) && !isNaN(parseInt(value))) return parseInt(value);
+                return value;
+            }}
+            {...props}
+        />
+    );
+};
```
lsjroberts, line 339: And here

grid-packages/ag-grid-docs/documentation/src/components/charts-api-explorer/Editors.tsx
```diff
@@ -299,26 +324,47 @@ export const JsonEditor = (props) => (
...
 );
 
-export const BooleanEditor = ({ value, onChange }) => <PresetEditor options={[false, true]} value={value} onChange={onChange} />;
+export const CoercedEditor = (props) => {
+    return (
+        <StringEditor
+            fromStringValue={(value) => {
+                if (value.toLowerCase() === 'false') return false;
+                if (value.toLowerCase() === 'true') return true;
+                if (!isNaN(value) && !isNaN(parseInt(value))) return parseInt(value);
```
alantreadway, line 333: Not sure how easy it is to access the actual union type here, but would it be possible to add an extra guard so that these coercions only apply if one of the union types is `boolean` or `number` respectively?

lsjroberts: Perhaps, but this is also used for types like `PixelSize` which are aliases to `number`. I don't think we could get the underlying type. Or the type at all since this is working on the generated json of types not the actual typescript?

alantreadway: Sorry for the late reply - I think you should be able to pass the filtered `options` from line 195 in via `editorProps`, similar to line 202?

You could then only enable the right sub-set of coercions?

lsjroberts: @alantreadway makes sense, done, PTAL

alantreadway (approved):
LGTM, thanks for the updates!

PR #6230, opened 2023-02-28
reuse existing padding headers

StephenCooper:


StephenCooper:
This does not seem to be the right approach to a ReactUI only issue. Looks like this fix should be done in our React header components.

PR #6233, opened 2023-02-28
AG-8047 - Fix-up ESLint violations, soft-enable rule checks.

alantreadway:
Aligns `ag-charts-community` setup with new modules; fixes many `eslint` violations.

Notable fixes:
- Adds `logger.ts` as the central utility for any console logging (enures `AG Charts` prefix is present, and offers common idioms like `warnOnce()` as extra features vs. `console.warn()`).

Additionally:
- Extends `fix` target to run ESLint fixes.
- Enables incremental linting on git commit via hooks.


charts-community-modules/ag-charts-community/package.json
```diff
@@ -106,6 +111,7 @@
   },
   "lint-staged": {
     "*.{ts,js,md,json}": "charts-community-modules/ag-charts-community/.lint-staged/run.sh prettier",
+    "*.{ts,js}": "charts-community-modules/ag-charts-community/.lint-staged/run.sh eslint",
```
lsjroberts, line 114: Does this have the potential to cause issues with generated `.js` files, or do you reckon it'll be ok?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -466,7 +466,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
                 await this.processData();
                 this.disablePointer(true);
                 splits.push(performance.now());
-            // Fall-through to next pipeline stage.
+            // eslint-disable-next-line no-fallthrough
```
lsjroberts, line 469: We do this in a few places, should we just allow it?

lsjroberts:
Great stuff, LGTM

manapeirov (approved):

charts-community-modules/ag-charts-community/src/chart/chart.ts
alantreadway: We need a comment for SonarCloud checks too, so I think in this case it's fine - but generally I agree with the premise that if we're routinely adding ignores something should change (either the rules, or our style).

https://sonarcloud.io/summary/new_code?id=ag-charts-community-latest

PR #6238, opened 2023-03-01
AG-8058 Add node double click

lsjroberts:

This adds double click handlers for each type of series.


lsjroberts:
Wrong ticket id...

PR #6239, opened 2023-03-01
AG-7121 node double click

lsjroberts:

This adds double click handlers for each type of series.


alantreadway (approved):
LGTM, nice work!

PR #6240, opened 2023-03-01
AG-8048 Adjusting legend for multiline labels

sryazanov:
Two changes:
1. Each line in a multiline lable truncated individually.
2. Label positioning takes variable height into account.


charts-community-modules/ag-charts-community/src/chart/legend.ts
```diff
@@ -607,14 +608,18 @@ export class Legend {
                 return;
             }
 
-            y = itemHeight * rowIndex;
-            x = rowSumColumnWidths[rowIndex] ?? 0;
+            const x = rowSumColumnWidths[rowIndex] ?? 0;
+            const itemHeight = columns[columnIndex].bboxes[rowIndex].height + paddingY;
+
+            if (rowIndex > 0) y[columnIndex] += itemHeight / 2;
```
alantreadway, line 614: Nit: Could you use brackets to surround the assignment statement here?

alantreadway (commented):
Generally LGTM - it might be worth adding a couple of visual test cases for paginated legend cases with multiline labels perhaps?

Existing test-cases:
https://github.com/ag-grid/ag-grid/blob/5cfd60b1d51faaec97a0a9a92ce37a48b553026b/charts-community-modules/ag-charts-community/src/chart/legend.test.ts#L88-L105

Accompanying visual snapshot:
https://github.com/ag-grid/ag-grid/blob/latest/charts-community-modules/ag-charts-community/src/chart/__image_snapshots__/legend-test-ts-legend-large-series-count-chart-legend-pagination-should-render-legend-correctly-at-position-bottom-1-snap.png

charts-community-modules/ag-charts-community/src/chart/legend.ts
sryazanov: Done

alantreadway (commented):
LGTM, thanks for the test additions!

I've asked @manapeirov to take a look as she's our Legend Expert (TM), for awareness and to help spot any further gotchas.

charts-community-modules/ag-charts-community/src/chart/legend.ts
```diff
@@ -608,14 +609,20 @@ export class Legend {
...
+            }
 
             rowSumColumnWidths[rowIndex] = (rowSumColumnWidths[rowIndex] ?? 0) + column.columnWidth;
 
             // Round off for pixel grid alignment to work properly.
             markerLabel.translationX = Math.floor(x);
-            markerLabel.translationY = Math.floor(y);
+            markerLabel.translationY = Math.floor(y[columnIndex]);
+
+            y[columnIndex] += itemHeight / 2;
```
manapeirov, line 625: Maybe try `y[columnIndex] += itemHeight` here and remove lines 615 to 617, I wonder if that would reduce the snapshot updates

sryazanov: This doesn't work quite well (and it becomes apparent on multiline labels). Because the text has middle vertical alignment, it grows up and down, thus more space needs to be added before and after the label.

sryazanov:
The problem gets more complicated than I thought. Using actual text BBox for makes the legend looks inconsistent (we tricked that using only one BBox, which is not perfect and becomes much worse in multiline scenario). Using font metrics (instead of actual metrics) could make the legend look stable, but it isn't so easy to do. For now, let's park this PR and forbid multiple lines in legends.

alantreadway:
> The problem gets more complicated than I thought. Using actual text BBox for makes the legend looks inconsistent (we tricked that using only one BBox, which is not perfect and becomes much worse in multiline scenario). Using font metrics (instead of actual metrics) could make the legend look stable, but it isn't so easy to do. For now, let's park this PR and forbid multiple lines in legends.

Ok, makes sense - thanks for exploring this.

PR #6243, opened 2023-03-01
AG-8051 Footnote

sryazanov:


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -182,6 +182,21 @@ export abstract class Chart extends Observable implements AgChartInstance {
         return this._subtitle;
     }
 
+    _footnote?: Caption = undefined;
+    set footnote(caption: Caption | undefined) {
```
alantreadway, line 186: This pattern is repeated a few times, I'm wondering if we should just write a decorator to get rid of some of the boilerplate here?

I just searched the codebase and we only seem to be doing it for these `Caption` cases - I may just drop-in a localised solution once this change is merged.

alantreadway (commented):
Looks good. Could you consider adjusting the documentation here to account for the new feature:
https://ag-grid.com/javascript-charts/layout/

charts-community-modules/ag-charts-community/src/chart/chart.ts
sryazanov: decorator could be a bit too much for a piece of code that repeats only 3 times. I introduced helper methods to reduce code duplication instead.

alantreadway: Ah, I already added it (the case is more generic that these three cases it turned out) - I'll wait for you to merge your change and I'll deal with the conflicts:
https://github.com/ag-grid/ag-grid/pull/6250

sryazanov:
Sounds good. Let's update the doc in a separate PR.

PR #6249, opened 2023-03-02
AG-8048 Multiple lines are disabled in legends

sryazanov:
Multi-line legends are more complicated than tickets and should be done separately.


alantreadway (approved):
LGTM, thanks!

PR #6250, opened 2023-03-02
Cleanup chart.ts set/get side-effect methods.

alantreadway:
Adds `@ActionOnWrite` decorator to try and clean-up the pattern of private property with set/get pair with side-effect on set. Applied to `chart.ts` for now to verify the pattern.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/util/proxy.ts
+++ b/charts-community-modules/ag-charts-community/src/util/proxy.ts
@@ -1,11 +1,5 @@
 import { addTransformToInstanceProperty } from './decorator';
 
-/**
- * Wraps the target property with a set method that will write assigned value into a nested object with the same
- * property key.
- *
- * @param proxyProperty property name of this, which has the child object to receive values.
- */
 export function ProxyOnWrite(proxyProperty: string) {
     return addTransformToInstanceProperty((target, _, value) => {
         target[proxyProperty] = value;
@@ -14,13 +8,6 @@ export function ProxyOnWrite(proxyProperty: string) {
     });
 }
 
-/**
- * Wraps the target property with a set method that will write assigned value into a nested object with a specified
- * property key.
- *
- * @param childName property name of this, which has the child object to receive values.
- * @param childProperty property key of the child object to receive values.
- */
 export function ProxyPropertyOnWrite(childName: string, childProperty?: string) {
     return addTransformToInstanceProperty((target, key, value) => {
         target[childName][childProperty ?? key] = value;
@@ -30,12 +17,12 @@ export function ProxyPropertyOnWrite(childName: string, childProperty?: string) 
 }
 
 /**
- * Allows side-effects to be triggered on property write. Exactly one callback is invoked on every
- * write, provided the assigned value changes.
+ * Allows side-effects to be triggered on property write.
  *
- * @param opts.addFn called when a new value is set
- * @param opts.removeFn called with the old value when a new value is set
- * @param opts.changeFn called on any change to the value
+ * @param opts.add called when a new value is set - never called for undefined values.
+ * @param opts.remove called with the old value when a new value is set - never called for undefined
+ *                    values.
+ * @param opts.change called on any change to the value - always called.
  */
 export function ActionOnWrite<T>(opts: {
     add?: (this: T, newValue: any) => void;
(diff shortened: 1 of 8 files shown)


charts-community-modules/ag-charts-community/src/util/proxy.ts
```diff
@@ -15,3 +15,34 @@ export function ProxyPropertyOnWrite(childName: string, childProperty?: string)
...
+    remove?: (this: T, oldValue: any) => void;
+    change?: (this: T, newValue?: any, oldValue?: any) => void;
+}) {
+    const { add: addFn, remove: removeFn, change: changeFn } = opts;
+    return addTransformToInstanceProperty((target, __, newValue, oldValue) => {
+        if (newValue === oldValue) {
+            return newValue;
+        }
+
+        if (oldValue !== undefined) {
```
sryazanov, line 38: If I understand correctly, it changes the order of operations. Caption setters do things in this order:
1. Unregister old value
2. Set property
3. Register new value.

With this change the order gonna be this:
1. Unregister old value.
2. Register new value.
3. Set the property.

The second order is easier to implement (with a decorator or just a singe helper). I tired to do it, but rolled back because of failing tests. Apparently the order makes some difference (I guess, registration is affected by the property set).

alantreadway: Yes that's right, we could change the decorator to have the original order of execution but I think it's unnecessary (we should ideally be doing the minimal work in the side-effect code to update model state, if we are triggering processing that relies on the new value of the property, that's probably gone too far).

I also made modifications in other part of the code to remove dependencies for a few cases (which may contribute to some of these order-of-execution issues you ran into), and tested the changes for side-effects with Integrated Charts (`titleEdit.ts` needed some tweaks for the same reason).

charts-community-modules/ag-charts-community/src/util/proxy.ts
```diff
@@ -15,3 +15,34 @@ export function ProxyPropertyOnWrite(childName: string, childProperty?: string)
...
+
+/**
+ * Allows side-effects to be triggered on property write.
+ *
+ * @param opts.add called when a new value is set - never called for undefined values.
+ * @param opts.remove called with the old value when a new value is set - never called for undefined
+ *                    values.
+ * @param opts.change called on any change to the value - always called.
+ */
+export function ActionOnWrite<T>(opts: {
```
lsjroberts, line 27: I'd be inclined to call this `ActionOnSet` to be clearer that it is decorating the setter function.

charts-community-modules/ag-charts-community/src/util/proxy.ts
```diff
@@ -15,3 +15,34 @@ export function ProxyPropertyOnWrite(childName: string, childProperty?: string)
...
+    add?: (this: T, newValue: any) => void;
```
lsjroberts, line 28: I don't mind the name `add` much, but it's not really adding, the property is always there. And similarly `remove` doesn't remove the property.

I understand that it is used in the context of adding and removing nodes, but this decorator in isolation doesn't show that. And it appears to be used in other contexts where it is not adding or removing anything.

Perhaps the naming should reflect the behaviour around the handling of undefined? e.g. `always` or `any` for `change`?

lsjroberts (changes requested):

lsjroberts (commented):
ignore me, haven't used github review features much, ignore the changes requested thing

charts-community-modules/ag-charts-community/src/util/proxy.ts
alantreadway: Fair - fixed.

charts-community-modules/ag-charts-community/src/util/proxy.ts
alantreadway: So I've renamed these options to:
- `newValue` - called when a new (non-undefined) value is about to be set.
- `oldValue` - called when the old (non-undefined) value is about to be replaced.
- `changeValue` - called whenever the value is about to change.

Does that naming fit better?

lsjroberts: Yeah that sounds much better 👍

alantreadway:
@lsjroberts PTAL

lsjroberts (approved):
LGTM

PR #6256, opened 2023-03-03
AG-8051 Footnote docs

sryazanov:
https://ag-grid.atlassian.net/browse/AG-8051

Add footnote option documentation.


lsjroberts (approved):
LGTM

grid-packages/ag-grid-enterprise/src/main.ts
```diff
@@ -395,20 +395,16 @@ IsColumnFuncParams,
 ISelectCellEditorParams,
 ISelectionHandle,
 ISelectionHandleFactory,
-ISelectionService,
```
alantreadway, line 398: These should be unnecessary, head of `latest` is accurate right now (I added a CI check to detect this case a little while ago)

alantreadway (commented):
Could you check if the updated image (removing series labels) is consistent with the documentation around it?

https://ag-grid.com/javascript-charts/getting-started/#labels-and-titles

^ it looks like the series labels should be present in the image?

grid-packages/ag-grid-enterprise/src/main.ts
sryazanov: Reverted

sryazanov:
Updated the two other images for consistency.

sryazanov:
As David suggested I reversed the order of the series to make the legend appear in the increasing order. But it also reversed the stacks.

alantreadway:
Do there need to be additions in the layout page for the new element in the diagram?

PR #6258, opened 2023-03-04
100x hover/scroll performance when having large buffer size

gabrielpetersson:
Not an active user of ag-grid anymore, but I had this gist snippet laying around that I thought could be of use

I tried 500 row buffer on a 150k row table, and it was unusable(2fps) which made me a bit confused so checked the code, and one of the core functions is poorly written. The below change ran x100 faster on my above table, and I think it should make any grid with 5k+ rows & larger array buffer ~5-20% faster

Most notably `res.concat(rowCtrl.getAllCellCtrls())` is extremely expensive on larger tables, as it would create huge amounts of unused intermediate arrays 

Found this old loom as well to prove it: https://www.loom.com/share/5a44c244c66141f192fb2cd6202c101e

### DIFF
--- a/grid-community-modules/core/src/ts/rendering/rowRenderer.ts
+++ b/grid-community-modules/core/src/ts/rendering/rowRenderer.ts
@@ -606,16 +606,22 @@ export class RowRenderer extends BeanStub {
     }
 
     public getAllCellCtrls(): CellCtrl[] {
-        let res: CellCtrl[] = [];
-        this.getAllRowCtrls().forEach(rowCtrl => res = res.concat(rowCtrl.getAllCellCtrls()));
+        const res: CellCtrl[] = [];
+        const all = this.getAllRowCtrls()
+        for (const rowCtrl of all) {
+            for (const cell of rowCtrl.getAllCellCtrls()){
+                res.push(cell)
+            }
+        }
         return res;
     }
 
     private getAllRowCtrls(): RowCtrl[] {
         const stickyRowCtrls = (this.stickyRowFeature && this.stickyRowFeature.getStickyRowCtrls()) || [];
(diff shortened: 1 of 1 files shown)


grid-community-modules/core/src/ts/rendering/rowRenderer.ts
```diff
@@ -606,16 +606,22 @@ export class RowRenderer extends BeanStub {
...
+            }
+        }
         return res;
     }
 
     private getAllRowCtrls(): RowCtrl[] {
         const stickyRowCtrls = (this.stickyRowFeature && this.stickyRowFeature.getStickyRowCtrls()) || [];
         const res = [...this.topRowCtrls, ...this.bottomRowCtrls, ...stickyRowCtrls];
-
-        Object.keys(this.rowCtrlsByRowIndex).forEach(key => res.push(this.rowCtrlsByRowIndex[key]));
```
gabrielpetersson, line 618: minor optimization but no reason to create an intermediate array of keys when building a high performance table

StephenCooper:
Hi @gabrielpetersson,

Thank you for identifying this for us. I have raised AG-8147 for us to follow up on these changes as I can see why this code is suboptimal.

gabrielpetersson:
@StephenCooper awesome! I'd also suggest checking through the rest of the code, have seen similar stuff sprinkled through the codebase

gabrielpetersson:
@StephenCooper Would you mind linking the PR? Cannot seem to find it

StephenCooper:
There was no PR, but you can see the commits [here](https://github.com/ag-grid/ag-grid/commits/latest/grid-community-modules/core/src/ts/rendering/rowRenderer.ts). (Excuse my mistakes...)

PR #6265, opened 2023-03-06
AG-8047 - Remove non-modularised navigator.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8047


charts-community-modules/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -432,34 +432,34 @@ export class ChartTheme {
...
-                fill: '#f2f2f2',
-                stroke: '#999999',
-                strokeWidth: 1,
-                width: 8,
-                height: 16,
-                gripLineGap: 2,
-                gripLineLength: 8,
-            },
-        },
+        // navigator: {
```
lsjroberts, line 435: delete?

alantreadway: I need to work out how this is going to work for modules first :-(

PR #6266, opened 2023-03-07
AG-3479 chart zoom module support

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-3479

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
@@ -60,7 +60,10 @@ export class CartesianChart extends Chart {
         const { seriesRoot } = this;
         if (clipSeries) {
             const { x, y, width, height } = seriesRect;
-            seriesRoot.setClipRectInGroupCoordinateSpace(new BBox(x, y, width, height));
+            const { pixelRatio } = this.scene.canvas;
+            seriesRoot.setClipRectInGroupCoordinateSpace(
+                new BBox(x / pixelRatio, y / pixelRatio, width, height / pixelRatio)
+            );
         } else {
             seriesRoot.setClipRectInGroupCoordinateSpace(undefined);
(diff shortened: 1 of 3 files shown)


charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -60,7 +60,10 @@ export class CartesianChart extends Chart {
         const { seriesRoot } = this;
         if (clipSeries) {
             const { x, y, width, height } = seriesRect;
-            seriesRoot.setClipRectInGroupCoordinateSpace(new BBox(x, y, width, height));
+            const { pixelRatio } = this.scene.canvas;
+            seriesRoot.setClipRectInGroupCoordinateSpace(
+                new BBox(x / pixelRatio, y / pixelRatio, width / pixelRatio, height / pixelRatio)
+            );
```
lsjroberts, line 66: @manapeirov Does this make sense? I haven't looked much into the clip rect stuff.

alantreadway: Hmmmm.... it's a bit odd to be apply this here, the theory is that we have a wrapper for `Canvas` that takes DPR into account and we shouldn't need to worry about (except when converting from browser-event coordinates to canvas coordinates perhaps)?
https://github.com/ag-grid/ag-grid/blob/e0c912ff0d220b0b8e0523c0e20e9d596b1b96db/charts-community-modules/ag-charts-community/src/canvas/hdpiCanvas.ts#L354

It maybe that the spot where we apply clipping needs to take this into account though?

lsjroberts: This bug has been tracked here https://ag-grid.atlassian.net/browse/AG-8142

alantreadway (approved):
LGTM

PR #6269, opened 2023-03-07
AG-7121 double click fixes

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-7121

Fixes toggling legend items when you double click on a legend item that was disabled, while one or more other items were enabled. It will now disable all others and enable only the item clicked.

Also some minor docs improvements.

### DIFF
--- a/grid-packages/ag-grid-docs/documentation/doc-pages/charts-legend/examples/legend-click-series-toggle/main.ts
+++ b/grid-packages/ag-grid-docs/documentation/doc-pages/charts-legend/examples/legend-click-series-toggle/main.ts
@@ -1,4 +1,8 @@
-import { AgChart, AgChartLegendClickEvent, AgChartOptions } from "ag-charts-community"
+import {
+  AgChart,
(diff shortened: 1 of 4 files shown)


grid-packages/ag-grid-docs/documentation/doc-pages/charts-legend/examples/legend-click-series-toggle/main.ts
```diff
@@ -157,7 +161,7 @@ const options: AgChartOptions = {
         itemId,
         enabled,
       }: AgChartLegendClickEvent) => {
-        window.alert(
+        console.log(
```
alantreadway, line 164: Side-note: I wonder if there is a half-way house between `alert()` and `console.log()` that could give visual feedback without blocking the user (maybe a temporary pop-out panel at the bottom of the example which disappears after a short timeout e.g. 2s).

alantreadway (approved):

PR #6270, opened 2023-03-07
AG-8036 Chart background image

sryazanov:


charts-community-modules/ag-charts-community/src/chart/background.ts
```diff
@@ -1,21 +1,40 @@
...
+
+    @Validate(NUMBER(0))
+    private _width: number = 0;
     set width(value: number) {
-        this.node.width = value;
+        this._width = value;
+        this.rectNode.width = this._width;
+        if (this._image) {
+            this._image.containerWidth = this._width;
+        }
```
alantreadway, line 21: We have a lot of code that follows this `set` pattern, but I would like to try and reduce or eliminate it (as you can see from your changes, this logic has to be repeated in several places to work correctly with this approach).

I think it would be preferable to have a `update()` method that is called from `Chart` during a `Chart.update()` processing call - if you could rework the side-effect handing in this and other `set` functions into an `update()` here, maybe we can chat later about where to bind that into `Chart`?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -656,6 +658,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         if (this.scene.resize(width, height)) {
             this.background.width = width;
             this.background.height = height;
+            this.background.image?.updatePosition();
```
alantreadway, line 661: Maybe this is the right spot to perform a `Background.update()`?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -209,6 +209,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         const background = this.background;
         background.fill = 'white';
         root.appendChild(background.node);
+        background.onimageload = this.onBackgroundImageLoad;
```
alantreadway, line 212: Maybe we could make this a constructor parameter, and remove the need for `onimageload` to be declared as optional? I would also consider using a simple inline lambda for this case (e.g. `new Background(() => this.update(ChartUpdateType.SCENE_RENDER))` perhaps).

charts-community-modules/ag-charts-community/src/scene/image.ts
```diff
@@ -0,0 +1,40 @@
...
+            if (stats) stats.nodesSkipped++;
+            return;
+        }
+        this.computeTransformMatrix();
+        this.matrix.toContext(ctx);
+
+        const image = this.sourceImage;
+        if (image) {
+            ctx.globalAlpha = this.opacity;
+            ctx.drawImage(image, 0, 0, image.width, image.height, this.x, this.y, this.width, this.height);
```
alantreadway, line 35: I would wrap this with a `ctx.save()` and `ctx.restore()` pair of calls, since otherwise the `ctx.globalAlpha` change might leak to other operations on the canvas that happen after this?

charts-community-modules/ag-charts-community/src/scene/image.ts
```diff
@@ -0,0 +1,40 @@
...
+    width: number = 0;
+
+    @SceneChangeDetection({ redraw: RedrawType.MAJOR })
+    height: number = 0;
+
+    @SceneChangeDetection({ redraw: RedrawType.MAJOR })
+    opacity: number = 1;
+
+    @SceneChangeDetection({ redraw: RedrawType.MAJOR })
+    sourceImage?: HTMLImageElement = undefined;
```
alantreadway, line 20: Maybe make this a constructor argument, as it's only set once immediately after creation?

charts-community-modules/ag-charts-community/src/chart/backgroundImage.ts
```diff
@@ -0,0 +1,126 @@
...
+            this._image.addEventListener('load', this.onImageLoad);
+        }
+    }
+
+    get complete() {
+        return this._image.complete;
+    }
+
+    containerWidth: number = 0;
+    containerHeight: number = 0;
```
alantreadway, line 38: Maybe consider if we need to keep these as copies of the background width/height? If we moved to a model where we have an `BackgroundImage.update()` call to process changes, could these just be parameters to the call (to reduce the amount of copies of state laying around)?

alantreadway (commented):
Comments inline - I think it would be work considering putting together what the doc changes around this look like, including some examples too?

charts-community-modules/ag-charts-community/src/scene/image.ts
alantreadway: Discussed offline, not needed

charts-community-modules/ag-charts-community/src/chart/backgroundImage.ts
sryazanov: There are two reasons to keep these:
1. Positioning tests need to set the values without image update (as they don't load images).
2. They used to re-layout when image loaded.

PR #6288, opened 2023-03-10
AG-7461 Add default cursor behaviour for nodes with click listeners 

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-7461

Checkout this branch and see this plunker for a demo - https://plnkr.co/edit/JSpdRMVodhtiEX6N?open=main.ts&preview

New behaviour is:

- If the cursor is within `tooltip.range` and the `cursor` option is defined, display that cursor
- If the cursor is within `series[].nodeClickRange` and a listener is attached to one of: `seriesNodeClick`, `seriesNodeDoubleClick`, `series[].nodeClick` or `series[].nodeDoubleClick`, then display a `pointer` cursor

In the plunker above it will display the `wait` cursor when within tooltip range of the `foo` series, the `pointer` cursor when within node click range of the `foo` or `qux` series, and finally it will not change the cursor for the `bar` series as that one has no listeners. Note `qux` is a double click listener.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/scene/node.ts
+++ b/charts-community-modules/ag-charts-community/src/scene/node.ts
@@ -389,14 +389,6 @@ export abstract class Node extends ChangeDetectable {
         return bbox;
     }
 
-    computeBBoxCenter(): [number, number] {
-        const bbox = this.computeBBox && this.computeBBox();
-        if (bbox) {
-            return [bbox.x + bbox.width * 0.5, bbox.y + bbox.height * 0.5];
-        }
-        return [0, 0];
-    }
-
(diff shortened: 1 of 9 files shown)


charts-community-modules/ag-charts-community/src/scene/node.ts
```diff
@@ -389,14 +389,6 @@ export abstract class Node extends ChangeDetectable {
         return bbox;
     }
 
-    computeBBoxCenter(): [number, number] {
```
lsjroberts, line 392: This function doesn't appear to be used anywhere in the project

alantreadway (approved):
LGTM

PR #6298, opened 2023-03-13
AG-7121 Rationalise legend series toggling and events docs

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-7121

Moves legend event docs to the events page, and the series toggling to the legend page.


alantreadway (approved):
LGTM other than one nit

PR #6302, opened 2023-03-13
AG-8142 - Fix clipRect high DPR cases

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8142

`ctx.getTransform()` was returning the raw Canvas matrix, including DPR multipliers - when using this as the basis for converting `clipRect` coordinates back to the canvas coordinate space, we were then getting a result that was out by a factor of the DPR.

I've removed use of `ctx.getTransform()` in favour of using the inverse of the locally managed `this.matrix`, which doesn't have DPR multipliers applied.

Bonus:
- Adds a verification that `save()` and `restore()` calls are balanced (discussed offline with @manapeirov)

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
@@ -64,11 +64,7 @@ export class CartesianChart extends Chart {
 
         const { seriesRoot } = this;
         if (clipSeries) {
-            const { x, y, width, height } = seriesRect;
-            const { pixelRatio } = this.scene.canvas;
-            seriesRoot.setClipRectInGroupCoordinateSpace(
-                new BBox(x / pixelRatio, y / pixelRatio, width / pixelRatio, height / pixelRatio)
-            );
+            seriesRoot.setClipRectInGroupCoordinateSpace(seriesRect);
         } else {
             seriesRoot.setClipRectInGroupCoordinateSpace(undefined);
         }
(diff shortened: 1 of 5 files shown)


charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -64,11 +64,7 @@ export class CartesianChart extends Chart {
 
         const { seriesRoot } = this;
         if (clipSeries) {
-            const { x, y, width, height } = seriesRect;
-            const { pixelRatio } = this.scene.canvas;
-            seriesRoot.setClipRectInGroupCoordinateSpace(
-                new BBox(x / pixelRatio, y / pixelRatio, width / pixelRatio, height / pixelRatio)
-            );
+            seriesRoot.setClipRectInGroupCoordinateSpace(seriesRect);
```
lsjroberts, line 67: Excellent 👍

charts-community-modules/ag-charts-community/src/scene/scene.ts
```diff
@@ -383,6 +380,9 @@ export class Scene {
             ctx.restore();
         }
 
+        // Check for save/restore depth of zero!
+        ctx.verifyDepthZero?.();
```
lsjroberts, line 384: Is the expectation that this is unlikely to throw the error unless we've written some bad code? And the user is not going to be able to cause this through their own efforts? Since there's no handling of the error here.

lsjroberts (approved):

charts-community-modules/ag-charts-community/src/scene/scene.ts
alantreadway: (to add: we were already checking for <0 cases, so this just adds checks for >0 cases)

PR #6317, opened 2023-03-14
AG-8171 - handle popup resizing when elements exist after resizable element

peterjrreynolds:


grid-community-modules/core/src/ts/rendering/features/positionableFeature.ts
```diff
@@ -289,11 +289,19 @@ export class PositionableFeature extends BeanStub {
...
-
-            if (clientHeight && (height + yPosition > clientHeight + parentTop)) {
-                height = clientHeight - yPosition;
+            if (clientHeight) {
+                const yPosition = popup ? this.position.y : elRect.top;
+                const parentTop = popup ? 0 : parentRect.top;
+
+                // When `forcePopupParentAsOffsetParent`, there may be elements that appear after the resizable element, but aren't included in the height.
+                // Take these into account here
+                const boundaryBottom = this.boundaryEl?.getBoundingClientRect()?.bottom;
```
gportela85, line 298: Not a big thing, but wouldn't it be better if we did

```
let additionalHeight = 0;
if (forcePopupParentAsOffsetParent && this.boundaryEl) {
    const { bottom } = this.boundaryEl.getBoundingClientRect();
    additionalHeight = bottom - elRect.bottom;
}

const availableHeight = clientHeight + parentTop - yPosition - additionalHeight;

if (height > availableHeight) {
    height = availableHeight;
}
```

This way we only calculate the boundingClienteRect of `this.boundaryEl` when necessary.

PR #6319, opened 2023-03-14
AG-8058 Interaction & highlight range

lsjroberts:

Adds `chart.highlightRange` option to toggle highlighting triggering in line with the tooltip or the node clicking. The default is `'tooltip'` since it would be rare for the tooltip range to be smaller than the node click range.

This has all got bit complect with the pointer, highlighting, tooltip and clicking ranges and toggling. A subject for refactor once behaviour has settled.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
@@ -547,6 +547,13 @@ export interface AgBaseChartListeners {
     doubleClick?: (event: AgChartDoubleClickEvent) => any;
 }
 
+type AgChartHighlightRange = 'tooltip' | 'node';
+
+export interface AgChartHighlightOptions {
+    /** By default, nodes will be highlighted when the cursor is within the `tooltip.range`. Set this to `'node'` to highlight nodes when within the `series[].nodeClickRange`. */
+    range?: AgChartHighlightRange;
+}
+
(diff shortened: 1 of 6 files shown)


charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -577,6 +577,8 @@ export interface AgBaseChartOptions {
     legend?: AgChartLegendOptions;
     /** A map of event names to event listeners. */
     listeners?: AgBaseChartListeners;
+    /** By default, nodes will be highlighted when the cursor is within the `tooltip.range`. Set this to `'node'` to highlight nodes when within the `series[].nodeClickRange`. */
+    highlightRange?: 'tooltip' | 'node';
```
alantreadway, line 581: Just wondering if we should make this an object rather than a standalone property (these top-level properties tend to be nested objects, and I'm wondering if there is other `highlight` behaviour we might want to control in the future here too)?

e.g.:
`highlight?: { range?: 'tooltip' | 'node' };`

(Make sure to not use an inline type though, as the Options Reference won't render it correctly)

lsjroberts: That's a very good point, will update and add a proper type.

alantreadway (approved):
LGTM - thanks!

PR #6341, opened 2023-03-17
AG-3479 support

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-3479

Adds a drag event listener on the chart that triggers an update, this lets us drag out and draw a box rect selection.


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -965,6 +966,12 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.update(ChartUpdateType.SCENE_RENDER);
     }
 
+    protected onDrag(event: InteractionEvent<'drag'>): void {
+        this.extraDebugStats['mouseX'] = event.offsetX;
+        this.extraDebugStats['mouseY'] = event.offsetY;
+        this.update(ChartUpdateType.PERFORM_LAYOUT);
+    }
```
alantreadway, line 973: I wouldn't go this route, as it means even if no features/modules that could be triggered on a drag are active, that we are unnecessarily triggering relayout/redraws? (Layout is comparatively expensive, since there isn't any short-circuit like there is for scene-graph rendering + change-detection).

Maybe what we should do instead is introduce a `ChartUpdateService` which can be passed as part of the `ModuleContext`, and we indirectly allow `Chart.update()` to be called from it?

It might be nice to move some of the responsibility for coordination and management of chart internal state/rendering updates vs. AgChartV2 option application there too, to further declutter `Chart`.

As a quick-win, perhaps just introduce the abstraction, and have it call directly to `Chart.update()` for the moment, and I can come back to the more detailed refactor?

lsjroberts: Agree on all your points. A `ChartUpdateService` would be nicer for sure. It would make it much clearer what is going on when you're in a module.

I did find it surprising that it required the `PERFORM_LAYOUT` level of update when dragging but worked perfectly fine with `SCENE_RENDER` when hovering, for exactly the same effect.

Will work on that initial abstraction.

alantreadway (approved):
LGTM

PR #6355, opened 2023-03-21
AG-3890 Avoid Pie labels overlap

alexanderby:


charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -856,10 +865,135 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+        avoidYCollisions(rightLabels);
+        avoidXCollisions(topLabels);
+        avoidXCollisions(bottomLabels);
+    }
+
     updateCalloutLabelNodes() {
         const { radiusScale, calloutLabel, calloutLine } = this;
         const calloutLength = calloutLine.length;
         const { offset, color } = calloutLabel;
+        this.calculateCalloutLabelCollisionOffsets();
```
alantreadway, line 996: Not sure if it conceptually makes more sense to do this as part of performLayout()? (It may not be practical though)

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -856,10 +865,135 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+        const collisionPadding = 4;
+
+        const boxesCollide = (a: BBox, b: BBox) => {
+            const p = collisionPadding / 2;
+            return (
+                a.x - p < b.x + b.width + p &&
+                a.x + a.width + p > b.x - p &&
+                a.y - p < b.y + b.height + p &&
+                a.y + a.height + p > b.y - p
+            );
```
alantreadway, line 926: Could we add this as a method on `BBox`, since determining if two `BBox` overlap seems like a useful general feature?

alantreadway (commented):
Could you add some visual tests for this new behaviour? We'll also need some docs (perhaps as a follow up)?

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
alexanderby: Oh yes, it is, it is called the 2nd time unnecessarily.

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
alexanderby: Sure

charts-community-modules/ag-charts-community/src/util/sector.ts
```diff
@@ -0,0 +1,96 @@
...
+    innerRadius: number;
+    outerRadius: number;
+}
+
+export interface LineCoordinates {
+    start: { x: number; y: number };
+    end: { x: number; y: number };
+}
+
+export function isPointInSector(x: number, y: number, sector: SectorBoundaries) {
```
alantreadway, line 16: Feels like we have some duplication with: https://github.com/ag-grid/ag-grid/blob/e583a309be26da97b4149f4fe36489586d7c017a/charts-community-modules/ag-charts-community/src/scene/shape/sector.ts#L71-L96

alantreadway: Could you refactor these so we don't have two copies of the same logic?

alantreadway (commented):
Generally LGTM - nice work! Let's try and remove the redundant code I spotted?

PR #6362, opened 2023-03-22
AG-8215 Add pie series toggling of series items not in legend

lsjroberts:
For https://ag-grid.atlassian.net/browse/AG-8215

A pie series that has no `calloutLabelKey` will not appear in the legend items. These should therefore be toggled when an item in a sibling series is clicked that shares the same `itemId`.


alantreadway (approved):
LGTM

Is it possible to easily add some test coverage for this case? (E.g. to `legend.test.ts`?)

PR #6363, opened 2023-03-22
Refactor Chart.performLayout() 

alantreadway:
Ad-hoc refactor to:
- reduce duplication between Chart sub-classes.
- declutter chart.ts by moving legend positioning to `legend.ts`.


lsjroberts (changes requested):
lgtm

lsjroberts (approved):

PR #6368, opened 2023-03-23
AG-6181 Update chart leave event handling to allow tooltip interaction

lsjroberts:
For https://ag-grid.atlassian.net/browse/AG-6181

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/chart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/chart.ts
@@ -283,7 +283,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.interactionManager.addListener('click', (event) => this.onClick(event));
         this.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event));
         this.interactionManager.addListener('hover', (event) => this.onMouseMove(event));
-        this.interactionManager.addListener('leave', () => this.disablePointer());
+        this.interactionManager.addListener('leave', (event) => this.onLeave(event));
         this.interactionManager.addListener('page-left', () => this.destroy());
 
         this.zoomManager.addListener('zoom-change', (_) =>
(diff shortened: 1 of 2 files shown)


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -972,6 +972,14 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.update(ChartUpdateType.SCENE_RENDER);
     }
 
+    protected onLeave(event: InteractionEvent<'leave'>): void {
+        if (this.tooltip.pointerLeftOntoTooltip(event)) {
```
manapeirov, line 976: maybe could just call this function `pointerOnTooltip`

manapeirov (approved):
LGTM

charts-community-modules/ag-charts-community/src/chart/chart.ts
lsjroberts: My thinking was that `pointerOnTooltip` implies it's looking at the `event.target` for a hover handler or similar. But the function looks at the `event.relatedTarget` instead.

charts-community-modules/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -291,4 +289,9 @@ export class Tooltip {
         }
         this.updateClass(visible, this.constrained);
     }
+
+    pointerLeftOntoTooltip(event: InteractionEvent<'leave'>): boolean {
+        const classList = (event.sourceEvent as MouseEvent).relatedTarget?.classList;
+        return classList && (classList.contains('ag-chart-tooltip') || classList.contains('ag-chart-tooltip-content'));
```
lsjroberts, line 295: Use `DEFAULT_TOOLTIP_CLASS` here

alantreadway (approved):
LGTM

PR #6373, opened 2023-03-24
AG-6166 Context menu support

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-6166


charts-community-modules/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -32,6 +32,14 @@ export class TooltipManager {
         this.applyStates();
     }
 
+    public clearAllTooltips() {
```
alantreadway, line 35: Hmmm.... interesting - this kinda breaks the conceptual model I had in mind, which is that each tooltip source has independent state and `TooltipManager` identifies which one wins where there are conflicts.

Could we instead have the `ContextMenu` set its own 'null' tooltip state and then retract it when the context menu closes? We could even introduce some better method of prioritisation in the conflict resolution in `TooltipManager`?

charts-community-modules/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -149,6 +152,9 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
             case 'dblclick':
                 return ['dblclick'];
 
+            case 'contextmenu':
```
alantreadway, line 155: Side-note: Didn't realise this was even an event type :D

charts-community-modules/ag-charts-community/src/chart/interaction/tooltipManager.ts
lsjroberts: Yeah, I wasn't too happy with it. Pushing a `null` state is a smart idea, I'll try that.

lsjroberts: I've changed it so that `updateTooltip()` can take a null state and not delete the state. Then added a `removeTooltip('id')` function that covers that purpose. Thoughts?

alantreadway (approved):
Thanks - LGTM!

PR #6398, opened 2023-03-29
AG-4371 - Crosshair docs

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-4371

Add docs page with examples for axes crosshairs

Add `format` string to crosshair label options
Use `crosshair.label.format` if provided, otherwise use `axis.label.format` if provided, otherwise use default format.

Fix `snap: true` for crosshairs on histogram series


alantreadway (approved):
LGTM

PR #6404, opened 2023-03-30
AG-7489 - Introduce shared DataModel - first set of cases.

alantreadway:
Spike for https://ag-grid.atlassian.net/browse/AG-7489

- Adds `DataModel` class to encapsulate all common data-processing behaviour.
- Converts these series types to use it:
  - `ScatterSeries`.
  - `BarSeries`
  - `AreaSeries`.


charts-community-modules/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        it('should generated the expected results', () => {
+            const data = examples.SIMPLE_LINE_CHART_EXAMPLE.data!;
+            const dataModel = new DataModel<any, any>({
+                dimensionKeys: ['date'],
+                valueKeys: ['petrol', 'diesel'],
+                dimensionKeyTypes: ['range'],
+                valueKeyTypes: ['range', 'range'],
+            });
+
+            expect(dataModel.processData(data)).toMatchSnapshot();
```
lsjroberts, line 18: It might be good to have some tests that don't rely on snapshots, with small inline data sets. So you can read the test and see what the code is supposed to do, and also because it is very easy to update a snapshot and not manually verify the change which could lead to bugs.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -155,33 +150,36 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
...
+        this.sizeScale.domain = marker.domain ? marker.domain : this.processedData.dataDomain.values[2];
     }
 
     getDomain(direction: ChartAxisDirection): any[] {
         if (direction === ChartAxisDirection.X) {
-            return this.xDomain;
+            return this.processedData?.dataDomain.values[0] ?? [];
         } else {
-            return this.yDomain;
+            return this.processedData?.dataDomain.values[1] ?? [];
```
lsjroberts, line 182: It might be worth making these indices as consts? I guess using a dict would be bad for other series.

lsjroberts (approved):
I think I get the idea 😅

charts-community-modules/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
alantreadway: Fair call, I've updated to use the `DataModel` index resolution method that I added so this is a little more readable.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -320,183 +316,81 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
...
+                    property: xKey,
+                    type: 'key',
+                    valueType: isContinuousX ? 'range' : 'category',
+                    validation: (v) => checkDatum(v, isContinuousX) != null,
+                },
+                ...[...seriesItemEnabled.entries()]
+                    .filter(([, enabled]) => enabled)
+                    .map(([yKey]) => ({
+                        property: yKey,
+                        type: 'value' as const,
```
lsjroberts, line 336: Another alternative to this would be to use an enum, e.g.

```ts
enum DataPropertyType = { Key = 'key', Value = 'value' };

{
    property: yKey,
    type: DataPropertyType.Value,
}
```

alantreadway: I've reworked the approach here, there are now some utility functions that remove most of the boilerplate and the need to do this 🎉

alexanderby (commented):
image

PR #6432, opened 2023-04-05
AG-4371 - Add hoverRect

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-4371

Add `hoverRect` to use in crosshairs

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
@@ -37,17 +37,17 @@ export class CartesianChart extends Chart {
 
         const { seriesRoot, seriesAreaPadding } = this;
 
-        const paddedX = seriesRect.x - seriesAreaPadding.left;
-        const paddedY = seriesRect.y - seriesAreaPadding.top;
-        const paddedWidth = seriesAreaPadding.left + seriesRect.width + seriesAreaPadding.right;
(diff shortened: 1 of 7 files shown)


charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -37,17 +37,17 @@ export class CartesianChart extends Chart {
...
 
-        const paddedX = seriesRect.x - seriesAreaPadding.left;
-        const paddedY = seriesRect.y - seriesAreaPadding.top;
-        const paddedWidth = seriesAreaPadding.left + seriesRect.width + seriesAreaPadding.right;
-        const paddedHeight = seriesAreaPadding.top + seriesRect.height + seriesAreaPadding.bottom;
+        const seriesPaddedRect = seriesRect.clone().grow(seriesAreaPadding);
 
-        const seriesPaddedRect = new BBox(paddedX, paddedY, paddedWidth, paddedHeight);
+        const hoverRectPadding = 20;
+        const hoverRect = seriesPaddedRect.grow(hoverRectPadding);
```
alantreadway, line 43: `BBox.grow()` mutates the state of the `BBox` and returns `this` for convenience/chaining, so is this doing the right thing? (I.e. `paddedRect` and `hoverRect` end up being the same thing in the `dispatchLayoutComplete()` call below.

manapeirov: fixed

PR #6435, opened 2023-04-06
AG-8264 Add initial motion and animation manager

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8264

Module example:

```ts
const rect = new _Scene.Rect();

ctx.animationManager.animate('example', {
    from: 0,
    to: Math.PI * 2,
    duration: 2000,
    ease: _Motion.linear,
    repeat: Infinity,
    repeatType: _Motion.RepeatType.Reverse,
    onUpdate: (v: number) => {
        rect.fill = `rgb(${255 - (v / (Math.PI * 2)) * 255}, 0, ${(v / (Math.PI * 2)) * 255})`;
        rect.x = 200 + v * 200;
        rect.y = 400 + Math.cos(v) * 200;
    },
});
```

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
@@ -1,0 +1,101 @@
+import { BaseManager } from './baseManager';
+import * as Motion from '../../motion/motion';
+
+type AnimationId = string;
+type AnimationType = 'animation-frame';
+
+interface AnimationEvent<AnimationType> {
+    type: AnimationType;
+    delta: number;
+}
+
+export class AnimationManager extends BaseManager<AnimationType, AnimationEvent<AnimationType>> {
+    private readonly states: Record<AnimationId, Motion.AnimationControls> = {};
+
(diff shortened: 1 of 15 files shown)


alexanderby (approved):
👍

charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -0,0 +1,101 @@
...
+                },
+                stop: () => {
+                    this.updaters.filter(([uid]) => uid !== id);
+                },
+            };
+        };
+    }
+
+    private requestAnimationFrame() {
+        this.requestId = requestAnimationFrame((time) => {
```
alexanderby, line 87: One potential problem is, when the tab is hidden (`document.hidden`), nothing will happen. So `requestIdleCallback` should be there for a fallback.

lsjroberts:
There are a few aspects to the integration with highlighting that are less than ideal and will need discussion before anything is merged.

lsjroberts:
Closing this draft to recreate with cleaned up code and focused on just highlighting.

PR #6455, opened 2023-04-12
AG-8006 - Warn on invalid axis type configuration and fallback to def…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8006


charts-community-modules/ag-charts-community/src/chart/chartAxisTypes.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const types: Record<string, AgCartesianAxisType> = {
+    number: 'number',
+    time: 'time',
+    log: 'log',
+    category: 'category',
+    groupedCategory: 'groupedCategory',
+};
+
+export const CHART_AXES_TYPES = {
```
alexanderby, line 11: We have `chartAxisTypes.ts` but `CHART_AXES_TYPES` (one singular and another plural form of `axis`).

alexanderby (approved):

charts-community-modules/ag-charts-community/src/chart/chartAxisTypes.ts
manapeirov: which one should it be do you think, couldn't make up my mind

PR #6456, opened 2023-04-12
AG-7600 Allow gradients for all shapes

alexanderby:


charts-community-modules/ag-charts-community/src/scene/gradient/linearGradient.ts
```diff
@@ -15,7 +18,17 @@ export class LinearGradient extends Gradient {
...
+            const d = Math.sqrt(h * h + w * w) / 2;
+            const a = Math.atan2(h, w);
+            const r =
+                angle <= 90
+                    ? radians
+                    : angle <= 180
+                    ? Math.PI - radians
+                    : angle <= 270
+                    ? radians - Math.PI
+                    : 2 * Math.PI - radians;
```
lsjroberts, line 30: This might be cleaner broken up into the different statements? Maybe?

manapeirov (approved):

PR #6466, opened 2023-04-17
fix: auto group column width changed when move column

iamnotblank:


StephenCooper:
Could you please provide a reproduction that shows what your suggested fix here is intended to do so that we have a code example to validate against?

iamnotblank:
Kapture 2023-05-06 at 17 20 36
The above gif video is that i record in official example page to explain this issue. 
Steps of the reproduction:

1. change the grouped column width
2. move the ungroued column to change grid column's order
3. the first column also grouped column's width changed to default 

This change can cause confusion for users, i think.

PR #6469, opened 2023-04-17
Ag 5094/axis label wrapping

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-5094


alantreadway (commented):
Looks reasonable, but let's think about how we properly isolate the axis collision avoidance strategies before merge?

charts-community-modules/ag-charts-community/src/axis.ts
```diff
@@ -149,6 +149,18 @@ export class AxisLabel {
     @Validate(BOOLEAN)
     enabled = true;
 
+    /** If set to `false`, axis labels will not be wrapped on multiple lines.*/
```
alexanderby, line 152: `lines.`**_**`*/`

charts-community-modules/ag-charts-community/src/axis.ts
```diff
@@ -149,6 +149,18 @@ export class AxisLabel {
...
+    @Validate(OPT_BOOLEAN)
+    autoWrap: boolean = false;
```
alexanderby, line 154: A bit unclear, what will it do with `autowrap=true` and unset `maxWidth` or `maxHeight`

charts-community-modules/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -280,6 +280,88 @@ export class Text extends Shape {
...
+        maxWidth: number,
+        maxHeight: number,
+        font: string,
+        fontSize: number,
+        truncate: boolean,
+        cumulativeHeight: number
+    ): { result: string; truncated: boolean; cumulativeHeight: number } {
+        const lines: string[] = [];
+        const guesstimate = Math.max(1, Math.round(maxWidth / fontSize));
+        const ellipsis = `...`;
```
alexanderby, line 317: There is a unicode ellipsis symbol `…` or `'\u2026'`.

charts-community-modules/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -280,6 +280,88 @@ export class Text extends Shape {
             offsetY += lineHeights[i];
         }
     }
+
+    static wrap(
```
alexanderby, line 284: Maybe it's worth adding visual tests for this cool feature.

alexanderby (commented):
Very nice approach, would be great to use `…` and have some tests for text only.

charts-community-modules/ag-charts-community/src/axis.ts
manapeirov: We use some default `maxWidth` which we calculate based on the available space

alexanderby (approved):
Still would be nice to have variations of the wrapped Text in a separate test file, like we have for `rect.test.ts`.

charts-community-modules/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -371,6 +374,13 @@ export class CartesianChart extends Chart {
         }
 
         let primaryTickCount = axis.nice ? primaryTickCounts[direction] : undefined;
+        const paddedBoundsCoefficient = 0.3;
+        axis.maxThickness =
+            axis.thickness > 0
+                ? axis.thickness
+                : direction === ChartAxisDirection.Y
+                ? paddedBounds.width * paddedBoundsCoefficient
+                : paddedBounds.height * paddedBoundsCoefficient;
```
alantreadway, line 383: Nit: Please avoid these nested ternary statements, they are being flagged up by SonarCloud all the time, and more importantly they can be a bit unreadable.

Restructure to something like:
```ts
let maxThickness =0; // default case
if (axis.thickness > 0) {
  maxThickness = axis.thickness;
} else if (direction === ChartAxisDirection.Y) {
  maxThickness = paddedBounds.width * paddedBoundsCoefficient;
} else {
  maxThickness =. paddedBounds.height * paddedBoundsCoefficient;
}
axis.maxThickness = maxThickness;
```

alantreadway (commented):
LGTM so far, just a couple of nits.

charts-community-modules/ag-charts-community/src/axis.ts
```diff
@@ -592,25 +629,70 @@ export class Axis<S extends Scale<D, number, TickInterval<S>>, D = any> {
...
+        this.updateLine();
+
+        const ticksResult = this.generateTicks({
+            primaryTickCount,
+            parallelFlipRotation,
+            regularFlipRotation,
+            labelX,
+            sideFlag,
+        });
+        const { tickData, combinedRotation, textBaseline, textAlign } = ticksResult;
```
alantreadway, line 647: Nit: Maybe remove `ticksResult` if it's only used for this destructuring? (Inline it above)

charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -732,6 +732,12 @@ export interface AgAxisLabelOptions {
     padding?: PixelSize;
     /** The colour to use for the labels */
     color?: CssColor;
+    /** If set to `true`, axis labels will be wrapped on multiple lines.*/
+    autoWrap?: boolean;
```
alantreadway, line 736: I'm not sure `autoWrap` is the right concept - should we allow selection of the wrapping strategy as well? (E.g. as per Alex's additions for titles)

We can follow-up on this aspect.

alantreadway (approved):
LGTM, no need to re-review for the nit.

PR #6476, opened 2023-04-17
AG-3227 - Automatically hide the tooltip arrow if offset is configured

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-3227


alantreadway (approved):
LGTM

PR #6486, opened 2023-04-18
AG-8372 - Allow per column sparkline tooltip customisation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8327

Allow all tooltip properties to be configurable on a per column basis.


alantreadway (approved):
LGTM

PR #6488, opened 2023-04-18
AG-7600 Legend modules support

alexanderby:


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -733,32 +764,32 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
             .filter((s) => s.showInLegend)
-            .forEach((series) => {
-                legendData.push(...series.getLegendData());
+            .forEach((series, i) => {
+                const data = series.getLegendData();
+                if (data && data.length > 0 && (i === 0 || data[0].legendType === legendType)) {
+                    legendData.push(...data);
+                    legendType = data[0].legendType;
+                }
             });
```
alantreadway, line 789: I wonder if it would be better to just call all `series.getLegendData()` and then inspect the results, rather than trying to keep track of the first legendType as we go?

It seems like if there are mismatching `legendType`s that we should throw an error perhaps? And if that check passes, we can simply concatenate all of the `data` items together?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -733,32 +764,32 @@ export abstract class Chart extends Observable implements AgChartInstance {
         return new Map(labels.map((l, i) => [visibleSeries[i], l]));
     }
 
+    private attachLegend(legendType: string) {
+        if (this.legendType === legendType || !(legendType in this.legendFactories)) {
+            return;
+        }
+        const legendFactory = this.legendFactories[legendType];
+        this.legend?.destroy();
```
alantreadway, line 772: Nit: Probably best to `this.legend = undefined` immediately after the destroy call, to avoid `destroy()` being called multiple times if the following steps fail?

charts-community-modules/ag-charts-community/src/chart/dataService.ts
```diff
@@ -0,0 +1,11 @@
+import { Series } from './series/series';
+
+type SeriesGetter = () => Series[];
```
alantreadway, line 3: `Series` is quite a broad contract (and an implementation type) - could we narrow the dependency here by introducing a minimal `interface` type that `Series` conforms to?

charts-community-modules/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -601,7 +602,7 @@ export class HistogramSeries extends CartesianSeries<SeriesNodeDataContext<Histo
                     strokeOpacity: strokeOpacity,
                 },
             },
-        ];
+        ] as CategoryLegendDatum[];
```
alantreadway, line 605: Avoid `as`, since it's essentially bypasses TS type-checking - could you instead either:
1) Assign the result to a const of type `CategoryLegendDatum[]` and then return it, as this doesn't bypass type-checks?
2) Or change the method signature return type to `CategoryLegendDatum[]`.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -577,7 +578,7 @@ export class LineSeries extends CartesianSeries<LineContext> {
                     strokeOpacity: marker.strokeOpacity ?? strokeOpacity ?? 1,
                 },
             },
-        ];
+        ] as CategoryLegendDatum[];
```
alantreadway, line 581: Ditto

charts-community-modules/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -495,7 +534,7 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
                     strokeOpacity: strokeOpacity ?? 1,
                 },
             },
-        ];
+        ] as CategoryLegendDatum[];
```
alantreadway, line 537: Ditto

charts-community-modules/ag-charts-community/src/chart/legendDatum.ts
```diff
@@ -1,10 +1,29 @@
...
+    data: any;
+    item: {
+        label: {
+            formatter?: (params: any) => string;
+        };
+    };
+    listeners: AgChartLegendListeners;
+}
 
 export interface LegendDatum {
```
alantreadway, line 17: Maybe this should be renamed to `ChartLegendDatum` for consistency with `ChartLegend`?

charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -311,6 +312,32 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
+    }
+
+    removeLegendModule(module: Module) {
+        this.legendModules[module.optionsKey]?.instance?.destroy();
+        delete this.legendModules[module.optionsKey];
+    }
+
+    isLegendModuleEnabled(module: LegendModule) {
+        return this.legendModules[module.optionsKey] != null;
+    }
```
alantreadway, line 339: Let's roll with this for now, but I'm thinking that these factory/DI-like behaviours will need to go elsewhere eventually 🤞

alantreadway (commented):
Nice, good progress forward - got some points to discuss though.

Regarding `Series` - we need to avoid exporting it and/or adding more dependencies, as it has a VERY broad contract and we need to decouple it from `Chart` in the future - minimising potential dependencies is probably a good idea.

alantreadway (approved):
LGTM

PR #6491, opened 2023-04-19
RTI-1426 Fix removing default Tooltip class name

alexanderby:
The value assigned through the theme is treated as the "previous", when another one is assigned through options.


alantreadway (approved):
LGTM

PR #6494, opened 2023-04-20
Ag 5568/axis ticks

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-5568

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/mapping/prepare.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/mapping/prepare.ts
@@ -165,6 +165,8 @@ export function prepareOptions<T extends AgChartOptions>(
         defaultOverrides = DEFAULT_CARTESIAN_CHART_OVERRIDES;
     }
 
+    removeDisabledOptions<T>(options);
+
     const { context, mergedOptions, axesThemes, seriesThemes } = prepareMainOptions<T>(defaultOverrides as T, options);
 
     // Special cases where we have arrays of elements which need their own defaults.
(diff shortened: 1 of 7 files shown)


charts-community-modules/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -371,6 +373,23 @@ function prepareAxis<T extends AxesOptionsTypes>(
...
+function removeDisabledOptions<T extends AgChartOptions>(options: T) {
+    // Remove configurations from all option objects with a `false` value for the `enabled` property.
+    jsonWalk(
+        options,
+        (_, visitingUserOpts) => {
+            if (!('enabled' in visitingUserOpts)) return;
+            if (visitingUserOpts.enabled === false) {
+                Object.entries(visitingUserOpts).forEach(([key]) => {
+                    if (key === 'enabled') return;
+                    delete visitingUserOpts[key];
```
alantreadway, line 385: Is this working? I would have thought that you need to mutate the `visitingMergedOpts` (unused 3rd parameter) as per `prepareEnabledOptions()`, I'm guessing `visitingUserOpts` is intended to be read-only since it's already been used to generate `visitingMergedOpts`?

alantreadway (commented):
One comment - could you add some (unit) test coverage for this too?

charts-community-modules/ag-charts-community/src/chart/mapping/prepare.ts
manapeirov: I was thinking we need the defaults to still be applied so this step happens before merging the options?

alantreadway (approved):
LGTM

PR #6517, opened 2023-04-26
AG-3479 Change zoom default behaviour and fix limits to area and min zoom

lsjroberts:
For https://ag-grid.atlassian.net/browse/AG-3479

### DIFF
--- a/charts-enterprise-modules/ag-charts-enterprise/src/zoom/zoom.ts
+++ b/charts-enterprise-modules/ag-charts-enterprise/src/zoom/zoom.ts
@@ -43,6 +43,9 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
     @Validate(NUMBER(0, 1))
     public scrollingStep = 0.1;
 
+    @Validate(STRING_UNION('pointer', 'start', 'end'))
+    public scrollingPivot: 'pointer' | 'start' | 'end' = 'end';
+
     @Validate(NUMBER(0, 1))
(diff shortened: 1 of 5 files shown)


charts-enterprise-modules/ag-charts-enterprise/src/zoom/zoom.ts
```diff
@@ -124,31 +130,38 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
     }
 
     private onDrag(event: _ModuleSupport.InteractionEvent<'drag'>) {
-        const isWithinSeriesRect = this.seriesRect?.containsPoint(event.offsetX, event.offsetY);
-        const isPrimaryMouseButton = (event.sourceEvent as MouseEvent).button === 0;
+        const sourceEvent = event.sourceEvent as DragEvent;
```
alantreadway, line 133: Side-note: Maybe we need better typings on `InteractionEvent` to avoid the need to case like this. Or maybe this is pointing to the need to extract some event metadata to be a first-class citizen of `InteractionEvent`?

alantreadway (approved):
LGTM

charts-enterprise-modules/ag-charts-enterprise/src/zoom/zoom.ts
lsjroberts: Yeah, for sure. I've had a look at this a couple of times but had some issues with out of date typescript inbuilt types. Maybe better now we've updated to a newer version.

PR #6522, opened 2023-04-26
Add animation manager

lsjroberts:
This is a port of https://github.com/ag-grid/ag-grid/pull/6435/ with experiments taken out and some tweaks.


alexanderby (approved):

charts-community-modules/ag-charts-community/src/motion/animate.ts
```diff
@@ -0,0 +1,209 @@
...
+    let requestId: number | undefined;
+    let lastTime: number | undefined;
+
+    function frame(time: number) {
+        if (lastTime === undefined) lastTime = time;
+        const delta = time - lastTime;
+        lastTime = time;
+        update(delta);
+
+        requestId = requestAnimationFrame(frame);
```
alantreadway, line 195: I have a small concern that maybe we could have an unbounded number of callbacks setup - would it be possible to have a single callback that coordinates all of the pending animations? This would allow us to perform more intelligent scheduling and prioritisation if they start to take significant time?

charts-community-modules/ag-charts-community/src/motion/easing.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// https://easings.net/
+export const easeIn = createEase<any>((x) => 1 - Math.cos((x * Math.PI) / 2));
+export const easeOut = createEase<any>((x) => Math.sin((x * Math.PI) / 2));
+export const easeInOut = createEase<any>((x) => -(Math.cos(x * Math.PI) - 1) / 2);
+export const easeOutElastic = createEase<any>((x) => {
+    const c4 = (2 * Math.PI) / 3;
+    if (x === 0 || x === 1) return x;
+    return Math.pow(2, -10 * x) * Math.sin((x * 10 - 0.75) * c4) + 1;
```
alantreadway, line 28: Any chance of introducing some constants to make these magic numbers a bit more readable?

charts-community-modules/ag-charts-community/src/motion/animate.ts
lsjroberts: That's exactly what this does. With a single`play()` call at in the chart it only creates this one callback.

charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -595,6 +600,8 @@ export interface AgBaseChartOptions {
     height?: PixelSize;
     /** By default, the chart will resize automatically to fill the container element. Set this to `false` to disable this behaviour. If either the `width` or `height` are set, auto-sizing will be disabled unless this is explicitly set to `true`.<br/><strong>Important:</strong> if this config is set to `true`, make sure to give the chart's `container` element an explicit size, otherwise you will run into a chicken and egg situation where the container expects to size itself according to the content and the chart expects to size itself according to the container. */
     autoSize?: boolean;
+    /** Configuration for chart animations. */
+    animation?: AgChartAnimationOptions;
```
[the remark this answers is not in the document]
lsjroberts: Do we want to disable animations entirely on community? I thought we were going to have the default ones and customisation is an enterprise feature?

charts-community-modules/ag-charts-community/src/chart/chartAnimation.ts
```diff
@@ -0,0 +1,23 @@
...
+    @Validate(BOOLEAN)
+    @ActionOnSet<ChartAnimation>({
+        newValue(enabled: boolean) {
+            if (enabled) {
+                this.animationManager?.play();
+            } else {
+                this.animationManager?.stop();
+            }
+        },
+    })
```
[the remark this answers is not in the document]
lsjroberts: Calling `play()` triggers the chart update method through the listener - https://github.com/ag-grid/ag-grid/pull/6522/files/5670b910694729bebfc461f5b1c47f262a626d58#diff-cd3867b053c5efd19d38f3dda54b506f65f61fd60de0f8fcc32cd5b68b52d646R296 - That would be an infinite loop?

It needs to call once when the chart is first rendered.

lsjroberts:
Changes for review:

- Added an animation enterprise module
- Call `play()` in chart constructor, but with `skipAnimations = true`. This forces the animation duration to always be `0`.
- When `animation.enabled: true` is set it sets `skipAnimations = false`.

I thought about doing this on `layout-complete` but that may cause issues down the line when doing initial load animations. I think this approach should work fine.

alantreadway:
Review comments: https://docs.google.com/document/d/1fhGpMJuCSwrRvtvSnt_o4RXvEW1JxOBjNx52wlW6FAs/edit

grid-community-modules/all-modules/generate-code-reference-files.js
```diff
@@ -233,8 +233,7 @@ function applyInheritance(extensions, interfaces, isDocStyle) {
                 // Omit: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
                 // Special logic to handle the removing of properties via the Omit utility when a type is defined via extension.
                 // e.g. export interface AgNumberAxisThemeOptions extends Omit<AgNumberAxisOptions, 'type'> { }
-                extended = a.params[0];
-                const fullInterface = interfaces[extended];
+                extended = a.params[0].replace(/<.*>/, '');
```
lsjroberts, line 236: This is to handle the case omitting keys with generics:

```ts
interface AnimationManyOptions<T> extends Omit<AnimationOptions<T>, 'from' | 'to' | 'onUpdate'>
```

alantreadway (approved):
LGTM, one tweak suggested, but otherwise go for merge 🏁

charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -0,0 +1,193 @@
...
+            const delta = time - this.lastTime;
+            this.lastTime = time;
+
+            this.updaters.forEach(([_, update]) => {
+                update(delta);
+            });
+
+            this.listeners.dispatch('animation-frame', { type: 'animation-frame', delta });
+
+            this.requestAnimationFrame();
```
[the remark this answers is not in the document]
lsjroberts: ohhh, I see, yep ok

PR #6526, opened 2023-04-27
AG-8207 - Make enterprise module definitions declarative

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8207

Changes:
- Adds a `factory/` folder to start to break factory responsibilities into, from `AgChartV2.ts` and other places.
- Removes module context callbacks to register and deregister new implementations; replaces with declarative definitions.
- Splits the lifecycle point for registration of new implementations from the lifecycle of module instances attached to a chart.
- Cleans up some elements of module definitions:
  - Removes the need for a wrapper type around module instances.
  - Removes the redundant `ModuleInstance.update()` method.
- Introduces the concept of theme templates to allow declarative theme extensions by modules.


lsjroberts (approved):
LGTM

charts-community-modules/ag-charts-community/src/chart/background/backgroundModule.ts
```diff
@@ -6,11 +6,7 @@ export const CHART_BACKGROUND_MODULE: Module = {
     optionsKey: 'background',
     packageType: 'community',
     chartTypes: ['cartesian', 'polar', 'hierarchy'],
-    initialiseModule(ctx) {
-        return {
-            instance: new Background(ctx),
-        };
-    },
+    instanceConstructor: Background,
```
alexanderby, line 9: Maybe it should be a factory just in case? Like `(ctx) => new Background(ctx)`.

charts-community-modules/ag-charts-community/src/util/module.ts
```diff
@@ -44,87 +41,55 @@ export interface AxisContext {
...
+
+    factoryConstructors?: JsonApplyParams['constructors'];
 }
 
 export interface RootModule<M extends ModuleInstance = ModuleInstance> extends BaseModule {
     type: 'root';
-    initialiseModule(ctx: ModuleContext): ModuleInstanceMeta<M>;
-    destroyModule?(ctx: ModuleContext): void;
+
+    instanceConstructor: new (ctx: ModuleContext) => M;
```
alexanderby, line 62: Maybe this one should be `rootFactory` or something, like `(ctx) => new Instance(ctx)` or `(ctx) => ({ destroy() })` to not be tied on classes?

charts-community-modules/ag-charts-community/src/util/module.ts
```diff
@@ -44,87 +41,55 @@ export interface AxisContext {
...
-export interface ModuleInstanceMeta<M extends ModuleInstance = ModuleInstance> {
-    instance: M;
-}
-
 interface BaseModule {
     optionsKey: string;
     packageType: 'community' | 'enterprise';
     chartTypes: ('cartesian' | 'polar' | 'hierarchy')[];
+
+    factoryConstructors?: JsonApplyParams['constructors'];
```
alexanderby, line 56: Maybe this one should be `optionFactories` or something to not be tied on classes?

charts-community-modules/ag-charts-community/src/chart/background/backgroundModule.ts
alantreadway: I didn't see a use-case for it right now, so I'd prefer to keep it lean for the moment.

charts-community-modules/ag-charts-community/src/util/module.ts
alantreadway: I think we can revisit the choice here if we needed to do something different easily, but for right now all of use-cases fit this pattern.

charts-community-modules/ag-charts-community/src/util/module.ts
alantreadway: I like the name, will change. Again, I think if we have non-class-based-cases later it is easy to retrofit - for now I would prefer the tighter contract.

charts-community-modules/ag-charts-community/src/chart/factory/seriesTypes.ts
```diff
@@ -0,0 +1,56 @@
...
+    const seriesConstructor = SERIES_FACTORIES[chartType] ?? BUILT_IN_SERIES_FACTORIES[chartType];
+    if (seriesConstructor) {
+        return new seriesConstructor();
+    }
+
+    throw new Error(`AG Charts - unknown series type: ${chartType}`);
+}
+
+export function getSeriesDefaults(chartType: string): {} {
+    return SERIES_DEFAULTS[chartType];
```
manapeirov, line 51: seriesType?

PR #6530, opened 2023-04-28
AG-8325 - clean up examples with legend: { enabled: false } and only …

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8325


alantreadway (approved):
LGTM

PR #6535, opened 2023-04-28
AG-8458 flush sync updates

StephenCooper:
Users are still running into excessive calls of `flushSync` so switch to only calling in the require scrolling use cases instead of attempting to switch off for problematic code paths. This should hopefully be more robust and not clash with user based api calls.


grid-community-modules/react/src/reactUi/agGridReactUi.tsx
```diff
@@ -43,8 +42,6 @@ export class AgGridReactUi<TData = any> extends Component<AgReactUiProps<TData>,
         if (this.state.context) {
             this.renderedAfterMount = true;
         }
-        FlushSyncToggle.off();
-        setTimeout(() => FlushSyncToggle.on(), 0);
```
StephenCooper, line 47: No longer needed to prevent flushSync during init of the Grid as we only enable flushSync in specific use cases.

grid-community-modules/core/src/ts/rendering/row/rowCtrl.ts
```diff
@@ -732,15 +732,15 @@ export class RowCtrl extends BeanStub {
...
+        this.updateColumnLists(true, true);
 
         if (this.beans.columnModel.wasAutoRowHeightEverActive()) {
             this.rowNode.checkAutoHeights();
         }
     }
 
     private onVirtualColumnsChanged(): void {
-        this.updateColumnLists();
+        this.updateColumnLists(false, true);
```
StephenCooper, line 743: Need this to ensure cells for the new columns are rendered before focus is applied.

PR #6542, opened 2023-05-02
AG-8110 Add chart event manager and decouple legend and series click event

lsjroberts:

The goal of this PR is to further the decoupling of legend from series, in an ideal world the legend would not know anything about the series. Previously it directly called the `series.toggleSeriesItem()` function but needed further knowledge of how each chart type handles internal series to coordinate the result of clicking related legend items.

This might not be the right approach, I'm on the fence.

If we do go with this I'll need to do the same work for the double click event in a follow up PR.


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -600,6 +605,9 @@ export abstract class Chart extends Observable implements AgChartInstance {
             series.data = this.data;
         }
         this.addSeriesListeners(series);
+
+        series.chartEventManager = this.chartEventManager;
+        series.addChartEventListeners();
```
lsjroberts, line 610: This `addChartEventListeners()` thing is kinda rubbish, but the listener can't be attached in the series constructor as the `chartEventManager` won't exist yet.

I'm very open to suggestions.

charts-community-modules/ag-charts-community/src/chart/interaction/chartEventManager.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseManager } from './baseManager';
+
+type ChartEventType = 'legend-item-click';
```
lsjroberts, line 3: For context, another likely event is `node-click` which will be used by the context menu module. And we could use it for other "higher order" events.

alantreadway (approved):
LGTM

PR #6544, opened 2023-05-02
AG-7489 - Apply DataModel to HistogramSeries.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7489

Heavily adjusts DataModel to handle HistogramSeries cases, and converts HistogramSeries to use DataModel for all heavy lifting (except automatic bin calculation):
- DataModel aggregate function no longer defaults to `sum()`.
- Aggregations can be calculated at datum or group level.
- Aggregations can perform more complex intermediate calculations (see `average()` use-case, which tracks negative sum, positive sum and count).


charts-community-modules/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -573,6 +560,34 @@ describe('DataModel', () => {
...
+            const dataModel = new DataModel<any, any, true>({
+                props: [
+                    { property: 'engine-size', type: 'key', valueType: 'category' },
+                    groupCount(),
+                    SORT_DOMAIN_GROUPS,
+                ],
+                groupByFn: () => {
+                    return (item) => {
+                        if (item.keys[0] < 100) {
+                            return ['<100'];
```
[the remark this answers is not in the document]
alantreadway: Ah, so the `groupByFn` returns an array of 'key' elements that combined are a full key - therefore there is only one bin returned. Multiple elements would denote a more complex key structure.

Maybe I can make this a little more self-documenting by adding a `DataKey` type of `DataKeyElement[]`?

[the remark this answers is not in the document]
lsjroberts: Ah of course. Yeh perhaps with that type, if it doesn't pollute things too much, but tbh it's good as is.

charts-community-modules/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -336,7 +338,7 @@ describe('DataModel', () => {
                         { property: 'kp', type: 'key', valueType: 'category' },
                         { property: 'vp1', type: 'value', valueType: 'range' },
                         { property: 'vp2', type: 'value', valueType: 'range' },
```
lsjroberts, line 340: Might be worth in the future adding `range('vp1')` and `category('kp')` functions similar to the `sum()` one?

charts-community-modules/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -417,57 +459,67 @@ export class DataModel<D extends object, K extends keyof D = keyof D, Grouped ex
...
             if (processedData.type === 'ungrouped') {
                 values = [values];
             }
 
             let resultIdx = 0;
-            for (const indices of resultSumValueIndices) {
-                const groupDomain = extendDomain([]);
+            for (const indices of resultAggValueIndices) {
+                let groupAggValues = resultGroupAggFns[resultIdx]?.() ?? extendDomain([]);
                 for (const distinctValues of values) {
```
lsjroberts, line 493: Three nested for loops on the data feels like it might be something to check for optimisation? Not urgent of course.

lsjroberts (approved):

charts-community-modules/ag-charts-community/src/chart/data/dataModel.test.ts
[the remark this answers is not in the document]
alantreadway: Had a go at improving the typings, but didn't manage to get anywhere satisfactory for right now - lets take it out of scope for now.

charts-community-modules/ag-charts-community/src/chart/data/dataModel.ts
alantreadway: Yup, this is actually the optimised version - the loop iterations are:
- Loop over each group.
- Loop over each aggregate function we need to calculate for that group.
- Loop over each value to include in the aggregate.

The intent is to only visit each group once, and perform each aggregation in turn. It should be minimal from a data-access perspective, and not jumping around the data-structure unnecessarily.

charts-community-modules/ag-charts-community/src/chart/data/dataModel.test.ts
alantreadway: Done - this was mostly a test-case issue as the series are already doing something similar in a series-specific way.

PR #6561, opened 2023-05-04
AG-8482 - Remove BarSeries.flipXY.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8482

Removes the pointless `flipXY` options from `BarSeries`, which was a proxy for `type` being set to `column` or `bar`.

Adds `ColumnSeries` which trivially overrides some methods to vary rendering behaviour.


lsjroberts (approved):
Much nicer, lgtm

PR #6572, opened 2023-05-09
AG-8505 Improve text wrapping algorithm

alexanderby:


charts-community-modules/ag-charts-community/src/scene/shape/text.test.ts
```diff
@@ -191,7 +191,10 @@ describe('Text', () => {
 
                     textNode.x = currX;
                     textNode.y = currY;
-                    textNode.text = Text.wrap(textNode.text, maxWidth, truncate ? maxHeight : Infinity, textNode);
+                    textNode.text = Text.wrap(textNode.text, maxWidth, truncate ? maxHeight : Infinity, textNode, {
+                        breakWord: true,
+                        hyphens: true,
+                    });
```
alantreadway, line 197: Should we test the various valid (configurable) range of break strategies?

charts-community-modules/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -312,78 +326,248 @@ export class Text extends Shape {
...
-                for (let len = lastLine.length; len >= 0; len--) {
-                    const truncatedLine = `${lastLine.substring(0, len)}${ellipsis}`;
-                    const lastLineWidth = HdpiCanvas.getTextSize(truncatedLine, font).width;
-                    if (lastLineWidth <= maxWidth) {
-                        lines.push(truncatedLine);
-                        return;
+            if (wordWidth <= maxWidth) {
+                // If the word is not too long, process once again but with the line break
+                newLine = true;
+                i--;
```
alantreadway, line 475: How confident are you that we won't infinitely loop, given `i` is no longer a loop variant? Should we add another safety measure to prevent this scenario (perhaps a loop variant which prevents more than `words.length * 2` iterations?).

charts-community-modules/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -312,78 +326,248 @@ export class Text extends Shape {
...
             }
         }
 
-        processText(text.trim());
-        return { result: lines.join('\n'), truncated, cumulativeHeight };
+        return lines;
+    }
+}
+
+const fontCaches: Map<string, Map<string, number>> = new Map();
```
alantreadway, line 555: Nit: This is a little more persistent than I was expecting (for the layout loop, we just cache the previous iteration input + result, and invalidate it if inputs don't match), how big is the overhead in more extreme cases? If it's <10 KBs it is fine.

charts-community-modules/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -312,78 +326,248 @@ export class Text extends Shape {
...
 
-            if (cumulativeHeight > maxHeight) {
-                truncated = true;
-                const lastLine = lines.pop();
-                if (!lastLine) {
-                    return;
-                }
+    private static punctuationMarks = ['.', ',', '-', ':', ';', '!', '?', `'`, '"', '(', ')'];
+
+    private static breakWord(
```
alantreadway, line 375: Thanks for breaking these algorithms up, the readability is much better now!

alantreadway (approved):
LGTM - thanks @alexanderby

PR #6578, opened 2023-05-10
AG-8110 Add bar series legend item name

lsjroberts:

Some of my wording on the docs/comments is a bit rubbish. Open to suggestions.

So any series defined in the chart config with a matching `legendItemName` will toggle together when one is clicked / double clicked. To hide the duplicates you need to use `showInLegend: false`. This behaviour is the same as in the pie series with its `legendItemKey`. We use `Name` here instead of `Key` as it is a static value due to how bar/column series data are defined.


charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
```diff
@@ -1296,12 +1296,14 @@ export interface AgBarSeriesOptions<DatumType = any> extends AgBaseSeriesOptions
...
     /** The key to use to retrieve x-values from the data. */
     xKey?: string;
-    /** The keys to use to retrieve y-values from the data. */
+    /** The key to use to retrieve y-values from the data. */
     yKey?: string;
     /** A human-readable description of the x-values. If supplied, this will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
     xName?: string;
     /** Human-readable description of the y-values. If supplied, a corresponding `yName` will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
     yName?: string;
+    /** Human-readable description of the y-values. If supplied, matching items with the same value will be toggled together. Falls back to `yName` if not provided, but does not merge. */
```
alantreadway, line 1305: Does this also de-duplicate legend items?

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -886,7 +889,7 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
                     seriesId: id,
                     enabled: seriesItemEnabled.get(yKey) || false,
                     label: {
-                        text: yNames[yKey] || yKey,
+                        text: legendItemNames[yKey] || yNames[yKey] || yKey,
```
alantreadway, line 892: Nit: Would it be more correct to use `??` rather than `||` here?

charts-community-modules/ag-charts-community/src/chart/agChartOptions.ts
lsjroberts: No, it's behaviour matches the pie series `legendItemKey` which is simply to trigger click & toggle events together. To hide duplicates a user needs to provide the `showInLegend: false` prop to the other series.

PR #6582, opened 2023-05-11
AG-8462 Pie initial load animation

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8462

This is the first of the series initial load animations. I've added a basic state machine to handle transitioning between different animation states. Happy to hear thoughts on this approach.

The pie series state machine currently starts in the `empty` state, then on the `load` event transitions to `ready`. The transition event is where the animation occurs. Once in the `ready` state it can no longer transition anywhere (currently), so the animation can't happen again even if the `load` event is called.

_edit: changed to `update` instead of `load`_

In the future this will also have a `highlight`, `add`, `remove` and `update` events with related states.

I may also look at adding event blocking and queue so that if the user adds a node in the middle of the initial loading animation it will queue that to happen after the load has finished.

https://github.com/ag-grid/ag-grid/assets/3817697/ee9ade09-df6f-42af-a4b2-f003a629211c

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/chart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/chart.ts
@@ -253,9 +253,8 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.layoutService = new LayoutService();
         this.updateService = new UpdateService((type = ChartUpdateType.FULL) => this.update(type));
 
-        // Delay playing animations by a 10ms to avoid the initial ~150ms jerk
-        setTimeout(() => this.animationManager.play(), 10);
         this.animationManager.skipAnimations = true;
+        this.animationManager.play();
 
         SizeMonitor.observe(this.element, (size) => {
             const { width, height } = size;
--- a/charts-community-modules/ag-charts-community/src/chart/chart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/chart.ts
@@ -253,8 +253,9 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.layoutService = new LayoutService();
         this.updateService = new UpdateService((type = ChartUpdateType.FULL) => this.update(type));
 
+        // Delay playing animations by a 10ms to avoid the initial ~150ms jerk
+        setTimeout(() => this.animationManager.play(), 10);
         this.animationManager.skipAnimations = true;
-        this.animationManager.play();
 
         SizeMonitor.observe(this.element, (size) => {
             const { width, height } = size;
--- a/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
@@ -74,7 +74,7 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
             ...opts,
             autoplay: this.isPlaying ? opts.autoplay : false,
             driver: this.createDriver(id),
-            duration: this.skipAnimations ? 0 : opts.duration,
+            duration: this.skipAnimations ? 1 : opts.duration,
         };
         const controller = baseAnimate(optsExtra);
 
--- a/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
@@ -83,7 +83,6 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
             ...opts,
             autoplay: this.isPlaying ? opts.autoplay : false,
             driver: this.createDriver(id),
-            duration: this.skipAnimations ? 1 : opts.duration,
         };
         const controller = baseAnimate(optsExtra);
 
@@ -93,6 +92,15 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
         }
 
         this.controllers[id] = controller;
+
+        if (this.skipAnimations) {
+            // Initialise the animation with the final values immediately and then stop the animation
+            opts.onUpdate?.(opts.to);
+            controller.stop();
+        } else {
+            // Initialise the animation immediately without requesting a frame to prevent flashes
+            opts.onUpdate?.(opts.from);
+        }
 
         return controller;
     }
--- a/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
@@ -34,8 +34,17 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
     private isPlaying = false;
     private requestId?: number;
     private lastTime?: number;
+    private readyToPlay = false;
 
     public skipAnimations = false;
+
+    constructor() {
+        super();
+
+        window.addEventListener('DOMContentLoaded', () => {
+            this.readyToPlay = true;
+        });
+    }
 
     public play() {
         if (this.isPlaying) return;
@@ -170,6 +179,11 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
 
     private startAnimationCycle() {
         const frame = (time: number) => {
+            if (!this.readyToPlay) {
+                this.requestId = requestAnimationFrame(frame);
+                return;
+            }
+
             if (this.lastTime === undefined) this.lastTime = time;
             const delta = time - this.lastTime;
             this.lastTime = time;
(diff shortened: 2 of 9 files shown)


charts-community-modules/ag-charts-community/src/chart/chart.ts
```diff
@@ -253,8 +253,9 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.layoutService = new LayoutService();
         this.updateService = new UpdateService((type = ChartUpdateType.FULL) => this.update(type));
 
+        // Delay playing animations by a 10ms to avoid the initial ~150ms jerk
+        setTimeout(() => this.animationManager.play(), 10);
```
lsjroberts, line 257: this breaks all the snapshots so I'll find a better solution

lsjroberts: If I can't, I'll take it out before merging and work on it separately as its not related to this ticket.

alexanderby (approved):
The code looks good to me. Maybe it needs some `ease-in-out` to roll smoothly, and be a bit quicker. Would be great to fade-in the labels after the spinning was complete.

lsjroberts:
> The code looks good to me. Maybe it needs some `ease-in-out` to roll smoothly, and be a bit quicker. Would be great to fade-in the labels after the spinning was complete.

I would agree, but https://ag-grid.atlassian.net/browse/AG-8460 says linear. Will see what feedback is from QA. Labels will be a future ticket I believe.

charts-community-modules/ag-charts-community/src/chart/chart.ts
alantreadway: LMK if there isn't a simple solution and we can try. and hash it out. If we can tie this into the `update()` cycle it might be best, rather than the object-construction part of the chart lifecycle.

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -1391,4 +1422,47 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+
+        this.groupSelection.selectByTag<Sector>(PieNodeTag.Sector).forEach((node) => {
+            const { datum } = node;
+
+            if (!this.animationManager || this.animationManager.skipAnimations) {
+                node.startAngle = datum.startAngle;
+                node.endAngle = datum.endAngle;
+                return;
+            }
+
```
alantreadway, line 1437: I'm thinking that we should avoid having these short-circuits as it duplicates code in the animation definition below - have the animation manager execute the final step immediately if `skipAnimations` is true perhaps?

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -1391,4 +1422,47 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
...
+            node.startAngle = rotation;
+            node.endAngle = rotation;
```
alantreadway, line 1439: Likewise, maybe it should be the animation managers responsibility to perform the step `0` of the animation immediately (unless `skipAnimation` is true perhaps)?

charts-community-modules/ag-charts-community/src/motion/states.ts
```diff
@@ -0,0 +1,56 @@
+import { Logger } from '../util/logger';
+
+interface StateDefinition<State extends string> {
+    actions?: {
+        onEnter?: () => void;
+        onExit?: () => void;
+    };
+    on: {
+        [key: string]: {
```
alantreadway, line 9: Shouldn't this `key` be bounded to an event type?

grid-packages/ag-grid-docs/documentation/doc-pages/charts-context-menu/examples/context-menu-actions/main.ts
```diff
@@ -10,7 +10,6 @@ const options: AgCartesianChartOptions = {
     text: "Sweaters made",
   },
   contextMenu: {
-    enabled: true,
```
alantreadway, line 13: Nit: For docs tweaks like this, it's ok to push to `latest` as long as the commit is linked back to the related JIRA. (I'd prefer that to tacking onto unrelated PRs if possible please)

alantreadway (commented):
Nice, a few feedback points to consider.

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
lsjroberts: In theory yes. That works in the browser. Unfortunately the snapshot tests disagree. They don't render. Perhaps something to do with `requestAnimationFrame` in the test environment. I'll have a look into it outside this PR.

charts-community-modules/ag-charts-community/src/motion/states.ts
lsjroberts: It should but typescript doesn't allow `[key: Event]`, it has to be a string or number.

charts-community-modules/ag-charts-community/src/chart/chart.ts
lsjroberts: I realised it's down to the content loading as it is slightly different behaviour in Chrome and Firefox. So popped in a `DOMContentLoaded` listener on the animation manager to set a flag.

charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -83,6 +93,15 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
...
         this.controllers[id] = controller;
 
+        if (this.skipAnimations) {
+            // Initialise the animation with the final values immediately and then stop the animation
+            opts.onUpdate?.(opts.to);
+            controller.stop();
+        } else {
+            // Initialise the animation immediately without requesting a frame to prevent flashes
+            opts.onUpdate?.(opts.from);
+        }
```
lsjroberts, line 103: This fixes the issue of skipping animations for tests and when disabled, and the else branch covers the initial flash. So hopefully we won't need any other checks and implementation specific initialisation.

charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -32,9 +34,18 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
...
 
     public skipAnimations = false;
 
+    constructor() {
+        super();
+
+        window.addEventListener('DOMContentLoaded', () => {
+            this.readyToPlay = true;
+        });
+    }
```
lsjroberts, line 47: This combined with the check in `startAnimationCycle()` prevents the jerky animation at the beginning caused by the large frame delta as the content loads.

charts-community-modules/ag-charts-community/src/chart/series/polar/pieSeries.ts
alantreadway: We should work out how to address these dependency issues, as we'll just end up adding redundant copies of code to handle these (non-existent) cases.

I think we should probably have `AnimationManager` as a constructor dependency, and have the factory code manage instantiation and/or wiring:
https://github.com/ag-grid/ag-grid/blob/255fb2ba29c649f76dc24479e46b7bcb4e86b73a/charts-community-modules/ag-charts-community/src/chart/factory/seriesTypes.ts#L41

For tests, `skipAnimation` should be true and the final 'tick' executed synchrnously, thus removing `requestAnimationFrame` from the equation?

alantreadway: ^ above has been resolved - thanks @lsjroberts

charts-community-modules/ag-charts-community/src/motion/states.ts
alantreadway: Discuss offline, here's a working example: https://www.typescriptlang.org/play?ts=4.3.5#code/JYOwLgpgTgZghgYwgAgMpjpAIhGphjAD2IAPOpihAB6QgAmAzso2FKAOYA0yAogG4RwAFQCeABwjMadJizacAfMgDeAKGSbkiQiUYB+AFyqNWsyV7hoR5AAoAlMgC8y-kWD0A3KbOaL1AhsHZ1d3Lx9kAF9vcxBjdV8tAG0AawhRZFA+QREJKQBdeIjEzQwoDggwYwpIGJKzHWI4u0cXZDcPOsToiJ7ItQGEPTBkSFZqjGxcfF0yAHIxsABGOeQAH2QFqTAAJjmeOYgcsABBVY3D44AhOeUnEz9mhK0joVOixMaSY2C2kABXAA2gK4xVGcHKlWMACJFktoaCzJFEVpTP1okA

alantreadway (approved):
LGTM if you can tweak the `StateDefinition`?

PR #6583, opened 2023-05-11
AG-8403 - Add DataModel support for accumulated groups + properties

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8403

Pre-requisite changes to support the `PieSeries` use-case.

Adds support for:
- Accumulative values across groups (rows - bar+area series stack case) and properties (columns - pie case).
- Normalisation across groups (rows) and properties (columns).
- Rounding to 4 significant figures for aggregate values, to mitigate cumulative error issues.

Refactors:
- Normalisation support to be a set of group/property processors, rather than an explicit built-in behaviour.
- Moved a bunch of processors into `processors.ts` to reduce the scope of `dataModel.ts`.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
+++ b/charts-community-modules/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
@@ -2289,6 +2289,664 @@ Object {
 }
 `;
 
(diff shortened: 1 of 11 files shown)


alexanderby (approved):

lsjroberts (approved):
Looks good as far as I can tell

PR #6593, opened 2023-05-15
AG-8364 - separate default tooltip show arrow behaviour and user conf…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8364


alantreadway (approved):
LGTM

PR #6596, opened 2023-05-15
AG-8460 Add disableInteractions option on animation manager

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8460

This adds the `disableInteractions: true` option when animating, which disables/enables interactions while that animation is running. With the intent on queueing some types of interaction to be run when re-enabled in the future.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
@@ -1,4 +1,5 @@
 import { BaseManager } from './baseManager';
+import { InteractionManager } from './interactionManager';
 import {
     animate as baseAnimate,
     AnimationControls,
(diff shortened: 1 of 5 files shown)


charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -176,20 +184,30 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
...
             return {
                 start: () => {
                     this.updaters.push([id, update]);
                     if (this.requestId == null) {
                         this.startAnimationCycle();
                     }
+
+                    if (disableInteractions) {
+                        this.interactionManager.disable();
+                        this.interactionDisablers.push(id);
```
alantreadway, line 198: Would it make sense to move state/management of tracking who requested disablement into `InteractionManager`? Just wondering if there maybe other use-cases where parts of charts might also want to do this, and if we have queued events they should only be released after all pausers have resumed, without explicitly knowledge of one another?

alantreadway (commented):
Generally happy with this, but I'm just wondering about the point mentioned?

charts-community-modules/ag-charts-community/src/chart/interaction/animationManager.ts
lsjroberts: Good shout, yeah that would make more sense. So I guess `interactionManager.disable(id)` and then `.enable(id)`. But given `.enable(id)` wouldn't necessarily actual enable it, it probably needs a different name like your saying with pause and... something

lsjroberts: Went with `disable(id)` and `removeDisable(id)`. Thoughts?

lsjroberts: Went with `pause(id)` and `resume(id)`

alantreadway (approved):
LGTM

PR #6598, opened 2023-05-15
AG-8567 Add debounce to scroll vis changes

StephenCooper:
This seems to avoid the resizer errors but would need to ensure this doesn't have other side effects.


StephenCooper:
> https://plnkr.co/edit/8ehW6ILj60gtIOoe?open=main.ts

This is resolved by updated fix

PR #6620, opened 2023-05-18
AG-8234 - data type object comparator col id

peterjrreynolds:


grid-community-modules/core/src/ts/columns/dataTypeService.ts
```diff
@@ -399,6 +404,35 @@ export class DataTypeService extends BeanStub {
...
+                const formatValue = (column: Column, colDef: ColDef, value: any) => dataTypeDefinition.valueFormatter!({
+                    api: this.gridOptionsService.api,
+                    columnApi: this.gridOptionsService.columnApi,
+                    context: this.gridOptionsService.context,
+                    column,
+                    colDef,
+                    value
+                });
+                colDef.comparator = (a: any, b: any) => {
+                    const column = this.columnModel.getPrimaryColumn(colId);
```
StephenCooper, line 416: Do we need to make this call inside the comparator as won't it return the same value for the given colId

peterjrreynolds: The column doesn't necessarily exist at the point `setColDefPropertiesForBaseDataType` runs (for a new grid, this runs before the column has been created). The only way to guarantee it does is to retrieve it inside the comparator.

PR #6622, opened 2023-05-18
AG-8566 Fix integrated multi-donut titles

alexanderby:


grid-enterprise-modules/charts/src/charts/chartComp/chartProxies/polar/pieChartProxy.ts
```diff
@@ -48,15 +48,24 @@ export class PieChartProxy extends ChartProxy {
...
+                    title: {
+                        text: f.displayName,
+                        showInLegend: numFields > 1,
+                        spacing: numFields > 1 ? 0 : undefined,
+                    },
                 } : undefined;
 
+                const calloutLabel = numFields > 1 ? { 
+                    calloutLabel: { enabled: false },
+                }: undefined;
```
alantreadway, line 60: This should probably be `{}` rather than `undefined`, otherwise the spread later on will error?

alexanderby:
Irrelevant, fixed in standalone.

PR #6632, opened 2023-05-19
AG-8603 - Fix legend toggle corrupting seriesItemEnabled tracking.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8603

Fixes corruption of `Series.seriesItemsEnabled` on legend toggle, which leads to side-effects such as axis labels being displayed even when there are no bound-series displayed.

Bonus:
- I made `Series.toggleSeriesItem()` protected since it no longer need to be exposed to `Legend`, a couple of tests use it but I'd rather lock it down and workaround in those situations.


lsjroberts (approved):
lgtm

PR #6646, opened 2023-05-23
AG-8461 AG-8463 Add resize animation event to skip animation on resize

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8461 and https://ag-grid.atlassian.net/browse/AG-8463


alantreadway (approved):
LGTM

PR #6656, opened 2023-05-24
AG-8512 - Add v30 ChartModel Migrations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8512

Adds `ChartModel` migrations for v30.0.0 release.

### DIFF
--- a/grid-enterprise-modules/charts/src/charts/chartModelMigration.ts
+++ b/grid-enterprise-modules/charts/src/charts/chartModelMigration.ts
@@ -26,6 +26,7 @@ export function upgradeChartModel(model: ChartModel): ChartModel {
     model = migrateIfBefore('28.0.0', model, migrateV28);
     model = migrateIfBefore('28.2.0', model, migrateV28_2);
     model = migrateIfBefore('29.0.0', model, migrateV29);
+    model = migrateIfBefore('30.0.0', model, migrateV30);
     model = cleanup(model);
 
     // Bump version to latest.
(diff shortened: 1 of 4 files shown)


grid-enterprise-modules/charts/src/charts/chartModelMigration.ts
```diff
@@ -180,6 +181,8 @@ function migrateV28(model: ChartModel) {
 function migrateV28_2(model: ChartModel) {
     model = jsonRename('chartOptions.pie.series.callout', 'calloutLine', model);
     model = jsonRename('chartOptions.pie.series.label', 'calloutLabel', model);
+    model = jsonRename('chartOptions.pie.series.labelKey', 'sectorLabelKey', model);
+    model = jsonRename('chartOptions.pie.series.labelName', 'sectorLabelName', model);
```
lsjroberts, line 185: Out of curiosity - are these things that were deprecated before v28 and now being removed, or things that were just missed at the time?

lsjroberts (approved):
LGTM

grid-enterprise-modules/charts/src/charts/chartModelMigration.ts
alantreadway: They were deprecated at 28.2.0 / 6.2.0, and have only just been removed - it's debatable if this is the right place to add the migration in retrospect I guess.

alantreadway: I've adjusted the migration to perform a 2nd attempt of this at v30 to avoid any missed updates too - thanks for calling this out.

PR #6657, opened 2023-05-24
AG-8110 Add warning for legend labels with multiple fills

lsjroberts:


charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -299,6 +300,27 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
 
     legendItemNames: { [key in string]: string } = {};
 
+    checkLegendItemNames() {
```
alantreadway, line 303: As per offline discussion: could we move this logic to be more generically applicable in the legend merge processing?

PR #6660, opened 2023-05-24
AG-8543 AG-8544 Fix no data overlay to re-use existing component

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8543 and https://ag-grid.atlassian.net/browse/AG-8544

The issue was that when `updateDelta()` is called it runs the chart constructor again so the overlays manager loses track of the no data overlay element. This fixes that by finding it each time from the class as a child of the parent.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/chart.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/chart.ts
@@ -367,6 +367,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.tooltipManager.destroy();
         this.tooltip.destroy();
         this.legend?.destroy();
+        this.overlays.noData.hide();
         SizeMonitor.unobserve(this.element);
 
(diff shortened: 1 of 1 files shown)


charts-community-modules/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -47,11 +41,28 @@ export class Overlay {
...
+        }
+
+        const newElement = document.createElement('div');
+        newElement.className = this.className;
+
+        return newElement;
+    }
+
+    private findElement(): HTMLElement | null {
+        return this.parentElement.querySelector(`.${this.className}`);
```
alantreadway, line 66: We don't search the dom for elements like this anywhere else, so I would like to avoid doing it and introducing the risk of something outside our control injecting DOM elements that we all of a sudden start interacting with?

alantreadway (approved):
LGTM

PR #6663, opened 2023-05-25
Dev server improvements

alantreadway:
- Adds debounce to incremental `generate-code-reference-files.js` invocations.
- Resurrects check for modification before write for `generate-code-reference-files.js` outputs.
- Removes redundant formatting for  `generate-code-reference-files.js` outputs.

### DIFF
--- a/grid-packages/ag-grid-docs/dev-server.js
+++ b/grid-packages/ag-grid-docs/dev-server.js
@@ -33,6 +33,21 @@ const lnk = require('lnk').sync;
 
 const EXPRESS_HTTPS_PORT = 8080;
 
+function debounce(cb, timeout = 500) {
+    let nextTimeout;
+
+    return () => {
+        if (nextTimeout) {
+            clearTimeout(nextTimeout);
+        }
+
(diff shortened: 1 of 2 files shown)


grid-packages/ag-grid-docs/dev-server.js
```diff
@@ -683,9 +696,7 @@ const watchAutoDocFiles = async () => {
     chokidar.watch(INTERFACE_GLOBS, {
         ignored: ignoredFolders,
         persistent: true
-    }).on('change', async (data) => {
-        await generateAutoDocFiles();
-    });
+    }).on('change', debounce(() => generateAutoDocFiles()));
```
StephenCooper, line 699: Looks good, I would suggest we debounce all of the watchers and not just these two

alantreadway: I'll look at that as a follow-up, since that case is more complex (paths are passed as parameters to the build functions).

alantreadway: https://github.com/ag-grid/ag-grid/pull/6667

PR #6667, opened 2023-05-25
More dev-server improvements

alantreadway:
- Add debounce for incremental example generation.
  - Debounce is tracked at docs page granularity, so large-scale example updates (`git pull` or `git checkout` cases) don't result in multiple re-generations.
- Trigger rebuild of specific example if number of changed examples < 5 for a docs page.
- Trigger rebuild of all docs page examples if >= 5 examples changed.
- Adjusts `example-generator-documentation.js` to avoid writing files if the content is unchanged.


StephenCooper (approved):
Looks ok to me. Definitely getting more complex but hopefully we will notice the improvement.

PR #6681, opened 2023-05-30
AG-7112 - Add axis title formatter.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-7112

Adds support for `axes[].title.formatter` callback function, allowing dynamic chart-axis title calculation. The primary use-case is Integrated Charts customisation.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/util/json.ts
+++ b/charts-community-modules/ag-charts-community/src/util/json.ts
@@ -287,7 +287,7 @@ export function jsonApply<Target, Source extends DeepPartial<Target>>(
         const propertyPath = `${path ? path + '.' : ''}${property}`;
         const targetClass = targetAny.constructor;
         const currentValue = targetAny[property];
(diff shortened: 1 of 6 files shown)


charts-community-modules/ag-charts-community/src/util/json.ts
```diff
@@ -287,7 +287,7 @@ export function jsonApply<Target, Source extends DeepPartial<Target>>(
         const propertyPath = `${path ? path + '.' : ''}${property}`;
         const targetClass = targetAny.constructor;
         const currentValue = targetAny[property];
-        let ctr = constructors[property] ?? constructors[propertyMatcherPath];
+        let ctr = constructors[propertyMatcherPath] ?? constructors[property];
```
lsjroberts, line 290: Interesting. If this was wrong before it seems like it could have had big implications? Surprised it didn't throw anything up sooner, unless I'm misunderstanding something.

lsjroberts (approved):
LGTM

PR #6699, opened 2023-06-01
AG-8537 Radar Line chart prototype

alexanderby:


charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+            ready: {
+                on: {
+                    update: {
+                        target: 'ready',
+                        action: () => this.animateUpdateReady(),
+                    },
+                },
+            },
+        });
+        this.animationState.debug;
```
alantreadway, line 222: Nit: Intended?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+            node.strokeWidth = 1;
+            node.fill = undefined;
+            node.pointerEvents = PointerEvents.None;
+        });
+    }
+
+    private updatePath() {
+        this.pathSelection.update(this.visible ? [true] : []).each((node) => {
+            const { path } = node;
+            path.clear();
```
alantreadway, line 377: Ditto - consider if `trackChanges: true` is needed here?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+        this.updateLabels();
+    }
+
+    private drawTempAxis() {
+        const { visible } = this;
+        const radius = this.radiusScale.range[1];
+        const cx = this.centerX;
+        const cy = this.centerY;
+        this.angleAxisSelection.update(visible ? this.nodeData : []).each((node, datum) => {
+            node.path.clear();
```
alantreadway, line 354: If the path is likely to be identical on most calls through here, it maybe worth passing `{ trackChanges: true }` as an option here, so that change detection has a change to decide there are actually no changed to (re)render?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+            const angle = this.angleScale.convert(datum.datum[this.angleKey]);
+            node.path.moveTo(cx, cy);
+            node.path.lineTo(cx + radius * Math.cos(angle), cy + radius * Math.sin(angle));
+            node.stroke = 'gray';
+            node.strokeWidth = 1;
+            node.pointerEvents = PointerEvents.None;
+        });
+        this.radiusAxisSelection.update(visible ? [true] : []).each((node) => {
+            node.path.clear();
```
alantreadway, line 363: Ditto - consider if `trackChanges: true` is needed here?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+            props: [
+                valueProperty(angleKey, false, { id: 'angleValue' }),
+                valueProperty(radiusKey, false, { id: 'radiusValue', invalidValue: undefined }),
+            ],
+        });
+        this.processedData = this.dataModel.processData(data ?? []);
+
+        const angleValueIdx = this.dataModel.resolveProcessedDataIndexById('angleValue')?.index ?? -1;
+        const radiusValueIdx = this.dataModel.resolveProcessedDataIndexById('radiusValue')?.index ?? -1;
+        this.angleScale.domain = this.processedData!.domain.values[angleValueIdx];
```
alantreadway, line 254: Nit: Do not use `!` (see my comment about `getDomain()`, if we follow that pattern then we avoid forcing types like this anyway)

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -0,0 +1,618 @@
...
+                    const labelY = y + sin * marker.size;
+                    labelNodeDatum = {
+                        text: labelText,
+                        x: labelX,
+                        y: labelY,
+                        hAlign: isNumberEqual(cos, 0) ? 'center' : cos > 0 ? 'left' : 'right',
+                        vAlign: isNumberEqual(sin, 0) ? 'middle' : sin > 0 ? 'top' : 'bottom',
+                    };
+                }
+            }
```
alantreadway, line 318: Side-note: I couldn't spot any similarity to `PieSeries` label handling, but if we do spot common processing we can break out utility functions to be shared like in this PR perhaps:
https://github.com/ag-grid/ag-grid/pull/6700

alantreadway (commented):
Nice, generally LGTM.

As per my side-note, let's keep an eye out for common code we can factor our as per Mana's refactors on `BarSeries`.

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
alexanderby: This prevents ESLint warning, I will remove when will be adding animations.

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
alexanderby: Noted, I will consider this when getting to label collisions.

alantreadway (approved):
LGTM

PR #6704, opened 2023-06-02
AG-8476 - Add general Charts formatter caching.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8476

Adds a `CallbackCache` that attempts to reduce the number of calls to user-provided callbacks in a consistent manner.

The cache strategy is:
- Allow at most one call per unique parameters to a user callback.
- The cache is invalidated on an `AgChart.update()` or `AgChart.updateDelta()` call.
- A unique key for parameters is generated using `JSON.stringify()`, so we may need to watch out for incorrect handling of some edge cases perhaps.

The `CallbackCache` is implemented for all `formatter()` options in this PR.

Bonus:
- We now have `ModuleContext` routed into all series and legend types.

***Plunkrs demoing axis label callback counts on resize***

Before:
https://github.com/ag-grid/ag-grid/assets/17544187/60412d18-ece4-4580-8892-4d63d6887c00

After:
https://github.com/ag-grid/ag-grid/assets/17544187/74d3b034-8357-4e1d-b9c0-6efadcf20dd9


lsjroberts:
Can probably refactor to get `animationManager` etc from ctx rather than directly passing in now. One for a future PR. https://github.com/ag-grid/ag-grid/pull/6704/files#diff-d94ba4a8c81e6529c57db3a1150e5554deb3f96bbc9eafca67fa2c1b560302d9R240

lsjroberts:
It may be worth mentioning in the docs that these functions are cached? Since if a user is (in a somewhat odd and bad move) relying on a side effect each time the function is called, it would be a breaking change.

lsjroberts (approved):
Looks great!

alantreadway:
> It may be worth mentioning in the docs that these functions are cached? Since if a user is (in a somewhat odd and bad move) relying on a side effect each time the function is called, it would be a breaking change.

We'll need to add it to the release notes as a breaking change at the least, I'm not sure where right now we would capture this in the docs since it's a cross-cutting concern, but will think about it.

PR #6721, opened 2023-06-06
AG-8531 AG-8532 Radar chart axes prototype

alexanderby:


charts-community-modules/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -432,7 +426,7 @@ function applyChartOptions(chart: Chart, processedOptions: Partial<AgChartOption
         applySeries(chart, processedOptions);
         forceNodeDataRefresh = true;
     }
-    if (isAgCartesianChartOptions(processedOptions) && processedOptions.axes) {
+    if ((processedOptions as any).axes) {
```
alantreadway, line 429: Nit: Could we avoid the `any` here by introducing a new type for charts with `axes`, and using a type-guard for that here (as per the pattern that was here before)?

charts-community-modules/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -520,7 +514,7 @@ function applySeries(chart: Chart, options: AgChartOptions) {
     chart.series = createSeries(chart, optSeries);
 }
 
-function applyAxes(chart: Chart, options: AgCartesianChartOptions) {
+function applyAxes(chart: Chart, options: any) {
```
alantreadway, line 517: Ditto `any`

charts-community-modules/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -576,34 +570,14 @@ function createSeries(chart: Chart, options: SeriesOptionsTypes[]): Series[] {
     return series;
 }
 
-function createAxis(chart: Chart, options: AgCartesianAxisOptions[]): ChartAxis[] {
+function createAxis(chart: Chart, options: any[]): ChartAxis[] {
```
alantreadway, line 573: Nit: Let's introduce a common interface type for axes rather than forcing `any` here?

charts-community-modules/ag-charts-community/src/chart/factory/axisTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+    [GroupedCategoryAxis.type]: GroupedCategoryAxis,
+    [LogAxis.type]: LogAxis,
+};
+
+export function registerAxis(axisType: string, ctor: AxisConstructor) {
+    AXIS_CONSTRUCTORS[axisType] = ctor;
+}
+
+export function getAxis(axisType: string, moduleCtx: ModuleContext) {
+    const seriesConstructor = AXIS_CONSTRUCTORS[axisType];
```
alantreadway, line 22: Nit: `series` in local variable names.

charts-community-modules/ag-charts-community/src/util/module.ts
```diff
@@ -47,6 +48,7 @@ export interface AxisContext {
     scaleInvert(position: number): any;
 }
 
+export type AxisConstructor = new (moduleContext: ModuleContext) => ChartAxis;
```
alantreadway, line 51: Let's avoid the dependency on `ChartAxis` here - `ChartAxis` is an implementation type, we already kinda screwed up by requiring `Series` on the next line (however, that is going to be much more difficult to untangle due to how closely coupled `Series` is to everything else).

`ChartLegend` is a good example of a minimal interface contract, can we try to do something similar here?

charts-community-modules/ag-charts-community/src/chart/axis/angleCategoryAxis.ts
```diff
@@ -0,0 +1,102 @@
+import { isEqual } from '../../util/number';
```
alantreadway, line 1: Guessing this needs to be removed from community?

charts-community-modules/ag-charts-community/src/chart/axis/radiusNumberAxis.ts
```diff
@@ -0,0 +1,49 @@
+import { LinearScale } from '../../scale/linearScale';
```
alantreadway, line 1: Guessing this needs to be removed from community?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/polar-axes/angle-category/typings.ts
```diff
@@ -0,0 +1,7 @@
+import { AgBaseCartesianAxisOptions, AgAxisCategoryTickOptions } from 'ag-charts-community';
+
+export interface AgAngleCategoryAxisOptions extends AgBaseCartesianAxisOptions {
```
alantreadway, line 3: Extending `AgBaseCartesianAxisOptions` seems wrong - can we introduce a non-cartesian axis type that is shared with that type perhaps?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/polar-axes/radius-number/typings.ts
```diff
@@ -0,0 +1,13 @@
+import { AgBaseCartesianAxisOptions, AgAxisNumberTickOptions } from 'ag-charts-community';
+
+export interface AgRadiusNumberAxisOptions extends AgBaseCartesianAxisOptions {
```
alantreadway, line 3: Ditto AgBaseCartesianAxisOptions?

charts-enterprise-modules/ag-charts-enterprise/src/radar-line/radarLineSeries.ts
```diff
@@ -278,10 +253,17 @@ export class RadarLineSeries extends _ModuleSupport.PolarSeries<RadarLineNodeDat
             return [];
         }
 
+        const angleScale = this.xAxis?.scale;
+        const radiusScale = this.yAxis?.scale;
```
alantreadway, line 257: Can we remove the concept of `xAxis` and `yAxis`?

charts-community-modules/ag-charts-community/src/chart/factory/axisTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+import { LogAxis } from '../axis/logAxis';
+import { NumberAxis } from '../axis/numberAxis';
+import { CategoryAxis } from '../axis/categoryAxis';
+import { GroupedCategoryAxis } from '../axis/groupedCategoryAxis';
+import { TimeAxis } from '../axis/timeAxis';
+
+const AXIS_CONSTRUCTORS: Record<string, AxisConstructor> = {
+    [NumberAxis.type]: NumberAxis,
+    [CategoryAxis.type]: CategoryAxis,
+    // @ts-ignore: Deprecated `createTick().count` type mismatch.
```
alantreadway, line 11: Is there anyway to avoid this, we don't commonly use this, and it could hide other contract mismatches?

charts-community-modules/ag-charts-community/src/util/module.ts
alexanderby: I tried making `ChartAxis` an interface, but it becomes pretty huge, and then there is a need for a `label` interface, `scale` interface. I will leave it without an interface for now, let's discuss when you are back?
```ts
export interface BoundSeries {
    type: string;
    getDomain(direction: ChartAxisDirection): any[];
    isEnabled(): boolean;
    getKeys(direction: ChartAxisDirection): string[];
    getNames(direction: ChartAxisDirection): (string | undefined)[];
    visible: boolean;
    getBandScalePadding?(): { inner: number; outer: number };
}

export interface ChartAxis {
    type: string;
    id: string;

    isModuleEnabled(module: any): boolean;
    addModule(module: any): void;
    removeModule(module: any): void;

    getLayoutState(): AxisLayout;
    position: AgCartesianAxisPosition;
    setCrossLinesVisible(visible: boolean): void;
    clipGrid(x: number, y: number, width: number, height: number): void;
    clipTickLines(x: number, y: number, width: number, height: number): void;
    crossLines?: any[];
    seriesAreaPadding: number;
    readonly direction: 'x' | 'y';
    range: number[];
    visibleRange: number[];
    gridLength: number;
    nice: boolean;
    thickness: number;
    maxThickness: number;

    update(primaryTickCount?: number): number | undefined;
    computeBBox(): BBox;
    gridPadding: number;
    translation: { x: number; y: number };
    updatePosition(position: { rotation: number; sideFlag: Flag }): void;
    rotation: number;
    label: { getSideFlag(): Flag };
    calculatePadding(min: number, _max: number): number;
    destroy(): void;
    keys: string[];
    attachAxis(node: Node): void;
    detachAxis(node: Node): void;
    linkedTo?: ChartAxis;
    boundSeries: BoundSeries[];
}
```

alantreadway (commented):
Nice, liking the `ChartAxis` change to an interface :+1:

alexanderby:
I made the requested changes. `ChartAxisDirection.X/Y` sounds a bit wrong for polar axes, but seems not a big deal, so I'm merging it to test all together and simplify the further work on radars.

PR #6728, opened 2023-06-07
RTI-1465 Fix Treemap tooltip arrow between tiles

alexanderby:


alexanderby:
Committed separately because of GitHub outage.

PR #6735, opened 2023-06-08
AG-8471 Add bar/column animation of adding, removing and updating data

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8471

This provides a new selection strategy if an optional `getDatumId()` function is provided. This allows the datum id to be tracked against the nodes created for them which in turn allows correct ordering and moving about when nodes are added and removed. If this is provided the series must handle the garbage collection itself by calling the `cleanup()` method on the selection, allowing the series time to animate the removed nodes out of view.

It adds a `diff` post-processor to the data model that can be given the previous data and provide back the list of keys that have been added, updated or removed. The animation handlers can then use this to move those nodes around.

Since a series `update()` method is called whenever a node is highlighted, I've added a special `waiting` animation state that the series goes into whenever the data itself is updated in the `processData()` method. This blocks any other transitions until the `update()` method finishes and transitions with the new processed data. Without this the animation would keep triggering with the new added/updated/removed nodes being re-drawn.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/data/processors.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/data/processors.ts
@@ -153,23 +153,42 @@ export function diff(previousData: ProcessedData<any>): ProcessorOutputPropertyD
                 removed: [] as any[],
             };
 
-            processedData?.data.forEach((datum) => {
-                const previous = previousData?.data.find((previous) => arraysEqual(datum.keys, previous.keys));
+            const added = new Map<string, any>();
+            const updated = new Map<string, any>();
+            const removed = new Map<string, any>();
+            const sep = '___';
 
-                if (previous && !arraysEqual(datum.values, previous.values)) {
-                    diff.updated.push(datum.keys);
-                } else if (!previous) {
-                    diff.added.push(datum.keys);
+            for (let i = 0; i < Math.max(previousData.data.length, processedData.data.length); i++) {
+                const prev = previousData.data[i];
+                const datum = processedData.data[i];
+
+                const prevId = prev?.keys.join(sep);
+                const datumId = datum?.keys.join(sep);
+
+                if (prevId === datumId) {
+                    if (!arraysEqual(prev.values, datum.values)) {
+                        updated.set(datumId, datum);
+                    }
+                    continue;
                 }
-            });
 
-            previousData?.data.forEach((previous) => {
-                const datum = processedData?.data.find((datum) => arraysEqual(previous.keys, datum.keys));
-
-                if (!datum) {
-                    diff.removed.push(previous.keys);
+                if (removed.has(datumId)) {
+                    updated.set(datumId, datum);
+                    removed.delete(datumId);
+                } else if (datum) {
+                    added.set(datumId, datum);
                 }
-            });
+                if (added.has(prevId)) {
+                    updated.set(prevId, prev);
+                    added.delete(prevId);
+                } else if (prev) {
+                    removed.set(prevId, prev);
+                }
+            }
+
+            diff.added = Array.from(added.values()).map((datum) => datum.keys);
+            diff.updated = Array.from(updated.values()).map((datum) => datum.keys);
+            diff.removed = Array.from(removed.values()).map((datum) => datum.keys);
 
             diff.changed = diff.added.length > 0 || diff.updated.length > 0 || diff.removed.length > 0;
 
--- a/charts-community-modules/ag-charts-community/src/chart/data/processors.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/data/processors.ts
@@ -4,6 +4,7 @@ import {
     PropertyId,
     PropertyValueProcessorDefinition,
     ReducerOutputPropertyDefinition,
+    ProcessedData,
 } from './dataModel';
 
 export const SMALLEST_KEY_INTERVAL: ReducerOutputPropertyDefinition<number> = {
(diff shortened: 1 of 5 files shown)


charts-community-modules/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -139,3 +140,52 @@ export function normalisePropertyTo(
...
+                }
+            });
+
+            previousData?.data.forEach((previous) => {
+                const datum = processedData?.data.find((datum) => arraysEqual(previous.keys, datum.keys));
+
+                if (!datum) {
+                    diff.removed.push(previous.keys);
+                }
+            });
```
alantreadway, line 172: Hmmm.... this is likely going to be quite expensive (O(2(n^2))) for larger data-sets. Could we make some simplifying assumptions like:
- Data is provided in the same order.
- Moved data is treated as distinct remove/add (obviously we'll need to tweak this later).

...and then just iterate through the `processedData` and `previousData` arrays in parallel and find differences as we go? (O(2n))

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -327,6 +327,10 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
             extraProps.push(normaliseGroupTo(activeSeriesItems, normaliseTo, 'sum'));
         }
 
+        if (this.processedData) {
+            extraProps.push(diff(this.processedData));
+        }
```
alantreadway, line 332: Feeling like we should short-circuit this if animations are disabled, as even `O(2n)` of extra overhead might be painful?

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -1014,6 +1035,79 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
...
+                if (keys && diff.added.find((a: string[]) => keyMatches(a, datum))) {
+                    props = [
+                        { from: startingX, to: datum.x },
+                        { from: 0, to: datum.width },
+                        { from: datum.y, to: datum.y },
+                        { from: datum.height, to: datum.height },
+                    ];
+                    delay += sectionDuration;
+                    duration = sectionDuration;
+                } else if (keys && diff.removed.find((r: string[]) => keyMatches(r, datum))) {
```
alantreadway, line 1092: Ditto?

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -1112,4 +1197,83 @@ export class ColumnSeries extends BarSeries {
...
         });
     }
+
+    animateReadyUpdate({ datumSelections }: { datumSelections: Array<Selection<Rect, BarNodeDatum>> }) {
+        datumSelections.forEach((datumSelection) => {
+            this.resetSelectionRects(datumSelection);
+        });
+    }
+
+    animateWaitingUpdateReady({ datumSelections }: { datumSelections: Array<Selection<Rect, BarNodeDatum>> }) {
```
alantreadway, line 1207: Seeing a lot of similar looking code in this method, any chance we could factor out the commonality?

charts-community-modules/ag-charts-community/src/scene/selection.ts
```diff
@@ -5,36 +5,66 @@ type NodeFactory<TNode extends Node, TDatum> = (datum: TDatum) => TNode;
...
         const old = this._data;
         const parent = this._parent;
         const factory = this._factory;
 
-        if (data.length > old.length) {
+        if (getDatumId) {
+            // Append any new datum ids and nodes to the end of the arrays
+            data.forEach((datum) => {
+                const datumId = getDatumId(datum);
+                if (this._ids.indexOf(datumId) === -1) {
```
alantreadway, line 56: Wondering if we should make `this._ids` a dictionary to try and keep the lookup time small, since the order of _ids doesn't seem to matter?

charts-community-modules/ag-charts-community/src/scene/selection.ts
```diff
@@ -43,14 +73,34 @@ export class Selection<TChild extends Node = Node, TDatum = any> {
...
 
+        // Copy the data into a new array to prevent mutation and duplicates
         this._data = data.slice(0);
-        for (let i = 0; i < data.length; i++) {
-            this._nodes[i].datum = this._data[i];
+
+        if (getDatumId) {
+            // Find and update the datum for each node by the index of the id within the set
+            for (let i = 0; i < this._ids.length; i++) {
+                const datum = this._data.find((d) => getDatumId(d) === this._ids[i]);
```
alantreadway, line 88: O(n^2) again :-(

charts-community-modules/ag-charts-community/src/chart/data/processors.ts
alantreadway: Demo Plunker for reference: https://plnkr.co/edit/sQRMpV8dU78tFNJl?open=script.js&preview

charts-community-modules/ag-charts-community/src/scene/selection.ts
lsjroberts: The ordering of the ids mattered as they related to the nodes by the index. But I've changed it to a dict using the datum id as the key and node index as the value.

charts-community-modules/ag-charts-community/src/chart/data/processors.ts
lsjroberts: I've updated it to use this approach. It saves a few ms with 8k nodes, but there's still a chunky hang for large data sets after animations finish that I need to look into separately.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -1014,6 +1035,79 @@ export class BarSeries extends CartesianSeries<SeriesNodeDataContext<BarNodeDatu
...
+                    { from: rect.x, to: datum.x },
+                    { from: rect.width, to: datum.width },
+                    { from: rect.y, to: datum.y },
+                    { from: rect.height, to: datum.height },
+                ];
+                let delay = diff.removed.length > 0 ? sectionDuration : 0;
+                let duration = sectionDuration;
+                let cleanup = false;
+
+                if (keys && diff.added.find((a: string[]) => keyMatches(a, datum))) {
```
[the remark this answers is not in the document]
lsjroberts: I've changed it to use a dict lookup and assume that a bar series never has more than one key, i.e. just use `this.processedData?.defs.keys?.[0]` instead of the whole array.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
lsjroberts: Yeah I had it split as I thought there might be more differences as I went, but I've combined them together now with a couple of checks on bar direction.

charts-community-modules/ag-charts-community/src/chart/series/cartesian/barSeries.ts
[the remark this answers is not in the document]
lsjroberts: Though looking at https://ag-grid.atlassian.net/browse/AG-7651 that assumption may not be possible? Is it multiple keys or is it a path for a single key i.e. `['foo', 'bar', 'baz'] === { foo: { bar: { baz: 'value' } } }`?

PR #6798, opened 2023-06-22
AG-8561 Add animation.duration option

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8561

This only adds a basic version of the `animation.duration` option for the total duration, mostly for QA purposes.


alantreadway (approved):
LGTM

PR #6804, opened 2023-06-23
AG-7933 - Implement DataModel reduction/merging.

alantreadway:

Implements merging of all series-specified `DataModel` `PropertyDefinition[]` so all processing is performed by the minimum number of `DataModel` instances.

Some things to note:
- `scope` has been introduced as an additional concept, which alongside `id` allows identification of which series originated a `PropertyDefinition` after reduction (and its subsequent output values in `ProcessedData`).
- Validation state is tracked per `scope`, and if there are mismatches in the set of valid rows per `scope` there is a final step to split results appropriately.
- `ProcessedData` no longer contains property->index information, since there is no longer a trivial mapping - `DataModel` exposes methods to allow indexes to be obtained in a correct manner.

Bonus:
- Cleaned up some of the APIs to be more consistent WRT argument ordering.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/data/dataModel.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/data/dataModel.ts
@@ -3,6 +3,8 @@ import { isNumber } from '../../util/value';
 import { windowValue } from '../../util/window';
 import { ContinuousDomain, extendDomain } from './utilFunctions';
 
+export type ScopeProvider = { id: string };
+
 export type UngroupedDataItem<D, V> = {
     keys: any[];
     values: V;
@@ -19,10 +21,6 @@ export interface UngroupedData<D> {
         groups?: any[][];
         aggValues?: [number, number][];
     };
-    indices: {
-        keys: Record<keyof D, number>;
-        values: Record<keyof D, number>;
-    };
     reduced?: Record<string, any>;
     defs: {
         keys: DatumPropertyDefinition<keyof D>[];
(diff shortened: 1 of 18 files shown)


charts-community-modules/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -264,25 +264,26 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
...
-                keys: [keys],
                 values: [yExtent],
                 aggValues: [ySumExtent] = [],
             },
         } = processedData;
 
+        const keyDef = dataModel.resolveProcessedDataDefById(this, `xValue`);
+        const keys = dataModel.getDomain(this, `xValue`, 'key', processedData);
+        // const yExtent = dataModel.getDomain(``);
+        // const ySumExtent = dataModel.getDomain(``);
```
lsjroberts, line 283: delete?

charts-community-modules/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -250,34 +264,70 @@ export class DataModel<
...
+        ];
+        for (const defs of allDefs) {
+            const def = defs.find(
+                ({ id, scopes }) => id === searchId && (scope == null || (scopes?.includes(scope.id) ?? false))
+            );
+            if (def) return { index: def.index, def };
+        }
 
-        return def;
+        throw new Error(`AG Charts - didn't find property definition for [${searchId}, ${scope.id}, ${type}]`);
```
lsjroberts, line 297: Much nicer than having to handle all those undefineds, nice.

lsjroberts (approved):
LGTM, very cool 🛳

charts-community-modules/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: Thanks for calling out, I needed to switch these to `DataModel.getDomain()` calls - now fixed.

lsjroberts (approved):

PR #6831, opened 2023-06-29
AG-8552 Change zoom panning to pan each axis separately

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8552

This changes zoom panning so that it will pan each series together such that they stay at the same relative positions to each other. But then if one series reaches the edge of it’s axis' domain, it will then stop while the others continue to move. Sliding over each other.

Previously it would reset the zoom of all the axes to match the last axis in the array.

https://github.com/ag-grid/ag-grid/assets/3817697/8b43a253-b437-4b21-9bb8-d888bcea2b08


alantreadway (approved):
LGTM

PR #6839, opened 2023-07-04
AG-8721 Restore Focus for React Async Rendering

StephenCooper:


gportela85 (changes requested):
This looks good, just one small detail, since we are only using `shouldRestoreFocus` in React, we should also find a way to only call `setRestoreFocus`  in React, otherwise we will end up with `this.restoreFocusedCellPosition` being set and never really cleared.

Maybe update the method names or add comments so we know for future reference that this is a React only functionality.

PR #6850, opened 2023-07-06
AG-7933 - Bugfixes for BarSeries/AreaSeries simplifications.

alantreadway:

Multiple fixes:
- Fixes cross-filtering bar/column series rendering so that filtered-out bar appears correctly.
- Fixes cross-filtering and animation updates not triggering redraw.
- Fixes incorrect grouping of ungrouped series types, leading to weird rendering.

Bouns:
- Adds regression tests for integrated-charts cases that triggered some of these cases.


charts-community-modules/ag-charts-community/src/scene/group.ts
```diff
@@ -271,7 +271,7 @@ export class Group extends Node {
         if (hasVirtualChildren) {
             // Mark virtual nodes as clean, but don't recurse (their children have already been updated).
             for (const child of this.virtualChildren) {
-                child.markClean({ recursive: false });
+                child.markClean({ recursive: 'virtual' });
```
lsjroberts, line 274: nit: The comment is not entirely accurate I guess? `..., and only recurse their virtual children ...` or something?

charts-community-modules/ag-charts-community/src/chart/__image_snapshots__/integrated-charts-examples-test-ts-integrated-charts-examples-changing-chart-type-for-histogram-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: Is this as expected? Not too familiar with histograms but looks a bit off?

lsjroberts (approved):
LGTM, great stuff

charts-community-modules/ag-charts-community/src/chart/__image_snapshots__/integrated-charts-examples-test-ts-integrated-charts-examples-changing-chart-type-for-histogram-it-should-render-to-canvas-as-expected-1-snap.png
alantreadway: Yeah I also paused on this one - I've just verified on production and the same data-set renders the same, so I'm ruling it out of scope to fix as part of this PR.

PR #6859, opened 2023-07-10
AG-7933 - Fix series layer management state updates.

alantreadway:

Fixes stale series layer zIndex on dynamic series changes causing rendering problems.

### DIFF
--- a/charts-community-modules/ag-charts-community/src/chart/series/seriesLayerManager.ts
+++ b/charts-community-modules/ag-charts-community/src/chart/series/seriesLayerManager.ts
@@ -99,7 +99,7 @@ export class SeriesLayerManager {
             delete this.groups[type][groupIndex];
             delete this.groups[type][id];
         } else if (groupInfo?.seriesIds.length > 0) {
-            // Update zIndexSubOrder to avoid it becoming state as series are removed and re-added
+            // Update zIndexSubOrder to avoid it becoming stale as series are removed and re-added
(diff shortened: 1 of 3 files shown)


manapeirov (approved):

charts-community-modules/ag-charts-community/src/chart/series/seriesLayerManager.ts
```diff
@@ -8,33 +8,39 @@ export type SeriesConfig = {
...
     }
 
-    public requestGroup(opts: SeriesConfig) {
-        const { id, seriesGrouping, type, rootGroup } = opts;
+    public requestGroup(seriesConfig: SeriesConfig) {
+        const { id, type, rootGroup: seriesRootGroup, seriesGrouping } = seriesConfig;
         const { groupIndex = id } = seriesGrouping ?? {};
 
+        if (this.series[id] != null) {
+            throw new Error(`AG Charts - series already has an allocated layer: ${this.series[id]}`);
```
lsjroberts, line 41: Would this be caused by the user duplicating `series.id` in the options? Is that not caught else where? Or is it only an internal error?

alantreadway: It's an internal error case. The issue I fixed would've been highlighted a lot sooner with this error in place, so figured it's worth adding (I also found a race-condition on `Chart.destroy()` as a result of these checks too).

PR #6874, opened 2023-07-13
AG-8471 Add animation throttling to bar updates

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8471

With the previous solution, bars after the first would have their animation cancelled. This solution creates animation throttle groups that signal that these animations should not throttle each other.

Using a `throttleGroup` in combination with a `throttleId` allows batches of animations to run normally but throttle later batches.

So in this example, each time the update animation is triggered a new `throttleGroup` is created for the whole datum selection, but the series always shared a single `throttleId` across all animations.

https://github.com/ag-grid/ag-grid/assets/3817697/4b617c67-8bff-4899-94e2-bbbf3942762f


alantreadway (approved):
LGTM, nice solution!

PR #6893, opened 2023-07-21
AG-8905 React Headers Reduce Renders

StephenCooper:


grid-community-modules/core/src/ts/headerRendering/cells/column/headerCellCtrl.ts
```diff
@@ -660,8 +658,8 @@ export class HeaderCellCtrl extends AbstractHeaderCellCtrl {
         listener();
     }
 
-    private setupColId(): void {
-        this.comp.setColId(this.column.getColId());
+    public getColId() {
```
StephenCooper, line 661: Only set once so might as well have the component read it in the initial render instead of triggering an update via `setState`

grid-community-modules/core/src/ts/headerRendering/row/headerRowCtrl.ts
```diff
@@ -35,40 +33,61 @@ export class HeaderRowCtrl extends BeanStub {
...
 
-        this.onRowHeightChanged();
-        this.onVirtualColumnsChanged();
-        this.setWidth();
+        this.setWidth(); // React comp not currently using state for width
 
         this.addEventListeners();
+    }
 
+    public getHeaderRowClass(): string {
```
StephenCooper, line 76: Turn these to getters so that the components can read their values on render instead of having them pushed in a second render, especially as these values are static.

grid-community-modules/react/src/reactUi/header/headerCellComp.tsx
```diff
@@ -1,50 +1,50 @@
...
                     eGui.current.style.width = width;
                 }
             },
             addOrRemoveCssClass: (name, on) => cssClassManager.addOrRemoveCssClass(name, on),
-            setColId: id => setColId(id),
-            setTitle: title => setTitle(title),
-
-            setAriaDescription: description => setAriaDescription(description),
-            setAriaSort: sort => setAriaSort(sort),
+            setTitle: title => _.addOrRemoveAttribute(eGui.current!, 'title', title),
```
StephenCooper, line 36: Update the dom directly as no real benefit of doing this via React state.

grid-community-modules/react/src/reactUi/header/headerRowComp.tsx
```diff
@@ -1,35 +1,33 @@
...
-    const [ height, setHeight ] = useState<string>();
-    const [top, setTop] = useState<string>();
-    const [ ariaRowIndex, setAriaRowIndex ] = useState<number>();
-    const [ cellCtrls, setCellCtrls ] = useState<AbstractHeaderCellCtrl[]>([]);
+    const { topOffset, rowHeight } = useMemo(() => ctrl.getTopAndHeight(), []);
+    const ariaRowIndex = ctrl.getAriaRowIndex();
+    const className = ctrl.getHeaderRowClass();
 
-    const eGui = useRef<HTMLDivElement>(null);
+    const transform = useMemo(() => ctrl.getTransform(), []);
```
StephenCooper, line 15: Using `useMemo` as the `getTransform` method does feature detection which we don't want to do on every render. Above for `className` and `ariaRowIndex` they simply return simple values so the memo overhead is probably more expensive then just reading the value again.

PR #6899, opened 2023-07-24
AG-8906 Row comps set ref

StephenCooper:


grid-community-modules/react/src/reactUi/rows/rowContainerComp.tsx
```diff
@@ -46,65 +46,107 @@ const RowContainerComp = (params: {name: RowContainerName}) => {
...
+
                 if (domOrderRef.current) {
                     return rowCtrls;
                 }
                 // if dom order not important, we don't want to change the order
                 // of the elements in the dom, as this would break transition styles
                 const oldRows = prev.filter(r => rowCtrls.indexOf(r) >= 0);
                 const newRows = rowCtrls.filter(r => oldRows.indexOf(r) < 0);
+
+                if (oldRows.length === prev.length && newRows.length === 0) {
```
StephenCooper, line 61: This extra logic maynot be required if we implement https://ag-grid.atlassian.net/browse/AG-8915

Although on the other hand it is a light check that is maybe safer to leave in to protect this code path from additional re-renderers.

grid-community-modules/react/src/reactUi/rows/rowContainerComp.tsx
```diff
@@ -46,65 +46,107 @@ const RowContainerComp = (params: {name: RowContainerName}) => {
...
-            setDomOrder: domOrder => {
-                if(domOrderRef.current != domOrder){
-                    domOrderRef.current = domOrder;
-                    updateRowCtrlsOrdered(false);
-                }
-            },
-            setContainerWidth: width => {
-                if (eContainer.current) {
-                    eContainer.current.style.width = width;
+    const areElementsReady = useCallback(() => {
```
StephenCooper, line 71: Because we have different templates we need to validate we have all the refs for the template that we are currently working with.

PR #6900, opened 2023-07-24
AG-8954 Grid Components

StephenCooper:


grid-community-modules/core/src/ts/context/context.ts
```diff
@@ -262,6 +262,10 @@ export class Context {
     public destroy(): void {
         if (this.destroyed) { return; }
 
+        // Set before doing the destroy, so if context.destroy() gets called via another bean
+        // we are marked as destroyed already to prevent running destroy() twice
+        this.destroyed = true;
```
StephenCooper, line 267: Noticed that the Context gets destroyed as part of the GridApi's destroy method so by bringing this flag above `this.destroyBeans(beanInstances);` means it actually prevents a double destruction of the context.

grid-community-modules/react/src/reactUi/agGridReactUi.tsx
```diff
@@ -102,10 +112,6 @@ export const AgGridReactUi = <TData,>(props: AgReactUiProps<TData>) => {
                         if (props.setGridApi) {
                             props.setGridApi(api, gridOptionsRef.current.columnApi!);
                         }
-                        destroyFuncs.current.push(() => {
```
StephenCooper, line 105: api is destroyed via Bean so we do not need to call this ourselves too.

PR #6980, opened 2023-08-22
Fix ag-row-selected. 

EugenGedroyc:
If a header of a group is selected, it is hidden after the mouse cursor is moved away from that header.

https://tinyurl.com/274okj88


StephenCooper:
Please report as an issue if this is still an issue.

PR #7033, opened 2023-09-11
AG-7953 Group total footers reactive

StephenCooper:


grid-enterprise-modules/row-grouping/src/rowGrouping/groupStage.ts
```diff
@@ -39,6 +39,7 @@ interface GroupingDetails {
     groupedColCount: number;
     transactions: RowNodeTransaction[];
     rowNodeOrder: { [id: string]: number; };
+    createGroupFooter: boolean;
```
StephenCooper, line 42: Add to GroupDetails so only queried once for this stage but always gets the latest value without the need for a managed property listener.

PR #7041, opened 2023-09-12
AG-684 Added gridPreDestroyed event

salmenus:


seanlandsman (commented):
The vue part looks fine to me.

Do we not need an equivalent change in the vue3 and angular wrappers?

grid-community-modules/core/src/ts/gridOptionsService.ts
```diff
@@ -73,6 +73,7 @@ export class GridOptionsService {
     // we store this locally, so we are not calling getScrollWidth() multiple times as it's an expensive operation
     private scrollbarWidth: number;
     private domDataKey = '__AG_' + Math.random().toString();
+    private static readonly alwaysSyncGlobalEvents: string[] = [Events.EVENT_GRID_PRE_DESTROYED];
```
StephenCooper, line 76: Can we make it a Set for performance reasons.

grid-community-modules/core/src/ts/gridOptionsService.ts
```diff
@@ -205,17 +207,29 @@ export class GridOptionsService {
...
+    globalEventHandlerFactory = (restrictToSyncOnly?: boolean) => {
+        return (eventName: string, event?: any) => {
+            // prevent events from being fired _after_ the grid has been destroyed
+            if (this.destroyed) {
+                return;
+            }
+
+            if (
+                GridOptionsService.alwaysSyncGlobalEvents.includes(eventName) && !restrictToSyncOnly ||
+                !GridOptionsService.alwaysSyncGlobalEvents.includes(eventName) && restrictToSyncOnly
```
StephenCooper, line 222: Nit: Would prefer to reduce the duplication of the sync check via extracting to a nicely named variable.

grid-community-modules/core/src/ts/gridOptionsService.ts
salmenus: Sure! it's a negligible gain given the size of the array, but let's do it for consistency with other implementations.

StephenCooper (approved):

PR #7060, opened 2023-09-14
AG-9218 - remove clipper containers from the grid

gportela85:
https://ag-grid.atlassian.net/browse/AG-9218


StephenCooper (approved):
LGTM

PR #7100, opened 2023-09-29
ag-grid-solid: Run onMount before createEffect

sin-ack:
Otherwise we access `gridOptions` before it's assigned.

### DIFF
--- a/grid-packages/ag-grid-solid/src/grid/agGridSolid.tsx
+++ b/grid-packages/ag-grid-solid/src/grid/agGridSolid.tsx
@@ -43,38 +43,6 @@ const AgGridSolid = (props: AgGridSolidProps) => {
         destroyFuncs.length = 0;
     });
 
-    // we check for property changes. to get things started, we take a copy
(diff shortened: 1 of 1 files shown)


StephenCooper:
Hi @sin-ack,

Thanks for this. Do you think you could provide a reproduction of this issue so that we can validate your changes?

sin-ack:
Sure, here's a StackBlitz example where the issue is exhibited: `https://stackblitz.com/edit/solidjs-templates-iyvhby?file=src%2FApp.tsx`

You should see "Error!" in the output window, and the error message will be printed to the console.

I couldn't reproduce it at first, but the specific issue seems to be with putting `columnDefs` behind a function (we need this in our project for i18n). If I use static `columnDefs` then the issue isn't exhibited. I'll look into it further tomorrow to see why this is the case.

StephenCooper:
AG Grid no longer actively maintains the Solid implementation. Please see https://github.com/solidjs-community/solid-ag-grid

PR #7149, opened 2023-10-16
AG-9564 - don't call api methods when setting grid options

peterjrreynolds:


grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -260,16 +256,7 @@ export class GridApi<TData = any> {
         force: boolean,
         changeSet: PropertyChangeSet | undefined = undefined
     ) {
-        // Ensure the GridOptions property gets updated and fires the change event as we
-        // cannot assume that the dynamic Api call will updated GridOptions.
         this.gos.set(propertyName, value, force, {}, changeSet);
-        // If the dynamic api does update GridOptions then change detection in the
```
StephenCooper, line 266: So all of the setters have now migrated to gos.set directly?

peterjrreynolds: Yes, there's a full list in the Jira

PR #7202, opened 2023-10-30
AG-9440 - Grid Option Validations

AndrewGlazier:


grid-community-modules/core/src/ts/validation/rules/colDefValidations.ts
```diff
@@ -0,0 +1,108 @@
...
+import { Deprecations, OptionsValidator, Validations } from "../validationTypes";
+import { ColDefUtil } from '../../components/colDefUtil';
+
+const COLUMN_DEFINITION_DEPRECATIONS: Deprecations<ColDef | ColGroupDef> = {
+};
+
+const COLUMN_DEFINITION_VALIDATIONS: Validations<ColDef | ColGroupDef> = {
+    // supported on all row models, but need module for client side.
+    enableRowGroup: (_options, gridOptions) => {
+        if (gridOptions.rowModelType ?? 'clientSide' === 'clientSide') {
```
StephenCooper, line 12: This looks like a bug? Should it be comparing against gridOptions.rowModelType

StephenCooper, line 12: This repetition of strings is something that I would like to wrap into a single function and call that.

grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -2316,5 +1820,815 @@ export class GridApi<TData = any> {
...
+        });
+    }
+
+    /**
+     * @deprecated v31 Use `api.updateGridOptions` or `api.updateGridOption` instead.
+     * Set the top pinned rows. Call with no rows / undefined to clear top pinned rows. 
+     **/
+    public setPivotMode(pivotMode: boolean) {
+        this.logSetGridOptionsDeprecation('pivotMode');
+        this.updateGridOption('pivotMode', pivotMode);
```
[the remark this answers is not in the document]
StephenCooper: I see the names are not quite the same.....

grid-community-modules/core/src/ts/components/componentUtil.ts
```diff
@@ -163,44 +113,8 @@ export class ComponentUtil {
         if (!changes || Object.keys(changes).length === 0) {
             return;
         }
-        this.changeSetId++;
-        const changesToApply = { ...changes };
-
-        // We manually call these updates so that we can provide a different source of gridOptionsChanged
-        // We do not call setProperty as this will be called by the grid api methods
-        if (changesToApply.columnTypes) {
-            api.setColumnTypes(changesToApply.columnTypes.currentValue, "gridOptionsChanged");
```
StephenCooper, line 172: Are we still able to maintain this different source for the api call vs grid options update?

grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -2316,5 +1820,815 @@ export class GridApi<TData = any> {
...
+        this.updateGridOptions({ [key]: value });
+    }
+
+    /**
+     * Updates the provided subset of gridOptions.
+     */
+    public updateGridOptions(options: Partial<GridOptions>): void {
+        this.gos.updateGridOptions({
+            options,
+            source: 'api',
```
StephenCooper, line 1847: Do we want to expose the source so that we can customise this? This would maybe also be a handy feature for users to be able to update to their own custom source perhaps

grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -2316,5 +1820,815 @@ export class GridApi<TData = any> {
...
 
+    /** Get the current state of the grid. Can be used in conjunction with the `initialState` grid option to save and restore grid state. */
+    public getState(): GridState {
+        return this.stateService.getState();
+    }
+
+
+
+    private logSetGridOptionsDeprecation(func: string) {
+        warnOnce(`${func} is deprecated. Please use api.updateGridOptions() instead.`);
```
StephenCooper, line 1831: Should this also include the individual method version?

AndrewGlazier: Have got this locally, so yes will do 👍

grid-community-modules/core/src/ts/gridApi.ts
AndrewGlazier: Have been discussing this with Pete today at some length. Basically the issue is that currently we allow setting source for some individual properties, namely the column def setters. These currently accept a source type of `ColumnEventType`.

The problem here is that this no longer makes sense, because we're asking people to set a partial gridOptions, so it's difficult to type this sensibly. Happy to add a source option with generic 'string' typing, but Pete suggested we just remove this for now, and see if it's requested. I'm somewhat ambivalent.

AndrewGlazier: Coming back to this, I'll add source but strictly type it to 'api' or 'gridOptionsChanged'

grid-community-modules/core/src/ts/components/componentUtil.ts
AndrewGlazier: Wasn't but will now.

grid-community-modules/core/src/ts/gridApi.ts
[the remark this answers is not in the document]
AndrewGlazier: Done

grid-community-modules/core/src/ts/gridApi.ts
StephenCooper: Will this then enable us to not make a breaking change for people who are maybe listening for the `gridOptionsChanged` event for columns? Otherwise we need to include this in the change log

StephenCooper:
Happy with the changes.

PR #7218, opened 2023-11-01
AG-6381 Initial implementation of page size selector

salmenus:


grid-community-modules/core/src/ts/grid.ts
```diff
@@ -347,6 +347,7 @@ export class GridCoreCreator {
             { componentName: 'AgHeaderRoot', componentClass: GridHeaderComp },
             { componentName: 'AgSortIndicator', componentClass: SortIndicatorComp },
             { componentName: 'AgPagination', componentClass: PaginationComp },
+            { componentName: 'AgPageSizeSelector', componentClass: OverlayWrapperComponent },
```
StephenCooper, line 350: Is this a copy and paste error with the OverlayWrapperComponent?

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+    private initComponentWrapper(): void {
+        this.wrapper = document.createElement('span');
+        this.wrapper.id =`ag-${this.getCompId()}`;
+        this.wrapper.classList.add('ag-paging-page-size-comp');
+        this.setTemplateFromElement(this.wrapper);
+    }
+
+    private mountPageSizeSelector(): void {
+        const eGui = loadTemplate(this.getTemplate());
+        while (eGui.firstChild) {
```
StephenCooper, line 49: What is this while loop doing?

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+            this.wrapper.appendChild(eGui.firstChild);
+        }
+
+        // Process rendered elements
+        this.setTemplateFromElement(this.wrapper);
+        if (!this.selectPageSize) {
+            throw new Error('ag-Grid: selectPageSize is not available');
```
StephenCooper, line 56: throwing seems a bit extreme, I don't think we have a pattern of throwing but instead I think we can warn / error log and fail graciously from this component. 

Also maybe this should have been handled within the validator.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        return /* html */`<div>` +
+            `<span ref="ePageLabel">${strPage}</span>` +
+            `<ag-select ref="eSelect" class="ag-paging-page-size-select"></ag-select>`
+        + `</div>`;
+    }
+
+    private getPageSizeSelectorValues(): number[] {
+        const paginationParams = this.gridOptionsService.get('paginationParams');
+        return paginationParams?.pageSizeSelectorValues ?? defaultPageSizeSelectorValues;
```
StephenCooper, line 155: Defaults are moving to central place.

grid-community-modules/core/src/ts/pagination/paginationComp.ts
```diff
@@ -41,8 +46,10 @@ export class PaginationComp extends Component {
         const isRtl = this.gridOptionsService.get('enableRtl');
         this.setTemplate(this.getTemplate());
 
-        const { btFirst, btPrevious, btNext, btLast } = this;
+        this.compPageSizeSelector = this.createManagedBean(new PageSizeSelectorComp());
```
StephenCooper, line 49: I wonder if we should conditionally render at this level instead of within the component. Then at least the dom will not have a hanging span that does nothing if the page size selector is not present.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    @RefSelector('eSelect') private selectPageSize: AgSelect | undefined;
+
+    private wrapper: HTMLElement;
+    private paginationParams: {
+        showPageSizeSelector?: boolean;
+        pageSizeSelectorValues?: number[];
+    } | undefined;
+
+    constructor() {
```
gportela85, line 25: Emtpy constructors don't need to call super, just remove the constructor altogether.

However, why are we not creating the template directly as a parameter of the constructor?

```
constructor() {
    super(`
        <div>
            <span ref="ePageLabel"></span>
            <ag-select ref="eSelect" class="ag-paging-page-size-select"></ag-select>
        </div>
    `)
}

A later in the init() method you set the value of this.ePageLabel to `strPage`.

File: grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts [hunk 7]
```diff
@@ -0,0 +1,162 @@
...
+
+    private initComponentWrapper(): void {
+        this.wrapper = document.createElement('span');
+        this.wrapper.id =`ag-${this.getCompId()}`;
+        this.wrapper.classList.add('ag-paging-page-size-comp');
+        this.setTemplateFromElement(this.wrapper);
+    }
+
+    private mountPageSizeSelector(): void {
+        const eGui = loadTemplate(this.getTemplate());
```
gportela85, line 48: We should get rid of this method and use the constructor as mentioned above
Also, `mount` and `unmount` is not a really a name convention we use.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        const pageSizesList = this.getPageSizeSelectorValues();
+        this.selectPageSize.resetOptions(pageSizesList.map(
+            value => ({ value: String(value), text: String(value) })
+        ));
+
+        const currentPageSize = this.paginationProxy.getPageSize();
+        if (!this.selectPageSize || !pageSizesList.includes(currentPageSize)) {
+            throw new Error('ag-Grid: selectPageSize is not included in the list of page sizes');
```
gportela85, line 66: No need to throw errors, if the data is not valid, just set it to a default value and issue a console warning.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+
+            if (newPageSize === this.paginationProxy.getPageSize()) {
+                return;
+            }
+
+            this.paginationProxy.setPageSize(newPageSize);
+        });
+    }
+
+    private unmountPageSizeSelector(): void {
```
gportela85, line 93: `unmount` is not a really a naming convention we use. You can do all this in the `destroy` method.

There is also a `clearElement` helper so you don't have to run this while loop here.

grid-community-modules/core/src/ts/widgets/agList.ts
```diff
@@ -78,6 +78,22 @@ export class AgList extends Component {
         return this;
     }
 
+    public clearOptions(): this {
+        this.options = [];
+        this.itemEls = [];
+        this.getGui().innerHTML = '';
```
gportela85, line 84: don't set innerHTML to = ''
if you need to clear an element, use the `clearElement` helper

grid-community-modules/core/src/ts/widgets/agList.ts
```diff
@@ -78,6 +78,22 @@ export class AgList extends Component {
...
+
+        if (this.value !== null) {
+            this.value = null;
+            this.fireChangeEvent();
+        }
+
+        this.displayValue = null;
+        this.clearHighlighted();
+
```
gportela85, line 93: a method to clear all the options should destroy all listeners that were added in the during the `addOption`.
As creating an AgList component is usually super fast, we tend to simply destroy it and recreate it.

grid-community-modules/core/src/ts/widgets/agSelect.ts
```diff
@@ -88,6 +88,15 @@ export class AgSelect extends AgPickerField<string | null, IPickerFieldParams, A
         return this;
     }
 
+    public resetOptions(options?: ListOption[]): this {
+        this.listComponent!.clearOptions();
+        if (options) {
+            this.listComponent!.addOptions(options);
+        }
+
+        return this;
```
gportela85, line 97: This should be replaced with a method that destroys the current list and creates a new one.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,162 @@
...
+            this.selectPageSize.setValue(String(newPageSize));
+        }
+    }
+
+    private getTemplate(): string {
+        const localeTextFunc = this.localeService.getLocaleTextFunc();
+        const strPage = localeTextFunc('pageSizeShowLabel', 'Show');
+
+        return /* html */`<div>` +
+            `<span ref="ePageLabel">${strPage}</span>` +
```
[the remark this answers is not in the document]
salmenus: removed as it's simpler now.

but in general - that could be done to bypass minifiers that would unnecessary whitespace and `\n` in string

grid-community-modules/core/src/ts/pagination/paginationComp.ts
```diff
@@ -27,6 +28,8 @@ export class PaginationComp extends Component {
     @RefSelector('lbCurrent') private lbCurrent: any;
     @RefSelector('lbTotal') private lbTotal: any;
 
+    @RefSelector('ePageSizeContainer') private ePageSizeContainer: PageSizeSelectorComp;
```
gportela85, line 31: this should be `@RefSelector('pageSizeComp') private pageSizeComp: PageSizeSelectorComp;`

The `e` at the beginning implies it's an HTMLElement.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,137 @@
...
+import {AgSelect} from "../../main";
+import {clearElement} from "../../utils/dom";
+
+export class PageSizeSelectorComp extends Component {
+
+    @Autowired('localeService') protected readonly localeService: LocaleService;
+    @Autowired('gridOptionsService') protected readonly gridOptionsService: GridOptionsService;
+    @Autowired('paginationProxy') private paginationProxy: PaginationProxy;
+
+    private eSelectPageSize: AgSelect | undefined;
```
gportela85, line 15: this should be `selectPageSizeComp`

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,137 @@
...
+            .setLabelAlignment('left')
+            .setAriaLabel(localisedLabel)
+            .onValueChange(() => this.handlePageSizeItemSelected());
+    }
+
+    private getPageSizeSelectorValues(): number[] {
+        const paginationPageSizeSelector = this.gridOptionsService.get('paginationPageSizeSelector');
+        if (Array.isArray(paginationPageSizeSelector)) {
+            return paginationPageSizeSelector;
+        } else {
```
gportela85, line 127: no need for else here

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        this.reloadPageSizesSelector();
+        if (!this.eSelectPageSize) {
+            return;
+        }
+
+        this.getGui().appendChild(this.eSelectPageSize.getGui());
+    }
+
```
gportela85, line 67: show() and hide() are basically the same when `this.isSelectorVisible` is true.
So we can probably combine them into one method

```
toggleSelectPageSizeComp(show: boolean) {
    if (this.isSelectorVisible) { this.reset(); }
    if (!show) { return; }

     this.reloadPageSizesSelector();
     if (!this.eSelectPageSize) { return; } 
     this.appendChild(this.selectPageSizeComp);
}
```

grid-community-modules/core/src/ts/pagination/paginationComp.ts
```diff
@@ -73,6 +76,16 @@ export class PaginationComp extends Component {
         this.setTotalLabels();
     }
 
+    private onPaginationPageSizeSelectorChange(): void {
+        const paginationPageSizeSelector = this.gridOptionsService.get('paginationPageSizeSelector');
+
+        if (!paginationPageSizeSelector) {
```
gportela85, line 82: this can be `this.pageSizeComp.toggleSelectDisplay(paginationPageSizeSelector)`

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (Array.isArray(paginationPageSizeSelector)) {
+            return paginationPageSizeSelector;
+        } else {
+            // Default page sizes when user provides `true` in the config.
+            return [10, 25, 50, 100];
+        }
+    }
+
+    public destroy() {
+        this.hide();
```
gportela85, line 134: this.toggleSelectDisplay(false)

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
salmenus: Good call 👌

PR #7228, opened 2023-11-06
AG-6381 Updated pageSizeSelector logic and documentation

salmenus:


grid-community-modules/core/src/ts/pagination/paginationComp.ts
```diff
@@ -76,9 +83,14 @@ export class PaginationComp extends Component {
...
-    private onPaginationPageSizeSelectorChange(): void {
+    private onPageSizeRelatedOptionsChange(): void {
         const paginationPageSizeSelector = this.gridOptionsService.get('paginationPageSizeSelector');
-        this.pageSizeComp.toggleSelectDisplay(!!paginationPageSizeSelector);
+        const paginationAutoPageSize = this.gridOptionsService.get('paginationAutoPageSize');
+        this.pageSizeComp.toggleSelectDisplay(!!paginationPageSizeSelector && !paginationAutoPageSize);
+    }
+
+    private onPaginationPageSizeChanged(): void {
+        this.paginationProxy.setPageSize(this.gridOptionsService.get('paginationPageSize'));
```
salmenus, line 93: **PS:** 

Instead of directly changing the page size in the selector component, we listen to prop changes - and update the page size here.

This also solves another bug related to `paginationAutoPageSize` ( it wasn't working because the implementation was just updating the `paginationPageSize` option with automatically calculated value), but that value wasn't applied to the UI.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -111,15 +121,44 @@ export class PageSizeSelectorComp extends Component {
...
+
+            return !isNumber || !isPositive;
+        });
+
+        if (hasInvalidValues) {
+            return PageSizeSelectorComp.defaultPageSizeOptions;
         }
 
-        return [10, 25, 50, 100];
+        return paginationPageSizeSelectorValues.sort((a, b) => a - b);
```
salmenus, line 161: Sort by numerical value instead of string value

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -111,15 +121,44 @@ export class PageSizeSelectorComp extends Component {
             .setLabelAlignment('left')
             .setAriaLabel(localisedLabel)
             .onValueChange(() => this.handlePageSizeItemSelected());
+
+        this.hasEmptyOption = shouldAddAndSelectEmptyOption;
```
salmenus, line 125: Following product discussion, we agreed to show an empty option and select it - when the current pageSize is not present in the pageSizeSelector options (regardless of whether they are default values or user-set).

When selects a value from the drop down for the first time, we remove the empty option.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
```diff
@@ -5,14 +5,17 @@ import { GridOptionsService } from "../../gridOptionsService";
 import { PaginationProxy } from "../paginationProxy";
 import { AgSelect } from "../../main";
 import { clearElement } from "../../utils/dom";
+import {warnOnce} from "../../utils/function";
```
gportela85, line 8: spacing

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
gportela85: if this is guaranteed to be a number, we don't need a callback function for sort. If we this is not guaranteed to be a number, then we need
`sort((a, b) => return Number(a) - Number(b))`

grid-community-modules/core/src/ts/pagination/paginationComp.ts
gportela85: set `this.gridOptionsService.get('paginationPageSize')` to a variable before calling `setPageSize`. makes it easier to debug later if necessary

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
salmenus: We still need this, even if it's guaranteed to be a number:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

`[10, 2].sort()` gives `[10, 2]` because default comparaison function converts everything to string.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
salmenus: Well. I thought about it and tried it before pushing.

That `pageSizeOptions.unshift('')` will change the content the `PageSizeSelectorComp.defaultPageSizeOptions` which can be returned by `this.getPageSizeSelectorValues()` .. and that will result defaults with an empty `''` all the time.

That's why I did recreate that array in `pageSizeOptions = ['', ...pageSizeOptions];`

It's a small one, usually just 3 ~ 4 options.
I don't think there is a significant impact on perf.

gportela85: In that case, `this.getPageSizeSelectorValues()` should be returning a copy of the default values, not the default object itself.

grid-community-modules/core/src/ts/pagination/pageSizeSelector/pageSizeSelectorComp.ts
gportela85: oh boy... ok.. let's leave it with the callback function

PR #7236, opened 2023-11-08
Export ColDefField from 'entities/colDef'

Maran23:
This enables developers to import and use the `ColDefField` when using AG Grid Modules.

Example: We currently have the following code in an Angular project:
```
localizeHeader(params: HeaderValueGetterParams<T>): string {
  const colDef: ColDef<T> = (params.colDef as ColDef<T>);
  const field: ColDefField<T> | undefined = colDef.field; // <- 'ColDefField' can not be imported
  ...
}
```
The following import does not work:
```
import {
  ColDefField
} from "@ag-grid-community/core";
```

This looks like an oversight to me. 
Related classes like `NestedFieldPaths` are exported as well, just `ColDefField` is not.

### DIFF
--- a/grid-community-modules/core/src/ts/main.ts
+++ b/grid-community-modules/core/src/ts/main.ts
@@ -394,6 +394,7 @@ export {
     IAggFuncParams,
     ColGroupDef,
     ColDef,
+    ColDefField,
     AbstractColDef,
     ColTypeDef,
     ValueSetterParams,


StephenCooper:
Thanks for pointing this out with an example. Was not considering this use case. I have made the change and pushed this. (There was another update needed too within enterprise.)

This will be resolve in our next release v31.

Maran23:
> Thanks for pointing this out with an example. Was not considering this use case. I have made the change and pushed this. (There was another update needed too within enterprise.)

Thank you for the quick response and for taking care of the rest!

PR #7311, opened 2023-11-30
Fix import error in Readme usage example

irreal:
The example inside readme shows the old method of importing a grid. This PR updates to the current approach of importing createGrid, which is already used in the rest of the example code.


StephenCooper:
Thanks for pointing this out. I have now committed this as well as some other places that had also been missed.

PR #7312, opened 2023-11-30
AG-5928 Added columns to sortChanged event

salmenus:


grid-community-modules/core/src/ts/sortController.ts
```diff
@@ -89,23 +92,29 @@ export class SortController extends BeanStub {
...
         this.eventService.dispatchEvent(event);
     }
 
-    private clearSortBarTheseColumns(columnsToSkip: Column[], source: ColumnEventType): void {
+    private clearSortBarTheseColumns(columnsToSkip: Column[], source: ColumnEventType): Column[] {
+        const clearedColumns: Column[] = [];
         this.columnModel.getPrimaryAndSecondaryAndAutoColumns().forEach((columnToClear: Column) => {
             // Do not clear if either holding shift, or if column in question was clicked
-            if (!columnsToSkip.includes(columnToClear)) {
+            if (!columnsToSkip.includes(columnToClear) && !!columnToClear.getSort()) {
```
salmenus, line 109: On reset columns that already have sort applied

grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -848,9 +848,22 @@ export class GridApi<TData = any> {
     /**
      * Gets the grid to act as if the sort was changed.
      * Useful if you update some values and want to get the grid to reorder them according to the new values.
+     *
+     * @param columns Optional list of columns sorted. This can either be an array of `Column` objects or `colKey`
+     * strings. The list will be passed to the `SortChanged` event.
      */
-    public onSortChanged() {
-        this.sortController.onSortChanged('api');
+    public onSortChanged(columns?: (string | Column)[]) {
```
salmenus, line 855: Users can optionally provide a list of columns when calling `onSortChanged` - which will be passed to the `OnSortChanged` event

grid-community-modules/core/src/ts/sortController.ts
```diff
@@ -50,14 +50,17 @@ export class SortController extends BeanStub {
         const doingMultiSort = (multiSort || this.gridOptionsService.get('alwaysMultiSort')) && !this.gridOptionsService.get('suppressMultiSort');
 
         // clear sort on all columns except those changed, and update the icons
+        const updatedColumns: Column[] = [];
         if (!doingMultiSort) {
-            this.clearSortBarTheseColumns(columnsToUpdate, source);
+            const clearedColumns = this.clearSortBarTheseColumns(columnsToUpdate, source);
+            updatedColumns.push(...clearedColumns);
```
salmenus, line 56: We also include the list of reset columns in the changed columns

grid-community-modules/core/src/ts/gridApi.ts
StephenCooper: I think that is misleading because it looks like the columns passed in will be the only ones to have their sort updated.

grid-community-modules/core/src/ts/events.ts
```diff
@@ -193,6 +193,10 @@ export interface AdvancedFilterBuilderVisibleChangedEvent<TData = any, TContext
 export interface SortChangedEvent<TData = any, TContext = any> extends AgGridEvent<TData, TContext> {
     /** Source of the sort change. */
     source: string;
+    /**
+     * The list of columns impacted by the sort change.
+     */
+    columns?: Column[];
```
StephenCooper, line 199: I am not sure about giving them the entire column. This might be fine but would just want to think about this a bit more. Are there other events that already give a column back?

salmenus: Yep. [Most column events](https://ag-grid.com/javascript-data-grid/column-events/) include array with impacted columns.

grid-community-modules/core/src/ts/sortController.ts
StephenCooper: Move the getSort check to only impact the clearedColumns.

grid-community-modules/core/src/ts/sortController.ts
```diff
@@ -89,23 +92,31 @@ export class SortController extends BeanStub {
...
     }
 
-    public dispatchSortChangedEvents(source: string): void {
+    public dispatchSortChangedEvents(source: string, columns?: Column[]): void {
         const event: WithoutGridCommon<SortChangedEvent> = {
             type: Events.EVENT_SORT_CHANGED,
             source
         };
+
+        if (columns) { event.columns = columns; }
```
StephenCooper, line 101: Nit pick, I would set this on the event directly or am I missing something?

StephenCooper (approved):

grid-community-modules/core/src/ts/sortController.ts
salmenus: * Without `if` check, `event.hasOwnProperty('columns')` still returns `true`   
_( even when `event.columns` is `undefined` )_
* In this case, the event object simply doesn't have `columns` prop.

I personally prefer not to have props at all if they are not needed, rather than setting them to `undefined` or `null`.

In practice: We will have similar outcome (unless we specifically check for `event.hasOwnProperty`) ... And ―in an extreme case―, slightly bigger mem overhead if we a large number of objects with a large number of props that are set to undefined ...

PR #7322, opened 2023-12-05
AG-10054 - fix setting of pivot mode via api and ui together

peterjrreynolds:


grid-enterprise-modules/column-tool-panel/src/columnToolPanel/pivotModePanel.ts
```diff
@@ -38,7 +38,7 @@ export class PivotModePanel extends Component {
     private onBtPivotMode(): void {
         const newValue = !!this.cbPivotMode.getValue();
         if (newValue !== this.columnModel.isPivotMode()) {
-            this.columnModel.setPivotMode(newValue, "toolPanelUi");
+            this.gridOptionsService.updateGridOptions({ options: { pivotMode: newValue} });
```
StephenCooper, line 41: Can we try and preserve the sources as before?

PR #7327, opened 2023-12-07
AG-9871 Updated logic related to row selection

salmenus:


grid-community-modules/core/src/ts/rendering/cell/cellKeyboardListenerFeature.ts
```diff
@@ -164,7 +164,7 @@ export class CellKeyboardListenerFeature extends BeanStub {
     private onSpaceKeyDown(event: KeyboardEvent): void {
         const { gridOptionsService } = this.beans;
 
-        if (!this.cellCtrl.isEditing() && gridOptionsService.isRowSelection()) {
+        if (!this.rowCtrl.hasDisabledRowSelectionCheckbox() && !this.cellCtrl.isEditing() && gridOptionsService.isRowSelection()) {
```
StephenCooper, line 167: I would put this as the last condition as it is the most expensive to calculate. We would want to avoid running this seeing as it loops through every column.

grid-community-modules/core/src/ts/rendering/row/rowCtrl.ts
```diff
@@ -253,6 +253,27 @@ export class RowCtrl extends BeanStub {
...
+
+        if (!colWithCheckboxSelectionFct) {
+            return false;
+        }
+
+        // When 'checkboxSelection' is a function, a checkbox is always rendered,
+        // but when the function returns false, the checkbox is disabled
+        const checkboxSelection = colWithCheckboxSelectionFct.getColDef().checkboxSelection;
+        const params = colWithCheckboxSelectionFct.createColumnFunctionCallbackParams(this.rowNode);
+        return typeof checkboxSelection === 'function' && !checkboxSelection(params);
```
StephenCooper, line 274: Isn't the check for if its a function redundant given the find above?

salmenus: added for Typescript, to avoid type casting.
i can do `(checkboxSelection as CheckboxSelectionCallbackParams)(params)` if you prefer that ?

salmenus: sure. we can cast to `CheckboxSelectionCallback` - done

grid-community-modules/core/src/ts/rendering/cell/cellKeyboardListenerFeature.ts
salmenus: good call. done

grid-community-modules/core/src/ts/selectionService.ts
```diff
@@ -94,14 +94,15 @@ export class SelectionService extends BeanStub implements ISelectionService {
...
+        const isApiSelection = source === 'api';
         let updatedCount = 0;
         for (let i = 0; i < nodes.length; i++) {
             const node = nodes[i];
             // when groupSelectsFiltered, then this node may end up intermediate despite
             // trying to set it to true / false. this group will be calculated further on
             // down when we call calculatedSelectedForAllGroupNodes(). we need to skip it
             // here, otherwise the updatedCount would include it.
-            const skipThisNode = groupSelectsFiltered && node.group;
+            const skipThisNode = !isApiSelection && groupSelectsFiltered && node.group;
```
AndrewGlazier, line 105: I'm struggling to work out why this needs changed - the behaviour we want is that api ignores `selectable` - but this seems to leave the scope of that?

AndrewGlazier (changes requested):
I think we need to consider more than what's being done here.

For example `detailNodes` historically haven't been selectable it seems, which was done by adding `selectable = false` this change means we can now select detail nodes via API, this seems like it'd be an unintended side effect possibly.

When doing transactions (CSRM and SSRM), it seems when we do an update, we deselect selected nodes which aren't selectable any more, this seems like it won't work with this new behaviour, as selectable and selected are less coupled, so this needs considered

It also looks like changes are needed in the serverSideSelectionService too

grid-community-modules/core/src/ts/selectionService.ts
salmenus: Logic updated

grid-community-modules/core/src/ts/entities/rowNode.ts
```diff
@@ -987,12 +987,19 @@ export class RowNode<TData = any> implements IEventEmitter, IRowNode<TData> {
...
+        // We only change in selection for the following scenarios:
+        // 1. when the row is selectable
+        // 2. when selection is done rowNode.setSelected(true/false) - [ but not via gridApi.selectAll() ]
+        // 3. when the user would like to deselect the node (newValue = false) regardless of the node being selectable or not.
+        //    Use case: API call to deselect all nodes should also deselect currently selected nodes that cannot be re-selected.
+        const changeSelectionAllowed = isSelectionViaRowNodeApi || this.selectable || !newValue;
         const selectionNotChanged = this.selected === newValue;
 
-        if (selectionNotAllowed || selectionNotChanged) { return false; }
+        if (this.detail || !changeSelectionAllowed || selectionNotChanged) { return false; }
```
salmenus, line 1002: When `this.detail` is set, `this.selectable` will be false ..

But since we want to override `this.selectable` when the call is made via `rowNode.setSelected(true)`, we will need to add a second check to exclude detail nodes.

PR #7331, opened 2023-12-08
Fix filter in ChangeLog page

sdidier-dev:
Hi AG Grid Team!
As I was taking a look at the [changelogs](https://www.ag-grid.com/changelog/?fixVersion=31.0.0), I noticed that the filter of the `Version` column seems broken, thus this PR to propose a fix.
image

I'm also proposing a modification for the `Issue` column even if it seems fine I'm not sure the comparator works as is.


StephenCooper:
Thank you for reporting this. I have made a few updates to both the changelog and pipeline pages which resolve these issues.

Issue #7291, opened 2023-11-26
AgGrid triggering Angular's Change Detection on almost every event


thekip:
```
[x] bug report => see 'Providing a Reproducible Scenario'
[] feature request => do not use Github for feature requests, see 'Customers of AG Grid'
[] support request => see 'Requesting Community Support'
```

**Providing a Reproducible Scenario**

Open an official example by [link](https://www.ag-grid.com/examples/data-update-row-data/simple-immutable-store/packages/angular/index.html ). I assume it uses latest version of AgGrid. 

* Open Chrome DevTools with installed [Angular Dev Tools](https://chromewebstore.google.com/detail/angular-devtools/ienfalfjdbdpebioblfackkekamfmbnh). 
* Go to Angular Tab -> Profiler. Hit the record button. 
* Start moving a cursor over a grid or prepend rows and start to scroll. 

See the recording: https://take.ms/DFEhr

**Current behavior**
Angular running change detection on many events produced by Ag-Grid. If you inspect the profile, you will see different "Sources" of CD cycle. Such as `setTimeOut`, `mouseOver`, `mouseOut`, `wheel` etc.   

**Expected behavior**
Ag grid should not trigger Angular's change detection on internal events. That cause serious performance issue in big Angular apps. 

**Please tell us about your environment:**

* **AG Grid version:** X.X.X
Latest

* **Browser:**
Chrome 119

thekip:
I also dug deeper in the root cause, and it seems ngZone is triggered by the "global" events (i don't know what that exactly means in the ag-grid internals) 
<img width="932" alt="image" src="https://github.com/ag-grid/ag-grid/assets/1586852/5e588839-9d24-432a-a3d0-28302435caf8">


All these events bypassing the `this.isEmitterUsed(eventType)` check, because they are `global = true`

AG-Zoheil:
Hello @thekip,

Thank you for the suggestion.

We agree this would be a valuable addition to the product and improve performance.

We have added this requirement to our backlog and we are tracking it with the following reference: **AG-10010**

We have no timeline for this at the moment, but this can still be picked up by our team when they're making modifications to this functional area of AG Grid. Please use the functionality currently provided by AG Grid until this is delivered at a later point.

See whether this item will be in the next release by checking the NEXT RELEASE checkbox on the product pipeline page:
https://www.ag-grid.com/ag-grid-pipeline/

[...]

PR #7350, opened 2023-12-14
AG-10010 Improve Angular change detection

StephenCooper:
Currently as reported in https://github.com/ag-grid/ag-grid/issues/7291 change detection is fired too much. 

This is happening because every internal event that gets raised asynchronously was using setTimeout which is patched by Zone.js.

The fix provided here resolves the majority of the issues by running AG Grid outside of zone.js via the method `ngZone.runOutsideAngular` as suggested here. 

https://angular.dev/best-practices/zone-pollution

So now AG Grid will be created out side of angular and any events and property changes will run outside of Angular. This dramatically reduces the amount of change detection and is a good first step. 

However, there are easy ways for an application to still cause lots of change detection. To make this a non breaking change we need to ensure that all user event listeners and callbacks still run within Angular. This is to ensure that if any application updates are made within those functions they get correctly picked up. 

So if the user has implemented `getRowId` this needs to be run inside Angular which means that change detection will be triggered for every row as part of the code that calls `getRowId` will be triggered by a scroll event. 

In a future upgrade we could update things as follows:

- event listeners run within Angular (likely where application updates are triggered)
- callbacks run outside of Angular (unlikely to update the application in a callback should just be for configuring the grid)

So in a major release we could bring callbacks outside of Angular and that would mean minimal change detection cycles. 

**EDIT: ** If the component is setup with onPush then this really does make a big difference now. So we likely will want to recommend that approach for best performance and maybe update our examples to support this.

### DIFF
--- a/grid-community-modules/angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
+++ b/grid-community-modules/angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
@@ -1,77 +1,29 @@
 import {Injectable, NgZone} from "@angular/core";
 import {VanillaFrameworkOverrides} from "@ag-grid-community/core";
-import {AgPromise} from "@ag-grid-community/core";
 
 @Injectable()
 export class AngularFrameworkOverrides extends VanillaFrameworkOverrides {
-    private isEmitterUsed: (eventType: string) => boolean;
-
     constructor(private _ngZone: NgZone) {
         super('angular');
     }
 
-    setEmitterUsedCallback(isEmitterUsed: (eventType: string) => boolean) {
-        this.isEmitterUsed = isEmitterUsed;
-
-    }
-
-    public setTimeout(action: any, timeout?: any): void {
+    dispatchEvent(listener: () => void): void {
+        // Make all events run outside Angular as they often trigger the setup of event listeners
+        // By having the event listeners outside Angular we can avoid triggering change detection
+        // Check for _ngZone existence as it is not present when Zoneless
         if (this._ngZone) {
-            this._ngZone.runOutsideAngular(() => {
-                window.setTimeout(() => {
-                    action();
-                }, timeout);
-            });
+            this._ngZone.runOutsideAngular(listener);
         } else {
-            window.setTimeout(() => {
-                action();
-            }, timeout);
+            listener();
         }
     }
 
-    public setInterval(action: any, interval?: any): AgPromise<number> {
-        return new AgPromise<number>(resolve => {
-            if (this._ngZone) {
-                this._ngZone.runOutsideAngular(() => {
-                    resolve(window.setInterval(() => {
-                            action();
-                        }, interval)
-                    );
-                });
-            } else {
-                resolve(window.setInterval(() => {
-                        action();
-                    }, interval)
-                );
-            }
-        });
-    }
-
-    addEventListener(element: HTMLElement, eventType: string, listener: EventListener | EventListenerObject, useCapture?: boolean): void {
-        if (this.isOutsideAngular(eventType) && this._ngZone) {
-            this._ngZone.runOutsideAngular(() => {
-                element.addEventListener(eventType, listener, useCapture);
-            });
-        } else {
-            element.addEventListener(eventType, listener, useCapture);
-        }
-    }
-
-    dispatchEvent(eventType: string, listener: () => {}, global = false): void {
-        if (this.isOutsideAngular(eventType)) {
-            if (this._ngZone) {
-                this._ngZone.runOutsideAngular(listener);
-            } else {
-                listener();
-            }
-        } else if (this.isEmitterUsed(eventType) || global) {
-            // only trigger off events (and potentially change detection) if actually used
-            if (!NgZone.isInAngularZone() && this._ngZone) {
-                this._ngZone.run(listener);
-            } else {
-                listener();
-            }
-        }
+    /**
+     * Make sure that any code that is executed outside of AG Grid is running within the Angular zone.
+     * This means users can update templates and use binding without having to do anything extra.
+     */
+    wrapOutgoing<T>( callback: () => T): T {
+        return this._ngZone ? this._ngZone.run(callback) : callback();
     }
 
     isFrameworkComponent(comp: any): boolean {
--- a/grid-community-modules/core/src/ts/context/beanStub.ts
+++ b/grid-community-modules/core/src/ts/context/beanStub.ts
@@ -1,6 +1,6 @@
 import { IEventEmitter } from "../interfaces/iEventEmitter";
 import { EventService } from "../eventService";
-import { AgEvent } from "../events";
+import { AgEvent, AgEventListener } from "../events";
 import { Autowired, Context, PreDestroy } from "./context";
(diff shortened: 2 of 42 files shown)


grid-community-modules/angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
```diff
@@ -1,77 +1,29 @@
...
-        if (this.isOutsideAngular(eventType) && this._ngZone) {
-            this._ngZone.runOutsideAngular(() => {
-                element.addEventListener(eventType, listener, useCapture);
-            });
-        } else {
-            element.addEventListener(eventType, listener, useCapture);
-        }
-    }
-
-    dispatchEvent(eventType: string, listener: () => {}, global = false): void {
```
StephenCooper, line 60: We don't need individual handling of setTimeout and setInterval anymore as we run outside of angular at a higher level.

grid-community-modules/core/src/ts/context/beanStub.ts
```diff
@@ -107,12 +107,19 @@ export class BeanStub implements IEventEmitter {
...
         }
 
         const destroyFunc: () => null = () => {
-            object.removeEventListener(event, listener);
+            if (object instanceof HTMLElement || object instanceof Window) {
+                object.removeEventListener(event, listener);
+            } else {
+                object.removeEventListener(event, listener);
+            }
+
```
StephenCooper, line 122: Change only made to support typing cleanly.

grid-community-modules/core/src/ts/events.ts
```diff
@@ -13,6 +13,16 @@ import { RowPinnedType, IRowNode } from './interfaces/iRowNode';
...
+export interface AgEvent {
+    /** Event identifier */
+    type: string;
+}
+
+export interface AgGridEvent<TData = any, TContext = any> extends AgGridCommon<TData, TContext>, AgEvent { }
+
+export type AgEventListener<TData = any,TContext = any> = (event: AgGridEvent<TData, TContext>) => void
+export type AgGlobalEventListener<TData = any, TContext = any> = (eventType: string, event: AgGridEvent<TData, TContext>) => void
+
```
StephenCooper, line 25: Added better types here to help tell the difference as global and single events have different arguements.

grid-community-modules/core/src/ts/gridOptionsService.ts
```diff
@@ -146,7 +148,7 @@ export class GridOptionsService {
                 mergedParams.columnApi = this.columnApi;
                 mergedParams.context = this.context;
 
-                return callback(mergedParams);
+                return this.frameworkOverrides.wrapOutgoing(() => callback(mergedParams));
```
StephenCooper, line 151: Maintain callbacks running within angular. In the future we may not do this wrapping.

grid-community-modules/core/src/ts/gridOptionsService.ts
```diff
@@ -294,7 +296,9 @@ export class GridOptionsService {
 
             const callbackMethodName = ComponentUtil.getCallbackForEvent(eventName);
             if (typeof (this.gridOptions as any)[callbackMethodName] === 'function') {
-                (this.gridOptions as any)[callbackMethodName](event);
+                this.frameworkOverrides.wrapOutgoing(() => {
+                    (this.gridOptions as any)[callbackMethodName](event);
+                })
```
StephenCooper, line 301: Ensure event listeners run within angular.

grid-community-modules/core/src/ts/misc/apiEventService.ts
```diff
@@ -1,14 +1,20 @@
...
+    private syncGlobalEventListeners: Set<AgGlobalEventListener> = new Set();
+    private asyncGlobalEventListeners: Set<AgGlobalEventListener> = new Set();
 
-    public addEventListener(eventType: string, listener: Function): void {
+    public addEventListener(eventType: string, userListener: AgEventListener): void {
+        const listener = (event: any) => {
+            this.getFrameworkOverrides().wrapOutgoing(
+                () => userListener(event),
+            );
+        }
```
StephenCooper, line 17: Ensure listeners registered via the gridApi are run within Angular.

thekip:
Thanks for a good job! 

> So if the user has implemented getRowId this needs to be run inside Angular which means that change detection will be triggered for every row as part of the code that calls getRowId will be triggered by a scroll event.

Is there a way to avoid it on the userland side? I mean if i exactly sure that i don't need a change detection in that place, is there any hook or something what i can use to disable ag grid firing these CD cycles?

StephenCooper:
> Is there a way to avoid it on the userland side? I mean if i exactly sure that i don't need a change detection in that place, is there any hook or something what i can use to disable ag grid firing these CD cycles?

 I am validating this now but I believe having the component setup as `OnPush` will avoid this issue.

thekip:
Well, onPush will not turn off change detection, it just excludes this exact component from the CD tree. So if AG Grid will execute its listeners in the Angular Zone it will trigger `Application.tick`, and the whole components tree would be re-checked. So onPush will help only if all my components would be onPush which is not true.

StephenCooper:
I was hoping that this would work but it does not seem to stop the tick. I think this is where I would want to make callbacks all run outside of Angular to avoid this issue. This however, is potentially breaking so cannot make that change until v32.

```ts
    getRowHeight = () => {
        return this.ngZone.runOutsideAngular(() => {
            return 30;
        });
    }
```

StephenCooper:
This code shows the issue  

```ts
        this.ngZone.runOutsideAngular(() =>

            // Running outside Angular when the setTimeout is run
            setTimeout(() => {
                // Inside the setTimeout we enter Zone
                this.ngZone.run(() =>
                    // Then leave it again but this is still ticked
                    this.ngZone.runOutsideAngular(() => console.log('inside'))
                );
            }, 4000)
        );
```

StephenCooper:
Just to wrap this up, following these changes it looks like instead of tens of CD cycles there will just be one per row created when scrolling. So this might be good enough anyway even if not the optimal solution.

avrahamappel-humi:
Hey folks, I was having an issue where my grid was slowing down due to change detection. I finally stumbled on this thread after two days of banging my head against the wall. Fortunately, switching to `OnPush` seems to have solved it for me.

If using `OnPush` is the current recommended solution for cases like this, can we add something to the docs [here](https://www.ag-grid.com/angular-data-grid/angular-ngzone/) (or at least elaborate a bit more)?

(for reference, my issue was expensive calls to `isExternalFilterPresent` and `doesExternalFilterPass`)

StephenCooper:
> Hey folks, I was having an issue where my grid was slowing down due to change detection. I finally stumbled on this thread after two days of banging my head against the wall. Fortunately, switching to `OnPush` seems to have solved it for me.
> 
> If using `OnPush` is the current recommended solution for cases like this, can we add something to the docs [here](https://www.ag-grid.com/angular-data-grid/angular-ngzone/) (or at least elaborate a bit more)?
> 

@avrahamappel-humi 
1. Is it possible to provide a reproduction for this that shows the difference in OnPush vs not. 
2. Which component did you apply OnPush too?
3. If you add the `debug` flag to `ag-grid-angular` does the logging output in the console change with OnPush and not?

PR #7359, opened 2023-12-18
Reduce use of innerHTML

StephenCooper:
There are some significant performance gains with these changes.

Manually creating the cell and row divs are safe changes so they should definitely land. Done in https://github.com/ag-grid/ag-grid/commit/f7165844ba97d253f7819b75f442d717005c0c37

The second change on how we render standard cells will need more validation and checks but can give a big performance gain.


StephenCooper:
The difference is massive. This is for 400 row buffer to exaggerate the rendering costs but shows how much of an impact this change can make if we can validate all use cases are safe. 

Screenshot 2023-12-19 at 10 48 48
Screenshot 2023-12-19 at 10 48 32

PR #7360, opened 2023-12-18
AG-6075 - Replace AgPromise with Promise

gportela85:
Initial attempt to remove AgPromise from the codebase


gportela85:
@peterjrreynolds I'd like to get this to work, but I probably need some help with the Filter bits, specially around the `isFilterActive` code.

gportela85:
Closing this for now as a complete refactor of the Grid component Factory and Filters will be required.
First task should be get rid of all calls to `resolveNow` and then remove `AgPromise`.

PR #7399, opened 2024-01-05
AG-9939 - Fix changelog versions dropdown

taktran:
Fix changelog versions dropdown not being clickable

## Changes

* Add aria-disabled styles to design system 
  * `aria-disabled` is useful if you want the styles of a disabled button, but still want pointer events to continue working eg, if the button contains other interactive elements.
* Fix changelog versions drop down not being clickable 
  * Use `aria-disabled` instead of `disabled` so that pointer events are still passed through
  * Clean up Collapsible button `color` and `background-color` to use design system buttons more


markdurrant (approved):
LGTM 👍

PR #7406, opened 2024-01-08
AG-10257 - Initial astro prototype

taktran:
WIP


taktran:
Merged to `latest` from `AG-10122/nx-astro-migration` branch

PR #7427, opened 2024-01-11
AG 10287 Add product dropdown

kylerphillips:
CleanShot 2024-01-11 at 16 54 34


markdurrant (approved):
Looking good 👍 I've refactored the component name from `product` to `productDropdown` as we might want to add a more generic dropdown further down the road.

PR #7497, opened 2024-01-24
AG-10453 Suppress col virtualisation in jsdom tests

StephenCooper:


StephenCooper:
Will not merge this and instead go with https://github.com/ag-grid/ag-grid/commit/dc53bbcfdf7a6e8cdf78740a543d395244eb4e32

PR #7514, opened 2024-01-30
AG-9559 Deprecated gridApi.getModel()

salmenus:


grid-community-modules/core/src/ts/gridApi.ts
```diff
@@ -447,6 +447,9 @@ export class GridApi<TData = any> {
      * Returns the row model inside the table.
      * From here you can see the original rows, rows after filter has been applied,
      * rows after aggregation has been applied, and the final set of 'to be displayed' rows.
+     *
+     * @deprecated As of v31.1.0, getModel() is deprecated in favor. You should use methods exposed by the grid API to
```
StephenCooper, line 451: I think this message is a bit confusing, can we reword it.

StephenCooper (changes requested):

grid-community-modules/core/src/ts/gridApi.ts
salmenus: Done

PR #7547, opened 2024-02-05
AG-8968 Radial chart options panels

timkendrick:


grid-enterprise-modules/charts/src/charts/chartComp/services/chartOptionsService.ts
rmc-software, line 1: @timkendrick there seems to be a some duplication here, can you try and reduce this?

PR #7671, opened 2024-03-01
AG-10776 Add Integrated Charts animation panel

timkendrick:


community-modules/core/src/widgets/agInputNumberField.ts
```diff
@@ -2,14 +2,25 @@ import { AgInputTextField, AgInputTextFieldParams } from "./agInputTextField";
...
     private min?: number;
     private max?: number;
 
-    constructor(config?: AgInputTextFieldParams) {
+    constructor(config?: AgInputNumberFieldParams) {
         super(config, 'ag-number-field', 'number');
+        this.precision = config?.precision;
+        this.step = config?.step;
+        this.min = config?.min;
+        this.max = config?.max;
```
peterjrreynolds, line 23: step, min and max only actually get set against the HTML input in the setter methods. Should move this to `postConstruct` and call the setters instead. Also note that `postConstruct` sets a default step against the input.

enterprise-modules/charts/src/charts/chartComp/menu/chartMenuUtils.ts
```diff
@@ -33,6 +33,39 @@ export class ChartMenuUtils extends BeanStub {
...
+                min: options?.min,
+                max: options?.max,
+            },
+            {
+                parseInputValue: value => {
+                    const numberValue = Number(value);
+                    return isNaN(numberValue) ? undefined : numberValue;
+                },
+                formatInputValue: value => {
+                    return value == null ? ',' : `${value}`;
```
peterjrreynolds, line 63: What's the purpose of the comma here? The input should be able to accept null/undefined values.

enterprise-modules/charts/src/charts/chartComp/menu/chartMenuUtils.ts
```diff
@@ -33,6 +33,39 @@ export class ChartMenuUtils extends BeanStub {
...
+            precision?: number,
+            step?: number,
+            min?: number,
+            max?: number,
+        }
+    ): AgInputNumberFieldParams {
+        return this.addValueParams<AgInputNumberFieldParams>(
+            expression,
+            {
+                label: this.chartTranslationService.translate(labelKey ?? 'color'),
```
peterjrreynolds, line 49: Fallback value shouldn't be here - `labelKey` should be mandatory

enterprise-modules/charts/src/charts/chartComp/services/chartTranslationService.ts
```diff
@@ -17,6 +17,7 @@ export class ChartTranslationService extends BeanStub {
         axis: 'Axis',
         radiusAxis: 'Radius Axis',
         navigator: 'Navigator',
+        animation: 'Animation',
```
peterjrreynolds, line 20: These also need adding to `locale.en.js`

community-modules/core/src/widgets/agInputNumberField.ts
timkendrick: thanks, will amend

PR #7705, opened 2024-03-08
AG-10905 Retain chart interactivity options when changing chart type

timkendrick:


enterprise-modules/charts/src/charts/chartComp/utils/object.ts
```diff
@@ -85,6 +85,8 @@ function isSpecial(value: any): boolean {
     return stringValue === '[object RegExp]' || stringValue === '[object Date]';
 }
 
+export function deepMerge<T>(target: T, source: T, options?: any): T;
+export function deepMerge(target: any, source: any, options?: any): any;
```
peterjrreynolds, line 89: Could we have a single type for this?
`deepMerge<TTarget, TSource = TTarget>(target: TTarget, source: TSource, options?: any): TTarget & TSource`?

enterprise-modules/charts/src/charts/chartComp/services/chartOptionsService.ts
```diff
@@ -371,37 +413,78 @@ export class ChartOptionsService extends BeanStub {
...
+    ): T | undefined {
+        // Determine the relevant series type theme override series keys for the current chart
+        const chartSeriesTypes = this.getChartThemeOverridesSeriesTypeKeys();
+
+        // Retrieve the first matching value
+        for (const seriesType of chartSeriesTypes) {
+            const value = this.retrieveChartOptionsSeriesThemeOverride<T>(chartOptions, seriesType, expression);
+            if (value === undefined) continue;
+            return value;
+        }
```
peterjrreynolds, line 427: `chartSeriesType.find`?

enterprise-modules/charts/src/charts/chartComp/utils/object.ts
timkendrick: I was wondering that, but I'm not sure what the interaction of heterogeneous types would be (e.g. `deepMerge([], {})` / `deepMerge({ foo: 'bar' }, { foo: 3 })` and given that that would require some digging through the implementation I reckoned it would be better to leave that to whoever wrote it.

I've left this as an overload that will fall back to the existing `: any` implementation overload, so to me this seemed like a straightforward improvement.

Feel free to amend the commit to add a more specific type 👍

enterprise-modules/charts/src/charts/chartComp/services/chartOptionsService.ts
timkendrick: `chartSeriesType.find` will also match `false` / `0` / `""` though right?

My natural inclination would be to write something like this:

```
return chartSeriesType
  .map(this.retrieveChartOptionsSeriesThemeOverride<T>(chartOptions, seriesType, expression))
  .find((value): value is Exclude<value, undefined> => value !== undefined)
```

...but based on your comments the other day about not using clever code I've been rewriting these more functional forms in terms of straightforward loops

Happy to adopt whatever style, let me know what you prefer here

PR #7710, opened 2024-03-11
AG-10689 Quick start style tweaks

kylerphillips:
**Changes**

- Removed Figma image (now uses DS button)

Screenshot 2024-03-11 at 12 14 11

- Moved theme grid into tab switcher

Screenshot 2024-03-11 at 12 14 25

- Uses real example for theme customisation

Screenshot 2024-03-11 at 12 14 43


markdurrant (approved):
Code all LGTM 👍

PR #7749, opened 2024-03-15
AG-10910 Updated pagination to remove unrelevant warning

salmenus:


community-modules/core/src/pagination/paginationProxy.ts
```diff
@@ -79,6 +79,12 @@ export class PaginationProxy extends BeanStub {
 
     private onPaginationGridOptionChanged(): void {
         this.active = this.gridOptionsService.get('pagination');
+        const autoPageSizeOption = this.gridOptionsService.get('paginationAutoPageSize');
+
+        if (this.active && this.pageSizeAutoCalculated !== undefined && !autoPageSizeOption) {
+            this.unsetAutoCalculatedPageSize();
```
salmenus, line 85: When pagination is disabled, [we set the page size to row count](https://github.com/ag-grid/ag-grid/blob/latest/community-modules/core/src/pagination/paginationProxy.ts#L430) via 'pageSizeAutoCalculated' prop.

Here we unset that, as it's not needed and may cause unnecessary warnings.

gportela85: can you add some comments to the code? It seems odd to have
`this.pageSizeAutoCalculated` and not undefined and `autoPageSizeOption` as false, forcing us to `unsetAutoCalculatedPageSize`.

Would be nice to get a comment in the code explaining what is happening,

PR #7794, opened 2024-03-26
AG-8904 Rename gridOptionsService to gos for bundle reduction

StephenCooper:


afrisgaard:
Hi @StephenCooper :=) We just realized this was released in the newest ag-grid versions.

Any idea why it's not tagging as a breaking change? It did break several trails of calls on the api on our app.

StephenCooper:
This was not marked as a breaking change as it is not part of our public interface. 

The gridOptionsService should not be used externally.

PR #7823, opened 2024-04-08
AG-7519 React state updates subject to memory leak

StephenCooper:
Avoid the following issue and also reduce the amount of work that React needs to do seeing as we are performing our own diff check anyway. 

https://github.com/facebook/react/issues/21706


community-modules/core/src/context/beanStub.ts
```diff
@@ -109,14 +111,17 @@ export class BeanStub implements IEventEmitter {
...
             (object as any).removeEventListener(event, listener);
-
-            this.destroyFunctions = this.destroyFunctions.filter(fn => fn !== destroyFunc);
             return null;
         };
 
         this.destroyFunctions.push(destroyFunc);
 
-        return destroyFunc;
+        return () => {
```
StephenCooper, line 119: Performance improvement by not removing each function from the array in a loop.

community-modules/react/src/reactUi/header/headerRowComp.tsx
```diff
@@ -29,9 +32,13 @@ const HeaderRowComp = (props: {ctrl: HeaderRowCtrl}) => {
...
             setTop: top => setTop(top),
             setHeaderCtrls: (ctrls, forceOrder, afterScroll) =>{
-                agFlushSync(afterScroll, () => {
-                    setCellCtrls(prev => getNextValueIfDifferent(prev, ctrls, forceOrder)!)
-                });
+                prevCellCtrlsRef.current = cellCtrlsRef.current;
+                cellCtrlsRef.current = ctrls;
+
+                const next = getNextValueIfDifferent(prevCellCtrlsRef.current, ctrls, forceOrder)!;
+                if (next !== prevCellCtrlsRef.current) {
```
StephenCooper, line 39: Perform the diff check outside of stateState call so that if there is no difference we don't update the state instead of doing this within React where more code will be run and potentially have objects leaked.

community-modules/react/src/reactUi/utils.tsx
```diff
@@ -134,7 +134,13 @@ export function getNextValueIfDifferent<T extends { getInstanceId: () => string
...
-    // Spread as we need to combine the old and new values
-    return [...oldValues, ...newValues];
+    // Spread as required to combine the old and new values
+    if(oldValues.length === 0) {
+        return newValues;
+    } else if(newValues.length === 0) {
+        return oldValues;
+    }else {
+        return [...oldValues, ...newValues];
+    }
```
StephenCooper, line 144: Bit of additional logic to avoid the cost of useless spreads

PR #7832, opened 2024-04-10
First pass at moving to Beans

StephenCooper:


StephenCooper:
This only looks to reduce by 1.5% so will investigate other approaches.

PR #7881, opened 2024-04-23
AG-9300 Move buttons and popover placement

kylerphillips:


documentation/ag-grid-docs/src/components/theme-builder/components/grid-config/GridConfigDropdown.tsx
```diff
@@ -47,12 +53,20 @@ const GridConfigDropdown = () => {
...
+            stroke="currentColor"
+            stroke-width="2"
```
taktran, line 60: You'll need to reactify these attributes with camel case eg, `stroke-width` -> `strokeWidth`, otherwise, there'll be errors in the console

taktran (approved):
Looks good, just some minor comment.

Also, I wonder why there is a margin for t
Screenshot 2024-04-23 at 2 54 10 pm
he download button, but not for the Grid features button to the popover? The popover kind of overlaps the button

PR #7967, opened 2024-05-14
AG-11304 - Use charts resolved interfaces for `ReferenceDocumentation`

taktran:


taktran:
Not doing this anymore, as it's too much work for grid to use it. May come back to this in future.

PR #7979, opened 2024-05-17
AG 11450 - Remove use of Charts autoSize.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11450

The Charts `autoSize` option is being removed for the 10.0.0 release - these updates should maintain backwards compatibility from a Grid perspective 🤞

### DIFF
--- a/enterprise-modules/charts/src/charts/__snapshots__/chartModelMigration.test.ts.snap
+++ b/enterprise-modules/charts/src/charts/__snapshots__/chartModelMigration.test.ts.snap
@@ -648,126 +648,6 @@ exports[`chartModelMigration upgradeChartModel should upgrade 22.1.0-doughnut-ca
 }
 `;
 
-exports[`chartModelMigration upgradeChartModel should upgrade 22.1.0-doughnut-callout successfully 1`] = `
-{
-  "cellRange": {
-    "columns": [
-      "country",
-      "sugar",
-      "fat",
-      "weight",
-    ],
-    "rowEndIndex": 19,
-    "rowEndPinned": null,
-    "rowStartIndex": 0,
-    "rowStartPinned": null,
-  },
-  "chartId": "id-760p8yfq12q",
-  "chartOptions": {
-    "common": {
-      "legend": {
-        "enabled": true,
-        "position": "right",
-      },
-    },
-    "donut": {
-      "background": {
-        "fill": "white",
-        "visible": true,
-      },
-      "legend": {
-        "enabled": true,
-        "item": {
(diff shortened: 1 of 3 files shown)


enterprise-modules/charts/src/charts/__snapshots__/chartModelMigration.test.ts.snap
```diff
@@ -648,126 +648,6 @@ exports[`chartModelMigration upgradeChartModel should upgrade 22.1.0-doughnut-ca
 }
 `;
 
-exports[`chartModelMigration upgradeChartModel should upgrade 22.1.0-doughnut-callout successfully 1`] = `
```
alantreadway, line 651: Note: Not sure why, but this was a redundant duplicate (see previous snapshot) - `nx test @ag-grid-enterprise/charts -u` removed it correctly.

enterprise-modules/charts/src/charts/chartComp/chartProxies/chartProxy.ts
```diff
@@ -222,6 +222,8 @@ export abstract class ChartProxy<TOptions extends AgChartOptions = AgChartOption
                 time: { crosshair }
             }
         } : {}
+        common.minHeight = 0;
+        common.minWidth = 0;
```
alantreadway, line 226: These are needed to match previous behaviour, and should ensure full backwards compatibility.

There were about 3 examples which showed layout issues without this change due to the popup / container HTML element being smaller than the 'safe' OOB defaults we've added, as a cross-check that this is working.

enterprise-modules/charts/src/charts/chartModelMigration.ts
```diff
@@ -289,6 +290,22 @@ function migrateV31(model: ChartModel) {
...
 }
 
+function migrateV32(model: ChartModel) {
+    model = jsonMutateProperty('chartOptions.*.autoSize', true, model, (parent, targetProp) => {
+        if (parent[targetProp] === true) {
+            // autoSize: true was the OOB default, so just use the new OOB default baked-in.
+        } else if (parent[targetProp] === false) {
+            // Fallback to legacy Charts defaults for autoSize: false.
+            parent['minHeight'] = 600;
+            parent['minWidth'] = 300;
```
alantreadway, line 300: This is a bit of an odd case, if anyone was setting `autoSize: false` before then I'm pretty sure the layout would have been borked, unless they also set `width` and `height` - if they were falling into that case, this mutation should yield no difference.

PR #8002, opened 2024-05-22
AG-11658 Deselection of indeterminate row with tree data

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-11658

TLDR: We couldn't rely on `RowNode.group` for tree data because it didn't get set for rows with user data. This flag was being relied on to determine the selection for group nodes.

### Changes
- Set the `node.group = true` in [GroupStage.addToParent](https://github.com/ag-grid/ag-grid/blob/86ca65867280595de7f21fb94d8c14dd618aa3c7/enterprise-modules/row-grouping/src/rowGrouping/groupStage.ts#L418) to make value of `group` consistent between tree data and normal row grouping.
- Don't remove an empty group node if using tree data and the node has associated tree data, simply mark the `.group` field `false`.
- Replace what I assume is an outdated reference to `calculatedSelectedForAllGroupNodes`
- A little bit of tidying

### Note
The original PR description is posted in a comment below for context on the initial solution (which you can see in [b49928a](https://github.com/ag-grid/ag-grid/pull/8002/commits/b49928a8b2d01008c62dfc242c5f9584cd32f567)).

The initial solution is a simple single line fix, but the current solution means that the semantics of `RowNode.group` are (more?) consistent between tree data and normal row grouping.

### DIFF
--- a/community-modules/core/src/entities/rowNode.ts
+++ b/community-modules/core/src/entities/rowNode.ts
@@ -956,8 +956,8 @@ export class RowNode<TData = any> implements IEventEmitter, IRowNode<TData> {
         return this.__hasChildren;
     }
 
-    public isEmptyRowGroupNode(): boolean | undefined {
-        return this.group && _missingOrEmpty(this.childrenAfterGroup);
+    public isEmptyRowGroupNode(): boolean {
+        return (this.group && _missingOrEmpty(this.childrenAfterGroup)) ?? false;
     }
 
     private dispatchCellChangedEvent(column: AgColumn, newValue: TData, oldValue: TData): void {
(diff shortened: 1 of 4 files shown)


eliasmalik:
Original PR description:

> TLDR: We can't rely on `RowNode.group` for tree data because it doesn't get set for rows with user data. This flag was being relied on to determine the selection for group nodes.

> ### Changes
> - Use `RowNode.hasChildren()` as a fallback/double-check when `RowNode.group` is `false` to handle tree data
> - Replace what I assume is an outdated reference to `calculatedSelectedForAllGroupNodes`

> ### Alternative solution
> The other way to handle this would be to set the `node.group = true` in [GroupStage.addToParent](https://github.com/ag-grid/ag-grid/blob/86ca65867280595de7f21fb94d8c14dd618aa3c7/enterprise-modules/row-grouping/src/rowGrouping/groupStage.ts#L418) to make `group` consistent between tree data and normal row grouping, which strikes me as a better solution, but I haven't established whether or not this breaks anything else yet.

> I'll continue testing before I mark this as ready for review, and may switch to the alternative solution.

PR #8004, opened 2024-05-23
AG-11668 Remove Bean, Autowire, Optional decorators

StephenCooper:


community-modules/core/src/context/context.ts
```diff
@@ -1,71 +1,162 @@
...
         if (!params || !params.beanClasses) {
             return;
         }
 
-        this.contextParams = params;
+        this.gridId = params.gridId;
+        this.beans.context = this;
 
-        this.logger = logger;
-        this.logger.log('>> creating ag-Application Context');
```
StephenCooper, line 56: Can we keep the logging in context?

community-modules/core/src/context/context.ts
```diff
@@ -253,125 +197,34 @@ export class Context {
...
+        ModuleRegistry.__unRegisterGridModules(this.gridId);
     }
 
-    public destroyBean<T extends BaseBean | null | undefined>(bean: T): undefined {
-        if (!bean) {
-            return;
-        }
-
-        this.destroyBeans([bean]);
+    public destroyBean(bean: BaseBean | null | undefined): void {
```
StephenCooper, line 209: Not sure about this change. It was nice being able to destroy and cleanup in a single line. See that some methods now have two steps to destroy the bean and then assign null. What's the motivation for this?

community-modules/core/src/ctrlsService.ts
```diff
@@ -47,9 +47,8 @@ interface ReadyParams {
 
 type CtrlType = keyof ReadyParams;
 
-@Bean(CtrlsService.NAME)
 export class CtrlsService extends BeanStub {
-    public static readonly NAME = 'ctrlsService';
```
StephenCooper, line 52: What was this NAME property? Guessing not used for something else?

community-modules/core/src/filter/provided/date/dateCompWrapper.ts
```diff
@@ -61,7 +61,8 @@ export class DateCompWrapper {
 
     public destroy(): void {
         this.alive = false;
-        this.dateComp = this.context.destroyBean(this.dateComp);
+        this.context.destroyBean(this.dateComp);
+        this.dateComp = undefined;
```
StephenCooper, line 65: This pattern seems less effecient.

community-modules/core/src/headerRendering/rowContainer/headerRowContainerCtrl.ts
```diff
@@ -97,7 +107,8 @@ export class HeaderRowContainerCtrl extends BeanStub {
         const refreshColumnGroups = () => {
             const groupRowCount = this.columnModel.getHeaderRowCount() - 1;
 
-            this.groupsRowCtrls = this.destroyBeans(this.groupsRowCtrls);
+            this.destroyBeans(this.groupsRowCtrls);
```
StephenCooper, line 110: Same thing here.

community-modules/core/src/rendering/cell/cellComp.ts
```diff
@@ -296,7 +305,8 @@ export class CellComp extends Component implements TooltipParentComp {
 
     private destroyRenderer(): void {
         const { context } = this.beans;
-        this.cellRenderer = context.destroyBean(this.cellRenderer);
+        this.cellRenderer = undefined;
```
StephenCooper, line 308: These are the wrong way round.

community-modules/core/src/ctrlsService.ts
AndrewGlazier: This was used in several places like the `agGridReactUi` comp as a form of type safety for getting the bean name, I added type safety to `getBean` so it was no longer required.

PR #8009, opened 2024-05-24
AG-9454 - introduce grid api interface

peterjrreynolds:


enterprise-modules/column-tool-panel/src/columnToolPanel/pivotModePanel.ts
```diff
@@ -33,10 +33,7 @@ export class PivotModePanel extends Component {
         const newValue = !!this.cbPivotMode.getValue();
         if (newValue !== this.columnModel.isPivotMode()) {
             this.gos.updateGridOptions({ options: { pivotMode: newValue }, source: 'toolPanelUi' as any });
-            const { api } = this;
-            if (api) {
-                api.refreshHeader();
-            }
+            this.ctrlsService.getHeaderRowContainerCtrls().forEach((c) => c.refresh());
```
StephenCooper, line 36: We might need the framework wrapping in this call stack. Maybe could be done around the pivot button. Otherwise Angular might trigger lots of change detection. Might be worth checking this.

PR #8020, opened 2024-05-29
AG-11711 Trim down BeanStub

StephenCooper:
Main changes:

- `context` is now private and removed `getContext`. Those classes that need the context can wire themselves. 
- `gridId` is now on BeanStub as that was the main use case of `getContext`
- `environment ` removed and wired when required

Future work could include removing frameworkOverrides but that might need some more care so leaving out of these changes.


enterprise-modules/master-detail/src/masterDetail/detailCellRenderer.ts
```diff
@@ -26,10 +28,13 @@ export class DetailCellRenderer extends Component implements ICellRenderer {
...
 
     private detailApi: GridApi;
-
     private params: IDetailCellRendererParams;
-
     private ctrl: DetailCellRendererCtrl;
+    private context: Context;
+
+    public wireBeans(beans: BeanCollection): void {
+        this.context = beans.context;
```
peterjrreynolds, line 36: missing super call

community-modules/core/src/gridOptionsService.ts
```diff
@@ -152,6 +155,78 @@ export class GridOptionsService extends BeanStub {
         return _exists(this.gridOptions[property]);
     }
 
+    private setupGridOptionListener<K extends keyof GridOptions>(
+        event: keyof GridOptions,
+        listener: PropertyValueChangedListener<K>
+    ): () => null {
+        this.gos.addPropertyEventListener(event, listener);
```
peterjrreynolds, line 162: Can remove the `.gos`

community-modules/core/src/widgets/component.ts
```diff
@@ -91,7 +91,7 @@ export class Component extends BeanStub implements ComponentBean {
         // ui exists if user sets template in constructor. when this happens,
         // We have to wait for the context to be autoWired first before we can create child components.
         this.agStackComponentsRegistry?.ensureRegistered(this.components);
-        if (element && this.getContext()) {
```
StephenCooper, line 94: Context is not actually being used here, but instead is just a proxy for auto wiring having run. We can tell this from the presence of the existing `agStackComponentsRegistry`

enterprise-modules/master-detail/src/masterDetail/detailCellRenderer.ts
StephenCooper: fixed.

PR #8023, opened 2024-05-29
AG-7026 Allow getRowId to be called for pinned rows

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-7026

### Changes
- Call `getRowId` if it's defined when setting the `RowNode.id` of pinned rows in the `pinnedRowModel`
  - I don't think we ever rely on pinned rows having a specific prefix, so I think this is a safe change.
- Add recycling logic to refresh checks in `RowRenderer` specifically for pinned rows. This addresses TC2 in the issue
- Tidying:
  - Renames `PinnedRowModel.getPinned[Top|Bottom]RowData` to `PinnedRowModel.getPinned[Top|Bottom]RowNodes` to better reflect the signature
  - Removes some type annotations that are inferred

This should probably be merged after #8022 and also then be updated to use the new `getRowIdCallback` method that it introduces.

### Question
- What is the reason for `RowCtrl.rowId` being an HTML escaped version of `RowNode.id`?

### DIFF
--- a/community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
+++ b/community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
@@ -114,13 +114,18 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
     private createNodesFromData(allData: any[] | undefined, isTop: boolean): RowNode[] {
         const rowNodes: RowNode[] = [];
         if (allData) {
+            const getRowId = this.gos.get('getRowId');
+            const goContext = this.gos.get('context');
+            const idPrefix = isTop ? RowNode.ID_PREFIX_TOP_PINNED : RowNode.ID_PREFIX_BOTTOM_PINNED;
+
             let nextRowTop = 0;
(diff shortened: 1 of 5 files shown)


community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -114,13 +114,18 @@ export class PinnedRowModel extends BeanStub {
...
+
             let nextRowTop = 0;
             allData.forEach((dataItem: any, index: number) => {
                 const rowNode = new RowNode(this.beans);
                 rowNode.data = dataItem;
 
-                const idPrefix = isTop ? RowNode.ID_PREFIX_TOP_PINNED : RowNode.ID_PREFIX_BOTTOM_PINNED;
-                rowNode.id = idPrefix + index;
+                rowNode.id =
+                    getRowId?.({ data: dataItem, level: 0, api: this.beans.gridApi, context: goContext }) ??
```
eliasmalik, line 127: Also I think `level: 0` is fine here because we can't group on pinned rows?

PR #8029, opened 2024-05-30
AG-11717 - filter bundle size

peterjrreynolds:


community-modules/core/src/components/framework/userComponentRegistry.ts
```diff
@@ -38,20 +30,11 @@ export class UserComponentRegistry extends BeanStub implements NamedBean {
     beanName = 'userComponentRegistry' as const;
 
     private agGridDefaults: { [key: string]: any } = {
-        //date
-        agDateInput: DefaultDateComponent,
```
StephenCooper, line 42: Is this only used when filtering? Does it not get used in editing?

community-modules/core/src/filter/provided/simpleFilterOptions.ts
```diff
@@ -0,0 +1,18 @@
+import type { ISimpleFilterModelType } from './iSimpleFilter';
+
+export class SimpleFilterOptions {
+    public static EMPTY: ISimpleFilterModelType = 'empty';
```
StephenCooper, line 4: Do we need this as an actual class or could this simply become a type and all the places this is used just use string values but in a type safe way?

community-modules/core/src/filter/provided/text/textFilter.ts
```diff
@@ -1,145 +1,39 @@
...
-        SimpleFilter.STARTS_WITH,
-        SimpleFilter.ENDS_WITH,
-        SimpleFilter.BLANK,
-        SimpleFilter.NOT_BLANK,
-    ];
-
-    static DEFAULT_FORMATTER: TextFormatter = (from: string) => from;
-
-    static DEFAULT_LOWERCASE_FORMATTER: TextFormatter = (from: string) =>
+    private static DEFAULT_FORMATTER: TextFormatter = (from: string) => from;
```
StephenCooper, line 15: Do these need to be static anymore seeing as they are private?

community-modules/core/src/filter/quickFilterConstants.ts
```diff
@@ -0,0 +1 @@
+export const EVENT_QUICK_FILTER_CHANGED = 'quickFilterChanged' as const;
```
StephenCooper, line 1: Is the motivation here to to able to include this const without then including an import path to the quick filter service? If that is the case we should evaluate if it does break treeshaking having it listed in a common file with the quick filter service.

If it does then we are vulnerable to tree shaking being broken in future releases.

community-modules/core/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -294,9 +294,17 @@ export class HeaderRowCtrl extends BeanStub {
...
-                case HeaderRowType.FLOATING_FILTER:
-                    headerCtrl = this.createBean(new HeaderFilterCellCtrl(headerColumn as AgColumn, this.beans, this));
+                case HeaderRowType.FLOATING_FILTER: {
+                    headerCtrl = this.createBean(
+                        this.beans.ctrlsFactory.getInstance<HeaderFilterCellCtrl>(
+                            'headerFilterCell',
+                            headerColumn as AgColumn,
+                            this.beans,
+                            this
+                        )!
```
StephenCooper, line 304: So this is the new pattern that will enable a class like HeaderFilterCellCtrl to be tree shaken even though it is constructed in-line with arguments.

I am guessing this would fail if the filter module has not been provided? Or the type would never hit this code path?

enterprise-modules/row-grouping/src/main.ts
```diff
@@ -1,4 +1,4 @@
-export { RowGroupingModule } from './rowGroupingModule';
+export * from './rowGroupingModule';
```
StephenCooper, line 1: Is this intentional? If it is what is this for?

enterprise-modules/row-grouping/src/rowGroupingModule.ts
```diff
@@ -13,14 +13,30 @@ import { PivotColDefService } from './rowGrouping/pivotColDefService';
...
+};
+
+export const GroupFloatingFilterModule: Module = {
+    version: VERSION,
+    moduleName: '@ag-grid-enterprise/group-floating-filter',
+    userComponents: [{ componentName: 'agGroupColumnFloatingFilter', componentClass: GroupFloatingFilterComp }],
+    dependantModules: [GroupFilterModule, FloatingFilterModule],
+};
+
+export const RowGroupingModule: Module = {
```
StephenCooper, line 38: This is an interesting idea, building out on top of the existing module setup.

community-modules/core/src/components/framework/userComponentRegistry.ts
peterjrreynolds: It is only used for filters

community-modules/core/src/headerRendering/row/headerRowCtrl.ts
peterjrreynolds: It never hits this code path (floating filter headers can only be created if `filterManager.hasFloatingFilters()` returns true)

enterprise-modules/row-grouping/src/main.ts
peterjrreynolds: It was intentional. We can use named exports, but why we would be exporting stuff from there that aren't modules?

community-modules/core/src/filter/quickFilterConstants.ts
peterjrreynolds: If they are in the same file, then `QuickFilterService` gets bundled based on the const being imported.

community-modules/core/src/filter/provided/text/textFilter.ts
peterjrreynolds: Done

community-modules/core/src/filter/provided/simpleFilterOptions.ts
peterjrreynolds: Done

enterprise-modules/row-grouping/src/main.ts
StephenCooper: We might want to stick to named imports for consistency. It might not impact things like packages, (seeing as this is specific module code) but that would also be potential concern if this pattern is used elsewhere.

community-modules/core/src/filter/quickFilterConstants.ts
StephenCooper: I wonder if this happens within an actual application. I get why it would keep the code as part of our bundled file, which is what you were validating against?

Let me try this against my test app.

enterprise-modules/row-grouping/src/main.ts
peterjrreynolds: We already have this pattern elsewhere, but I have updated this and the others in this PR to named exports.

community-modules/core/src/filter/quickFilterConstants.ts
peterjrreynolds: I was testing against our built bundle. I have moved the const back and removed the private static property.

PR #8033, opened 2024-05-31
AG-11731 - editing bundle size

peterjrreynolds:


community-modules/core/src/edit/editModules.ts
```diff
@@ -0,0 +1,77 @@
...
+export const EditCoreModule: Module = {
+    version: VERSION,
+    moduleName: '@ag-grid-community/edit-core',
+    beans: [EditService],
+};
+
+export const UndoRedoEditModule: Module = {
+    version: VERSION,
+    moduleName: '@ag-grid-community/undo-redo-edit',
+    beans: [UndoRedoService],
```
StephenCooper, line 23: Does this not need to depend on the EditCoreModule?

community-modules/core/src/edit/cellEditors/largeTextCellEditor.ts
```diff
@@ -1,27 +1,10 @@
...
-    /**
-     * Number of character columns to display.
-     * @default 60
-     */
-    cols: number;
-}
+import type { ILargeTextEditorParams } from './iLargeTextCellEditor';
 
 export class LargeTextCellEditor extends PopupComponent implements ICellEditorComp {
     private static TEMPLATE /* html */ = `<div class="ag-large-text">
```
StephenCooper, line 10: I am checking if these static templates prevent tree shaking.

community-modules/core/src/edit/editModules.ts
peterjrreynolds: Technically no, but I've added it for consistency

community-modules/core/src/edit/cellEditors/largeTextCellEditor.ts
StephenCooper: This does. I have pushed a change to inline the template for this one. All other editors already do not have the static templates.

PR #8054, opened 2024-06-06
AG-11783 Remove EventKeys consts with Typed events

StephenCooper:
1. Removed all const `Events.EVENT_NAME`, for grid, column and row
2. Introduced `AgEventType` which is a typed version of all the global events used via the global EventService
3. Every Event interface is locked into a specific `AgEventType`
4. Remove individual component local static event names
5. Typed all local event services via component / service level event types 
6. BeanStub addManagedListeners now has different versions for global events, Dom elements, grid services / components


community-modules/angular/updateGridAndColumnProperties.js
```diff
@@ -203,7 +203,7 @@ function getGridPropertiesAndEventsJs() {
 const updateGridProperties = (getGridPropertiesAndEvents) => {
     // extract the grid properties & events and add them to our angular grid component
     const { code: gridPropertiesAndEvents, types } = getGridPropertiesAndEvents();
-    const importsForProps = `import {${EOL}    ${types.join(',' + EOL + '    ')}${EOL}} from "@ag-grid-community/core";`;
+    const importsForProps = `import type {${EOL}    ${types.join(',' + EOL + '    ')}${EOL}} from "@ag-grid-community/core";`;
```
StephenCooper, line 206: Updated Angular generation to use type imports

community-modules/core/src/context/beanStub.ts
```diff
@@ -76,48 +77,72 @@ export abstract class BeanStub<TLocalEvent extends string = string>
...
+    ) {
+        return this._setupListeners<keyof HTMLElementEventMap>(object, handlers);
+    }
+    public addManagedEventListeners(handlers: EventHandlers<AgEventType>) {
+        return this._setupListeners<AgEventType>(this.eventService, handlers);
+    }
+    public addManagedListeners<TEvent extends string>(object: IEventEmitter<TEvent>, handlers: EventHandlers<TEvent>) {
+        return this._setupListeners<TEvent>(object, handlers);
+    }
+
```
StephenCooper, line 124: So now have separate managed setup functions for

1. Elements - so we can type those accurately
2. EventListeners - so can reduce duplication and enforce strict types for global events
3. Other event emitters

community-modules/core/src/interfaces/iRowNode.ts
```diff
@@ -39,24 +39,41 @@ export interface SetSelectedParams {
     source: SelectionEventSourceType;
 }
 
-export interface RowNodeEvent<TData = any> extends AgEvent {
-    /** Event identifier */
-    type: RowNodeEventType;
+export interface RowNodeEvent<T extends RowNodeEventType, TData = any> extends AgEvent<T> {
```
StephenCooper, line 42: Adding all the sub interfaces as users may want them.

enterprise-modules/row-grouping/src/rowGrouping/groupFilter/groupFilter.ts
```diff
@@ -273,11 +268,8 @@ export class GroupFilter extends TabGuardComp implements IFilterComp {
 
     private onColumnRowGroupChanged(): void {
         this.updateGroups().then(() => {
-            this.dispatchEvent({
-                type: GroupFilter.EVENT_COLUMN_ROW_GROUP_CHANGED,
-            });
-            this.eventService.dispatchEvent({
```
StephenCooper, line 279: `filterAllowedUpdated` is not used anywhere and is not public so it has been removed.

community-modules/core/src/gridApiService.ts
```diff
@@ -911,23 +912,23 @@ export class GridApiService<TData = any> extends BeanStub implements GridApi, Na
         return value;
     }
 
-    public override addEventListener(eventType: string, listener: (...args: any[]) => any): void {
+    public override addEventListener(eventType: AgPublicEventType, listener: (...args: any[]) => any): void {
```
peterjrreynolds, line 915: Should this also be updated on the interface?

community-modules/core/src/gridApiService.ts
```diff
@@ -942,10 +943,10 @@ export class GridApiService<TData = any> extends BeanStub implements GridApi, Na
         }
 
         const event: WithoutGridCommon<GridPreDestroyedEvent<TData>> = {
-            type: Events.EVENT_GRID_PRE_DESTROYED,
+            type: 'gridPreDestroyed',
             state: this.getState(),
         };
-        this.dispatchEvent(event);
+        this.dispatchLocalEvent(event);
```
peterjrreynolds, line 949: This shouldn't be local. `GridApiService` overrides `dispatchEvent`

community-modules/core/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -94,25 +92,26 @@ export class HeaderRowCtrl extends BeanStub {
...
-        this.addManagedPropertyListener('headerHeight', this.onRowHeightChanged.bind(this));
-        this.addManagedPropertyListener('pivotHeaderHeight', this.onRowHeightChanged.bind(this));
-        this.addManagedPropertyListener('groupHeaderHeight', this.onRowHeightChanged.bind(this));
-        this.addManagedPropertyListener('pivotGroupHeaderHeight', this.onRowHeightChanged.bind(this));
-        this.addManagedPropertyListener('floatingFiltersHeight', this.onRowHeightChanged.bind(this));
+        this.addManagedPropertyListener('headerHeight', onHeightChanged);
+        this.addManagedPropertyListener('pivotHeaderHeight', onHeightChanged);
+        this.addManagedPropertyListener('groupHeaderHeight', onHeightChanged);
+        this.addManagedPropertyListener('pivotGroupHeaderHeight', onHeightChanged);
+        this.addManagedPropertyListener('floatingFiltersHeight', onHeightChanged);
```
peterjrreynolds, line 114: Can't this use `addManagedPropertyListeners` now?

community-modules/core/src/gridApiService.ts
StephenCooper: yes it should be. We still have more improvements to do in another ticket for these types.

PR #8057, opened 2024-06-06
AG-11780 break up grid api

peterjrreynolds:


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModelApi.ts
```diff
@@ -0,0 +1,74 @@
...
+export function forEachNodeAfterFilterAndSort<TData = any>(
+    beans: BeanCollection,
+    callback: (rowNode: IRowNode<TData>, index: number) => void
+): void {
+    beans.rowModelHelperService?.getClientSideRowModel()?.forEachNodeAfterFilterAndSort(callback);
+}
+
+export function resetRowHeights(beans: BeanCollection): void {
+    if (beans.columnModel.isAutoRowHeightActive()) {
+        _warnOnce('calling gridApi.resetRowHeights() makes no sense when using Auto Row Height.');
```
StephenCooper, line 41: I think this warning is a bit over the top. We should just return if its not safe to call the method and probably not worry about telling the user this.

community-modules/core/src/api/gridApi.test.ts
```diff
@@ -0,0 +1,149 @@
...
+//     age: number;
+// }
+
+describe('dummyTest', () => {
+    it('always true', () => {
+        expect('hello').toBe('hello');
+    });
+});
+
+// describe('GridApi setGridOption, updateGridOptions types', () => {
```
StephenCooper, line 17: So yeah, these tests were to make sure about the typing. Not too worried if we drop them now as its unlikely we are making changes there again.

community-modules/core/src/validation/apiFunctionValidator.ts
```diff
@@ -0,0 +1,281 @@
+import type { ApiFunction, ApiFunctionName } from '../api/iApiFunction';
+import type { BeanCollection } from '../context/context';
+import type { RowModelType } from '../interfaces/iRowModel';
+import { ModuleNames } from '../modules/moduleNames';
+import { ModuleRegistry } from '../modules/moduleRegistry';
+import { _warnOnce } from '../utils/function';
+
+const functionModules: { [name in ApiFunctionName]?: ModuleNames } = {
```
StephenCooper, line 8: Can we invert these maps when defining them to remove all the duplication? The current format is not bundle size friendly with all those repeated strings.

StephenCooper: Although this will be less of an issue when we actually make it possible to drop validations.

community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModelApi.ts
peterjrreynolds: This was an existing warning added intentionally under AG-6161

community-modules/core/src/validation/apiFunctionValidator.ts
peterjrreynolds: Can't invert due to the lookups, but I have come up with an alternative solution that works

PR #8066, opened 2024-06-10
AG-11758 Reduce use of static

StephenCooper:
Static properties and methods work against tree shaking so we want to reduce our use of them where we can.

This PR is the start of that work, along with some general cleanup to reduce duplication in some of our components. 

General approach to these changes:

- private static variables moved to plain consts outside of the class. (means names will be mangled)
- TEMPLATES are either inlined or just moved above the class
- grid options coercion is now just standalone config and functions 
- removed wrapper classes around purely public static methods
- combined RowContainer options into a single map to reduce duplication across many methods


community-modules/core/src/gridOptionsService.ts
```diff
@@ -80,6 +80,44 @@ export interface PropertyValueChangedEvent<K extends keyof GridOptions> extends
 export type PropertyChangedListener = (event: PropertyChangedEvent) => void;
 export type PropertyValueChangedListener<K extends keyof GridOptions> = (event: PropertyValueChangedEvent<K>) => void;
 
+/**
+ * Handles value coercion including validation of ranges etc. If value is invalid, undefined is set, allowing default to be used.
+ */
+const PROPERTY_COERCIONS: Map<keyof GridOptions, (value: any) => GridOptions[keyof GridOptions]> = new Map([
```
peterjrreynolds, line 86: Didn't we discuss removing these this release?

community-modules/react/src/reactUi/rows/rowContainerComp.tsx
```diff
@@ -11,7 +11,7 @@ const RowContainerComp = (params: { name: RowContainerName }) => {
     const { context } = useContext(BeansContext);
 
     const { name } = params;
-    const containerType = useMemo(() => getRowContainerTypeForName(name), [name]);
+    const containerOptions = getRowContainerOptions(name);
```
peterjrreynolds, line 14: Should this be memoized?

community-modules/core/src/main.ts
```diff
@@ -263,13 +264,14 @@ export {
     IRowContainerComp,
     RowContainerCtrl,
     RowContainerType,
-    getRowContainerTypeForName,
+    RowContainerOptions,
+    getRowContainerOptions,
```
peterjrreynolds, line 268: Underscore prefix?

community-modules/core/src/main.ts
```diff
@@ -263,13 +264,14 @@ export {
...
 } from './gridBodyComp/rowContainer/rowContainerCtrl';
 
 // headerRendering
 export { BodyDropPivotTarget } from './headerRendering/columnDrag/bodyDropPivotTarget';
 export { BodyDropTarget } from './headerRendering/columnDrag/bodyDropTarget';
-export { CssClassApplier } from './headerRendering/cells/cssClassApplier';
+export { getHeaderClassesFromColDef, getToolPanelClassesFromColDef } from './headerRendering/cells/cssClassApplier';
```
peterjrreynolds, line 274: Underscore prefix?

community-modules/core/src/gridOptionsService.ts
StephenCooper: Yes, vaguely remember this so have raised https://ag-grid.atlassian.net/browse/AG-11845 for that as a separate item.

community-modules/react/src/reactUi/rows/rowContainerComp.tsx
StephenCooper: Well it will return the same object reference everytime it is called from the map so not technically required to. Maybe safer to leave the useMemo in in case of a change in the implementation of getRowContainerOptions that means this isn't true.

PR #8068, opened 2024-06-10
AG-11846 replace static selectors

StephenCooper:
To remove the static `selector` properties from components have had to introduce sibling selector objects which are used to map a component class to a selector such as:

```ts
export const GridHeaderSelector: ComponentSelector = {
    selector: 'AG-HEADER-ROOT',
    Component: GridHeaderComp,
};
```

This unblocks tree shaking but still gives us locality of selector definition next to the component that it represents. 

With this change have removed the agStackRegistry and instead fully rely on component selectors being supplied for the templates that they are used within. I feel this is a simpler mental model and should avoid bugs where a component was working because something else registered a component it was depending on.


community-modules/core/src/widgets/agRadioButton.ts
```diff
@@ -1,13 +1,10 @@
 import type { CheckboxChangedEvent } from '../events';
 import type { AgCheckboxParams } from '../interfaces/agFieldParams';
 import { AgCheckbox } from './agCheckbox';
-import type { AgComponentSelector } from './component';
 
 export interface AgRadioButtonParams extends AgCheckboxParams {}
 
 export class AgRadioButton extends AgCheckbox<AgRadioButtonParams> {
-    static override selector: AgComponentSelector = 'AG-RADIO-BUTTON';
```
StephenCooper, line 9: The radio button is never used in a component so it does not require a selector.

community-modules/core/src/widgets/component.ts
```diff
@@ -402,23 +403,20 @@ export class Component<TLocalEvent extends string = ComponentEvent>
     }
 }
 
+/** All the AG Grid components that are used within internal templates via <ag-autocomplete> syntax */
 export type AgComponentSelector =
-    | 'AG-ADVANCED-FILTER'
```
StephenCooper, line 406: Removed all components that do not actually use a selector.

PR #8076, opened 2024-06-12
AG-11861 Clear beans and component refs automatically

StephenCooper:
Components take references to either other components or HTML elements via the `data-ref` selector. This PR adds destroy functions to set those to `undefined` as part of the destruction logic so that we don't keep references to any DOM nodes via the `data-refs`.


community-modules/core/src/rendering/rowRenderer.ts
```diff
@@ -954,13 +954,13 @@ export class RowRenderer extends BeanStub implements NamedBean {
     }
 
     public override destroy(): void {
-        this.removeAllRowComps();
+        this.removeAllRowComps(true);
```
StephenCooper, line 957: We should not spend time animating rows when the entire row renderer is being destroyed.

community-modules/core/src/widgets/component.ts
```diff
@@ -236,6 +238,7 @@ export class Component<TLocalEvent extends string = ComponentEvent>
             newComponent.setParentComponent(this as Component);
 
             this.createBean(newComponent, null, afterPreCreateCallback);
+            this.addDestroyFunc(this.destroyBean.bind(this, newComponent));
```
StephenCooper, line 241: Moved this to ensure it is the first destroy function

PR #8085, opened 2024-06-13
AG-10420 event listener typings

StephenCooper:
Improve the event typings so that based on the event type the event parameters will be correctly inferred. 

```ts
// params is now correctly inferred as RowDataUpatedEvent
this.addEventListener('rowDataUpdated', (params) => {})

this.addGlobalEventListerner((eventType, event) => {
  if(event.type === 'rowDataUpdated'){
      // event is inferred as RowDataUpdatedEvent 
  }
})
```
This does make me wonder if we should go ahead and remove the `eventType` from the global event listener seeing as it does not add any value and it does not work with inferring.


community-modules/core/src/context/beanStub.ts
```diff
@@ -82,16 +85,18 @@ export abstract class BeanStub<TEventType extends string = BeanStubEvent>
         this.dispatchLocalEvent({ type: 'destroyed' } as { type: BeanStubEvent } as any);
     }
 
+    // The typing of AgEventListener<any, any, any> is not ideal, but it's the best we can do at the moment to enable
+    // eventService to have the best typing at the expense of BeanStub local events
     /** Add a local event listener against this BeanStub */
-    public addEventListener<T extends TEventType>(eventType: T, listener: AgEventListener<any, any, T>): void {
+    public addEventListener<T extends TEventType>(eventType: T, listener: AgEventListener<any, any, any>): void {
```
StephenCooper, line 91: The `T` was in the wrong place, that was the TConext value. Removing and setting the event type to any as a best step. 

Full fix will be to make event param maps compulsory across all local event listeners.

community-modules/core/src/eventTypes.ts
```diff
@@ -137,7 +137,6 @@ export const INTERNAL_EVENTS = [
     'alignedGridScroll',
     'alignedGridColumn',
     'gridOptionsChanged',
-    'gridStylesChanged',
```
StephenCooper, line 140: Value was duplicated.

community-modules/core/src/events.ts
```diff
@@ -2,19 +2,160 @@ import type { AgChartThemeOverrides } from 'ag-charts-types';
...
+import type { BuildEventTypeMap } from './interfaces/iEventEmitter';
 import type { IFilterComp } from './interfaces/iFilter';
 import type { IRowNode, RowPinnedType } from './interfaces/iRowNode';
 import type { RowNodeTransaction } from './interfaces/rowNodeTransaction';
 import type { ServerSideTransactionResult } from './interfaces/serverSideTransaction';
 
-export const ALWAYS_SYNC_GLOBAL_EVENTS: Set<string> = new Set(['gridPreDestroyed', 'fillStart', 'pasteStart']);
+export const ALWAYS_SYNC_GLOBAL_EVENTS: Set<AgEventType> = new Set(['gridPreDestroyed', 'fillStart', 'pasteStart']);
+
+export type AgEventTypeParams<TData = any, TContext = any> = BuildEventTypeMap<
```
StephenCooper, line 20: This is the main type that enables us to get the inference. It is basically self writing and will fail if new events are added but not included in this type.

community-modules/core/src/interfaces/iEventEmitter.ts
```diff
@@ -1,15 +1,23 @@
...
+export type IEventListener<TEventType extends string> = (params: AgEvent<TEventType>) => void;
+export type IGlobalEventListener<TEventType extends string> = (
+    eventType: TEventType,
+    event: AgEvent<TEventType>
+) => void;
 export interface IEventEmitter<TEventType extends string> {
     addEventListener(
         eventType: TEventType,
-        listener: AgEventListener<any, any, TEventType>,
+        listener: IEventListener<TEventType>,
```
StephenCooper, line 14: New interface IEventListener does not depend on the global events like AgEventListener which was a common source of typing pain.

PR #8101, opened 2024-06-17
[AG-11710] Add a context property to column definitions to support custom meta-data on the column-level

SalvatorePreviti:
Background:

There is already a `context` property of type any, see https://localhost:4610/react-data-grid/context
Users have requested a context property also per column.

Changes:

- Adds an additional property `context` to AbstractColDef (and consequently to ColDef and ColGroupDef) of type any that allows to assign and passing custom data to to callbacks.
- Augment the error message in such a way it includes a reference to the "context" property to the valid properties validator so users know what is the proper way to pass custom data to the grid definition or columns and column groups definitions.
- Fix the documentation by adding the context property to colDef and colGroupDef
- Merge FUNCTIONAL_PROPERTIES and CALLBACK_PROPERTIES in FUNCTION_PROPERTIES for the column definitions, this is required to fix typescript confusing coldef and grid def "context" properties.
- Fix the constraint for the type WithoutGridCommon, this is required to fix typescript confusing coldef and grid def "context" properties.

image

### DIFF
--- a/community-modules/core/src/entities/colDef.ts
+++ b/community-modules/core/src/entities/colDef.ts
@@ -57,7 +57,7 @@ export interface AbstractColDef<TData = any, TValue = any> {
     cellAriaRole?: string;
 
     /**
-     * Provides a context object that is provided to different callbacks the column uses. Used for passing additional information to the callbacks by your application.
+     * Context property can be used to associate arbitrary application data with this column definition.
      */
     context?: any;
 }
--- a/community-modules/core/src/entities/colDef.ts
+++ b/community-modules/core/src/entities/colDef.ts
@@ -59,7 +59,7 @@ export interface AbstractColDef<TData = any, TValue = any> {
     /**
      * Provides a context object that is provided to different callbacks the column uses. Used for passing additional information to the callbacks by your application.
      */
-    colContext?: any;
+    context?: any;
 }
 
 /** Configuration options for column groups in AG Grid.  */
--- a/community-modules/core/src/entities/colDef.ts
+++ b/community-modules/core/src/entities/colDef.ts
@@ -55,6 +55,11 @@ export interface AbstractColDef<TData = any, TValue = any> {
      * @default 'gridcell'
      */
     cellAriaRole?: string;
+
+    /**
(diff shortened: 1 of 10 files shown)


community-modules/core/src/entities/colDef.ts
```diff
@@ -55,6 +55,11 @@ export interface AbstractColDef<TData = any, TValue = any> {
      * @default 'gridcell'
      */
     cellAriaRole?: string;
+
+    /**
+     * Provides a context object that is provided to different callbacks the column uses. Used for passing additional information to the callbacks by your application.
+     */
+    context?: any;
```
StephenCooper, line 62: We need this to be different to the Grid context docs as this is not quite right for what we want here. Maybe something along these lines.

`Context property can be used to associate arbitrary application data with this column definition.`

community-modules/core/src/interfaces/iCommon.ts
```diff
@@ -4,7 +4,7 @@ import type { GridApi } from '../api/gridApi';
  * Enables types safe create of the given type without the need to set the common grid properties
  * that will be merged with the object in a centralised location.
  */
-export type WithoutGridCommon<T extends AgGridCommon<any, any>> = Omit<T, keyof AgGridCommon<any, any>>;
+export type WithoutGridCommon<T> = Omit<T, keyof AgGridCommon<any, any>>;
```
StephenCooper, line 7: Can we still constrain to at least `object` to catch the case of this being used against string literals types which are present around the grid.

SalvatorePreviti: Mhhh so, arrays and strings are objects as well, so writing the constraint "T extends object" would still pass for strings and numbers and anything really, including null. If I remember correctly, "T extends object" or "T extends {}" is unfortunately not giving us any advantage in this case

SalvatorePreviti: My reasoning was: As this is a constraint specific to the usage of this type and is not use in any inference but just when using this type, does it makes sense to guard it with a constraint?

StephenCooper: Ok fair enough. :)

StephenCooper (approved):

PR #8107, opened 2024-06-18
[AG-5497] [Overlay] Allow showing the loading overlay using a loading property

SalvatorePreviti:
1/ Add a new grid loading property - boolean to override built-in logic showing loading overlay depending on rowData value.

Changes:

- Add in grid options a boolean property "loading" that can also be set to undefined
- Fix type coercion to support undefined boolean values
- Modify the overlayService to support the new logic based on loading property
- Exhaustive behavioural unit test that covers the behaviour described in the ticket and edge cases

DO NOT MERGE - work in progress

### DIFF
--- a/community-modules/core/src/entities/gridOptions.ts
+++ b/community-modules/core/src/entities/gridOptions.ts
@@ -770,11 +770,11 @@ export interface GridOptions<TData = any> {
      *
      * - When set to true, the grid will show the 'loading' overlay, regardless of the status of the grid.
      * - When set to false, the grid will never show any overlay, regardless of the status of the grid.
-     * - When not set, or set to undefined, the grid will show the the grid will show the loading overlay if rows are not set, and will show 'no rows' overlay if rows are set to empty (classic behavior).
+     * - When not set, or set to undefined, the grid will show the loading overlay if rows are not set, and will show 'no rows' overlay if rows are set to empty (classic behavior).
      *
      * When set, this property has priority over overlay APIs (showLoadingOverlay, showNoRowsOverlay, hideOverlay).
      */
-    loading?: boolean;
+    loading?: boolean | undefined;
     /**
      * Provide a template for 'loading' overlay.
      */
--- a/community-modules/core/src/entities/gridOptions.ts
+++ b/community-modules/core/src/entities/gridOptions.ts
@@ -766,6 +766,16 @@ export interface GridOptions<TData = any> {
 
     // *** Overlays *** //
     /**
+     * Forces the grid to show or hide the loading overlay.
+     *
+     * - When set to true, the grid will show the 'loading' overlay, regardless of the status of the grid.
+     * - When set to false, the grid will never show any overlay, regardless of the status of the grid.
+     * - When not set, or set to undefined, the grid will show the the grid will show the loading overlay if rows are not set, and will show 'no rows' overlay if rows are set to empty (classic behavior).
+     *
+     * When set, this property has priority over overlay APIs (showLoadingOverlay, showNoRowsOverlay, hideOverlay).
+     */
+    loading?: boolean;
+    /**
      * Provide a template for 'loading' overlay.
      */
     overlayLoadingTemplate?: string;
--- a/community-modules/core/src/propertyKeys.ts
+++ b/community-modules/core/src/propertyKeys.ts
@@ -388,6 +388,8 @@ export class PropertyKeys {
         'suppressAdvancedFilterEval',
     ];
 
+    public static BOOLEAN_OR_UNDEFINED_PROPERTIES: KeysOfType<boolean | undefined>[] = ['loading'];
+
     // If property does not fit above, i.e union that should not be coerced.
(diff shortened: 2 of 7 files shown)


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -765,6 +765,16 @@ export interface GridOptions<TData = any> {
...
+    /**
+     * Forces the grid to show or hide the loading overlay.
+     *
+     * - When set to true, the grid will show the 'loading' overlay, regardless of the status of the grid.
+     * - When set to false, the grid will never show any overlay, regardless of the status of the grid.
+     * - When not set, or set to undefined, the grid will show the loading overlay if rows are not set, and will show 'no rows' overlay if rows are set to empty (classic behavior).
+     *
+     * When set, this property has priority over overlay APIs (showLoadingOverlay, showNoRowsOverlay, hideOverlay).
+     */
+    loading?: boolean | undefined;
```
StephenCooper, line 777: I think this is too detailed and could be simplified. We should focus on the simplicity of using this property. When its true, we will show the loading overlay.

Also the no rows overlay should be shown if there are no rows and loading = false. 

We shouldn't worry about the classic behavior here. That can just be in our docs as a helpful default if not setting the loading property explicitly.

SalvatorePreviti: Agreed

StephenCooper: Also I think this should just be `loading?: boolean` as adding the explicit undefined is not something we have done yet across the grid.

community-modules/core/src/propertyKeys.ts
```diff
@@ -388,6 +388,8 @@ export class PropertyKeys {
         'suppressAdvancedFilterEval',
     ];
 
+    public static BOOLEAN_OR_UNDEFINED_PROPERTIES: KeysOfType<boolean | undefined>[] = ['loading'];
```
StephenCooper, line 391: I would prefer if we could just handle this within the loading overlay service itself. Even if that means that if loading is set at all then you are locked into using it and there is no default loading spinner for the first data load.

community-modules/core/src/rendering/overlays/overlayService.ts
```diff
@@ -31,20 +32,23 @@ export class OverlayService extends BeanStub implements NamedBean {
...
+        this.onLoadingChanged();
     }
 
     public showLoadingOverlay(): void {
         if (this.gos.get('suppressLoadingOverlay')) {
             return;
         }
+        const loading = this.gos.get('loading');
+        if (loading !== undefined && !loading) {
+            _warnOnce('AG Grid: showLoadingOverlay has no effect when loading=false');
```
StephenCooper, line 49: _warnOnce adds the `AG Grid: ` prefix so that should not be included here.

community-modules/core/src/entities/gridOptions.ts
SalvatorePreviti: I will remove it for now, as the code base is currently not configured to handle this properly.

https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes

When this option is enabled, and I worked in codebases where this is enabled, `x?:type` is different from `x?:type | undefined` - the compiler gets angry if an undefined is passed to an optional property that is not marked as possibly undefined

community-modules/core/src/test/overlays.test.ts
```diff
@@ -0,0 +1,401 @@
...
+            api.setGridOption('loading', false);
+            expect(document.querySelector(OVERLAY)).toBeFalsy();
+        });
+
+        test('suppressLoadingOverlay has priority', () => {
+            const api = createMyGrid({ columnDefs, loading: true, suppressLoadingOverlay: true });
+            expect(document.querySelector(OVERLAY)).toBeFalsy();
+            expect(document.querySelector(NO_ROWS)).toBeFalsy();
+
+            api.setGridOption('loading', undefined);
```
StephenCooper, line 311: Would be nice to wrap this in a helper method to make everything more concise

community-modules/core/src/rendering/overlays/overlayService.ts
```diff
@@ -27,24 +28,28 @@ export class OverlayService extends BeanStub implements NamedBean {
...
-        }
+        this.updateOverlayVisibility();
     }
 
     public showLoadingOverlay(): void {
         if (this.gos.get('suppressLoadingOverlay')) {
             return;
         }
+        const loading = this.gos.get('loading');
+        if (loading !== undefined && !loading) {
```
StephenCooper, line 50: Can we change this to `loading === false`

SalvatorePreviti: yes, since we are coercing - but if you remove coercion this would be weird - you were telling me coercion has to be removed in the future or?

SalvatorePreviti: ok replacing it to false comparison, the tests cover the case of passing undefined (so they test coercion) so they will break in any case when we change that, so we can safely test for false and go back to fix this if we need to

PR #8122, opened 2024-06-21
AG-11719 Update Column Interfaces Docs

StephenCooper:
1. Adding missing JsDoc comments to IColumn interfaces files. 
2. Fixed type of `getMinWidth` `getMaxWidth` to only `number` as they are always defined. Cleaned up usages to reflect this.
3. Added a `ColumnGroup` api page for group instances which also includes `ProvidedColumnGroup`

Still decide: how many methods on `Column` should be officially supported on the new interface? Have deprecated some on the interface as they seem obviously private but I suspect there could be more.


documentation/ag-grid-docs/src/components/reference-documentation/documentation-helpers.ts
```diff
@@ -73,7 +73,7 @@ export function getTypeUrl(type, framework) {
 
 export function getLinkedType(type, framework) {
     if (!Array.isArray(type)) {
-        type = [type];
+        type = [...type.split('|')];
```
StephenCooper, line 76: Fixes applying type links to union types such as `Column | ColumnGroup`

community-modules/core/src/entities/agColumn.ts
```diff
@@ -274,7 +271,7 @@ export class AgColumn<TValue = any> extends BeanStub<ColumnEventName> implements
             width = 200;
         }
 
-        return Math.max(Math.min(width, maxColWidth), minColWidth);
+        return Math.max(Math.min(width, this.maxWidth), this.minWidth);
```
StephenCooper, line 274: This can use the properties as when this is called they would have already been set so no need to duplicate the default logic here.

documentation/ag-grid-docs/src/content/api-documentation/column-object/properties.json
```diff
@@ -24,102 +24,10 @@
         "addEventListener": {},
         "removeEventListener": {}
     },
-    "events": {
```
StephenCooper, line 27: Column events have their own page so removing the duplication from column reference page.

documentation/ag-grid-docs/src/content/api-documentation/column-object-group/properties.json
```diff
@@ -0,0 +1,20 @@
+{
+    "_config_": {
+        "codeSrc": "columnGroup.AUTO.json"
+    },
+    "ColumnGroup": {
```
StephenCooper, line 5: Just adding all public for now.

documentation/ag-grid-docs/src/content/api-documentation/column-object-group/providedProperties.json
```diff
@@ -0,0 +1,18 @@
+{
+    "_config_": {
+        "codeSrc": "providedColumnGroup.AUTO.json"
+    },
+    "ProvidedColumnGroup": {
```
StephenCooper, line 5: Just adding all public for now.

documentation/ag-grid-docs/src/content/api-documentation/column-object/properties.json
```diff
@@ -4,7 +4,7 @@
     },
     "methods": {
         "meta": {
-            "displayName": "Column Methods"
+            "displayName": "Methods"
         },
         "getColId": {},
```
StephenCooper, line 9: This is still missing a lot of events that are on the `Column` interface. Need to decide if we just put them all on and deal with any clean up via deprecations in future releases or if we want to deprecate more from the interface.

PR #8130, opened 2024-06-24
AG-9318 overlays react unmount fix

SalvatorePreviti:
Work in progress

When rendering overlays, the last showXXX or hide should win over the previous showXXX or hide no matter when the promises that creates the component are resolved.

The original implementation does not handle properly the case when the order of completion of promises is different than the order of creation of those promises. By keeping a reference to the last started promise we can handle the situation better.

Also, moving the "gridOption" to the overlay manager simplifies a lot the handling of the promises for registering the handler

- overlayWrapperComponent race condition free promise handling
- new common interfaces and component class for Overlay, LoadingOverlay and NoRowsOverlay will just extend those
- minor optimizations in reactComponent.ts
- fuzz test for mount/unmount of react components, just to be sure that also changing the order of execution the behavior is consistent and all components gets unmounted
- fix jest configuration for behavioural unit tests


community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
```diff
@@ -56,48 +57,59 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
...
+            if (!comp) {
+                return; // Error handling
+            }
+
+            this.eOverlayWrapper.appendChild(comp.getGui());
+            this.activeOverlay = comp;
+
+            if (gridOption) {
+                this.updateListenerDestroyFunc = this.addManagedPropertyListener(gridOption, ({ currentValue }) => {
+                    (comp as any)?.refresh(this.gos.addGridCommonParams({ ...(currentValue ?? {}) }));
```
peterjrreynolds, line 89: If we pull an interface out of `ILoadingOverlay` and `INoRowsOverlay` overlay, can type the component properly.

PR #8135, opened 2024-06-26
AG-11977 Delay auto sizing if column headers are not rendered

StephenCooper:


community-modules/core/src/columns/columnAutosizeService.ts
```diff
@@ -57,6 +61,15 @@ export class ColumnAutosizeService extends BeanStub implements NamedBean {
         // hasn't fully drawn out all the cells yet (due to cell renderers in animation frames).
         this.animationFrameService.flushAllFrames();
 
+        if (this.timesDelayed < 5 && !this.renderStatusService?.areHeadersRendered(colKeys)) {
```
peterjrreynolds, line 64: This check will always return true for non-React.

I think you want `this.timesDelayed < 5 && this.renderStatusService && !this.renderStatusService.areHeadersRendered(colKeys)`

community-modules/react/src/reactUi/renderStatusService.tsx
```diff
@@ -0,0 +1,36 @@
...
+        let allColsRendered = true;
+        colKeys.forEach((key) => {
+            if (!allColsRendered) {
+                return;
+            }
+
+            const column = this.columnModel.getCol(key);
+            if (!column) {
+                return;
+            }
```
peterjrreynolds, line 23: Can just remove the column check/loop completely - it's not needed

community-modules/react/src/reactUi/renderStatusService.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+            this.ctrlsService.getHeaderRowContainerCtrls().forEach((container) => {
```
peterjrreynolds, line 25: if you use every, can short-circuit

PR #8137, opened 2024-06-26
AG-11965 New way to define modules that is type safe and verifies that all APIs are defined correctly

SalvatorePreviti:
- Split GridApi in multiple base interfaces so we can know exactly which modules export which APIs
- Add a new _defineModule that helps checking that all and only the exported API methods are present, to define APIs it requires a generic argument with the interface containing all API functions
- Fix generate-code-reference-files to handle base interfaces, and, add validation to it so compilation and CI will fail if the same API is declared in multiple interfaces or if a type is not correctly defined in the file


StephenCooper (approved):
Happy with the changes.

community-modules/core/src/main.ts
```diff
@@ -560,7 +560,48 @@ export { CellNavigationService } from './cellNavigationService';
...
+    SideBarGridApi,
+    StatusBarGridApi,
+    InfiniteRowModelGridApi,
+    CsvExportGridApi,
+    RowGroupingGridApi,
+    RangeSelectionGridApi,
+    ServerSideRowModelGridApi,
+    MenuGridApi,
+    MasterDetailGridApi,
+    _ExcelExportGridApi,
```
peterjrreynolds, line 600: This shouldn't have an underscore prefix

community-modules/core/src/main.ts
```diff
@@ -560,7 +560,48 @@ export { CellNavigationService } from './cellNavigationService';
 export { KeyCode } from './constants/keyCode';
 export { VerticalDirection, HorizontalDirection } from './constants/direction';
 export { Grid, GridParams, Params, GridCoreCreator, createGrid, provideGlobalGridOptions } from './grid';
-export { GridApi, DetailGridInfo, StartEditingCellParams, GetCellValueParams } from './api/gridApi';
+export {
+    GridApi,
+    CoreGridApi,
```
peterjrreynolds, line 565: Do we need to export the types that are only used within core?

plugins/ag-grid-generate-code-reference-files/src/executors/generate/generate-code-reference-files.ts
```diff
@@ -444,13 +444,49 @@ export function getColumnOptions(colDefFile: string, filterFile: string) {
 
 export function getGridApi(gridApiFile: string) {
     const srcFile = parseFile(gridApiFile);
-    const gridApi = findNode('GridApi', srcFile);
+    const gridApi = findNode('GridApi', srcFile) as ts.InterfaceDeclaration;
+
+    // console.log(gridApi.heritageClauses);
```
peterjrreynolds, line 449: Can remove

community-modules/core/src/main.ts
SalvatorePreviti: I think we need to export all the types here, if not we would need to import a type from a core subfolder when we define a module. Would be more convenient to export * from really, as it would reduce the amount of a developer need to change when adding a new module, but is fine for now I think.

Exporting all base types of GridApi seems to be a good thing in general as developers that are writing wrappers over AG Grid can decide what to expose if they want to.

community-modules/core/src/main.ts
SalvatorePreviti: I left it as it was, let me have a look and a think

plugins/ag-grid-generate-code-reference-files/src/executors/generate/generate-code-reference-files.ts
SalvatorePreviti: removed

PR #8187, opened 2024-06-28
RTI-2176 Support multiple Vue grids

StephenCooper:
Previously the `changes` and `timeout` properties where shared across all instances of the grid which meant that any property changes performed by one grid would be prevented from being fired for the other grid.

This change brings those properties onto the component instance so that they each have their own changes and timeout and are not overriding each other any more.

### DIFF
--- a/community-modules/vue3/src/Utils.ts
+++ b/community-modules/vue3/src/Utils.ts
@@ -52,8 +52,7 @@ export const getAgGridProperties = (): [Properties, Properties, Properties] => {
             deep: true,
         },
     };
-    let timeout: number | null = null;
-    let changes: { [key: string]: any } = {};
+
     ComponentUtil.ALL_PROPERTIES.filter((propertyName: string) => propertyName != 'gridOptions') // dealt with in AgGridVue itself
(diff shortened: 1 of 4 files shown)


community-modules/vue3/src/Utils.ts
```diff
@@ -62,13 +61,13 @@ export const getAgGridProperties = (): [Properties, Properties, Properties] => {
...
+                    this.batchChanges[propertyName] =
                         currentValue === ComponentUtil.VUE_OMITTED_PROPERTY ? undefined : currentValue;
-                    if (timeout == null) {
-                        timeout = setTimeout(() => {
-                            _processOnChange(changes, this.api);
-                            timeout = null;
-                            changes = {};
+                    if (this.batchTimeout == null) {
+                        this.batchTimeout = setTimeout(() => {
+                            _processOnChange(this.batchChanges, this.api);
```
SalvatorePreviti, line 68: there is a problem here - if we don't set this.batchTimeout = null before calling _processOnChange, if a change happens in _processOnChange will be ignored as it won't start the timer and will not be processed. Is this intentional?

PR #8233, opened 2024-07-01
AG-11849 Case insensitive sorting

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-11849

Trying out Salvo's testing project 👀 

Note that `String.localeCompare` already gives us the behaviour that we want here, but we only use it when doing `accentedSort: true`. Otherwise we fallback to default string comparison in JS, presumably for performance reasons?

Actually looking at the performance, Chrome, Firefox and Safari are unfortunately very different (see benchmark: https://jsbm.dev/Heo4NDKG1ysMV) in different cases.

For ASCII-only strings, `String.localeCompare` is very fast in Chrome (possibly because of [this optimisation for ASCII strings](https://stackoverflow.com/a/77295491)). Also fast in Safari. It is 3x slower in Firefox.

For non-ASCII strings, `String.localeCompare` is quite slow across the board, but using `toLocaleLowerCase` before comparing is slow in Firefox but not Chrome (and seems to not work at all in Safari).

I'm not able to test on Edge (but it's Chromium so should be the same as Chrome)

Given the above, and the fact that we already don't guarantee locale-accurate sorting by default, I think we should just go ahead and use `String.toLowerCase` to give us case-insensitive sorting.


eliasmalik:
Given the chat on JIRA, I'm closing this now

PR #8235, opened 2024-07-01
Update apiUtils.ts

### LINKED ISSUE #8229: [CRITICAL] V32 error with zone.js framewroks like Angular
```
[X] bug report => see 'Providing a Reproducible Scenario'
```

**Current behavior**

Usage of ag-grid v32 with zone.js broke the app with error
`main.esm.mjs:1492 AG Grid: Unknown API function: 'then' on GridApi.`

looking at implementation i see that you now use a `Proxy` for checking if the method exists, well there should be a safelist of methods that should return undefined like `then`.

Zone.js that throws error from Proxy handler get method is

```
function resolvePromise(promise, state, value) {
            const onceWrapper = once();
            if (promise === value) {
                throw new TypeError(TYPE_ERROR);
            }
            if (promise[symbolState] === UNRESOLVED) {
                // should only get value.then once based on promise spec.
                let then = null;
                try {
                    if (typeof value === 'object' || typeof value === 'function') {
                        then = value && value.then;
                    }
```

value.then executes the new Proxy get method

**Expected behavior**

Should not throw an error for safe methods like `.then` for checking if it's a Promise, or not throw error just a console warning and return undefined

Ag v32

Typescript Angular v18

Repro: https://stackblitz.com/edit/stackblitz-starters-mtzipd?file=src%2Fmain.ts ( open in new window if you don't see the "console" button )

leonelvsc:
Fix #8229 , if property is not a function should not return a wrappedFn it should return undefined

### DIFF
--- a/community-modules/core/src/api/apiFunctionService.ts
+++ b/community-modules/core/src/api/apiFunctionService.ts
@@ -40,9 +40,7 @@ export class ApiFunctionService extends BeanStub implements NamedBean {
         if (this.isDestroyed) {
             return this.destroyedHandler(functionName);
         }
-        if (this.isFrameworkMethod(functionName)) {
-            return undefined;
-        }
+        
         return undefined;
     }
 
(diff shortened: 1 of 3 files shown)


leonelvsc:
note: this will make code that relies on `api.propertyNotFn` work again but it will still show an error on the console

leonelvsc:
Implemented a simple safeList that avoids triggering console warnings and errors

StephenCooper:
Thanks, for the PR. Am closing as we have committed a change for this fix.

PR #8259, opened 2024-07-04
AG-12085 Fix auto id for pinned rows

StephenCooper:


community-modules/core/src/gridOptionsService.ts
```diff
@@ -607,7 +607,7 @@ export class GridOptionsService extends BeanStub implements NamedBean {
             let id = getRowId(params);
 
             if (typeof id !== 'string') {
-                _warnOnce(`The getRowId callback must return a string. The ID ${id} is being cast to a string.`);
+                _warnOnce(`The getRowId callback must return a string. The ID `, id, ` is being cast to a string.`);
```
eliasmalik, line 610: I think this change means that only the first non-string key will cause a log to be printed, rather than one for all of them. Is this deliberate? 

Concretely, if `getRowId` returns `13` for one row and `15` for another row, the user will only see:
> The getRowId callback must return a string. The ID 13 is being cast to a string.

testing/behavioural/src/rows/pinned-rows.test.ts
```diff
@@ -0,0 +1,117 @@
+import { ClientSideRowModelModule } from '@ag-grid-community/client-side-row-model';
+import type { GetRowIdParams, GridOptions } from '@ag-grid-community/core';
+import { ModuleRegistry, createGrid } from '@ag-grid-community/core';
+import exp from 'constants';
```
eliasmalik, line 4: Unused I think?

testing/behavioural/src/rows/pinned-rows.test.ts
```diff
@@ -0,0 +1,117 @@
...
+                columnDefs,
+                pinnedTopRowData: topData,
+                getRowId: (p) => {
+                    getRowIdParams = p;
+                    return p.data.athlete;
+                },
+            });
+
+            assertPinnedRowData(topData, 'top');
+            expect(getRowIdParams.data).toBe(topData[0]);
```
eliasmalik, line 69: Alternative approach is to make `getRowId` a `jest.fn`, then we could 
```
expect(...).toHaveBeenLastCalledWith(expect.objectContaining({ 
  data: topData[0], 
  rowPinned: 'top' 
});
```

community-modules/core/src/gridOptionsService.ts
StephenCooper: Good point, that is correct. We may need to update _warnOnce as there are other places that could benefit from this fix.

PR #8266, opened 2024-07-04
AG-12106 Fix localisation related example generation

seanlandsman:


StephenCooper (commented):
Are there no changes to React?

seanlandsman:
> Are there no changes to React / Vue?

No, fixed by changes in parser-utils

PR #8275, opened 2024-07-08
codemods docs

SalvatorePreviti:
Improve codemods documentation and add the command line arguments and configuration details

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
@@ -169,10 +169,10 @@ export default defineUserConfig({
         // This interceptor will be called only if a previous call to
         // `matchGridImport` returned true for the same import path
         if (importPath === "@my-org/my-grid" && agGridExportName === "createGrid") {
-        // Handle`import { createGridWithCustomOption } from "@my-org/my-grid"`
-        return importName === "createGridWithCustomOption";
-    }
-    // Handle everything else, including `import { createGrid } from "@my-org/my-grid"`
+            // Handle`import { createGridWithCustomOption } from "@my-org/my-grid"`
+            return importName === "createGridWithCustomOption";
+        }
+        // Handle everything else, including `import { createGrid } from "@my-org/my-grid"`
         return importName === agGridExportName;
     },
 });
--- a/documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
@@ -4,10 +4,10 @@ title: "Codemods"
 
 ## How to Upgrade AG Grid Version
 
-All major AG Grid releases from v31 onwards come with an accompanying Codemod to help automate the upgrade process.
-Codemods are scripts that fix your project's source files to address the majority of breaking changes and deprecations when upgrading from an older version.
+All AG Grid releases from v31 onwards come with an accompanying Codemod to help automate the upgrade process.
+Codemods are scripts that fix the project's source files to address the majority of breaking changes and deprecations when upgrading from an older version.
 
-This is the easiest way to make sure your projects stay up-to-date with the latest AG Grid changes.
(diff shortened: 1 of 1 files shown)


documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
```diff
@@ -4,7 +4,8 @@ title: "Codemods"
 
 ## How to Upgrade AG Grid Version
 
-All major AG Grid releases from v31 onwards come with an accompanying Codemod to help automate the upgrade process. Codemods are small scripts that fix your project's source files to address the majority of breaking changes and deprecations when upgrading from an older version.
+All major AG Grid releases from v31 onwards come with an accompanying Codemod to help automate the upgrade process.
```
StephenCooper, line 7: We can remove the word "major" as we actually release code mods for each minor version too.

documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
```diff
@@ -29,3 +30,170 @@ Some points to bear in mind when using Codemods:
 * While we attempt to automate as many upgrade paths as possible, unusual use cases may still require some manual intervention.
 
 See the relevant migration documentation page for more information on the Codemod for a specific AG Grid version.
+
+
+## Usage
+
+Typically to run the migration you would run the following command, where `$FROM_VERSION` refers to your project's existing AG Grid version.
```
StephenCooper, line 37: Aim for the passive voice, and avoid words like Typically as we should just focus on the normal use case at this point.

documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
```diff
@@ -29,3 +30,170 @@ Some points to bear in mind when using Codemods:
...
+
+    Loads a custom configuration file.
+    See the [Configuration file](#configuration-file) section for more information.
+
+### Configuration file
+
+A configuration file is a [commonJS](https://en.wikipedia.org/wiki/CommonJS) JavaScript or TypeScript module that exports a configuration object.
+- A JavaScript file, with extension `.cjs` - ```npx @ag-grid-devtools/cli migrate --config=./my-migration-config.cjs```
+- A [TypeScript](https://www.typescriptlang.org/) file with extension `.cts` - ```npx @ag-grid-devtools/cli migrate --config=./my-migration-config.cts```
+- A commonJS module from a package installed in node_modules with the `module:` prefix - ```npx @ag-grid-devtools/cli migrate --config=module:my-npm-package/my-grid-migration.cjs```
```
StephenCooper, line 111: Do we really expect users to be doing this?

SalvatorePreviti: The use case was, a big corp has a react package that many teams use, they can put their migration script in their react package. But I agree, let's keep it simple and let's not document it

documentation/ag-grid-docs/src/content/docs/codemods/index.mdoc
```diff
@@ -24,8 +25,278 @@ Codemods are partially supported for the following frameworks (full support comi
...
+
+## Usage
+
+To run the migration, where `$FROM_VERSION` refers to the project's existing AG Grid version:
+
+```
+npx @ag-grid-devtools/cli@latest migrate --from=$FROM_VERSION
+```
+
+For example, to migrate from AG Grid v30 to v31:
```
StephenCooper, line 43: Is the to value automatically inferred? In which case this command on latest would actually go to v32?

SalvatorePreviti: No this comes from the previous version of the documentation and I didn't change it, also if I find it confusing as well. I think we can just use "30" as there is no codemod before 31.
"latest" should work as is supported by semver comparison (also if I didn't try it) but this is a "from" not a "to" so saying "from=latest" doesn't make sense I think

StephenCooper (approved):
Approved.

PR #8278, opened 2024-07-08
AG-11856 React 19 Support

StephenCooper:


community-modules/core/src/ctrlsService.ts
```diff
@@ -98,6 +98,12 @@ export class CtrlsService extends BeanStub implements NamedBean {
     public register<K extends CtrlType, T extends ReadyParams[K]>(ctrlType: K, ctrl: T): void {
         this.params[ctrlType] = ctrl;
         this.checkReady();
+
+        ctrl.addDestroyFunc(() => {
+            // Enables going back into waiting state during extra React StrictMode render.
+            this.params[ctrlType] = null!;
+            this.checkReady();
+        });
```
StephenCooper, line 106: This is a key part of the fix, where when a ctrl is destroyed in unregisters itself from the CtrlService which puts the service back into waiting state. This enables ctrls, (on their second render) to queue callbacks like the first time while waiting for the elements to be set again.

community-modules/core/src/headerRendering/cells/column/headerCellCtrl.ts
```diff
@@ -65,6 +65,10 @@ export class HeaderCellCtrl extends AbstractHeaderCellCtrl<IHeaderCellComp, AgCo
     ): void {
         this.comp = comp;
 
+        if (_isSameElement(this.eGui, eGui)) {
+            return;
+        }
```
StephenCooper, line 70: For ctrls that are managed by AG Grid, i.e they are not created in the React component, then we need to handle the fact that `setComp` can be called twice now. React 19 now double calls ref callbacks. React 18 did not do this.

We are basing the detection on the fact that the same html element is passed again.

community-modules/core/src/context/beanStub.ts
```diff
@@ -218,9 +218,12 @@ export abstract class BeanStub<TEventType extends string = BeanStubEvent>
      * @param events Array of GridOption properties to listen for changes too.
      * @param listener Shared listener to run if any of the properties change
      */
-    public addManagedPropertyListeners(events: (keyof GridOptions)[], listener: PropertyChangedListener): void {
+    public addManagedPropertyListeners(
```
StephenCooper, line 221: Exposing the cleanup functions for consistency even though this is not actually required for these changes now.

PR #8280, opened 2024-07-08
AG-11965 (part 2) new type safe way to define api fns validation

SalvatorePreviti:
In the previous PR https://github.com/ag-grid/ag-grid/pull/8137 we introduced a new way to define modules in such a way we won't forget API methods or we don't forget to expose them in interfaces.
The validator was not however checking enforcing that the functions are associated to the right module interface. This PR solves this.

- Change the validator so that adding a new API is more type safe, and a compilation error is raised if a method is forgot or in the wrong interface.


StephenCooper (approved):
LGTM

PR #8284, opened 2024-07-09
AG-11965 - fix regression in code reference files generate

SalvatorePreviti:
The implementation of getGridApi in docs generator was not recursive and was not processing all base interfaces.
The previous change merged in https://github.com/ag-grid/ag-grid/pull/8280 introduced a regression as it introduces a new base interface, so all core APIs were not included in the documentation generated json.

This fixes the issue by making getGridApi recursive


StephenCooper (approved):
Approved.

PR #8322, opened 2024-07-11
AG-11982 allow overlays to be focusable and prevent header cells from focus when using loading overlay

SalvatorePreviti:
- add a new utility method isGridFocused in FocusService that returns true if an element in the grid has focus
- change focusable containers logic in grid controller, so that the exclusive loading overlays are the only focusable thing, and the non exclusive no-rows overlay is focusable
- change the overlay wrapper component so when the exclusive loading overlay is visible, and the grid was focused, the focus goes to the loading overlay
- change the overlay wrapper component so that when the exclusive loading overlay is closed, the focus goes back to the previous focused element, if still exists

### DIFF
--- a/community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
+++ b/community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
@@ -148,7 +148,7 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
         _clearElement(this.eOverlayWrapper);
 
         // Focus the element that was focused before the exclusive overlay was shown
-        elementToFocus?.focus();
+        elementToFocus?.focus({ preventScroll: true });
     }
 
     public hideOverlay(): void {
--- a/community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
+++ b/community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
@@ -1,5 +1,6 @@
 import type { BeanCollection } from '../../context/context';
 import type { GridOptions } from '../../entities/gridOptions';
+import type { FocusService } from '../../focusService';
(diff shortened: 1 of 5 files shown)


community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
```diff
@@ -120,6 +146,9 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
         this.destroyBean(activeOverlay);
 
         _clearElement(this.eOverlayWrapper);
+
+        // Focus the element that was focused before the exclusive overlay was shown
+        elementToFocus?.focus();
```
peterjrreynolds, line 151: Want `focus({ preventScroll: true })` here, otherwise if the user has scrolled the grid out of view it will jump back when loading completes.

SalvatorePreviti: fixed

community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
```diff
@@ -59,14 +63,29 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
...
-        overlayComponentPromise?.then((comp) => {
+        if (!overlayComponentPromise) {
+            return;
+        }
+
+        this.setDisplayed(true, { skipAriaHidden: true });
+
+        if (exclusive && this.focusService.isGridFocused()) {
+            const activeElement = this.gos.getActiveDomElement();
+            if (activeElement && 'focus' in activeElement) {
```
gportela85, line 83: better to use `if (activeElement && !this.gos.isNothingFocused())`

PR #8327, opened 2024-07-12
AG-12193 Correctly update Pinned rows on data change

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12193

There was previously no update logic for pinned rows, which means they were either re-created entirely (if there was no matching ID found) or they were left intact (without updating associated row data).

This now adds update logic analogous to that in `ClientSideRowModel`, but a bit simpler.

### DIFF
--- a/community-modules/core/src/entities/rowNode.ts
+++ b/community-modules/core/src/entities/rowNode.ts
@@ -457,13 +457,6 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
         this.setDisplayed(rowTop !== null);
     }
 
-    public setRowTopAndRowIndex(rowTop: number, rowIndex: number): number {
-        this.setRowTop(rowTop);
-        this.setRowHeight(this.beans.gos.getRowHeightForNode(this).height);
-        this.setRowIndex(rowIndex);
-        return this.rowHeight!;
-    }
-
     public clearRowTopAndRowIndex(): void {
         this.oldRowTop = null;
         this.setRowTop(null);
--- a/community-modules/core/src/entities/rowNode.ts
+++ b/community-modules/core/src/entities/rowNode.ts
@@ -457,6 +457,13 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
         this.setDisplayed(rowTop !== null);
     }
 
+    public setRowTopAndRowIndex(rowTop: number, rowIndex: number): number {
+        this.setRowTop(rowTop);
+        this.setRowHeight(this.beans.gos.getRowHeightForNode(this).height);
+        this.setRowIndex(rowIndex);
+        return this.rowHeight!;
+    }
+
     public clearRowTopAndRowIndex(): void {
         this.oldRowTop = null;
         this.setRowTop(null);
--- a/community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
+++ b/community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
@@ -138,14 +138,6 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
         nodes.setOrder(newOrder);
     }
 
-    public getPinnedTopRowNodes(): RowNode[] {
-        return this.pinnedTopRows.asArray();
-    }
-
-    public getPinnedBottomRowNodes(): RowNode[] {
-        return this.pinnedBottomRows.asArray();
-    }
-
     public getPinnedTopTotalHeight(): number {
         return this.getTotalHeight(this.pinnedTopRows);
     }
@@ -170,12 +162,15 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
         return this.pinnedBottomRows.getByIndex(index);
     }
 
-    public forEachPinnedTopRow(callback: (rowNode: RowNode, index: number) => void): void {
-        this.pinnedTopRows.forEach(callback);
+    public getPinnedRowById(id: string, floating: NonNullable<RowPinnedType>): RowNode | undefined {
+        return floating === 'top' ? this.pinnedTopRows.getById(id) : this.pinnedBottomRows.getById(id);
     }
 
-    public forEachPinnedBottomRow(callback: (rowNode: RowNode, index: number) => void): void {
-        this.pinnedBottomRows.forEach(callback);
+    public forEachPinnedRow(
+        floating: NonNullable<RowPinnedType>,
+        callback: (node: RowNode, index: number) => void
+    ): void {
+        return floating === 'top' ? this.pinnedTopRows.forEach(callback) : this.pinnedBottomRows.forEach(callback);
     }
 
     private getTotalHeight(rowNodes: OrderedCache<RowNode>): number {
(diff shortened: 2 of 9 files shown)


community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -31,28 +29,32 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
         const rows = floating === 'top' ? this.pinnedTopRows : this.pinnedBottomRows;
-        return _missingOrEmpty(rows);
+        return rows.isEmpty();
     }
 
     public isRowsToRender(floating: RowPinnedType): boolean {
         return !this.isEmpty(floating);
     }
 
+    /** @deprecated */
```
eliasmalik, line 39: As far as I can see this method isn't used anywhere in the codebase. My preference would be to delete it, but this class is exposed in the public interface (via `community-modules/core/src/main.ts`).

It is, however, un-documented. Would it count as a breaking change to remove an undocumented method from an undocumented class that happens to be exposed?

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -100,90 +102,190 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
+        return node.rowTop! + node.rowHeight!;
+    }
+}
+
+/**
+ * Cache that maintains record of insertion order
+ *
+ * Allows lookup by key as well as insertion order (which is why we didn't use Map)
+ */
+class OrderedCache<T extends { id: string | undefined }> {
```
eliasmalik, line 234: Added this because the new update logic is simplified by allowing lookup by ID but we still also need to maintain the order of rows.

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
StephenCooper: No, this would be fine to delete if unused. It is only exported so it can be used by other modules. We only have to respect publicly documented api surfaces.

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -91,7 +93,7 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
 
     private setPinnedTopRowData(): void {
```
StephenCooper, line 94: We could probably merge `setPinnedTopRowData` and `setPinnedBottomRowData` if we pass in the relevant details to reduce the duplication.

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -100,90 +102,190 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
+                // new node
+                const rowNode = new RowNode(this.beans);
+                rowNode.id = id;
+                rowNode.data = data;
+                rowNode.rowPinned = floating;
                 rowNode.setRowTop(nextRowTop);
                 rowNode.setRowHeight(this.gos.getRowHeightForNode(rowNode).height);
-                rowNode.setRowIndex(index);
+                rowNode.setRowIndex(i);
                 nextRowTop += rowNode.rowHeight!;
```
StephenCooper, line 161: Again do you think we can tweek this code so that we don't repeat this same logic in both branches seeing as it is consistent.

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -100,90 +102,190 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
+
+    public forEach(callback: (item: T, index: number) => void): void {
+        this.ordering.forEach((id, index) => {
+            const node = this.cache[id];
+            node && callback(node, index);
+        });
+    }
+
+    public asArray(): T[] {
+        return this.ordering.map((id) => this.cache[id]!);
```
StephenCooper, line 272: Does this get called a lot? Should we be concerned about performance of creating an array each time its called?

eliasmalik: Well, this ends up being called in the row renderer when the rows are redrawn, so I think you're right that it's worth considering the performance.

The array here is used in `RowRenderer.refreshFloatingRows` to:
1. Build a map of ID -> `RowNode`
2. Iterate through the nodes in order

We can do both these things with the existing datastructures in `OrderedCache` so we probably don't need to recreate everything.

I'll work on implementing this, I think it'll end up a bit simpler anyway.

community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -20,10 +20,14 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
 
     public postConstruct(): void {
-        this.setPinnedTopRowData();
-        this.setPinnedBottomRowData();
-        this.addManagedPropertyListener('pinnedTopRowData', () => this.setPinnedTopRowData());
-        this.addManagedPropertyListener('pinnedBottomRowData', () => this.setPinnedBottomRowData());
+        this.setPinnedRowData(this.gos.get('pinnedTopRowData'), 'top');
+        this.setPinnedRowData(this.gos.get('pinnedBottomRowData'), 'bottom');
+        this.addManagedPropertyListener('pinnedTopRowData', () =>
+            this.setPinnedRowData(this.gos.get('pinnedTopRowData'), 'top')
```
StephenCooper, line 26: The property listener has params which have the new value in it so you don't need to look it up again. `e.currentValue`

community-modules/core/src/entities/rowNode.ts
```diff
@@ -457,6 +457,13 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
         this.setDisplayed(rowTop !== null);
     }
 
+    public setRowTopAndRowIndex(rowTop: number, rowIndex: number): number {
```
StephenCooper, line 460: The issue with this change is that it means this code cannot be dropped as part of a "pinned" module. I know its not much but these are the kind of things that we should start considering going forward for bundle size.

So I think this should live in the pinnedRowModel

community-modules/core/src/rendering/rowRenderer.ts
```diff
@@ -479,15 +479,15 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
      * @param rowNodes The canonical list of row nodes that should have associated controllers
      */
-    private refreshFloatingRows(rowCtrls: RowCtrl[], rowNodes: RowNode[]): void {
-        const nodeMap = Object.fromEntries(rowNodes.map((node) => [node.id!, node]));
+    private refreshFloatingRows(rowCtrls: RowCtrl[], floating: NonNullable<RowPinnedType>): void {
         const rowCtrlMap = Object.fromEntries(rowCtrls.map((ctrl) => [ctrl.getRowNode().id!, ctrl]));
 
-        for (let i = 0; i < rowNodes.length; i++) {
-            const node = rowNodes[i];
+        this.pinnedRowModel.forEachPinnedRow(floating, (node, i) => {
```
StephenCooper, line 485: When accessing properties from `this.` repeatedly we aim to do the following. This is better for bundle size and also performs better. 

```ts
const { pinnedRowModel } = this;
```

StephenCooper (approved):
Happy with these changes.

PR #8329, opened 2024-07-12
AG-12201 - more standard gridApi object - change the way Proxy is used

SalvatorePreviti:
With this PR, gridApi is now an object that inherits a Proxy
This means that if a method is not found, JS will just walk up in the prototype chain and use the proxy getter only for unregistered methods.

Implement also a cache so the same function instance is always returned instead of creating a new function at every get.


SalvatorePreviti:
Abandoned - we decided to follow this approach instead - https://github.com/ag-grid/ag-grid/pull/8344

PR #8344, opened 2024-07-16
AG-12201 - gridApi without Proxy

SalvatorePreviti:
Proxy in the grid api was creating a series of issues, so we decided to remove the proxy and use a plain JS object.

- Move the grid api function names from the validator module to the core module, keeping it private. Note that minified this is currently less than 6kb and also adding more methods in the future we do not expect it to grow enough to pose a problem with bundle size.
- Change the gridApiFunctionService to build an object every time with all the grid api wrappers, also from external modules.
- Ensure that every method in gridApi have a valid name, that makes debugging and stack trace nicer
- Ensure that the grid api object has a class name "GridApi"
- Add tests to avoid regressions and to verify warnings
- Move the missing method validation logic outside of the validator module, as now is much smaller and doesn't need to live in a separate module


community-modules/core/src/api/apiFunctionService.ts
```diff
@@ -3,24 +3,50 @@ import { BeanStub } from '../context/beanStub';
...
+import type { GridApi } from './gridApi';
+import { gridApiFunctionNames, gridApiFunctionsMap } from './gridApiFunctionNames';
 import type { ApiFunction, ApiFunctionName } from './iApiFunction';
 
-function dispatchEvent(beans: BeanCollection, event: AgEvent<AgEventType>): void {
+const dispatchEvent = (beans: BeanCollection, event: AgEvent<AgEventType>): void =>
     beans.eventService.dispatchEvent(event);
-}
+
+const destroyed = {
```
StephenCooper, line 15: Nit pick, name `destroyed` should probably be updated to reflect all these properties not just the destroy ones.

community-modules/core/src/api/apiUtils.ts
```diff
@@ -1,22 +1,9 @@
...
-            // allow api to work with async/await
-            if (prop === 'then') {
-                return;
-            }
-            return (...args: any[]) => target.callFunction(prop as any, args);
-        },
-    }) as any;
-}
-
 export function createApiProxy(context: Context): { beanName: 'gridApi'; bean: GridApi } {
```
StephenCooper, line 4: Nit pick: Its not a proxy anymore.

peterjrreynolds (approved):

StephenCooper (approved):
LGTM

PR #8345, opened 2024-07-16
AG-12257 Grid related patch notes

adam-wang-uk:


documentation/ag-grid-docs/public/changelog/releaseVersionNotes.json
```diff
@@ -1,8 +1,4 @@
 [
-    {
```
StephenCooper, line 2: Why is 32.0.2 removed?

StephenCooper (approved):

PR #8351, opened 2024-07-17
AG-12285 Selection API changes

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12285

### DIFF
--- a/community-modules/core/src/entities/gridOptions.ts
+++ b/community-modules/core/src/entities/gridOptions.ts
@@ -2428,33 +2428,37 @@ export interface LoadingCellRendererSelectorResult {
 export type DomLayoutType = 'normal' | 'autoHeight' | 'print';
 
 /** Configuration options for selection */
-export type SelectionOptions<TData = unknown, TValue = unknown> =
+export type SelectionOptions<TData = any, TValue = any> =
     | RowSelectionOptions<TData, TValue>
     | CellSelectionOptions<TData>;
 
 /** Cell selection options */
-export interface CellSelectionOptions<TData = unknown> {
+export interface CellSelectionOptions<TData = any> {
     mode: 'cell';
-    /**
-     * Configuration options for the fill handle
-     */
-    fillHandleOptions?: FillHandleOptions<TData>;
     /**
      * If `true`, only a single range can be selected
      * @default false
      */
     suppressMultiRanges?: boolean;
     /**
-     * Set to `true` to enable the Range Handle
-     * @default false
+     * Determine the selection handle behaviour. Can be used to configure the range handle and the fill handle.
+     * Set to `true` to enable the range handle with default configuration.
      */
-    enableRangeHandle?: boolean;
+    handle?: boolean | RangeHandleOptions | FillHandleOptions<TData>;
+}
+
+/**
+ * Configuration options for the range handle
+ */
+export interface RangeHandleOptions {
+    mode: 'range';
 }
 
 /**
  * Configuration options for the fill handle
  */
-export interface FillHandleOptions<TData> {
+export interface FillHandleOptions<TData = any> {
+    mode: 'fill';
     /**
      * Set this to `true` to prevent cell values from being cleared when the Range Selection is reduced by the Fill Handle.
      * @default false
@@ -2471,11 +2475,11 @@ export interface FillHandleOptions<TData> {
     setFillValue?: <TContext = any>(params: FillOperationParams<TData, TContext>) => any;
 }
 
-export type RowSelectionOptions<TData = unknown, TValue = unknown> =
+export type RowSelectionOptions<TData = any, TValue = any> =
     | SingleRowSelectionOptions<TData, TValue>
     | MultiRowSelectionOptions<TData, TValue>;
 
-interface CommonRowSelectionOptions<TData, TValue> {
+interface CommonRowSelectionOptions<TData = any, TValue = any> {
     /**
      * If `true`, rows will not be deselected if you hold down `Ctrl` and click the row or press `Space`.
      * @default false
@@ -2497,16 +2501,17 @@ interface CommonRowSelectionOptions<TData, TValue> {
     isRowSelectable?: IsRowSelectable<TData>;
 }
 
-export interface SingleRowSelectionOptions<TData = unknown, TValue = unknown>
-    extends CommonRowSelectionOptions<TData, TValue> {
+/**
+ * Determines selection behaviour when only a single row can be selected at a time
+ */
+export interface SingleRowSelectionOptions<TData = any, TValue = any> extends CommonRowSelectionOptions<TData, TValue> {
     mode: 'singleRow';
 }
 
 /**
  * Determines selection behaviour when multiple rows can be selected at once.
  */
-export interface MultiRowSelectionOptions<TData = unknown, TValue = unknown>
-    extends CommonRowSelectionOptions<TData, TValue> {
+export interface MultiRowSelectionOptions<TData = any, TValue = any> extends CommonRowSelectionOptions<TData, TValue> {
     mode: 'multiRow';
     /**
      * Determine group selection behaviour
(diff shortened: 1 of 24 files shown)


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+    mode: 'cell';
+    /**
+     * Configuration options for the fill handle
+     */
+    fillHandleOptions?: FillHandleOptions<TData>;
+    /**
+     * If `true`, only a single range can be selected
+     * @default false
+     */
+    suppressMultiRangeSelection?: boolean;
```
AndrewGlazier, line 2446: I think 'selection' is implict now, suppressMultipleRanges might be a better name here

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+    /**
+     * Set to `true` to enable the Range Handle
+     * @default false
+     */
+    enableRangeHandle?: boolean;
```
AndrewGlazier, line 2451: I'm not too clear on this feature, are range handle/fill handle mutually exclusive? If so might be nice to conflate these into one configuration property to prevent them being enabled simultaneously

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+/**
+ * Determines selection behaviour when multiple rows can be selected at once.
+ */
+export interface RowSelectionOptions<TData, TValue> {
+    mode: 'row';
+    /**
+     * If `true`, rows will not be deselected if you hold down `Ctrl` and click the row or press `Space`.
+     * @default false
+     */
+    suppressRowDeselection?: boolean;
```
AndrewGlazier, line 2483: A few properties in here specify row, I think this isn't needed anymore due to explicitly stating mode is row

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+    /**
+     * If `true`, row selection won't happen when rows are clicked. Use when you only want checkbox selection.
+     * @default false
+     */
+    suppressRowClickSelection?: boolean;
+    /**
+     * If `true`, only a single row can be selected at a time. Selecting a row will de-select any other rows.
+     * @default false
+     */
+    suppressMultipleRowSelection?: boolean;
```
AndrewGlazier, line 2493: I think 'Multi' is more used in the API than Multiple

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+     * @default 'none'
+     */
+    groupSelection?: GroupSelectionOptions;
+    /**
+     * Determine checkbox selection behaviour
+     * @default false
+     */
+    checkboxSelection?: CheckboxSelectionOptions<TData, TValue>;
+    /**
+     * Determines how "select all" behaviour works both via the API (i.e. `selectAll`) and via header checkbox selection.
```
AndrewGlazier, line 2505: I'm not understanding the via API bit of this comment, will the API not accept a parameter determining how we'll select? It's good in my opinion for the API to add this flexibility in case people want to add buttons in their own UI with this flexibility

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+    /**
+     * Callback to be used to determine which rows are selectable. By default rows are selectable, so return `false` to make a row un-selectable.
+     */
+    isRowSelectable?: IsRowSelectable<TData>;
+}
+
+/**
+ * Determines whether checkboxes are displayed for selection
+ */
+export type CheckboxSelectionOptions<TData, TValue> =
```
AndrewGlazier, line 2527: I feel like this has mostly niche options, i.e showDisabledCheckboxes + CheckboxSelectionCallback aren't really that common, it would probably be good to have just `true` as an option, to bypass providing an object for a simple enable

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+    /**
+     * If `true`, only a single row can be selected at a time. Selecting a row will de-select any other rows.
+     * @default false
+     */
+    suppressMultipleRowSelection?: boolean;
+    /**
+     * Determine group selection behaviour
+     * @default 'none'
+     */
+    groupSelection?: GroupSelectionOptions;
```
AndrewGlazier, line 2498: A note on this, we're kind of creating a convention where `Options` was making me expect the options to be objects, this threw me for a loop when I realised this one wasn't - maybe this shouldn't be named `Options`?

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2414,3 +2426,138 @@ export interface LoadingCellRendererSelectorResult {
...
+/**
+ * Determines how "select all" behaviour works both via the API (i.e. `selectAll`) and via header checkbox selection.
+ */
+export interface SelectAllOptions {
+    /**
+     * When selecting all via either the API or a header checkbox, only those rows that satisfy the currently
+     * active filter will be selected.
+     * @default false
+     */
+    filteredOnly?: boolean;
```
AndrewGlazier, line 2556: filteredOnly & currentPageOnly are mutually exclusive, shouldn't they be merged?

community-modules/core/src/gridOptionsService.ts
```diff
@@ -249,6 +249,7 @@ export class GridOptionsService extends BeanStub implements NamedBean {
         });
 
         this.validationService?.processGridOptions(this.gridOptions);
+        this.validationService?.processLegacyOptions(options);
```
AndrewGlazier, line 252: This is doing a lot of mutation of the grid options, but won't be reactive in this case as all of these are hooked up after the events are generated.

Also it'd be better if these mutations happened here in the GOS, as we will want the validation service to be treeshakable so will need to refactor this out before it can be, as it's quite a large bit of code with a small amount of benefit for prod. (I'm aware the validation service does a couple of mutations already, but we'll remove those too at some point)

I might recommend instead using value getters , e.g  see getGrandTotalRow, rather than mutations, as this will have less risk of causing issues with reactivity, as we do some shallow checks etc in places which mutation will impact

community-modules/core/src/gridOptionsService.ts
```diff
@@ -615,3 +616,83 @@ export class GridOptionsService extends BeanStub implements NamedBean {
...
+                break;
+            case 'none':
+                port('groupSelectsChildren', undefined);
+                port('groupSelectsFiltered', undefined);
+                break;
+            default:
+                break;
+        }
+
+        const colDefs = go.columnDefs;
```
AndrewGlazier, line 669: I don't know about this, it looks like we're saying that in the next major the first col def will be the one with the checkbox, but this isn't the behaviour we're aiming for. If the column is hidden due to grouping (common in a lot of our examples) we're going to just lose the checkbox column entirely.

I'm not sure we should introduce this one until we're certain of how we want checkbox columns to work and ready to introduce them - this one might be better to leave as a col property until we have the fully fledged feature

eliasmalik: Yeah that's fair, I think this is the most awkward one and I struggled to think of how to handle it without actually just implementing AG-9484. I'll remove this part from this PR for now.

With that said,
> it looks like we're saying that in the next major the first col def will be the one with the checkbox, but this isn't the behaviour we're aiming for.

I'm imagining that this code will only exist in v32.x, and be removed before we do the v33 release. This is just saying "if you choose to use `SelectionOptions` in v32.x, you'll have to live with this behaviour until v33" (which I agree is not great. As you say, better to leave as a col option until we have another discussion about 9484).

community-modules/core/src/gridOptionsService.ts
eliasmalik: Yeah, probably safer. I was trying to avoid making any changes at all to code doing the "getting" of selection options, but possibly not worth it.

PR #8369, opened 2024-07-18
AG-12273 improve and optimize tree strategy

SalvatorePreviti:
changed enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy.ts

Generally:

- inline some methods
- use strings path instead of group info
- reduce the number of allocated objects
    - replace forEach with for loops
    - reduce recursion
    - reduce allocation of temporary objects and functions
    - optimize topologicalSort

Cache:

- simplify so that we need just a method cacheUpsert that returns the TreeNode.
- move the cache and its method inside the bean
- optimize so that we don't create the internal subtree for leafs (reduce allocations and GC work)
- avoid replacing nodes with null so inserting order [a,b,c] [a,b] can work

Also:
- added more tests and improved the tests


enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy.ts
```diff
@@ -676,60 +592,70 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
+            sortedNodes.push(found);
+            stillToFind.delete(id);
+        }
 
-    public get(path: GroupInfo[], level: number, key: string): RowNode | null | undefined {
-        const cache = this.traverse(path, level - 1);
-        return cache[key]?.node;
-    }
+        // we can stop early if we've already found all the nodes
+        if (stillToFind.size === 0) {
```
AndrewGlazier, line 649: Would we be able to drop the `stillToFind` set if we just looked for sortedNodes.length === rowNodes.length?

SalvatorePreviti: I really think it can be dropped, but I still need to do some more reasoning and testing and ask you more questions. I was thinking to drop it, but to do it in the next stage

PR #8376, opened 2024-07-22
AG-12336 - fix - tabtonextcell false with pagination

peterjrreynolds:


community-modules/core/src/gridComp/gridCtrl.ts
```diff
@@ -204,6 +202,35 @@ export class GridCtrl extends BeanStub {
...
+    public allowFocusForNextCoreContainer(up?: boolean): void {
+        const coreContainers = this.view.getFocusableContainers();
+        const { nextIndex, indexWithFocus } = this.getNextFocusableIndex(coreContainers, up);
+        if (indexWithFocus === -1 || nextIndex < 0 || nextIndex >= coreContainers.length) {
+            return;
+        }
+        const comp = coreContainers[nextIndex];
+        comp.setAllowFocus?.(true);
+        setTimeout(() => {
+            comp.setAllowFocus?.(false);
```
gportela85, line 214: Why do we need this to be async? Is there something we need to wait for?
Is there an event we can use instead?

PR #8377, opened 2024-07-22
AG-11923 suppressCellFocus updates

StephenCooper:
The aim is to update the `tabIndex` property for cells and fullwidth rows without triggering a React rerender. 

This is achieved by moving the management of the `tabIndex` attribute into the ctrls and out of the components.

### DIFF
--- a/community-modules/core/src/utils/dom.ts
+++ b/community-modules/core/src/utils/dom.ts
@@ -458,8 +458,8 @@ export function _iterateNamedNodeMap(map: NamedNodeMap, callback: (key: string, 
     }
 }
 
-export function _addOrRemoveAttribute(element: HTMLElement, name: string, value: any) {
-    if (value == null) {
+export function _addOrRemoveAttribute(element: HTMLElement, name: string, value: string | number | null | undefined) {
+    if (value == null || value === '') {
         element.removeAttribute(name);
     } else {
         element.setAttribute(name, value.toString());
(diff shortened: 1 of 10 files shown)


community-modules/core/src/utils/dom.ts
```diff
@@ -458,8 +458,8 @@ export function _iterateNamedNodeMap(map: NamedNodeMap, callback: (key: string,
     }
 }
 
-export function _addOrRemoveAttribute(element: HTMLElement, name: string, value: any) {
-    if (value == null) {
+export function _addOrRemoveAttribute(element: HTMLElement, name: string, value: string | number | null | undefined) {
+    if (value == null || value === '') {
```
StephenCooper, line 462: Makes it possible to use this in more places to reduce code duplication. 

Does not break the existing use cases of this method as they do not pass in the empty string.

community-modules/react/src/reactUi/cells/cellComp.tsx
```diff
@@ -501,7 +500,7 @@ const CellComp = (props: { cellCtrl: CellCtrl; printLayout: boolean; editingRow:
     );
 
     return (
-        <div ref={setRef} style={userStyles} tabIndex={tabIndex} role={cellAriaRole} col-id={colId}>
+        <div ref={setRef} style={userStyles} role={cellAriaRole} col-id={colId}>
```
StephenCooper, line 503: `tabindex` now managed directly via dom manipulation so not part of the template

PR #8381, opened 2024-07-22
AG-12177 Fix footer sharing same event service as sibling

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12177

### DIFF
--- a/community-modules/core/src/entities/rowNode.ts
+++ b/community-modules/core/src/entities/rowNode.ts
@@ -1170,7 +1170,7 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
 
         // we don't copy these properties as they cause the footer node
         // to have properties which should be unique to the row.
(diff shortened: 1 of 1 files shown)


community-modules/core/src/entities/rowNode.ts
```diff
@@ -1170,7 +1170,7 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
 
         // we don't copy these properties as they cause the footer node
         // to have properties which should be unique to the row.
-        const ignoredProperties = new Set(['eventService', '__objectId', 'sticky']);
+        const ignoredProperties = new Set(['localEventService', '__objectId', 'sticky']);
```
StephenCooper, line 1173: Can we apply types to this please to avoid this happening again in the future.

StephenCooper (approved):
Approved

PR #8389, opened 2024-07-24
AG-12309 overlays and sidebar focus fix

SalvatorePreviti:
In the previous release we introduced focusable overlays.
This however didn't work as intended as the situation is more complex than anticipated.

- There is a preexisting bug in the sidebar backward focus that makes it enter in an infinite focus loop if there is no cell or no column selectable, see [AG-12389](https://ag-grid.atlassian.net/browse/AG-12389), that this PR solves as well
- The Overlay cannot be a Focusable container because it lives inside the grid view, that is already a focusable container, and focusNextGridCoreContainer would not work on it, its focus must be managed instead inside the grid view
- Is not true that when the loading overlay other things cannot be focused: the sidebar and statusbar can be focused. When loading overlay, that is an exclusive overlay, only what is inside the grid view cannot be focused


community-modules/core/src/focusService.ts
```diff
@@ -707,7 +738,9 @@ export class FocusService extends BeanStub implements NamedBean {
         const nextRow = backwards ? this.rowPositionUtils.getLastRow() : this.rowPositionUtils.getFirstRow();
 
         if (!nextRow) {
-            return false;
+            return backwards
+                ? (canFocusOverlay && this.focusOverlay(backwards)) || this.focusLastHeader()
+                : canFocusOverlay && this.focusOverlay(backwards);
```
peterjrreynolds, line 743: Would be good to simplify this:
```
const overlayFocused = canFocusOverlay && this.focusOverlay(backwards);
return overlayFocused || (backwards && this.focusLastHeader());
```

community-modules/core/src/focusService.ts
```diff
@@ -718,7 +751,9 @@ export class FocusService extends BeanStub implements NamedBean {
         }
 
         if (rowIndex == null || !column) {
-            return false;
+            return backwards
+                ? (canFocusOverlay && this.focusOverlay(backwards)) || this.focusLastHeader()
+                : canFocusOverlay && this.focusOverlay(backwards);
```
peterjrreynolds, line 756: Duplication

community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
```diff
@@ -37,6 +42,31 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
             </div>`);
     }
 
+    private handleKeyDown(e: KeyboardEvent): void {
+        if (!this.isDisplayed() || !this.eOverlayWrapper || e.defaultPrevented) {
```
peterjrreynolds, line 46: Although it's not currently going to happen here, should generally check `_isStopPropagationForAgGrid(e)`.

Usually we do this by using the `ManagedFocusFeature` to add the event handling logic.

gportela85: I don't think we need this check at all, maybe
```
if (e.defaultPrevented || _isStopPropagationForAgGrid(e)) {
    return;
}
```

but how would the keydownEvent be fired within the overlayWrapper when it's not displayed?

community-modules/core/src/rendering/overlays/overlayWrapperComponent.ts
```diff
@@ -37,6 +42,31 @@ export class OverlayWrapperComponent extends Component implements LayoutView {
...
+            return;
+        }
+
+        if (e.key !== KeyCode.TAB) {
+            return;
+        }
+
+        // If we are pressing shift+tab on the first or last (if backwards) focusable element in the overlay, we need to move focus outside
+        const activeElement = this.gos.getActiveDomElement();
```
gportela85, line 55: It think this method should be (I have not tested this but seems right 😶‍🌫️  ):
```
private handleKeyDown(e: KeyboardEvent): void {
   if (e.defaultPrevented || _isStopPropagationForAgGrid(e) || e.key !== KeyCode.TAB) {
        return; 
    }

    const nextEl = this.focusService.findNextFocusableElement(this.eOverlayWrapper, false, e.shiftKey);
    if (nextEl) {
        return; 
     }
     let isFocused = false;
     if (e.shiftKey) {
          isFocused = this.focusService.focusGridView(_last(this.visibleColsService.getAllCols()), true, false));
     } else {
         isFocused = this.focusService.focusNextGridCoreContainer(false);
     }

     if (isFocused) {
          e.preventDefault();
     }
}
```

SalvatorePreviti: I tried this before writing this but it didn't work, however, findNextFocusableElement does call findFocusableElements inside. Let me try again

SalvatorePreviti: fixed! thank you

PR #8393, opened 2024-07-24
AG-12395 tree data wrong level

SalvatorePreviti:
The logic that computed the tree node level was inconsistent.
Tree nodes should start with level 0 (as -1 is the root)

- fix the treeStrategy to compute and update the level correctly
- fix the snapshot test data to match the correct result
- improve the tests by adding a checkTreeDiagram that checks nodes level and parent and prints the tree structure if invalid
image


enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -448,6 +448,7 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
             }
 
             parentGroup = row;
+            row.level = level;
```
AndrewGlazier, line 451: Just curious, why do we need this one? I see the issue on 470 but not sure what this is solving

SalvatorePreviti: for the "move" - without this move transactions don't work. Considering we are going to rewrite everything this simple fix here solves the issue, do you see problems with it?

PR #8401, opened 2024-07-25
AG-12406 use shared attribute method

StephenCooper:
Reduce duplicated code logic when there is an existing utils method that can be reused.


community-modules/core/src/headerRendering/cells/column/headerCellComp.ts
```diff
@@ -33,15 +33,7 @@ export class HeaderCellComp extends AbstractHeaderCellComp<HeaderCellCtrl> {
...
-        const setAttribute = (name: string, value: string | null | undefined) => {
-            if (value != null && value != '') {
-                eGui.setAttribute(name, value);
-            } else {
-                eGui.removeAttribute(name);
-            }
-        };
-
-        setAttribute('col-id', this.column.getColId());
+        eGui.setAttribute('col-id', this.column.getColId());
```
StephenCooper, line 36: getColId is never null, so we do not need the remove logic.

PR #8412, opened 2024-07-26
AG-12399 Support nested properties in API reference parser

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12399


documentation/update-algolia-indices/src/generators/api-refs.ts
```diff
@@ -83,6 +107,24 @@ export const parseApiPageData = (details: APIPageData): AlgoliaRecord[] => {
...
             });
+
+            if (addNestedFields) {
+                for (const key in data.type.nested ?? {}) {
+                    if (IGNORE_FIELDS.has(key)) {
+                        continue;
+                    }
+                    records.push({
+                        source: 'api',
+                        objectID: path,
```
AndrewGlazier, line 118: I think this might need to be unique

PR #8413, opened 2024-07-26
AG-12500 TreeStrategy rewrite

SalvatorePreviti:
TreeStrategy rewrite.

We keep a secondary tree data structure together with the rows.
We associate a node with a TreeNode, both storing the row in node.row and by storing the TreeNode in a hidden field in the row.
We break the association when the row is removed or the TreeStrategy destroyed.

TreeStrategy uses a two stage approach both for first time creation and updates.
Multiple updates interact with the tree, and a commit stage commits all updates reducing expensive computations.

The operations are:
  - create a path with a filler node with a null row, calling TreeStrategy.upsertPath
  - replace an existing node row with another, or null, with TreeStrategy.overwriteRow

Insert and overwrite will invalidate the affected paths with node.invalidate(), so that the commit operation will only
update the affected paths without traversing the whole tree.
Consider that order of invalidated items is not deterministic, so the commit operation should be able to handle any order.
The subtrees that don't have children after a move or remove operation will be marked as ghosts and removed during commit.

During commit, the childrenAfterGroup and allLeafChildren arrays are rebuilt, and the updates are applied.
The ghost nodes are removed.
Before commit those arrays are NOT representing the truth, so they should not be used.

TODO:

At the moment row ordering cannot be consistent in some scenarios, and we need to do some broad brainstorming around it. The main problem is that filler rows do not appear and cannot currently appear in the rowOrdering map. And there is currently no way to properly reorder rows especially filler rows, the current implementation based on __objectId is incorrect.
We need to review how the rowOrdering works in general and for group and especially trees, if we need to maintain it or not, and especially decide a new strategy for ordering those filler nodes. Is a bigger problem and once solved it will probably prompt further improvements and refactoring in the TreeStrategy and TreeNode classes.
The proposed TreeStrategy implementation can be improved, in performance and reduced complexity, once we agree on how to handle row reordering and ordering

Note: there are some changed tests that are not related to trees, this is because I introduced strict null checks in the tsconfig file for tests


community-modules/core/src/entities/rowNode.ts
```diff
@@ -158,7 +158,7 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
     public allChildrenCount: number | null;
 
     /** Children mapped by the pivot columns. */
-    public childrenMapped: { [key: string]: any } | null = {};
+    public childrenMapped: { [key: string]: any } | null = null;
```
SalvatorePreviti, line 161: NOTE: this is to save some memory per node, especially when not using grouping or treeData

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -171,527 +212,302 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
-                this.setTreeNodeRow(treeNode, row);
+        node: TreeNode,
+        parent: TreeNode | null,
+        level: number
+    ): CommitFlags {
+        // First, commit children
+
+        let childrenFlags: CommitFlags = CommitFlags.None;
+        const childLevel = level + 1;
+        for (;;) {
```
gportela85, line 325: As standard we use `while(true)` instead of `for (;;)`

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -171,527 +148,440 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
-                batchRemover.removeFromChildrenAfterGroup(child.parent, child);
-            } else {
-                _removeFromArray(child.parent.childrenAfterGroup!, child);
-                child.parent.updateHasChildren();
+    /**
+     * Gets the last node of a path. Inserts filler nodes where needed.
+     * Note that invalidate() is not called, is up to the caller to call it if needed.
+     */
+    private upsertPath(path: string[]): TreeNode | null {
+        for (let level = 0, parent: TreeNode | null = this.root, stop = path.length - 1; level <= stop; ++level) {
```
peterjrreynolds, line 234: nit - I think this would be more readable with `parent` and `stop` defined above this (scoping doesn't matter in this particular case).

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -171,527 +148,440 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
+        const { row: oldRow, ghost } = node;
 
-    private shotgunResetEverything(details: TreeGroupingDetails, afterColumnsChanged: boolean): void {
-        if (afterColumnsChanged || this.oldGroupDisplayColIds === undefined) {
-            const newGroupDisplayColIds =
-                this.showRowGroupColsService
-                    .getShowRowGroupCols()
-                    ?.map((c) => c.getId())
-                    .join('-') ?? '';
+        const prevNode = newRow?.treeNode as TreeNode | null;
```
peterjrreynolds, line 265: This could be `undefined`

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -171,527 +148,440 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
-    // performance: keep track of the nodes we haven't found yet so we can return early
-    const stillToFind = new Set<string>();
-
-    for (let i = 0; i < rowNodes.length; i++) {
-        const row = rowNodes[i];
-        const id = row.id!;
-        idLookup.set(id, row);
-        stillToFind.add(id);
+        let id = node.level + '-' + node.key;
+        for (let p = node.parent; p?.parent; p = p.parent) {
```
peterjrreynolds, line 500: nit - not a fan of using `for` loops like this

peterjrreynolds:
Title needs updating to the correct Jira (current one has been closed out now)

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeNode.ts
```diff
@@ -0,0 +1,306 @@
...
+ * Insert and overwrite will invalidate the affected paths with node.invalidate(), so that the commit operation will only
+ * update the affected paths without traversing the whole tree.
+ * Consider that order of invalidated items is not deterministic, so the commit operation should be able to handle any order.
+ * The subtrees that don't have children after a move or remove operation will be marked as ghosts and removed during commit.
+ *
+ * During commit, the childrenAfterGroup and allLeafChildren arrays are rebuilt, and the updates are applied.
+ * The ghost nodes are removed.
+ * Before commit those arrays are NOT representing the truth, so they should not be used.
+ */
+export class TreeNode {
```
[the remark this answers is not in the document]
SalvatorePreviti: In this case we need to export it from core - should it be named _ITreeNode as is for internal use only?

PR #8423, opened 2024-07-29
AG-12433 Angular Consistent event firing order

StephenCooper:
The use of a promise for GridReady meant that event got fired async even though the grid was ready for it to be fired. Swapping the code to hold all events until ngAfterViewInit has been run so that we maintain a consistent ordering of events.

### DIFF
--- a/community-modules/core/src/entities/gridOptions.ts
+++ b/community-modules/core/src/entities/gridOptions.ts
@@ -5,7 +5,7 @@ import type { AgChartTheme, AgChartThemeOverrides } from 'ag-charts-types';
 
 import type {
     AdvancedFilterBuilderVisibleChangedEvent,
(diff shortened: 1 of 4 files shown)


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2190,7 +2190,7 @@ export interface GridOptions<TData = any> {
     /**
      * Async transactions have been applied. Contains a list of all transaction results.
      */
-    onAsyncTransactionsFlushed?(event: AsyncTransactionsFlushed<TData>): void;
+    onAsyncTransactionsFlushed?(event: AsyncTransactionsFlushedEvent<TData>): void;
```
StephenCooper, line 2193: This was using the old deprecated event interface.

PR #8429, opened 2024-07-29
AG-9947 - forced menuUtils to restore focus async

gportela85:
https://ag-grid.atlassian.net/browse/AG-9947


enterprise-modules/menu/src/menu/menuUtils.ts
```diff
@@ -54,7 +54,14 @@ export class MenuUtils extends BeanStub implements NamedBean {
             return;
         }
 
-        this.focusHeaderCell(restoreFocusParams);
+        // requires async for react applications
+        setTimeout(() => {
```
StephenCooper, line 58: Can we pull this logic into the focus service itself? That way these setTimeouts are not littered across the entire code base?

gportela85: Humm.. this is a complicated one to change in the `focusService`. `MenuUtils.focusHeaderCell` calls `FocusService.focusHeaderPosition` which returns a boolean. In the MenuUtils we don't do anything with this return value, but in the `HeaderNavigationService` we do use it.
To fix this in the `FocusService` we would have to do one of:
1 - Change `focusHeaderPosition` to return a Promise and update every method that uses it (breaking change because it's used by the API).
2 - Create a new method `focusHeaderPositionAsync` that I'm not sure it's much help. But it would keep the code within the FocusService.

Do you have a preference?

StephenCooper: Option 2 please. Main idea is to keep them all together in the focus service.

StephenCooper: Ok, happy for this to stay here in that case then.

StephenCooper (approved):

PR #8432, opened 2024-07-30
AG-11247 - fix no click editing lose focus

peterjrreynolds:


community-modules/core/src/edit/editApi.ts
```diff
@@ -54,8 +54,17 @@ export function startEditingCell(beans: BeanCollection, params: StartEditingCell
...
-        beans.focusService.setFocusedCell(cellPosition);
+    const { focusService, gos } = beans;
+    const isFocusWithinCell = () => {
+        const activeElement = gos.getActiveDomElement();
+        const eCell = cell.getGui();
+        return activeElement !== eCell && eCell.contains(activeElement);
+    };
+    if (gos.get('stopEditingWhenCellsLoseFocus') && isFocusWithinCell()) {
+        // force focus back on the cell so it doesn't get lost
+        focusService.setFocusedCell({ ...cellPosition, forceBrowserFocus: true });
```
gportela85, line 65: since we already call `ensureColumnVisible` and `ensureIndexVisible` we should probably set `preventScrollOnBrowserFocus: true` here.

gportela85: Also, since we don't call this code many times in a row, it might be good to simplify it to:

```
focusService.setFocusedCell({
   ...cellPosition,
   forceBrowserFocus: gos.get('stopEditingWhenCellsLoseFocus') && isFocusWithinCell()
   preventScrollOnBrowserFocus: true
});
```

peterjrreynolds: Done. Still need to wrap it in an if block so that we don't fire the focus event when we don't need it.

PR #8442, opened 2024-07-31
AG-12283 Cannot edit tree data group rows without setting `enableGroupEdit`

SalvatorePreviti:
Tree groups should be editable by default.
The check for editable tree group was not correct.

It seems that in version 32 the "group" field of the row is now correct for tree row groups, but it wasn't before


StephenCooper (commented):
Makes sense.

StephenCooper (approved):

PR #8459, opened 2024-08-04
AG-12485 Protect against duplicate data passed to pinned rows

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12485


community-modules/core/src/pinnedRowModel/pinnedRowModel.ts
```diff
@@ -97,10 +98,28 @@ export class PinnedRowModel extends BeanStub implements NamedBean {
...
+                _warnOnce(
+                    'Duplicate ID',
+                    id,
+                    'found for pinned row with data',
+                    data,
+                    'When `getRowId` is defined, it must return unique IDs for all pinned rows. Use the `rowPinned` parameter.'
+                );
+                continue;
+            } else {
+                i++;
```
gportela85, line 119: if there is a `continue` in the `if` above, there is no need for `else`.
The reest of the code looks good.

PR #8509, opened 2024-08-07
AG-9484 Dedicated checkbox column

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-9484

### DIFF
--- a/community-modules/core/src/columns/columnModel.ts
+++ b/community-modules/core/src/columns/columnModel.ts
@@ -1091,26 +1091,10 @@ export class ColumnModel extends BeanStub implements NamedBean {
         prev: SelectionOptions | undefined,
         source: ColumnEventType
     ) {
-        if (!prev) {
-            if (current && current.mode !== 'cell' && current.checkboxColumn) {
-                // Selection options were previously undefined but now contain checkboxColumn options
-                return this.refreshAll(source);
-            }
-        } else if (prev.mode === 'cell') {
-            if (current && current.mode !== 'cell' && current.checkboxColumn) {
-                // Selection options were in cell mode but now are in row mode with checkboxColumn options
-                return this.refreshAll(source);
-            }
-        } else if (!prev.checkboxColumn) {
-            if (current && current.mode !== 'cell' && current.checkboxColumn) {
-                // Selection options did not have checkboxColumn options but now do have them
-                return this.refreshAll(source);
-            }
-        } else if (prev.checkboxColumn) {
-            if (!current || current.mode === 'cell') {
-                // Selection options had checkboxColumn enabled but now are in cell mode or disabled
-                return this.refreshAll(source);
-            }
+        const prevCheckbox = prev?.mode !== 'cell' ? prev?.checkboxes : undefined;
+        const currCheckbox = current?.mode !== 'cell' ? current?.checkboxes : undefined;
+        if (prevCheckbox !== currCheckbox) {
+            this.refreshAll(source);
         }
     }
 }
--- a/community-modules/core/src/columns/columnModel.ts
+++ b/community-modules/core/src/columns/columnModel.ts
@@ -34,7 +34,7 @@ import type { ColumnSizeService } from './columnSizeService';
 import { GROUP_AUTO_COLUMN_ID, isColumnControlsCol } from './columnUtils';
 import { destroyColumnTree, getColumnsFromTree, isColumnGroupAutoCol } from './columnUtils';
 import type { ColumnViewportService } from './columnViewportService';
-import type { IControlsColService } from './controlColService';
+import type { IControlsColService } from './controlsColService';
 import type { FuncColsService } from './funcColsService';
 import type { PivotResultColsService } from './pivotResultColsService';
 import type { VisibleColsService } from './visibleColsService';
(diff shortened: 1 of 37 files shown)


community-modules/core/src/columns/controlColService.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        if (!so) {
+            return [];
+        }
+
+        if (so.mode === 'cell') {
+            return [];
+        }
+
+        if (so.checkboxSelection || (so.mode === 'multiRow' && so.headerCheckbox)) {
```
AndrewGlazier, line 27: Unsure about whether we need `so.mode === 'multiRow' && so.headerCheckbox`, or if this would simply be better as a validationService warning message

community-modules/core/src/columns/columnModel.ts
```diff
@@ -142,7 +147,7 @@ export class ColumnModel extends BeanStub implements NamedBean {
         }
 
         this.addManagedPropertyListeners(
-            ['groupDisplayType', 'treeData', 'treeDataDisplayType', 'groupHideOpenParents'],
+            ['groupDisplayType', 'treeData', 'treeDataDisplayType', 'groupHideOpenParents', 'selectionOptions'],
```
AndrewGlazier, line 150: Is there any way we can make this more selective? `selectionOptions` themselves may not impact the columns at all, only the subset options right?

community-modules/core/src/columns/controlColService.ts
eliasmalik: This is more about typescript not letting me access `headerCheckbox` unless I assert we are in `'multiRow'` mode

AndrewGlazier: I get that, but my point is, should `headerCheckbox` be implicitly enabling `checkboxSelection`, or maybe `checkboxSelection` should be an object with headerCheckbox as a child if this is a dependent relationship?

community-modules/core/src/columns/columnModel.ts
```diff
@@ -1038,6 +1085,34 @@ export class ColumnModel extends BeanStub implements NamedBean {
             this.autoColService!.updateAutoCols(this.autoCols.list, source);
         }
     }
+
+    private onSelectionChanged(
+        current: SelectionOptions | undefined,
+        prev: SelectionOptions | undefined,
+        source: ColumnEventType
+    ) {
+        if (!prev) {
```
AndrewGlazier, line 1094: Would this be able to simplify to something like:
const prevCheckbox = prev?.mode !== 'cell' ? prev?.checkboxColumn : undefined;
const currCheckbox = curr?.mode !== 'cell' ? curr?.checkboxColumn : undefined;
if (prevCheckbox !== currCheckbox) {
   refreshAll();
}

community-modules/core/src/columns/controlsColService.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const CONTROLS_COLUMN_ID_PREFIX = 'ag-Grid-ControlsColumn' as const;
+
+export class ControlsColService extends BeanStub implements NamedBean, IControlsColService {
+    beanName = 'controlsColService' as const;
+
+    public createControlsCols(): AgColumn[] {
+        const so = this.gos.get('selection');
+        const controlsColDef = this.gos.get('controlsColDef');
```
AndrewGlazier, line 18: Can you move this/enableRTL down to the relevant scope

community-modules/core/src/columns/columnModel.ts
```diff
@@ -1038,6 +1085,18 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
         }
     }
+
+    private onSelectionChanged(
+        current: SelectionOptions | undefined,
+        prev: SelectionOptions | undefined,
+        source: ColumnEventType
+    ) {
+        const prevCheckbox = prev?.mode !== 'cell' ? prev?.checkboxes : undefined;
+        const currCheckbox = current?.mode !== 'cell' ? current?.checkboxes : undefined;
```
AndrewGlazier, line 1095: Add headerCheckbox reactivity

community-modules/core/src/headerRendering/cells/column/selectAllFeature.ts
```diff
@@ -31,7 +34,24 @@ export class SelectAllFeature extends BeanStub {
...
+        this.selectionOptions = this.gos.get('selection');
+        this.addManagedPropertyListener('selection', (e) => (this.selectionOptions = e.currentValue));
+    }
+
+    /**
+     * Select all is enabled by default when using the legacy selection API.
+     * When using the new selection API, select all is only enabled for the checkbox column
+     */
+    private isEnabled(): boolean {
+        return this.selectionOptions === undefined || isColumnControlsCol(this.column);
```
AndrewGlazier, line 47: Remove check so we can support new API with deprecated col def options until we remove or undeprecate

community-modules/core/src/columns/controlsColService.ts
```diff
@@ -0,0 +1,60 @@
...
+                comparator(valueA, valueB, nodeA, nodeB) {
+                    const aSelected = nodeA.isSelected();
+                    const bSelected = nodeB.isSelected();
+                    return aSelected && bSelected ? 0 : aSelected ? 1 : -1;
+                },
+                suppressFillHandle: true,
+                // overrides
+                ...controlsColDef,
+                // non-overridable properties
+                editable: false,
```
AndrewGlazier, line 50: Default, but overridable

PR #8515, opened 2024-08-07
AG-12503 - tidy up gos

peterjrreynolds:


community-modules/core/src/dragAndDrop/dragAndDropService.ts
```diff
@@ -6,6 +6,7 @@ import type { IAggFunc } from '../entities/colDef';
 import type { Environment } from '../environment';
 import type { MouseEventService } from '../gridBodyComp/mouseEventService';
 import type { RowDropZoneParams } from '../gridBodyComp/rowDragFeature';
+import { _getDocument, getRootNode } from '../gridOptionsUtils';
```
StephenCooper, line 9: Is the pattern only use an _ prefix for the utils that have to be exported from main.ts?

Would it be simpler just to always use the _ prefix?

enterprise-modules/server-side-row-model/src/serverSideRowModel/stores/storeUtils.ts
```diff
@@ -142,7 +142,7 @@ export class StoreUtils extends BeanStub implements NamedBean {
     }
 
     private assertRowModelIsServerSide(key: keyof GridOptions) {
-        if (!this.gos.isRowModelType('serverSide')) {
+        if (!_isRowModelType(this.gos, 'serverSide')) {
```
StephenCooper, line 145: Wonder if we should make the specific methods so don't have to pass the string each time these are called. Might not make enough of a difference though.

community-modules/core/src/gridOptionsService.ts
```diff
@@ -591,27 +305,11 @@ export class GridOptionsService extends BeanStub implements NamedBean {
...
-        }
-
-        return (params) => {
-            let id = getRowId(params);
-
-            if (typeof id !== 'string') {
-                _warnOnce(`The getRowId callback must return a string. The ID `, id, ` is being cast to a string.`);
-                id = String(id);
-            }
+    public assertModuleRegistered(moduleName: ModuleNames, reason: string): boolean {
```
StephenCooper, line 308: Is this not a util, so that gridId is not part of the public interface to make that interface easier to share?

Otherwise, these would be good as utils too.

StephenCooper (commented):
Some comments, this drops 2kb off core due to those method names now being minifiyable

community-modules/core/src/dragAndDrop/dragAndDropService.ts
peterjrreynolds: Updated all to use the prefix

enterprise-modules/server-side-row-model/src/serverSideRowModel/stores/storeUtils.ts
peterjrreynolds: Added specific methods. Also updated everything that was checking `this.rowModel.getType()` to use the new functions.

community-modules/core/src/gridOptionsService.ts
peterjrreynolds: Yes, trying to make `GridOptionsService` a generic property service, so didn't want to publicly expose grid ID.

StephenCooper (approved):
Looking good.

PR #8522, opened 2024-08-08
AG-12497 fix row ordering for tree data

SalvatorePreviti:
By adding a new property indexInRowData we can avoid passing the rowNodeOrder to the grouping stage.
Furthermore, rowNodeOrder wasn't passed consistently, while indexInRowData can be kept consistent and always contains the index in of the row in root.allLeafChildren

Changes in IRowNode:

Now we expose the new readonly field indexInRowData to the user.

Changes in CSRM related classes:

Remove thje rowNodeOrder map.

Provide instead a new flagrowNodesOrderChanged to the grouping stage that is true when the order changes because rows were inserted (and not appended) or because a new rowData was set with a different ordering.

Simplify the logic that executes the new rowData

Keep the indexInRowData always up to date for any transaction

Optimizes sorting after rowData is overwritten, as we can order the rows in O(n) instead of O(n log n)

Changes in GroupStrategy:
we now use sortGroupChildren in the GroupStage to sort childrenAfterGroup, that uses indexInRowData

Changes in TreeStrategy:

Unfortunately we do NOT receive any transaction update for when the order changes but rows are not updated (for example, const rowData = [ ... ]; api.setGridOption('rowData', rowData); rowData.reverse(); api.setGridOption('rowData', rowData); does not send any update, but the tree needs to invalidate all the children that have changed their position.

Remove ghost and ghosts counter, as it was used to try to keep the order consistent when rowNodeOrder was undefined. Since we always have the ordering, this is now not needed and we can remove it. Instead, empty filler nodes are removed in the commit stage.

Since we have the order in indexInRowData, we can simplify sorting logic and childrenAfterGroup update

As agreed, the behaviour of ordering is: group indexInRowData specify the order in which the group and all its children appear. Filler nodes instead inherit the indexInRowData from the leftmost group/leaf

The order of duplicate rows matters, so now we swap the row with duplicate row with the smallest indexInRowData on commit.

Fixes AG-12591, we copy the expanded state if a new row is replacing an old row and isGroupOpenByDefault callback is not provided

Testing:

A specific behavioural unit test was added for row data ordering, written before the changes, and verified that it produces the same output with the changes in CSRM related files.
Optimizations and refactoring:

More optimizations and refactoring:
- Reuses the filler nodes, as now the order of creation is not important anymore, we are not using __objectId to compare them
- Move the logic to compute childrenAfterGroup and allLeafChildren to the node itself.
- Execute the sorting only if strictly needed, at the commit stage
- Update the allLeafChildren only if rows are added, updated or deleted but not if only sorting changes
- Improve performance by using a frozen single empty array, as leafs won't have children (this reduces the allocations considerably)
- Improve performance by using the same array for allLeafChildren if is the same as the only child present
- Simplify clearing and destruction, simplify code
- Change RowNode class type so that childrenAfterGroup and allLeafChildren are readonly, and use subtyping only where those are supposed to be set.
- Fix the computation of expanded, so leaves are not expanded

Tested that this example https://plnkr.co/edit/6VjWZDfjOqlK2UKu?open=main.js now takes between 1.2 to 2.2 seconds on my machine for 1200000 rows


community-modules/client-side-row-model/src/clientSideRowModel/immutableService.ts
```diff
@@ -52,79 +52,85 @@ export class ImmutableService extends BeanStub implements NamedBean, IImmutableS
...
 
-    public setRowData(rowData: any[]): void {
-        const transactionAndMap = this.createTransactionForRowData(rowData);
-        if (!transactionAndMap) {
-            return;
+    public setRowData<TData>(rowData: TData[]): void {
+        // convert the data into a transaction object by working out adds, removes and updates
+
+        // If true, we will not apply the new order specified in the rowData, but keep the old order.
+        const suppressSortOrder = this.gos.get('suppressMaintainUnsortedOrder');
```
AndrewGlazier, line 59: Can this move down to where it's used

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -14,14 +14,31 @@ import {
...
 } from '@ag-grid-community/core';
 
 const ROOT_NODE_ID = 'ROOT_NODE_ID';
 const TOP_LEVEL = 0;
+
+interface ClientSideRowNode extends RowNode {
+    indexInRowData: number;
+}
+
+interface ClientSideRootNode extends RowNode {
```
peterjrreynolds, line 27: Would be good to add a comment on this and `ClientSideRowNode` so that we know what it's used for

community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -48,6 +48,10 @@ enum RecursionType {
     PivotNodes,
 }
 
+interface RootNode extends RowNode {
```
peterjrreynolds, line 51: Also comment here

community-modules/core/src/entities/rowNode.ts
```diff
@@ -147,11 +153,25 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
     /** Used by server side row model, true if this node needs refreshed by the server when in viewport */
     public __needsRefreshWhenVisible: boolean;
 
-    /** All lowest level nodes beneath this node, no groups. */
-    public allLeafChildren: RowNode<TData>[] | null;
+    /**
+     * This is the index of this row in rootNode.allLeafChildren. It need to be kept consistent.
+     * If is -1, the row is not in rootNode.allLeafChildren, for example, the root node or filler nodes for tree data.
+     */
+    public readonly indexInRowData: number = -1;
```
StephenCooper, line 160: Can we have the comment that explains that this can actually be modified outside of RowNode via the extended interfaces. 

Idea being to avoid confusion / incorrect assumption in the future that this is only set in the constructor of the RowNode

enterprise-modules/row-grouping/src/rowGrouping/groupStage/groupStrategy/groupRow.ts
```diff
@@ -0,0 +1,6 @@
+import type { RowNode } from '@ag-grid-community/core';
+
+export interface GroupRow extends RowNode {
```
peterjrreynolds, line 3: Also here

community-modules/client-side-row-model/src/clientSideRowModel/immutableService.ts
```diff
@@ -52,79 +52,85 @@ export class ImmutableService extends BeanStub implements NamedBean, IImmutableS
...
+        this.clientSideRowModel.afterImmutableDataChange(rowNodeTransaction, orderChanged || rowsInserted);
     }
 
-    // converts the setRowData() command to a transaction
-    private createTransactionForRowData(
-        rowData: any[]
-    ): [RowDataTransaction, { [id: string]: number } | undefined] | undefined {
+    /** Converts the setRowData() command to a transaction */
+    private createTransactionForRowData<TData>(rowData: TData[]): RowDataTransaction | null {
         if (_missing(this.clientSideRowModel)) {
```
StephenCooper, line 83: We have a util function `_isClientSideRowModel(this.gos)` that could be used here.

community-modules/core/src/interfaces/iRowNode.ts
```diff
@@ -162,6 +162,13 @@ interface BaseRowNode<TData = any> {
     /** Used by server side row model, `true` if this row node failed a load. */
     failedLoad: boolean | undefined;
 
+    /**
+     * This is the index of the row in the source rowData array and it is kept consistent also during transactions.
+     * It does not change when sorting, filtering, grouping, pivoting or any other UI related operation.
+     * It is -1 if this node is a filler node in tree data or grouping.
+     */
+    readonly indexInRowData: number;
```
StephenCooper, line 170: Can we rename to `sourceRowIndex`.

Suggested update to the comment.

```
     * The index of the row in the source rowData array including any updates via transactions.
     * It does not change when sorting, filtering, grouping, pivoting or any other UI related operations.
     * If this is a filler node (a visual row created by AG Grid in tree data or grouping) the value will be `-1`.
```

PR #8545, opened 2024-08-14
AG-12457 handle duplicates gracefully in trees

SalvatorePreviti:
When a group contains duplicate keys, we need to keep them in the tree node, to allow proper warning and updates.
This is especially needed when we have transactions that can for a short time a duplicate (for example, swapping the path of two nodes).

- add a nullable set of duplicates in the node
- properly manage the state of replacing a filler with a real node, adding a duplicate, setting the main row, and popping a row from duplicates when the main row is moved or removed


enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeNode.ts
```diff
@@ -33,20 +33,30 @@ const getRowIndex = (row: RowNode | null | undefined, rowNodeOrder: RowNodeOrder
 /** Compare two RowNode by the TreeNode rowPosition. Assumes TreeNode to be set and valid. */
 const rowPositionComparer = (a: RowNode, b: RowNode): number => a.treeNode!.rowPosition - b.treeNode!.rowPosition;
 
+const rowUnlinked = (row: TreeRow, root: boolean): void => {
```
AndrewGlazier, line 36: Can you rename this? Is this 'orphaning' a row?

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeNode.ts
```diff
@@ -169,40 +185,82 @@ export class TreeNode implements ITreeNode {
...
+            } else {
+                this.row = null;
+            }
+        } else {
+            // Delete from the duplicate rows
+            if (!duplicateRows?.delete(rowToRemove)) {
+                return false; // Not found
+            }
+            if (duplicateRows.size === 0) {
+                this.duplicateRows = null; // Free memory
```
AndrewGlazier, line 260: Not sure when we reach this, is this meant to be inside of the delete?

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeStrategy.ts
```diff
@@ -211,12 +215,11 @@ export class TreeStrategy extends BeanStub implements IRowNodeStage {
...
-                const node = rows![i].treeNode as TreeNode | null;
+                const row = rows![i];
+                const node = row.treeNode as TreeNode | null;
                 if (node) {
-                    const oldRow = this.clearRow(node);
-                    if (oldRow !== null) {
-                        setTreeRowExpandedInitialized(oldRow, false);
-                    }
+                    this.removeRow(node, row);
+                    setTreeRowExpandedInitialized(row, false);
```
AndrewGlazier, line 222: Don't think we need to worry about resetting state on this node as it is a zombie if it has been removed and shouldn't be recoverable

PR #8559, opened 2024-08-15
AG-12583 remove row node order from RowNode transactions

SalvatorePreviti:
By adding a new property `indexInRowData` we can avoid passing the `rowNodeOrder` to the grouping stage.
Furthermore, rowNodeOrder wasn't passed consistently, while indexInRowData can be kept consistent and always contains the index in of the row in root.allLeafChildren

Changes in IRowNode:

Now we expose the new readonly field indexInRowData to the user.

Changes in CSRM related classes:

- Remove thje rowNodeOrder map.
- Provide instead a new flag`rowNodesOrderChanged` to the grouping stage that is true when the order changes because rows were inserted (and not appended) or because a new rowData was set with a different ordering.
- Simplify the logic that executes the new rowData
- Keep the indexInRowData always up to date for any transaction
- Optimizes sorting after rowData is overwritten, as we can order the rows in O(n) instead of O(n log n)

Changes in GroupStrategy:
 we now use sortGroupChildren in the GroupStage to sort childrenAfterGroup, that uses indexInRowData

Changes in TreeStrategy:

- Unfortunately we do NOT receive any transaction update for when the order changes but rows are not updated (for example, `const rowData = [ ... ];   api.setGridOption('rowData', rowData); rowData.reverse(); api.setGridOption('rowData', rowData);` does not send any update, but the tree needs to invalidate all the children that have changed their position.
- Remove ghost and ghosts counter, as it was used to try to keep the order consistent when rowNodeOrder was undefined. Since we always have the ordering, this is now not needed and we can remove it. Instead, empty filler nodes are removed in the commit stage.
- Since we have the order in indexInRowData, we can simplify sorting logic and childrenAfterGroup update
- As agreed, the behaviour of ordering is: group indexInRowData specify the order in which the group and all its children appear. Filler nodes instead inherit the indexInRowData from the leftmost group/leaf
- The order of duplicate rows matters, so now we swap the row with duplicate row with the smallest indexInRowData on commit.
- Fixes AG-12591, we copy the expanded state if a new row is replacing an old row and isGroupOpenByDefault callback is not provided

Testing:

A specific behavioural unit test was added for row data ordering, written before the changes, and verified that it produces the same output with the changes in CSRM related files.


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1095,13 +1095,14 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
                 this.groupStage.execute({
                     rowNode: this.rootNode,
                     rowNodeTransactions: rowNodeTransactions,
-                    rowNodeOrder: rowNodeOrder,
+                    rowNodeOrderChanged: !!rowNodeOrderChanged,
```
peterjrreynolds, line 1098: If the typing is correct, this should already be `boolean` (and the one below)

community-modules/core/src/interfaces/rowNodeTransaction.ts
```diff
@@ -7,4 +7,6 @@ export interface RowNodeTransaction<TData = any> {
     remove: IRowNode<TData>[];
     /** Row nodes updated */
     update: IRowNode<TData>[];
+    /** Is true if update caused node to be moved or new nodes are added */
+    orderChanged: boolean;
```
peterjrreynolds, line 11: Do we want to expose this to users, or should we have a separate internal interface?

enterprise-modules/row-grouping/src/rowGrouping/groupStage/treeStrategy/treeNode.ts
```diff
@@ -160,12 +118,20 @@ export class TreeNode implements ITreeNode {
...
 
+    public isEmptyFillerNode(): boolean {
+        const row = this.row;
+        if (row?.data) {
+            return false;
+        }
+        if (this.children?.size) {
+            return false;
+        }
+        return true;
```
peterjrreynolds, line 132: For the sake of brevity - `return !this.row?.data && !this.children?.size;`

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -102,7 +104,11 @@ export class ClientSideNodeManager {
             // we use rootNode as the parent, however if using ag-grid-enterprise, the grouping stage
             // sets the parent node on each row (even if we are not grouping). so setting parent node
             // here is for benefit of ag-grid-community users
-            rootNode.allLeafChildren = rowData.map((dataItem) => this.createNode(dataItem, this.rootNode, TOP_LEVEL));
+            rootNode.allLeafChildren = rowData.map((dataItem, index) => {
+                const node = this.createNode(dataItem, this.rootNode, TOP_LEVEL);
+                node.positionInRootChildren = index;
```
AndrewGlazier, line 109: This can live inside createNode, in both instances where we calling `createNode` we should have or be able to easily infer the index

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -129,6 +135,7 @@ export class ClientSideNodeManager {
             remove: [],
             update: [],
             add: [],
+            orderChanged: false,
```
AndrewGlazier, line 138: We shouldn't add this to this interface, as the transaction is a public interface

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -139,10 +146,20 @@ export class ClientSideNodeManager {
 
         this.updateSelection(nodesToUnselect, 'rowDataChanged');
 
+        let orderChanged = false;
         if (rowNodeOrder) {
```
AndrewGlazier, line 150: Is there a reason we aren't flattening this out? Weren't we going to stop generating this map seeing as we're updating the position on the node now?

community-modules/core/src/entities/rowNode.ts
```diff
@@ -153,8 +150,22 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
...
     public __needsRefreshWhenVisible: boolean;
 
+    /**
+     * This number represents the ordering of this row in root.allLeafChildren.
+     * This is NOT the index of the row, is a monotonic increasing number, so it may have gaps,
+     * furthermore, we try to not modify it if the order of rows did not change, to avoid unnecessary updates.
+     *
+     * This is used by the GroupStage to maintain the order of the nodes in the tree consistent (for trees and groups).
+     *
+     * This replaces the old rowNodeOrder map that was originally used for a similar purpose.
```
AndrewGlazier, line 160: No point commenting reference to old pattern

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -139,10 +146,20 @@ export class ClientSideNodeManager {
...
-            _sortRowNodesByOrder(this.rootNode.allLeafChildren, rowNodeOrder);
+            if (sortRowNodesByRowNodeOrderMap(this.rootNode.allLeafChildren, rowNodeOrder)) {
+                orderChanged = true;
+            }
+        }
+
+        if (orderChanged || rowDataTran.add?.length) {
+            if (updatePositionsInRootChildren(this.rootNode.allLeafChildren)) {
```
AndrewGlazier, line 157: We should only need to do this in two scenarios:
- Inside of `executeAdd`, and only for the rows occurring after the inserted rows. (we can calculate the index at insertion time for new rows)
- When provided a rowNodeOrder

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: It is a too big of a change to remove the rowNowOrder from this part of the clientSideNodeManager - there is some complex logic that, whilst is clear, and should be fine to refactor, is too much for now and should be part of a complete refactoring of CSRM

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: I agree, but I think we should do this in another refactoring of the CSRM. I tested that the performance impact of calling updatePositionsInRootChildren for even 1 million rows is negligible, is very fast indeed, compared to all the other expensive operations we are doing here

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -139,10 +146,20 @@ export class ClientSideNodeManager {
 
         this.updateSelection(nodesToUnselect, 'rowDataChanged');
 
+        let orderChanged = false;
         if (rowNodeOrder) {
-            _sortRowNodesByOrder(this.rootNode.allLeafChildren, rowNodeOrder);
+            if (sortRowNodesByRowNodeOrderMap(this.rootNode.allLeafChildren, rowNodeOrder)) {
```
AndrewGlazier, line 151: This change seems a bit of an over-optimization, why do we need extra logic here to restrict sorting, this was fine before - I'd prefer if we could cut this back to just the requirements as by the PR until there's a reason we need to optimize. I don't see much value in passing down the orderChanged flag as well - I'd rather we cut back on this and focused on the solution first

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: Defining a new interface `ClientSideRowNodeTransaction`

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: Tried to replicate the same behaviour as before, as GroupStrategy was performing the sort only if rowNodeOrder was present

SalvatorePreviti: I removed it - but this MIGHT have side effects with row grouping, and groups moving around as they will be sorted for every transaction - if this is desiderable I do not know

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -182,49 +222,83 @@ export class ClientSideNodeManager {
...
-            const len = allLeafChildren.length;
-            let normalisedAddIndex = addIndex;
+        const allChildrenCount = allLeafChildren.length;
+        let addIndex = rowDataTran.addIndex;
 
+        if (typeof addIndex !== 'number' || addIndex > allChildrenCount || Number.isNaN(addIndex)) {
+            addIndex = allChildrenCount;
+        }
+
+        if (addIndex < 0) {
```
AndrewGlazier, line 239: These checks are excessive, and if we were going to sanitize this is probably not the right place for it.

It's fine to use `addIndex` as we receive it

enterprise-modules/row-grouping/src/rowGrouping/groupStage/groupStrategy/groupStrategy.ts
```diff
@@ -161,6 +162,7 @@ export class GroupStrategy extends BeanStub implements IRowNodeStage {
                 this.moveNodesInWrongPath(tran.update as RowNode[], details, batchRemover);
             }
             if (_existsAndNotEmpty(tran.add)) {
+                nodesAdded = true;
```
AndrewGlazier, line 165: This seems redundant, don't we already set `rowNodesOrderChanged` from any transaction that added rows? It'll always be true anyway?

SalvatorePreviti: Only if inserted as insertion changes order, I am not sure if group would break if things were added at the end, this is why this is here, let me have a look!

SalvatorePreviti: all right i believe it should work, removing nodesAdded

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -118,32 +127,63 @@ export class ClientSideNodeManager {
...
+
+        let orderChanged = false;
+        let lastPosition = -1;
+        for (let rowIdx = 0, rowsLength = rows.length; rowIdx < rowsLength; ++rowIdx) {
+            const row = rows[rowIdx];
+
+            const newPosition = newRowDataIndexMap.get(row.data) ?? rowIdx;
+            row.positionInRootChildren = newPosition; // has to match the index of the new row data
+
+            if (newPosition < lastPosition) {
```
AndrewGlazier, line 175: Should this just be rowIdx !== newPosition?

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -118,32 +127,63 @@ export class ClientSideNodeManager {
...
+        rowData.forEach((data: TData, index) => {
+            newRowDataIndexMap.set(data, index);
+        });
+
+        let orderChanged = false;
+        let lastPosition = -1;
+        for (let rowIdx = 0, rowsLength = rows.length; rowIdx < rowsLength; ++rowIdx) {
+            const row = rows[rowIdx];
+
+            const newPosition = newRowDataIndexMap.get(row.data) ?? rowIdx;
```
AndrewGlazier, line 172: I think the ?? rowIdx is redundant and we'd be better using `newRowDataIndexMap.get(row.data)!`

In scenarios like this we tend to let exceptions throw, as it helps to catch bad scenarios - and I can't see a scenario where we should ever _not_ have an index here

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: Yeah you are right, assuming the positionInRootChildren is always correct, it must match the row index to be in order. simplifying the code

SalvatorePreviti: Ah no wait, that's not correct, if there are removal the ID could be different but we don't need to sort

community-modules/client-side-row-model/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: Fixed the behaviour to handle negative values as historically correct (append too).

PR #8576, opened 2024-08-19
fix: migrated fixes for CVE-2024-38996 to b23.2.1 to avoid prototype pollution 

Manthravadi:
Migrated the fixes present in PR #8250 so that it can be backported to version 23.2.1

Here is the link to the original PR that was merged:
https://github.com/ag-grid/ag-grid/commit/ff731699453f2632d4852b3a3c34b479c406068c

I made similar changes in branch `b23.2.1` as this is where the fix needs to be backported to and a patch needs to be released as `23.2.2`

### DIFF
--- a/community-modules/core/src/ts/utils/object.test.ts
+++ b/community-modules/core/src/ts/utils/object.test.ts
@@ -1,0 +1,39 @@
+import { cloneObject, deepCloneObject, mergeDeep } from './object';
+
+describe('object', () => {
+    test('mergeDeep does not allow prototype pollution', () => {
+        const BAD_JSON = JSON.parse('{"__proto__":{"polluted":true}}');
+        const victim = {};
+        try {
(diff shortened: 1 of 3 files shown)


Manthravadi:
@StephenCooper : Is there any update on this? It would be kind of you to review these changes and help us merge this PR. We are dependent on this version specifically and cannot update beyond this version due to large breaking change in the community package.

AG-Zoheil:
Hi @Manthravadi,

Thanks for getting in touch about this.

We have provided patches for our latest version (`v32`) and the version it was reported in (`v31`) and we will not be patching any older versions.
 
As usual, we'd recommend upgrading the version of AG Grid in your application so you can benefit from all the improvements in functionality, stability, security, theming and performance we've made since the version you're using.

If you are unable to do so, the only other option would be to fork and apply your own patch manually.

Thanks for understanding!

PR #8580, opened 2024-08-20
AG-9464 API changes to row click selection

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-9464


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2499,15 +2499,12 @@ export type RowSelectionOptions<TData = any, TValue = any> =
...
     /**
-     * If `true`, rows will not be deselected if you hold down `Ctrl` and click the row or press `Space`.
-     * @default false
-     */
-    suppressDeselection?: boolean;
-    /**
-     * If `true`, row selection won't happen when rows are clicked. Use when you only want checkbox selection.
+     * If `true` or `'supressSelection'`, rows will not be selected when clicked.
+     *
+     * If `true` or `'suppressDeselection'`, rows will not be deselected if you hold down `Ctrl` and click the row or press `Space`.
```
AndrewGlazier, line 2504: The space condition is true for both selection/deselection, this comment makes it feel a bit like it only applies for deselection, and also makes me think I have to hold ctrl and press space

Could be improved with something like (I'm not sure if the ctrl instruction is needed in the jsdoc, as it still requires 'clicking'?, also property names might be self explanatory enough):
Modifies the selection behaviour when clicking a row, or pressing `Space` while focusing a row.

documentation/ag-grid-docs/src/content/docs/row-selection-multi-row/index.mdoc
```diff
@@ -52,8 +52,16 @@ Use the example below to explore row range behaviour. For example, follow these
...
+When checkboxes are enabled, selected rows can be de-selected by clicking the checkbox. 
+When checkboxes are disabled, rows can be deselected by holding {% kbd "^ Ctrl" /%} and clicking the row. 
+This is also possible when checkboxes are enabled, as long as `selection.suppressClickSelection` is not enabled.
+Hold {% kbd "^ Ctrl" /%} and click the selected row in the example below. 
+
+{% gridExampleRunner title="Deselecting Rows" name="deselection" /%}
+
 #### Deselecting Ranges
-Typically, rows are deselected individually by holding {% kbd "^ Ctrl" /%} and clicking the row or unchecking the checkbox. Users may also deselect multiple rows at once by using holding {% kbd "⇧ Shift" /%} and clicking a row. 
+Users may deselect multiple rows at once by using holding {% kbd "⇧ Shift" /%} and clicking a row.
```
AndrewGlazier, line 64: using holding

documentation/ag-grid-docs/src/content/docs/row-selection-single-row/index.mdoc
```diff
@@ -76,47 +84,38 @@ const gridOptions = {
...
-This behaviour is enabled in the example below using the `selection.suppressClickSelection` flag.
+In some cases, such as when checkbox selection is enabled, it can make sense to only permit rows to be selected when its checkbox is clicked, rather than when the row itself is clicked. 
+This can be achieved by setting `suppressClickSelection` to `true` or `'suppressClick'`.
 
 ```{% frameworkTransform=true %}
 const gridOptions = {
     selection: { 
         mode: 'singleRow',
-        suppressClickSelection: true
+        suppressClickSelection: 'suppressClick',
```
AndrewGlazier, line 100: 'suppressSelection'

documentation/ag-grid-docs/src/content/docs/row-selection-single-row/index.mdoc
```diff
@@ -76,47 +84,38 @@ const gridOptions = {
 
 Note that we have enabled `suppressClickSelection` in this example, to prevent users from affecting selection with row clicks.
 
-## Disabling Click Selection
+## Prevent Click Selection & Deselection
+
+The `selection.suppressClickSelection` configuration setting controls the suppression of the default selection behaviour when clicking.
```
AndrewGlazier, line 89: This is a bit of a mouthful - how about:

The `selection.suppressClickSelection` property configures whether a rows selection will be impacted when the row is clicked, or when space is pressed while focused.

documentation/ag-grid-docs/src/content/docs/row-selection-single-row/index.mdoc
```diff
@@ -76,47 +84,38 @@ const gridOptions = {
...
 
-When checkbox selection is enabled, in many cases it makes sense to only permit rows to be selected when its checkbox is clicked, rather than when the row itself is clicked.
+{% interfaceDocumentation interfaceName="SingleRowSelectionOptions" names=["suppressClickSelection"] config={"description": ""}/%}
 
-This behaviour is enabled in the example below using the `selection.suppressClickSelection` flag.
+In some cases, such as when checkbox selection is enabled, it can make sense to only permit rows to be selected when its checkbox is clicked, rather than when the row itself is clicked. 
+This can be achieved by setting `suppressClickSelection` to `true` or `'suppressClick'`.
```
AndrewGlazier, line 94: I think we can remove the specific explanation of `suppressDeselection` is we reword these two lines a little

How about:
In some scenarios, such as when using [Checkbox Selection](link) it can be preferable to disable other methods of selection, such as keyboard (`Spacebar`), selection via row click, and [Deselection via Row Click](link). `suppressClickSelection` can be entirely enabled via `true` or granularly enabled via the values `'suppressSelection'` and `'suppressDeselection'`.

documentation/ag-grid-docs/src/content/docs/row-selection-single-row/index.mdoc
```diff
@@ -76,47 +84,38 @@ const gridOptions = {
...
-
-In some circumstances you may want to prevent users from deselecting a selected row. This can be done using the `selection.suppressDeselection` flag.
+It may also be necessary to prevent users from deselecting rows via {% kbd "^ Ctrl" /%}+click. This can be achieved by setting `suppressClickSelection` to `true` or `'suppressDeselection'`.
 
 ```{% frameworkTransform=true %}
 const gridOptions = {
     selection: { 
         mode: 'singleRow',
-        suppressDeselection: true,
+        suppressClickSelection: 'suppressDeselection',
```
AndrewGlazier, line 111: this snippet/comments around it can be dropped if we reword the above section as suggested/similar to suggestion

documentation/ag-grid-docs/src/content/docs/row-selection-single-row/index.mdoc
```diff
@@ -76,47 +84,38 @@ const gridOptions = {
...
     }
 }
 ```
 
-Observe that in the example below, it is not possible to deselect a row. Only by selecting another row will the currently selected row be deselected. That is, there is always one selected row.
+All of the configuration states for this option can be explored in the example below. Use the select element to switch between modes.
```
AndrewGlazier, line 116: The example below demonstrates the three possible configurations for this property. (maybe we should add an example for disabling it too, for completeness?)

PR #8582, opened 2024-08-20
AG-12593 Deprecate range selection events for new API

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12593


community-modules/core/src/misc/state/stateService.ts
```diff
@@ -248,15 +249,19 @@ export class StateService extends BeanStub implements NamedBean {
...
         } = this.gos.get('initialState') ?? {};
         if (focusedCellState) {
             this.setFocusedCellState(focusedCellState);
         }
-        if (rangeSelectionState) {
-            this.setRangeSelectionState(rangeSelectionState);
+        if (cellSelectionState) {
+            this.setCellSelectionState(cellSelectionState);
+        } else if (rangeSelectionState) {
+            _warnOnce('As of v32.2, `rangeSelection` is deprecated. Use `cellSelection` instead.');
```
AndrewGlazier, line 263: I'm not sure about this one - can you sync with Pete on this?
The problem with this is we don't encourage devs to configure selection state - this is most likely coming from a generated state we provide and they've serialized.

It might be that we want to keep this artifact around a lot longer than our normal deprecations, so that users can move versions without losing state. It may be better to just not warn here.

Again though, I'd defer to @peterjrreynolds

community-modules/core/src/misc/state/stateService.ts
```diff
@@ -267,13 +272,15 @@ export class StateService extends BeanStub implements NamedBean {
         this.updateCachedState('sideBar', this.getSideBarState());
         this.updateCachedState('focusedCell', this.getFocusedCellState());
         this.updateCachedState('rangeSelection', this.getRangeSelectionState());
```
AndrewGlazier, line 274: Is range selection state the same as cell selection state? If so we may want to drop this line, as no point generating it twice, and we can push towards the new model

AndrewGlazier, line 274: Is range selection state the same as cell selection state? If so we may want to drop this line, as no point generating it twice, and we can push towards the new model

eliasmalik: I kept this cache key because I think the cache is exposed via the grid API (`getState`), so if users are expecting to find a `rangeSelection` field, I should probably leave it there?

community-modules/core/src/interfaces/gridState.ts
```diff
@@ -23,6 +23,14 @@ export interface RangeSelectionCellState {
     startColId: string;
 }
 
+/** @deprecated v32.2 Use `CellSelectionCellState` instead. */
+export type RangeSelectionCellState = CellSelectionCellState;
```
peterjrreynolds, line 27: For backwards-compatibility, this should be an interface not a type (we wouldn't expect users to extends this, but it's technically a breaking change otherwise).

community-modules/core/src/misc/state/stateService.ts
AndrewGlazier: We don't encourage users to manipulate state ( I think we actively discourage it ) check this with Pete too I think, but if we're not removing it, might be better to only call `getRangeSelectionState` once for perf, and update the cache state with the same ref in both places

community-modules/core/src/misc/state/stateService.ts
peterjrreynolds: We expect applications may be serialising initial state to a permanent store. This means we need to be able to support old versions.

Take a look at `chartModelMigration.ts` for how we do this for integrated chart models.

We'll need to do a phased approach here.

Firstly, will need to introduce versions to initial state.

For now (next minor), we want to convert from the old model to the new model on input (so the processing logic only has to handle the new style). Should return both when retrieving state.

For the next major, we can probably be quite aggressive and remove the old property. The migration logic will still support it, so we can just document accordingly that it's not in the typing but it will still work (which is essentially what integrated does).

Regarding deprecations, we can deprecate in the interface, but we don't want to log it.

eliasmalik: > For now (next minor), we want to convert from the old model to the new model on input (so the processing logic only has to handle the new style). Should return both when retrieving state.

To be clear then, does this mean in the migration we copy `rangeSelection` into `cellSelection` but _don't_ delete `rangeSelection`? And if we want to return both when retrieving state, then we still need to maintain a `rangeSelection` cache entry?

peterjrreynolds (approved):

PR #8583, opened 2024-08-20
AG-1372 Remove "mixed" copy behaviour when using new selection API

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-1372


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -2522,6 +2524,10 @@ interface CommonRowSelectionOptions<TData = any, TValue = any> {
      * Callback to be used to determine which rows are selectable. By default rows are selectable, so return `false` to make a row un-selectable.
      */
     isRowSelectable?: IsRowSelectable<TData>;
+    /**
+     * When enabled and a row is selected, the copy action should copy the entire row, rather than just the focussed cell
```
AndrewGlazier, line 2528: Am aware this is a correct spelling of focussed, and it seems we use both spellings, but we use `focused` significantly more, so probs better to stick with that spelling

documentation/ag-grid-docs/src/content/docs/clipboard/index.mdoc
```diff
@@ -32,52 +32,25 @@ In the below example `copyHeadersToClipboard` has been enabled, try:
 
 ## Copying Rows
 
-When [Row Selection](./row-selection) is enabled by setting `gridOptions.selection.mode` to either `"singleRow"` or `"multiRow"`, then copying while a row is selected will add the whole row's contents to your clipboard.
+When [Row Selection](./row-selection) is enabled by setting `gridOptions.selection.mode` to either `"singleRow"` or `"multiRow"`, the default behaviour is to copy only the focussed cell to the clipboard.
```
AndrewGlazier, line 35: focussed -> focused

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
         }
     }
 
+    private shouldCopyCells(selection?: SelectionOptions) {
+        const suppressCopySingleCellRanges = this.gos.get('suppressCopySingleCellRanges');
```
AndrewGlazier, line 763: This func might be a bit easier to follow if we start with:
if (!this.rangeService) {
   return false;
}

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
+    }
+
+    private shouldCopyRows(selection?: SelectionOptions) {
+        const suppressCopyRowsToClipboard = this.gos.get('suppressCopyRowsToClipboard');
+
+        const hasSelection = !this.selectionService.isEmpty();
+        if (selection) {
+            return selection.mode !== 'cell' && (selection.copySelectedRows ?? true) && hasSelection;
+        } else {
+            return hasSelection && !suppressCopyRowsToClipboard;
```
AndrewGlazier, line 781: same as above, is it possible to have a selection if we have no selection options? If we do would we actually want to copy it?

Maybe start with if (!selection) { return false }

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
         }
     }
 
+    private shouldCopyCells(selection?: SelectionOptions) {
+        const suppressCopySingleCellRanges = this.gos.get('suppressCopySingleCellRanges');
+
+        const hasSelection = this.rangeService && !this.rangeService.isEmpty();
+        const shouldSkip = this.rangeService && !this.rangeService.isMoreThanOneCell() && suppressCopySingleCellRanges;
+        if (selection) {
```
AndrewGlazier, line 767: I misunderstood selection here, might be valuable to add some comment outlining the fact that this is diff between new and legacy api

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
eliasmalik: This branch is trying to preserve the existing behaviour. We only implement the new behaviour if people are using the new selection API.

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
+            return selection.mode === 'cell' && hasSelection;
+        } else {
+            return hasSelection && !shouldSkip;
+        }
+    }
+
+    private shouldCopyRows(selection?: SelectionOptions) {
+        const suppressCopyRowsToClipboard = this.gos.get('suppressCopyRowsToClipboard');
+
+        const hasSelection = !this.selectionService.isEmpty();
```
AndrewGlazier, line 777: if (!hasSelection) {
return false
}

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
         }
     }
 
+    private shouldCopyCells(selection?: SelectionOptions) {
+        const suppressCopySingleCellRanges = this.gos.get('suppressCopySingleCellRanges');
+
+        const hasSelection = this.rangeService && !this.rangeService.isEmpty();
+        const shouldSkip = this.rangeService && !this.rangeService.isMoreThanOneCell() && suppressCopySingleCellRanges;
```
AndrewGlazier: Can we move shouldSkip into the second else block

enterprise-modules/clipboard/src/clipboard/clipboardService.ts
```diff
@@ -758,6 +759,29 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
+        if (selection) {
+            return selection.mode === 'cell' && hasSelection;
+        } else {
+            return hasSelection && !shouldSkip;
+        }
+    }
+
+    private shouldCopyRows(selection?: SelectionOptions) {
+        const suppressCopyRowsToClipboard = this.gos.get('suppressCopyRowsToClipboard');
```
AndrewGlazier, line 786: Second else block

PR #8584, opened 2024-08-20
AG-12599 - maintainColumnOrder extended and modified

AndrewGlazier:


community-modules/core/src/entities/gridOptions.ts
```diff
@@ -335,9 +335,9 @@ export interface GridOptions<TData = any> {
     };
     /**
      * Keeps the order of Columns maintained after new Column Definitions are updated.
-     * @default false
+     * @default 'pivotResultColumns'
      */
-    maintainColumnOrder?: boolean;
+    maintainColumnOrder?: boolean | 'primaryAndPivotResultColumns' | 'primaryColumns' | 'pivotResultColumns' | 'none';
```
peterjrreynolds, line 340: Can change `primaryAndPivotResultColumns` to `all`

Also, add deprecation note on boolean to the comment here

documentation/ag-grid-docs/src/content/docs/column-updating-definitions/index.mdoc
```diff
@@ -132,7 +132,7 @@ In other words, to have the grid correctly match Columns make sure each Column h
 The example below demonstrates the different matching strategies. Note the following:
 
 * All columns, with the exception of Country, are matched correctly. This means any column width, sort etc
-  will be kept between changes to the columns. Country will have its state reset, as it will be treated as a new column
+  will be kept between changes to the columns. Country will have its state reset, §as it will be treated as a new column
```
peterjrreynolds, line 135: Typo

community-modules/core/src/columns/columnModel.ts
```diff
@@ -237,7 +231,22 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
+        if (maintainColumnOrder === true) {
+            maintainColumnOrder = 'primaryAndPivotResultColumns';
+        } else if (maintainColumnOrder === false) {
+            maintainColumnOrder = 'pivotResultColumns';
+        }
+
+        const shouldSortNewColDefs =
+            maintainColumnOrder !== 'none' &&
+            (maintainColumnOrder === 'primaryAndPivotResultColumns' ||
+                maintainColumnOrder === (this.showingPivotResult ? 'pivotResultColumns' : 'primaryColumns'));
```
StephenCooper, line 246: Can we have this in a grid options utils function.

community-modules/core/src/entities/gridOptions.ts
AndrewGlazier: done - deprecation note done without @deprecated due to lack of deprecation of entire property

documentation/ag-grid-docs/src/content/docs/column-updating-definitions/index.mdoc
AndrewGlazier: done

community-modules/core/src/columns/columnModel.ts
AndrewGlazier: done

community-modules/core/src/entities/gridOptions.ts
```diff
@@ -335,9 +335,11 @@ export interface GridOptions<TData = any> {
     };
     /**
      * Keeps the order of Columns maintained after new Column Definitions are updated.
-     * @default false
+     *
+     * The use of boolean values with `maintainColumnOrder` have been deprecated as of v32.2.0
+     * @default 'pivotResultColumns'
      */
-    maintainColumnOrder?: boolean;
+    maintainColumnOrder?: boolean | 'all' | 'primaryColumns' | 'pivotResultColumns' | 'none';
```
peterjrreynolds, line 342: Should we export a type for the string values?

AndrewGlazier: I don't think so - we don't for other string unions. Lots of precedent to not, and not much (if any) to do so

PR #8588, opened 2024-08-20
AG-3776 - added onDragCancel

gportela85:
https://ag-grid.atlassian.net/browse/AG-3776
https://ag-grid.atlassian.net/browse/AG-12446

### DIFF
--- a/community-modules/core/src/dragAndDrop/dragService.ts
+++ b/community-modules/core/src/dragAndDrop/dragService.ts
@@ -1,3 +1,4 @@
+import { KeyCode } from '../constants/keyCode';
 import type { NamedBean } from '../context/bean';
 import { BeanStub } from '../context/beanStub';
 import type { BeanCollection } from '../context/context';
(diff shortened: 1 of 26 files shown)


community-modules/core/src/dragAndDrop/dragService.ts
```diff
@@ -177,12 +178,18 @@ export class DragService extends BeanStub implements NamedBean {
         const mouseMoveEvent = (event: MouseEvent) => this.onMouseMove(event, params.eElement);
         const mouseUpEvent = (event: MouseEvent) => this.onMouseUp(event, params.eElement);
         const contextEvent = (event: MouseEvent) => event.preventDefault();
+        const keydownEvent = (event: KeyboardEvent) => {
+            if (event.key === KeyCode.ESCAPE) {
+                this.cancelDrag(params.eElement);
+            }
```
peterjrreynolds, line 184: Do we need to stop the event propagating here?

community-modules/core/src/events.ts
```diff
@@ -433,10 +435,14 @@ export interface DragStartedEvent<TData = any, TContext = any> extends AgDragEve
     type: 'dragStarted';
 }
 
-export interface DragStoppedEvent<TData = any, TContext = any> extends AgDragEvent<`dragStopped`, TData, TContext> {
+export interface DragStoppedEvent<TData = any, TContext = any> extends AgDragEvent<'dragStopped', TData, TContext> {
     type: 'dragStopped';
 }
 
+export interface DragCancelledEvent<TData = any, TContext = any> extends AgDragEvent<'dragCancelled', TData, TContext> {
+    type: 'dragCancelled';
```
peterjrreynolds, line 443: Don't need the `type` here - it's passed through the generic param.

In fact, don't need the type for the above two events either. Looks like these got left in by accident when the generic param was added.

community-modules/core/src/gridBodyComp/rowDragFeature.ts
```diff
@@ -368,6 +383,12 @@ export class RowDragFeature extends BeanStub implements DropTarget {
                       events.onDragStop!(this.draggingToRowDragEvent('rowDragEnd', e as any));
                   }
                 : onDragStop,
+            onDragCancel: events.onDragCancel
```
peterjrreynolds, line 386: Could de-duplicate the code for all these events

community-modules/core/src/events.ts
gportela85: fixed

community-modules/core/src/dragAndDrop/dragService.ts
gportela85: I don't think we should stop the propagation here, we can perform whatever actions we need and still allow the customer to add their own listeners.

PR #8591, opened 2024-08-21
AG-12109 Support deep / shallow merging of global grid options

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-12109


community-modules/core/src/grid.ts
```diff
@@ -111,15 +111,58 @@ export interface Params {
...
+            mergedGridOps.context = providedOptions.context;
+        }
+
+        return mergedGridOps;
+    }
 }
 
+/**
+ * When providing global grid options, specify how they should be merged with the grid options provided to individual grids.
+ * - `deep` will merge the global options into the provided options deeply
```
peterjrreynolds, line 148: I know it's kind of implicit, but we should specifically mention that local options take precedence during the deep merge.

community-modules/core/src/grid.ts
```diff
@@ -224,15 +267,8 @@ export class GridCoreCreator {
...
-        let mergedGridOps: GridOptions = {};
-        if (GlobalGridOptions.gridOptions) {
-            // Merge deep to avoid leaking changes to the global options
-            _mergeDeep(mergedGridOps, GlobalGridOptions.gridOptions, true, true);
-            // Shallow copy to ensure context reference is maintained
-            mergedGridOps = { ...mergedGridOps, ...providedOptions };
-        } else {
-            mergedGridOps = providedOptions;
-        }
+        const mergedGridOps = GlobalGridOptions.applyGlobalGridOptions({}, providedOptions);
```
peterjrreynolds, line 270: I know this replicates mergeDeep, but the first argument here seems redundant

PR #8593, opened 2024-08-21
Ag 12227 inline code element improvements

kylerphillips:


kylerphillips:
@markdurrant @taktran worth us seeing if there are any huge speed implications with using JetBrains as the code font family

taktran (approved):
👍🏼

PR #8601, opened 2024-08-22
AG-12049 theme grid option

BernieSumption:
This PR allows Theming API themes to be supplied to grids through a grid option.

### Summary of implementation.

`@ag-grid-community/core` defines and exports the `GridTheme` interface:

```
interface GridTheme {
    startUse(args: GridThemeUseArgs): void;
    stopUse(): void;
    getCssClass(): string;
}
type GridThemeUseArgs = {
    loadThemeGoogleFonts: boolean | undefined;
    container: HTMLElement;
};
```

* When a theme is passed as a grid option, the grid calls `startUse(...)` passing the grid's parent element and the value of the `loadThemeGoogleFonts` grid option, and then sets the class returned by `getCssClass()` on the root wrapper element of the grid.
* `@ag-grid-community/theming` themes now implement this interface, the relevant implementation is in `Theme.ts`
* A number of changes to the Theming API and Theme Builder were required for compatibility with the new model, this accounts for most of the diff.
* There are changes to the example generator and runner to handle the Theming API in docs


community-modules/core/src/propertyKeys.ts
```diff
@@ -392,7 +393,7 @@ export class PropertyKeys {
     ];
 
     // If property does not fit above, i.e union that should not be coerced.
-    public static OTHER_PROPERTIES: GridOptionKey[] = ['suppressStickyTotalRow'];
+    public static OTHER_PROPERTIES: GridOptionKey[] = ['suppressStickyTotalRow', 'loadThemeGoogleFonts'];
```
StephenCooper, line 396: This can be a boolean property. boolean | undefined does not need this special treatment.

community-modules/theming/src/styles/core/core-css.ts
```diff
@@ -612,7 +612,7 @@ export const coreDefaults: CoreParams = {
     },
     wrapperBorder: true,
     rowBorder: true,
-    browserColorScheme: 'inherit',
+    browserColorScheme: 'light',
```
StephenCooper, line 615: Just checking this was intended?

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/grid-vanilla-to-angular.ts
```diff
@@ -63,14 +64,17 @@ function addModuleImports(
...
     imports.push("import { AgGridAngular } from '@ag-grid-community/angular';");
-    imports.push('// NOTE: Angular CLI does not support component CSS imports: angular-cli/issues/23273');
-    imports.push("import '@ag-grid-community/styles/ag-grid.css';");
 
-    // to account for the (rare) example that has more than one class...just default to quartz if it does
-    // we strip off any '-dark' from the theme when loading the CSS as dark versions are now embedded in the
-    // "source" non dark version
-    const theme = inlineGridStyles.theme ? inlineGridStyles.theme.replace('-dark', '') : 'ag-theme-quartz';
-    imports.push(`import "@ag-grid-community/styles/${theme}.css";`);
+    if (!usesThemingApi(bindings)) {
```
StephenCooper, line 68: Oh nice :) glad to see the back of that warning in our Angular examples

community-modules/theming/src/styles/core/core-css.ts
BernieSumption: Yes, the default theme is light coloured, so if you load it in a dark mode page it looks bad if you let it inherit the dark scrollbars from the page.

community-modules/core/src/environment.ts
```diff
@@ -222,6 +224,33 @@ export class Environment extends BeanStub implements NamedBean {
         }
         return Object.freeze(allThemeClasses);
     }
+
+    private handleThemeGridOptionChange(): void {
+        const theme = this.gos.get('theme') || null;
```
gportela85, line 229: We should assign variables at the top so we don't keep reading values from `this` multiple times.

`const { gos, gridTheme, eMeasurementContainer, ancestorThemeClasses } = this`

BernieSumption: Fixed

PR #8617, opened 2024-08-28
AG-12464 remove ResizeObserver polyfill

BernieSumption:
This PR:

1. Removes the ResizeObserver polyfill
2. Deprecates gridOptions.suppressBrowserResizeObserver and turns it into a no-op

### DIFF
--- a/community-modules/core/src/misc/resizeObserverService.ts
+++ b/community-modules/core/src/misc/resizeObserverService.ts
@@ -2,83 +2,14 @@ import type { NamedBean } from '../context/bean';
 import { BeanStub } from '../context/beanStub';
 import { _getWindow } from '../gridOptionsUtils';
 
-const DEBOUNCE_DELAY = 50;
 export class ResizeObserverService extends BeanStub implements NamedBean {
     beanName = 'resizeObserverService' as const;
 
-    private polyfillFunctions: (() => void)[] = [];
-    private polyfillScheduled: boolean;
-
     public observeResize(element: HTMLElement, callback: () => void): () => void {
         const win = _getWindow(this.gos);
-        const useBrowserResizeObserver = () => {
-            const resizeObserver = new win.ResizeObserver(callback);
-            resizeObserver.observe(element);
-            return () => resizeObserver.disconnect();
-        };
-
-        const usePolyfill = () => {
-            // initialise to the current width and height, so first call will have no changes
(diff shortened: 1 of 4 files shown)


community-modules/core/src/misc/resizeObserverService.ts
```diff
@@ -1,84 +0,0 @@
-import type { NamedBean } from '../context/bean';
-import { BeanStub } from '../context/beanStub';
-import { _getWindow } from '../gridOptionsUtils';
-
-const DEBOUNCE_DELAY = 50;
-export class ResizeObserverService extends BeanStub implements NamedBean {
```
StephenCooper, line 6: Would suggest keeping the service and just dropping the polyfill to maintain the window retrieval logic.

community-modules/core/src/misc/resizeObserverService.ts
```diff
@@ -1,84 +0,0 @@
...
-            };
-
-            periodicallyCheckWidthAndHeight();
-
-            // the callback function we return sets running to false
-            return () => (running = false);
-        };
-
-        const suppressResize = this.gos.get('suppressBrowserResizeObserver');
-        const resizeObserverExists = !!win.ResizeObserver;
```
StephenCooper, line 51: Need to keep this check for JSDOM

StephenCooper (changes requested):
Added some comments

community-modules/core/src/misc/resizeObserverService.ts
BernieSumption: Done

community-modules/core/src/misc/resizeObserverService.ts
BernieSumption: Done

StephenCooper (changes requested):
Need to remove from https://github.com/ag-grid/ag-grid/blob/8f603eb2f41181a6092ddc63739199dc2fa249d1/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json#L790

BernieSumption:
> Need to remove from
> 
> https://github.com/ag-grid/ag-grid/blob/8f603eb2f41181a6092ddc63739199dc2fa249d1/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json#L790

Done

StephenCooper (approved):
Approved.

PR #8619, opened 2024-08-28
AG-12637 - fix rich select focus issue

peterjrreynolds:


gportela85:
Instead of adding these flags, I think it might be better to add `this.hidePicker();` at the end of `AgRichSelect#onTabKeyDown`.

PR #8627, opened 2024-08-30
AG-12663 - fix column tool panel nested split groups

peterjrreynolds:


gportela85:
@peterjrreynolds I don't think we should be making `mergeLeafPathTrees` public just because of unit testing here.
Maybe we can use `syncLayoutWithGrid` that is already public to test this?

PR #8633, opened 2024-09-02
AG-12701 [Bug] Flex minWidth columns do not size correctly when to the left of maxWidth columns

BernieSumption:
In flex sizing, total width is divided into:

1. Reserved space: space that is already taken up
2. Available space: space to be distributed between flex columns (total width - used space)

The previous flex implementation calculated reserved space from non-flex columns only. It distributed available space among flex columns, taking into account their minWidth and maxWidth. This lead to the edge case bugs described in ticket AG-12701.

This implementation calculates reserved space from both non-flex columns, _and the minimum width of flex columns_. It then distributes available space among flex columns, taking into account their maxWidth (there is no need to take minWidth into account because it is already accounted for in the reserved width).

### DIFF
--- a/community-modules/core/src/columns/columnSizeService.ts
+++ b/community-modules/core/src/columns/columnSizeService.ts
@@ -325,20 +325,19 @@ export class ColumnSizeService extends BeanStub implements NamedBean {
             }
         }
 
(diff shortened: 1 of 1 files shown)


gportela85 (approved):
Looks good

PR #8645, opened 2024-09-04
allow behavioural tests to use source code instead of bundles

SalvatorePreviti:


SalvatorePreviti:
Discussed with Sean, he approved, we can review this in the future if/when needed

PR #8655, opened 2024-09-05
AG-11856 Provisional React 19 support

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-11856


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1234,7 +1234,10 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             this.rowDataTransactionBatch = [];
             const waitMillis = this.gos.get('asyncTransactionWaitMillis');
             this.applyAsyncTransactionsTimeout = window.setTimeout(() => {
-                this.executeBatchUpdateRowData();
+                if (this.isAlive()) {
+                    // Handle case where grid is destroyed before timeout is triggered
+                    this.executeBatchUpdateRowData();
+                }
```
StephenCooper, line 1240: Destroying a grid that has an outstanding batch update triggers a call even though the grid is destroyed.

community-modules/core/src/ctrlsService.ts
```diff
@@ -47,57 +48,95 @@ interface ReadyParams {
...
+            params.stickyTopRight?.isAlive() &&
+            params.stickyBottomCenter?.isAlive() &&
+            params.stickyBottomLeft?.isAlive() &&
+            params.stickyBottomRight?.isAlive() &&
+            params.centerHeader?.isAlive() &&
+            params.leftHeader?.isAlive() &&
+            params.rightHeader?.isAlive() &&
+            params.fakeHScrollComp?.isAlive() &&
+            params.fakeVScrollComp?.isAlive() &&
+            params.gridHeaderCtrl?.isAlive();
```
StephenCooper, line 108: Checking for `isAlive()` is a way for us to go back into waiting state in React double render but without having to clear the controllers. 

This is a safer change because it means the controllers are still present when the grid is actually destroyed and we have logic that is run during cleanup that can access these.

community-modules/core/src/headerRendering/cells/abstractCell/abstractHeaderCellCtrl.ts
```diff
@@ -425,16 +425,19 @@ export abstract class AbstractHeaderCellCtrl<
         });
     }
 
+    protected clearComponent(): void {
+        this.removeDragSource();
+        (this.resizeFeature as any) = null;
+        (this.comp as any) = null;
+        (this.eGui as any) = null;
```
StephenCooper, line 432: These are tied to the lifecycle of the component and not the controller so they have their own method that is cleared by the component instead of the controller.

documentation/ag-grid-docs/public/example-runner/grid-react-boilerplate/systemjs.config.dev.js
```diff
@@ -1,5 +1,7 @@
 (function (global) {
     var sjsPaths = {};
+    // eslint-disable-next-line no-global-assign
```
StephenCooper, line 3: These changes must be reverted before the code changes are merged.

community-modules/core/src/rendering/rowRenderer.ts
```diff
@@ -993,13 +993,13 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
-        this.removeAllRowComps();
+        this.removeAllRowComps(true);
         super.destroy();
     }
 
-    private removeAllRowComps(): void {
+    private removeAllRowComps(suppressAnimation: boolean = false): void {
         const rowIndexesToRemove = Object.keys(this.rowCtrlsByRowIndex);
-        this.removeRowCtrls(rowIndexesToRemove);
+        this.removeRowCtrls(rowIndexesToRemove, suppressAnimation);
```
StephenCooper, line 1002: Is a waste animating rows when they are all being destroyed.

community-modules/core/src/ctrlsService.ts
peterjrreynolds: Could iterate here?

community-modules/core/src/headerRendering/rowContainer/headerRowContainerCtrl.ts
```diff
@@ -74,7 +74,13 @@ export class HeaderRowContainerCtrl extends BeanStub implements ScrollPartner {
             advancedFilterEnabledChanged: onDisplayedColsChanged,
         });
 
-        this.ctrlsService.registerHeaderContainer(this, this.pinned);
+        let headerType: 'centerHeader' | 'leftHeader' | 'rightHeader' = 'centerHeader';
+        if (this.pinned === 'left') {
+            headerType = 'leftHeader';
+        } else if (this.pinned === 'right') {
+            headerType = 'rightHeader';
+        }
```
peterjrreynolds, line 82: ```const headerType = `${typeof this.pinned === 'string' ? this.pinned : 'center'}Header```

community-modules/core/src/interfaces/iFrameworkOverrides.ts
```diff
@@ -66,4 +66,9 @@ export interface IFrameworkOverrides {
 
     /** Used by `RowRenderer` when releasing lock. Allows React async refreshes to work. */
     releaseLockOnRefresh?(): void;
+
+    /** Used by the CtrlsService to decide whether to trigger the whenReady callbacks asynchronously.
+     * Required for React to work with StrictMode from v19 with the current implementation of the CtrlsService.
+     */
+    initGridAsync?(): boolean;
```
peterjrreynolds, line 73: `isInitGridAsync` as method isn't initialising the grid?

StephenCooper: Renamed to `runWhenReadyAsync`

eero-dev:
This PR seems to have broken bunch of our tests running on react 18.

Will attempt to produce a minimal repro.

TypeError: Cannot read properties of undefined (reading 'getGridId')
 ❯ node_modules/@ag-grid-community/react/dist/package/index.esm.mjs:2640:40
    2638|       apiRef.current = void 0;
    2639|     });
    2640|     gridIdRef.current = apiRef.current.getGridId();

PR #8659, opened 2024-09-05
AG-12624 - Pivoting docs

AndrewGlazier:


enterprise-modules/row-grouping/src/rowGrouping/pivotColDefService.ts
```diff
@@ -170,60 +170,63 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
                     const columnName: string | null = this.columnNameService.getDisplayNameForColumn(
                         valueColumn,
                         'header'
                     );
-                    const totalColDef = this.createColDef(valueColumn, columnName, groupDef.pivotKeys);
+                    const totalColDef = this.createColDef(valueColumn, columnName, def.pivotKeys);
                     totalColDef.pivotTotalColumnIds = childAcc.get(valueColumn.getColId());
 
-                    totalColDef.columnGroupShow = 'closed';
+                    totalColDef.columnGroupShow = !isSuppressExpand ? 'closed' : 'open';
```
AndrewGlazier, line 200: This is more than a style change, but due to line 174 we will never reach this actual change. This is here to enable AG-12739 when we remove the flag on 174

PR #8671, opened 2024-09-09
AG-12656 - react full row edit update

peterjrreynolds:


community-modules/react/src/reactUi/cells/cellComp.tsx
```diff
@@ -231,7 +231,10 @@ const CellComp = ({
                         cellCtrl.focusCell(true);
                     });
                 } else {
-                    cellCtrl.cellEditorAttached();
+                    // need to ensure that render has completed
+                    setTimeout(() => {
+                        cellCtrl.cellEditorAttached();
+                    });
```
gportela85, line 237: Since `if` and `else` use setTimeout, we should put the condition itself and use a single `setTimeout`.

PR #8696, opened 2024-09-13
AG-12835 strict bind call apply

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-12835


community-modules/core/src/gridBodyComp/rowDragFeature.ts
```diff
@@ -347,52 +373,55 @@ export class RowDragFeature extends BeanStub implements DropTarget {
         const onDragStop = this.onDragStop.bind(this);
         const onDragCancel = this.onDragCancel.bind(this);
 
+        let params: InternalRowDropZoneParams;
```
StephenCooper, line 376: No actual code changes here, just making params against internal interface and then casting to the external interface that users would code against.

PR #8701, opened 2024-09-13
AG-12847 - column move module

peterjrreynolds:


community-modules/core/src/columnMove/columnDrag/moveColumnFeature.ts
```diff
@@ -34,7 +34,7 @@ export class MoveColumnFeature extends BeanStub implements DropListener {
     public wireBeans(beans: BeanCollection) {
         this.columnModel = beans.columnModel;
         this.visibleColsService = beans.visibleColsService;
-        this.columnMoveService = beans.columnMoveService;
+        this.columnMoveService = beans.columnMoveService!;
```
gportela85, line 37: why is the `!` required here? seems odd

PR #8703, opened 2024-09-13
AG-12737 - allow syncing of fractional scrollLeft

gportela85:
https://ag-grid.atlassian.net/browse/AG-12737

### DIFF
--- a/community-modules/core/src/utils/dom.ts
+++ b/community-modules/core/src/utils/dom.ts
@@ -243,7 +243,7 @@ export function _getScrollLeft(element: HTMLElement, rtl: boolean): number {
         scrollLeft = Math.abs(scrollLeft);
 
         if (_isBrowserChrome() && !_isRtlNegativeScroll()) {
-            scrollLeft = element.scrollWidth - element.clientWidth - scrollLeft;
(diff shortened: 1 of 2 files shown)


community-modules/core/src/utils/dom.ts
```diff
@@ -243,7 +243,7 @@ export function _getScrollLeft(element: HTMLElement, rtl: boolean): number {
         scrollLeft = Math.abs(scrollLeft);
 
         if (_isBrowserChrome() && !_isRtlNegativeScroll()) {
-            scrollLeft = element.scrollWidth - element.clientWidth - scrollLeft;
+            scrollLeft = element.scrollWidth - element.getBoundingClientRect().width - scrollLeft;
```
StephenCooper, line 246: @gportela85 why does this change fix the issue?

gportela85: The change that actually fixes the issue is the one above, where we no longer `Math.round()` the value of scrollLeft.

This change here is an addition to it, because since we are using floating points for scrollLeft, we should also use floating points for the `width`. `clientWidth` is always an integer, so we are now grabbing the value from the BoundingClientRect which is more granular.

PR #8728, opened 2024-09-16
RTI-2283 move refreshing headers docs section

SalvatorePreviti:
Current: Refreshing headers docs section is on the updating definitions page - [React Grid: Updating Column Definitions | AG Grid](https://www.ag-grid.com/archive/32.2.0/react-data-grid/column-updating-definitions/#refreshing-headers) .

Desired: This should be on the column headers page next to the custom header components section.

 

 

Additionally, update the following:

Current: All Header Components that still exist after the new Column Definitions are applied (in other words, the Column still exists after the update, it was not removed) will have its refresh method called.

Desired: Only show this as is for non-React. For React, replace will have its refresh method called with will be re-rendered.

 

Current: For frameworks where possible, the Header Component returns false which results in the component getting destroyed and recreated.

Desired: Remove For frameworks where possible and instead display this text only for non-React.


documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
```diff
@@ -180,6 +180,27 @@ The following props are passed to the Custom Component (`CustomHeaderProps` inte
 {% interfaceDocumentation interfaceName="CustomHeaderProps" exclude=["template"] config={ "description": "" } /%}
 {% /if %}
 
+### Refreshing Headers
```
peterjrreynolds, line 183: Can move this section down slightly and replace the existing "Refresh" section.

documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
```diff
@@ -180,6 +180,27 @@ The following props are passed to the Custom Component (`CustomHeaderProps` inte
...
+
+If you are creating your own [Header Components](./column-headers/) then you will need to be aware of
+how Header Components are refreshed.
+
+All Header Components that still exist after the new Column Definitions are applied (in other words, the Column still
+exists after the update, it was not removed) {% if isFramework("react") %}will be re-rendered.{% else/ %}will have its `refresh` method called.{% /if %}
+
+
+It is up to the Header Component to update based on any changes it may find in the Column Definition.
```
peterjrreynolds, line 192: For non-React can add in the text from the existing "Refresh" section here - 
`If the refresh was successful ... consistent with the refresh method of Cell Renderers.`

PR #8737, opened 2024-09-16
RTI-2312 remove auto description on igroupcellrenderparams interface component

SalvatorePreviti:
- add a _config_.description = "" in documentation/ag-grid-docs/src/content/interface-documentation/group-cell-renderer/group-cell-renderer.json
- fix documentation/ag-grid-docs/src/components/reference-documentation/utils/getInterfaceDocumentationModel.ts to use the override description in the _config_ if present

I verified there is no other "description" field override in any _config_


taktran (commented):
The other way to do this, which is present in other places is to add `config={ "description": "" }` to the markdoc tag ie, in `documentation/ag-grid-docs/src/content/docs/group-cell-renderer/index.mdoc`

```
{% interfaceDocumentation interfaceName="IGroupCellRendererParams" overrideSrc="group-cell-renderer/group-cell-renderer.json" config={ "description": "" } /%}
```

eliasmalik (approved):

PR #8761, opened 2024-09-17
AG-12883 Correct behaviour when moving selection column

eliasmalik:
https://ag-grid.atlassian.net/browse/AG-12883

### DIFF
--- a/community-modules/core/src/columns/columnModel.ts
+++ b/community-modules/core/src/columns/columnModel.ts
@@ -380,7 +380,7 @@ export class ColumnModel extends BeanStub implements NamedBean {
         }
 
         destroyPrevious();
-        const [tree, treeDepth] = this.columnFactory.balanceTreeForAutoCols(list, this.cols?.tree);
+        const [tree, treeDepth] = this.columnFactory.balanceTreeForAutoCols(list, this.cols.tree);
         this.autoCols = {
             list,
             tree,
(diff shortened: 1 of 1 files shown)


community-modules/core/src/columns/columnModel.ts
```diff
@@ -402,10 +402,17 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
+        }
 
         const list = this.controlsColService?.createControlsCols() ?? [];
+        const areSame = areColIdsEqual(list, this.controlsCols?.list ?? []);
+
+        if (areSame) {
+            return;
+        }
 
         const [tree, treeDepth] = this.columnFactory.balanceTreeForAutoCols(list, this.cols.tree);
```
AndrewGlazier, line 427: I think skipping this is problematic, the controls col will only ever be generated once in this scenario (unless toggled I imagine) but we'll never rebalance the col if the tree depth changes, I think we still need to do this, but not the sorting at the end maybe?

eliasmalik: You're right, I validated this with a plunkr: https://plnkr.co/edit/hm7ehaW3gqn4yzE5?open=main.js&preview

I'm now using the same approach to setting `lastOrder` as is used for the autoCols. I didn't realise originally that we use reference equality to restore the sort order. The `filter` approach inserts the new object references into `lastOrder`, but the sorting approach I was using didn't.

PR #8764, opened 2024-09-17
AG-12272 - step2 - named bean client side node manager

SalvatorePreviti:
- Adds a new grid option "treeDataChildrenField"
- Makes clientSideNodeManager a named bean and provide two implementation with a base abstract class, clientSideNodeManager and clientSideTreeNodeManager
- Modify the clientSideRowModel to load the correct client side node manager instance depending on "treeDataChildrenField". If present, the clientSideTreeNodeManager is used
- Adds a new makeFieldPathGetter function

In this PR, the clientSideTreeNodeManager is not yet processing the children hierarchical tree data, that is another PR.
Also, the module is temporarily defined in CSRM module, but we are going to change soon for bundle size the structure of our projects so it does not make sense for now to create a new package for it.


community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1210,7 +1229,12 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         // - clears selection, done before we set row data to ensure it isn't readded via `selectionService.syncInOldRowNode`
         this.selectionService.reset('rowDataChanged');
 
-        this.nodeManager.setRowData(rowData);
+        if (typeof rowData === 'string') {
+            _warnOnce('rowData must be an array.');
+        } else {
+            this.nodeManager.setRowData(rowData);
+            this.rowNodesCountReady = true;
```
AndrewGlazier, line 1236: We should ask Pete to verify this specific bit, he introduced this flag due to a race condition and I don't know the implications off the top of my head (specifically the fact we set it lower down too)

SalvatorePreviti: I will ask, but to make it simpler and clearer probably this need some refactoring. For now the behaviour should be 1:1 to what it was the previous implementation, also switching the node managers.  For context, having two different node managers implies we will have two different rowCountReady, so I just moved it into the row model instead, maintaining exactly the same behaviour and setting it to true in the same points it was set by the node manager 

fiy @peterjrreynolds

community-modules/client-side-row-model/src/treeData/clientSideTreeNodeManager.ts
```diff
@@ -163,7 +111,7 @@ export class ClientSideNodeManager {
     }
 
     /** Converts the setRowData() command to a transaction */
-    private createTransactionForRowData<TData>(rowData: TData[]): RowDataTransaction | null {
+    private createTransactionForRowData(rowData: TData[]): RowDataTransaction<TData> | null {
         const getRowIdFunc = _getRowIdCallback(this.gos);
         if (getRowIdFunc == null) {
             _errorOnce('ImmutableService requires getRowId() callback to be implemented, your row data needs IDs!');
```
AndrewGlazier, line 117: "ImmutableService" is an internal reference, we should make this more user friendly

SalvatorePreviti: this was the original warning, i am going to do another PR to remove the immutable service and fix the warning

community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
peterjrreynolds: The only thing that is different now is that it's set after the `rowDataUpdateStarted` event is fired. I can't remember if there was an ordering requirement around these.

If all the test cases in AG-10437 pass, then it's fine.

SalvatorePreviti: Thank you, changed the code, just to be safe, to maintain the order of operations now. In the other branch

PR #8796, opened 2024-09-23
AG-12847 - extract sorting

peterjrreynolds:


community-modules/client-side-row-model/src/clientSideRowModel/sortStage.ts
```diff
@@ -1,24 +1,63 @@
...
+function updateChildIndexes(rowNode: RowNode): void {
+    if (_missing(rowNode.childrenAfterSort)) {
+        return;
+    }
+
+    const listToSort = rowNode.childrenAfterSort;
+    for (let i = 0; i < listToSort.length; i++) {
+        const child = listToSort[i];
+        const firstChild = i === 0;
+        const lastChild = i === rowNode.childrenAfterSort.length - 1;
```
StephenCooper, line 31: Can we extract the length from `listToSort` above the loop.

StephenCooper (approved):
LGTM

community-modules/client-side-row-model/src/clientSideRowModel/sortStage.ts
peterjrreynolds: Existing code. Will address in later change

PR #8800, opened 2024-09-23
AG-12677 update crossfiltering for integrated charts

StephM-W:


enterprise-modules/charts/src/charts/chartComp/services/chartCrossFilterService.ts
```diff
@@ -36,7 +36,14 @@ export class ChartCrossFilterService extends BeanStub implements NamedBean {
             return;
         }
 
-        const colId = ChartCrossFilterService.extractFilterColId(event);
+        let colId = ChartCrossFilterService.extractFilterColId(event);
+
+        if (colId.endsWith('Filter')) {
+            // remove 'Filter' suffix if present - this handles the area/line chart
+            // highlighting rather than filtering behaviour
+            colId = colId.replace('Filter', '');
```
peterjrreynolds: It might be worthwhile making this string a bit more distinct to avoid clash with user-defined IDs.

PR #8808, opened 2024-09-24
AG-12272 - step 5 - simplify master detail initialization

SalvatorePreviti:
At the moment, CSRM recreates all the rowNodes when masterDetail is toggled.
However, this seems to not be necessary, we just need to update the master flag and rerun refreshModel

### DIFF
--- a/community-modules/client-side-row-model/src/clientSideNodeManager/clientSideNodeManager.ts
+++ b/community-modules/client-side-row-model/src/clientSideNodeManager/clientSideNodeManager.ts
@@ -375,7 +375,16 @@ export class ClientSideNodeManager<TData>
         return node;
     }
 
-    protected setMasterForRow(rowNode: RowNode<TData>, data: TData, level: number, isNewRow: boolean): void {
+    public updateRowsMasterDetail(): void {
+        this.rootNode.allLeafChildren?.forEach((rowNode) => {
+            const data = rowNode.data;
+            if (data) {
+                this.setMasterForRow(rowNode, data, TOP_LEVEL, true);
+            }
+        });
+    }
+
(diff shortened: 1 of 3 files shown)


community-modules/client-side-row-model/src/clientSideNodeManager/clientSideNodeManager.ts
```diff
@@ -391,7 +400,7 @@ export class ClientSideNodeManager<TData>
             rowNode.setMaster(false);
         }
 
-        if (isNewRow) {
```
AndrewGlazier, line 394: The name `canExpandOrCollapse` isn't quite correct - isNewRow is more accurate as we _can_ expand and collapse when it's a transaction update, but don't want to apply the initial expanded state in that scenario

community-modules/client-side-row-model/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -293,6 +293,10 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             }
 
             if (arePropertiesImpacted(groupStageRefreshProps)) {
+                if (properties.includes('masterDetail')) {
+                    this.nodeManager.updateRowsMasterDetail?.();
```
AndrewGlazier, line 297: This is too keen, we don't want to expand master rows when grouping props changes, only when new row is created via transaction or setRowData

SalvatorePreviti: Ah, let's chat about this, but this will happen only if masterDetail property changes

PR #8816, opened 2024-09-25
AG-12929 Allow cell selection and range selection together

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12929 and https://ag-grid.atlassian.net/browse/AG-12902


community-modules/core/src/propertyKeys.ts
```diff
@@ -402,6 +401,7 @@ export class PropertyKeys {
         'suppressStickyTotalRow',
         // needs avoid coercion to boolean so that we can warn if it's unset
         'loadThemeGoogleFonts',
+        'rowSelection',
```
eliasmalik, line 404: Have to make `rowSelection` an "other" property because it's unioned with string values.

PR #8820, opened 2024-09-25
AG-12975 move masterDetail logic from RowNode+CSRM to MasterDetail module

SalvatorePreviti:
To improve tree shaking bundle size, and simplify the modifications to CSRM to handle different types of tree data, we agreed that would be simpler to move master detail logic outside of CSRM and RowNode into the MasterDetail module.

We are going to slowly apply the same refactoring to other methods present in RowNode in other PRs, but this is the first one.

The final behaviour is identical and we do have also unit tests for it.


packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -382,7 +375,9 @@ export class ClientSideNodeManager {
                 nodesToUnselect.push(rowNode);
             }
 
-            this.setMasterForRow(rowNode, item, TOP_LEVEL, false);
+            if (!this.gos.get('treeData')) {
```
SalvatorePreviti, line 378: Note: this is fine, as, we do reload all the rowNodes in any case if treeData or masterDetail grid options change - so we do not need to handle the case where an updated node is coming from a treeData option toggle or a masterDetail option toggle - it will never happen

packages/ag-grid-community/src/interfaces/iDetailGridApiService.ts
```diff
@@ -1,6 +1,10 @@
 import type { DetailGridInfo } from '../api/gridApi';
+import type { RowNode } from '../entities/rowNode';
 
 export interface IDetailGridApiService {
+    /** Called by CSRM to initialize a node as master */
+    setMasterForRow(rowNode: RowNode, data: any, shouldSetExpanded: boolean): void;
```
peterjrreynolds: Could get rid of the any here - `setMasterForRow<TData = any>(rowNode: RowNode<TData>, data: TData, shouldSetExpanded: boolean)`

PR #8837, opened 2024-09-27
AG-12847 - extract feature modules

peterjrreynolds:


packages/ag-grid-community/src/api/apiModule.ts
```diff
@@ -57,9 +32,7 @@ export const CoreApiModule = _defineModule<_CoreGridApi>({
     },
 });
 
-export const RowApiModule = _defineModule<_RowGridApi<any>>({
-    version: VERSION,
-    moduleName: '@ag-grid-community/row-api',
+export const RowApiModule = defineCommunityModule<_RowGridApi<any>>('RowApiModule', {
```
StephenCooper, line 35: Did you validate that the typing still works the same to ensure these methods are required and match the api interface?

packages/ag-grid-community/src/api/sharedApiModule.ts
```diff
@@ -1,27 +1,25 @@
...
 import { collapseAll, expandAll, onRowHeightChanged } from './csrmSsrmSharedApi';
 import type { _CsrmSsrmSharedGridApi, _SsrmInfiniteSharedGridApi } from './gridApi';
 import { getCacheBlockState, isLastRowIndexKnown, setRowCount } from './ssrmInfiniteSharedApi';
 
 // these modules are not used in core, but are shared between multiple other modules
 
-export const CsrmSsrmSharedApiModule = _defineModule<_CsrmSsrmSharedGridApi>({
-    version: VERSION,
-    moduleName: '@ag-grid-community/csrm-ssrm-shared-api',
+export const CsrmSsrmSharedApiModule = defineCommunityModule<_CsrmSsrmSharedGridApi>('CsrmSsrmSharedApiModule', {
```
StephenCooper, line 8: I wonder if there is a way to ensure we get these strings matching the module that we export.

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -249,37 +219,11 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
-            }
-
-            if (arePropertiesImpacted(filterAggStageRefreshProps)) {
-                this.refreshModel({ step: ClientSideRowModelSteps.FILTER_AGGREGATES });
-                return;
-            }
-
-            if (arePropertiesImpacted(flattenStageRefreshProps)) {
-                this.refreshModel({ step: ClientSideRowModelSteps.MAP });
+            for (const { refreshProps, step } of this.orderedStages) {
```
AndrewGlazier, line 222: Rob was opposed to these living inside of the stages, and wanted them in the CSRM - this was the original design I had though so not personally opposed to it

To add, you'll need to re-verify all the properties now they're moved into modules - some were used in multiple places but moved to the top-level refresh as they were always available, if the refresh now lives in the modules, and the modules are optional, these dependencies will need to live in all of the modules that rely on them as a dependency, not just the top level one

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -631,7 +631,11 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
 
         const event = { ...this.createGlobalRowEvent('rowGroupOpened'), expanded, event: e || null };
 
-        this.beans.rowNodeEventThrottle.dispatchExpanded(event, forceSync);
+        const { rowNodeEventThrottle } = this.beans;
+
+        rowNodeEventThrottle
+            ? rowNodeEventThrottle.dispatchExpanded(event, forceSync)
```
AndrewGlazier: can you add a comment here clarifying that this is for CSRM, also if statement is more appropriate than ternary for ops like this

packages/ag-grid-community/src/api/apiModule.ts
peterjrreynolds: Yes

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
peterjrreynolds: Everything that runs at the top level is a group property so won't do anything without grouping. All the other properties also align correctly with the feature stage with the exception of:
- `groupDisplayType` in the sort stage. This is also in the group stage, so no impact here. Could probably remove from the sort stage, but I've left it for now.
- All the flatten stage properties. These are all to do with grouping, so could probably exclude them if not using grouping. The flatten stage is mandatory, so I've left as is for now. Could probably refactor later to make the flatten stage optional.

packages/ag-grid-community/src/entities/rowNode.ts
peterjrreynolds: Done

packages/ag-grid-community/src/api/sharedApiModule.ts
peterjrreynolds: Have added this as something to look into on the module validation ticket.

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -139,9 +139,7 @@ export class ValueService extends BeanStub implements NamedBean {
             rowNode.field &&
             (column.getColDef().showRowGroup === true || column.getColDef().showRowGroup === rowNode.field);
 
-        if (forFilter && colDef.filterValueGetter) {
-            result = this.executeFilterValueGetter(colDef.filterValueGetter, data, column, rowNode);
```
StephenCooper, line 141: I don't see how this works?

packages/ag-grid-community/src/clientSideRowModel/rowNodeEventThrottle.ts
```diff
@@ -2,34 +2,24 @@ import type { NamedBean } from '../context/bean';
...
     beanName = 'rowNodeEventThrottle' as const;
 
     private animationFrameService?: AnimationFrameService;
-    private rowModel: IRowModel;
+    private rowModel: IClientSideRowModel;
 
     public wireBeans(beans: BeanCollection): void {
         this.animationFrameService = beans.animationFrameService;
-        this.rowModel = beans.rowModel;
+        this.rowModel = beans.rowModel as IClientSideRowModel;
```
StephenCooper, line 16: Is this and the logic below equivalent to ensure this only runs clientside?

peterjrreynolds: Before, every method in this class checked if it was client side before running. Now we wire it in the module with the row model set to client side. The check for dispatching directly to the event service is now in row node. So the behaviour is equivalent (but more efficient now).

packages/ag-grid-community/src/valueService/valueService.ts
peterjrreynolds: Before, filters used to call the value service with the `forFilter` argument set to true. Now the filter value getter logic just lives in the filter code.

PR #8830, opened 2024-09-26
AG-12915 - Add error codes on website

taktran:
Work moved to https://github.com/ag-grid/ag-grid/pull/8838

## Changes

* Generate error pages on website on url: `/[framework]-data-grid/errors/[code]?[params?]
* Get errors from export of `ag-grid-community`
* Allow error pages to be extended with markdoc content in `content/errors/[code].mdoc`

## References

Related to https://github.com/ag-grid/ag-grid/pull/8705

PR #8705, opened 2024-09-13
Extract error content from code

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-7440

PR #8838, opened 2024-09-27
AG-12915 - Add error codes page to the website

taktran:
## Changes

* Generate error pages on website at: `/[framework]-data-grid/errors/[code]?[params?]`
* Get errors from export of `ag-grid-community`
* Allow error pages to be extended with markdoc content in `content/errors/[code].mdoc`
* Do some transformation of objects when logging and showing the errors
* Integrate with https://github.com/ag-grid/ag-grid/pull/8705

## References

Previous work https://github.com/ag-grid/ag-grid/pull/8830

### DIFF
--- a/documentation/ag-grid-docs/src/content/errors/1.mdoc
+++ b/documentation/ag-grid-docs/src/content/errors/1.mdoc
@@ -1,0 +1,1 @@
+This is an error about row data. See [Row Data docs](./row-ids/).
--- a/documentation/ag-grid-docs/src/content/errors/2.mdoc
+++ b/documentation/ag-grid-docs/src/content/errors/2.mdoc
@@ -1,0 +1,12 @@
+---
+title: "Row data error #2"
+description: Row data description
+---
+
+This is an error about duplicate node ids. See [Row IDs docs](./row-ids/#row-ids).
+
+## Why is this error happening?
+
+Because you have a duplicate node id
+
+
--- a/documentation/ag-grid-docs/src/content/errors/2.mdoc
+++ b/documentation/ag-grid-docs/src/content/errors/2.mdoc
@@ -1,5 +1,4 @@
 ---
-title: "Row data error #2"
 description: Row data description
 ---
 
--- a/documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
+++ b/documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
@@ -12,6 +12,7 @@ import { SideNavigation } from '@components/pages-navigation/components/SideNavi
 import { getHeadings, getTopHeading } from '@utils/markdoc/getHeadings';
 import { DOCS_TAB_ITEM_ID_PREFIX } from '@ag-website-shared/constants';
 import { AG_GRID_ERRORS } from '../../../../../../packages/ag-grid-community/src/validation/errorMessages/errorText';
+import Note from '@ag-website-shared/components/alert/Note';
 
 interface Params {
     framework: Framework;
@@ -97,6 +98,7 @@ const headings = errorPage
 
                 <h2 id={fullErrorTextHeading.slug}>{fullErrorTextHeading.text}</h2>
                 <pre id="errorCodeText">{defaultErrorText}</pre>
+                <Note>To show this error text in the dev console, import the <code>ValidationModule</code>.</Note>
             </div>
         </div>
 
--- a/documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
+++ b/documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
@@ -76,16 +76,7 @@ const headings = errorPage
 <Layout title={title} description={description} showSearchBar={true} showDocsNav={true}>
     <div class:list={[styles.contentViewport, 'layout-grid']}>
         <PagesNavigation client:load menuSections={menuSections} framework={framework} pageName={pageName} />
-        <div
-            id="doc-content"
-            class:list={[
-                styles.docPage,
-                styles.errorPage,
-                {
-                    noSideMenu: styles.hideSideMenu,
-                },
-            ]}
-        >
+        <div id="doc-content" class:list={[styles.docPage, styles.errorPage]}>
             <Header
                 client:load
                 title={title}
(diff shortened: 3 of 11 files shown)


packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -63,5 +64,7 @@ export function getError<TId extends ErrorId, TParams extends GetErrorParams<TId
     }
 
     const errorBody = msgOrFunc(args as any);
-    return Array.isArray(errorBody) ? errorBody : [errorBody];
+    const errorLink = getErrorLink(errorId, args);
+    const errorSuffix = `\nSee ${errorLink}`;
```
taktran, line 68: Adding error link in dev console too, so users can potentially get more help on the error page

documentation/ag-grid-docs/src/content/errors/1.mdoc
```diff
@@ -0,0 +1 @@
+This is an error about row data. See [Row Data docs](./row-ids/).
```
taktran, line 1: These are placeholder mdoc files that are able to replace the default error page for the error code

packages/ag-grid-community/src/validation/logging.ts
```diff
@@ -27,13 +27,46 @@ function getMsgOrDefault<TId extends ErrorId>(logger: LogFn, id: TId, args: GetE
...
 }
 
-const minifiedLog = (errorNum: number, args: GetErrorParams<any>) => {
+/**
+ * Stringify object, removing any circular dependencies
+ */
+function stringifyObject(inputObj: any) {
+    const object: Record<string, any> = {};
+
+    for (const prop in inputObj) {
```
StephenCooper: Can we not just use `Object.entries`?

documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
```diff
@@ -0,0 +1,124 @@
...
+<Layout title={title} description={description} showSearchBar={true} showDocsNav={true}>
+    <div class:list={[styles.contentViewport, 'layout-grid']}>
+        <PagesNavigation client:load menuSections={menuSections} framework={framework} pageName={pageName} />
+        <div
+            id="doc-content"
+            class:list={[
+                styles.docPage,
+                styles.errorPage,
+                {
+                    noSideMenu: styles.hideSideMenu,
```
Comment (markdurrant) [inline documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro]: Do we need this line if we'll always have a side menu?

packages/ag-grid-community/src/validation/logging.ts
```diff
@@ -27,13 +27,46 @@ function getMsgOrDefault<TId extends ErrorId>(logger: LogFn, id: TId, args: GetE
...
+        if (
+            Object.prototype.hasOwnProperty.call(inputObj, prop) &&
+            typeof inputObj[prop] !== 'object' &&
+            typeof inputObj[prop] !== 'function'
+        ) {
+            object[prop] = inputObj[prop];
```
StephenCooper, line 38: Maybe we should put a limit on how many properties we stringify? Guard against very big objects

markdurrant (approved):
Everything that I understand looks good 👍 

One very minor question/comment.

documentation/ag-grid-docs/src/content/errors/2.mdoc
```diff
@@ -0,0 +1,12 @@
+---
+title: "Row data error #2"
```
StephenCooper: Should we let this be changed? Seems like this part should be consistent across the error pages.

packages/ag-grid-community/src/validation/logging.ts
taktran: Nope, tried that, but the issue is circular dependencies eg, when params has a `RowNode`

```
JSON.stringify(Object.entries(inputObj))
VM5840:1 Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'context' -> object with constructor 'Context'
    --- property 'beans' closes the circle
    at JSON.stringify (<anonymous>)
    at <anonymous>:1:6

JSON.stringify(Object.fromEntries(Object.entries(inputObj)))
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'context' -> object with constructor 'Context'
    --- property 'beans' closes the circle
    at JSON.stringify (<anonymous>)
    at <anonymous>:1:6
```

packages/ag-grid-community/src/validation/logging.ts
taktran: We were thinking we'll look at all the errors first, and then look into whether we need to restrict big objects. It could be that it'd be better to limit what is logged

documentation/ag-grid-docs/src/content/errors/2.mdoc
taktran: Good shout - have updated.

documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
Comment (taktran) [reply documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro]: Good point. Updated.

packages/ag-grid-community/src/validation/logging.ts
StephenCooper: I meant just for line 36, then you wouldn't have to have the hasOwnProperty check.

packages/ag-grid-community/src/validation/logging.ts
StephenCooper: Well, some of the users own objects could end up being the ones that are logged and that is where we have no control. Would be a shame for the link not to work, because it went past some limit.

packages/ag-grid-community/src/validation/logging.ts
taktran: Ah right 👍🏼  Updated.

packages/ag-grid-community/src/validation/logging.ts
taktran: Chatted in person, and we'll come back to this as we have more errors to look at.

taktran: Fixed this here: https://github.com/ag-grid/ag-grid/pull/9183

PR #8854, opened 2024-09-30
AG-12994 - Remove full store

AndrewGlazier:


packages/ag-grid-enterprise/src/serverSideRowModel/serverSideRowModel.ts
```diff
@@ -571,10 +568,7 @@ export class ServerSideRowModel extends BeanStub implements NamedBean, IServerSi
 
         const states: any = {};
         root.forEachStoreDeep((store) => {
-            if (store instanceof FullStore) {
-                const { id, state } = store.getBlockStateJson();
-                states[id] = state;
-            } else if (store instanceof LazyStore) {
+            if (store instanceof LazyStore) {
```
peterjrreynolds: If we only support the `LazyStore` now, should be able to tidy this up and remove the check

AndrewGlazier: done

packages/ag-grid-community/src/gridOptionsUtils.ts
```diff
@@ -316,6 +316,19 @@ export function _canSkipShowingRowGroup(gos: GridOptionsService, node: RowNode):
     return false;
 }
 
+export function _getMaxConcurrentDatasourceRequests(gos: GridOptionsService): number | undefined {
+    const res = gos.get('maxConcurrentDatasourceRequests');
```
StephenCooper, line 317: Should the default not be part of GRID OPTIONS DEFAULTS?

PR #8861, opened 2024-10-01
AG-13028 move updateSelectable to selection service

SalvatorePreviti:
As a result of splitting core modules the SelectableService was created in grouping module, but, it does not belong there as it needs to be in the Selection module.


packages/ag-grid-community/src/selection/baseSelectionService.ts
```diff
@@ -19,4 +41,76 @@ export abstract class BaseSelectionService extends BeanStub {
...
+
+    /**
+     * Updates the selectable state for a node by invoking isRowSelectable callback.
+     * If the node is not selectable, it will be deselected.
+     *
+     * Callers:
+     *  - property isRowSelectable changed
+     *  - after grouping / treeData
+     */
+    public updateSelectable(skipLeafNodes: boolean) {
```
AndrewGlazier, line 58: Can you split this out if you're now sharing this between both services? Most of this logic doesn't belong to the SSRM

SalvatorePreviti: @eliasmalik

PR #8871, opened 2024-10-02
AG-13035 - Switch to using ag-charts-types where possible.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13035


packages/ag-grid-enterprise/src/charts/main.ts
```diff
@@ -2,7 +2,7 @@ import { AgCharts, time } from 'ag-charts-community';
 
 export { GridChartsModule } from './gridChartsModule';
 
-export * from 'ag-charts-community';
```
alantreadway, line 5: This is the main point of the change, for tree-shaking purposes - updating the other non-implementation references is just a follow-through cleanup.

PR #8872, opened 2024-10-02
AG-12272 - step 9 - tree construction moved from grouping stage to client side node manager

SalvatorePreviti:
- Rename treeStage/treeStrategy into TreeNodeManager
- Move the generate transaction from row data logic into the abstract client side node manager, as it is used by flat client side node manager and getDataPath client side node manger
- Make the tree node manager work with the get data path node manager (all tests green)


packages/ag-grid-enterprise/src/treeData/clientSidePathTreeNodeManager.ts
```diff
@@ -1,335 +1,92 @@
...
-                } else {
-                    add.push(data);
-                }
-            });
+    public commitTransactions(
+        transactions: RowNodeTransaction<TData>[],
+        changedPath: ChangedPath | undefined,
+        rowNodesOrderChanged: boolean
+    ): void {
+        this.treeNodeManager.initRootNode(this.rootNode);
```
AndrewGlazier, line 39: Is this needed?

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1077,29 +1095,8 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
-                        afterColumnsChanged: afterColumnsChanged,
-                    });
-                }
-
-                this.updateSelectableAfterGrouping(changedPath);
-            }
-        }
-
-        if (!treeData) {
+            this.updateSelectableAfterGrouping(changedPath);
```
AndrewGlazier: This probably isn't needed as grouping isn't changing the structure of selection in tree data any more (still needed, but in the stage where we build the hierarchy)

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1349,20 +1346,27 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             return;
         }
 
+        const changedPath = this.createChangePath(rowNodeTransactions);
```
AndrewGlazier, line 1324: Can we rename the parent func (commonUpdateRowData) to be more clear that it's only for handling transactions

PR #8883, opened 2024-10-03
AG-12847 - update module names

peterjrreynolds:


packages/ag-grid-community/src/modules/moduleRegistry.ts
```diff
@@ -94,14 +93,12 @@ export function _assertModuleRegistered(
         return true;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const modName = Object.entries(ModuleNames).find(([k, v]) => v === moduleName)?.[0];
-    const warningMessage = `AG Grid: unable to use ${reason} as the ${modName} is not registered${areGridScopedModules ? ` for gridId: ${gridId}` : ''}. Check if you have registered the module:
-       
+    const isEnterprise = false; // TODO - enterprise check
```
StephenCooper: Use module enterprise flag

packages/ag-grid-community/src/syncService.ts
```diff
@@ -51,7 +50,7 @@ export class SyncService extends BeanStub implements NamedBean {
         this.eventService.dispatchEvent({
             type: 'gridReady',
         });
-        const isEnterprise = this.gos.isModuleRegistered(ModuleNames.EnterpriseCoreModule);
+        const isEnterprise = this.gos.isModuleRegistered('EnterpriseCoreModule');
```
StephenCooper: Move inside of the debug

StephenCooper (approved):
Approved with small changes.

packages/ag-grid-community/src/modules/moduleRegistry.ts
peterjrreynolds: Can't do this as we only have the module name, not the module itself. Will have to have a list of enterprise module names. Need to move this into the validation service

peterjrreynolds: Will do this as a separate task

peterjrreynolds: Done as part of this

packages/ag-grid-community/src/syncService.ts
peterjrreynolds: Done

PR #8903, opened 2024-10-04
AG-12915 - Allow base url to be changed based on environment

taktran:
Follow up to https://github.com/ag-grid/ag-grid/pull/8838

## Changes

* Add `baseUrl.ts` file to store the base url for error messages. Default to localhost
  * Update tests to use localhost url
* Add script to update `baseUrl.ts` for different environments
* Add script to check `baseUrl.ts` against different environments
* Clean up deployment scripts


taktran (commented):
Looks good from Sean

PR #8919, opened 2024-10-07
AG-13057 Split out non-SSRM updateSelectable code

eliasmalik:
https://ag-grid.atlassian.net/browse/AG-13057


packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -664,4 +666,71 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
...
+     *  - property isRowSelectable changed
+     *  - after grouping / treeData
+     */
+    public override updateSelectable(skipLeafNodes: boolean): void {
+        const { gos } = this;
+
+        const isRowSelecting = _isRowSelection(gos);
+        const isRowSelectable = _getIsRowSelectable(gos);
+
+        if (!isRowSelecting || !isRowSelectable) {
```
AndrewGlazier: This is interesting, we don't update status if no isRowSelectable func, seems like a bug, surely if we remove func then everything is selectable?

Also if `isRowSelecting` is false we ignore reactive changes to selectable?

Can you check if these two scenarios ^ have reproducible bugs we can make tickets for?

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -664,4 +666,71 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
...
+        const nodesToDeselect: RowNode[] = [];
+
+        const nodeCallback = (node: RowNode) => {
+            if (skipLeafNodes && !node.group) {
+                return;
+            }
+
+            // Only in the CSRM, we allow group node selection if a child has a selectable=true when using groupSelectsChildren
+            if (isCsrmGroupSelectsChildren && node.group) {
+                const hasSelectableChild = node.childrenAfterGroup!.some((rowNode) => rowNode.selectable === true);
```
AndrewGlazier: I'm not sure the implications of this - but in CSRM we allow selection if one of the descendants are selectable - what if groupSelects: 'filteredDescendents'? Should it not be selectable if there's no selectable children? This complicates the solution but can you highlight this scenario to Zoheil and see if we want to capture this as requirement?

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -664,4 +666,71 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
...
+            if (!rowSelectable && node.isSelected()) {
+                nodesToDeselect.push(node);
+            }
+        };
+
+        // Needs to be depth first in this case, so that parents can be updated based on child.
+        if (isCsrmGroupSelectsChildren) {
+            const csrm = this.rowModel as IClientSideRowModel;
+            const changedPath = new ChangedPath(false, csrm.getRootNode());
+            changedPath.forEachChangedNodeDepthFirst(nodeCallback, true, true);
```
AndrewGlazier: I wonder if (skipLeafNodes = true) scenario should pass the actual changed path in, so we can reduce unnecessary calls to unchanged groups

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -664,4 +666,71 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
...
+
+    /**
+     * Updates the selectable state for a node by invoking isRowSelectable callback.
+     * If the node is not selectable, it will be deselected.
+     *
+     * Callers:
+     *  - property isRowSelectable changed
+     *  - after grouping / treeData
+     */
+    public override updateSelectable(skipLeafNodes: boolean): void {
```
AndrewGlazier: I think we can get rid of the skipLeafNodes flag, it looks there's distinct scenarios for updating selectable:
1. isRowSelectable changed, for CSRM we want this to check all leaf nodes `isRowSelectable` callback, then if `groupSelectsChildren` we update parents based on children using DFS. Otherwise we iterate over all group nodes using a changed path, for this see point 2
2. Iterate over all group nodes (no leaf nodes) after grouping has changed - currently we created a changed path to do a full DFS, but the groupStage already has a changed path (which will be correct to impacted nodes via transactions), if we create a new func we could pass the changed path through, or fall back on a new one where omitted, this func could then be called from (1) for the last scenario

packages/ag-grid-community/src/selection/selectionService.ts
eliasmalik: Yup this is a bug, but is now fixed in this PR.

packages/ag-grid-community/src/selection/selectionService.ts
eliasmalik: Desired behaviour is currently being debated betwee Zoheil and Kiril, will do any necessary work under whatever JIRA ends up being generated from that.

packages/ag-grid-community/src/selection/selectionService.ts
eliasmalik: This is being done now after rebasing on Salvo's changes.

packages/ag-grid-community/src/selection/selectionService.ts
eliasmalik: I tried to explore a way to do this but I think we end up with more code overall, since we still need to account for infinite and viewport row models here.

packages/ag-grid-community/src/selection/selectionService.ts
SalvatorePreviti: Maybe just to be safe can we use ?. instead of ! to avoid it to blow up

SalvatorePreviti (approved):

PR #8943, opened 2024-10-08
AG-13079 refactor for treeshaking

StephM-W:


packages/ag-grid-community/src/main.ts
```diff
@@ -302,6 +303,7 @@ export { _requestAnimationFrame } from './misc/animationFrameService';
 export type { AnimationFrameService } from './misc/animationFrameService';
 export { AlignedGrid } from './interfaces/iAlignedGrid';
 export type { MenuService } from './misc/menu/menuService';
+export * from './interfaces/iColsService';
```
SalvatorePreviti: We do not do export * usually in main, especially if this is only types it might not compile in the right way. Explicit export would be better

packages/ag-grid-community/src/columns/baseColsService.ts
```diff
@@ -0,0 +1,435 @@
...
+import type { IColsService } from '../interfaces/iColsService';
+import type { WithoutGridCommon } from '../interfaces/iCommon';
+import { _removeFromArray } from '../utils/array';
+import { _attrToBoolean, _attrToNumber, _exists, _missingOrEmpty } from '../utils/generic';
+import { dispatchColumnChangedEvent } from './columnEventUtils';
+import type { ColKey, ColumnModel, Maybe } from './columnModel';
+import type { ColumnState, ColumnStateParams } from './columnStateService';
+import type { VisibleColsService } from './visibleColsService';
+
+export abstract class _BaseColsService extends BeanStub implements IColsService {
```
SalvatorePreviti: As Pete told me once, we do not need to add underscores for classes and interfaces we export. Also, in tree module I did use Abstract prefix, not Base - I think we should standardise and decide:
1) do we want underscores for abstract base classes?
2) shall we call them Base or Abstract?

I am fine with any approach, but would like to be consistent

@AndrewGlazier your input would be appreciated

packages/ag-grid-community/src/columns/columnUtils.ts
```diff
@@ -4,8 +4,9 @@ import { isColumn } from '../entities/agColumn';
 import type { AgProvidedColumnGroup } from '../entities/agProvidedColumnGroup';
 import { isProvidedColumnGroup } from '../entities/agProvidedColumnGroup';
 import type { ColumnInstanceId } from '../interfaces/iColumn';
-import { _warn } from '../validation/logging';
+import { _exists } from '../main-umd-noStyles';
```
SalvatorePreviti: this is not the right import - vscode mistake, it should be import from '../utils/generic' i think

SalvatorePreviti: @peterjrreynolds can we add an eslint error when we import the wrong things?

packages/ag-grid-community/src/interfaces/iColsService.ts
```diff
@@ -0,0 +1,33 @@
+import type { ColKey, Maybe } from '../columns/columnModel';
+import type { ColumnState, ColumnStateParams } from '../columns/columnStateService';
+import type { AgColumn } from '../entities/agColumn';
+import type { ColumnEventType } from '../events';
+
+// WIP during FuncColsService refactor
+interface IColsServiceExtraMethods {
```
SalvatorePreviti: mhhh do we really need this?
Another option is to define optional methods.

```
interface XXX {
  methodThatAlwaysExists(): void;
  methodThatCouldBeImplementedBySomeSubClassesButNotall?(): void
}```

Comment (SalvatorePreviti) [reply packages/ag-grid-community/src/interfaces/iColsService.ts]: Or maybe I am not sure I understand what this is for yet :)

File: packages/ag-grid-community/src/columns/baseColsService.ts (hunk 2 above)
Comment (StephM-W) [reply packages/ag-grid-community/src/columns/baseColsService.ts]: I used underscore here to indicate a private base class, rather than its abstract aspect. Happy to follow conventions

File: packages/ag-grid-enterprise/src/rowGrouping/rowGroupingApi.ts [hunk 5]
```diff
@@ -17,7 +19,7 @@ export function setColumnAggFunc(
     key: string | ColDef | Column,
     aggFunc: string | IAggFunc | null | undefined
 ): void {
-    beans.funcColsService.setColumnAggFunc(key, aggFunc, 'api');
+    (beans.valueColsService as ValueColsService).setColumnAggFunc(key, aggFunc, 'api');
```
SalvatorePreviti: instead of casting cannot we add the method in the interface? even if is an optional method

packages/ag-grid-community/src/columns/baseColsService.ts
```diff
@@ -0,0 +1,435 @@
...
+
+    public sortColumns(compareFn?: (a: AgColumn, b: AgColumn) => number): void {
+        this.columns.sort(compareFn);
+    }
+
+    protected abstract getEventName(): 'columnValueChanged' | 'columnRowGroupChanged' | 'columnPivotChanged';
+
+    public abstract setColumns(colKeys: ColKey[], source: ColumnEventType): void;
+
+    protected _setColumns(
```
SalvatorePreviti: I did not see the pattern of underscore anywhere for private and protected members, that I usually prefer but it does not follow what I've seen around in this codebase - I would suggest to remove the _ and if there is a conflict name to doSetColumns

packages/ag-grid-enterprise/src/rowGrouping/rowGroupingApi.ts
StephM-W: Good question - it means we then have that method on the other two related services, happy to follow conventions

SalvatorePreviti: I don't know if we have convention here, but in this case instead of casting I would add a nullable method interface FuncColsService { setColumnAggFunc?() : void; }

Same for all the other "as" for similar scenarios

packages/ag-grid-community/src/columns/baseColsService.ts
SalvatorePreviti: Ok, I'll happily wait for Andrew comment here, and, change Tree stuff accordingly in case.

packages/ag-grid-community/src/columns/columnUtils.ts
StephM-W: Makes sense - playing with eslint now

packages/ag-grid-community/src/interfaces/iColsService.ts
StephM-W: As we discussed, your suggestion of nullable methods can work here

packages/ag-grid-community/src/columns/funcColsService.ts
```diff
@@ -2,564 +2,96 @@ import type { NamedBean } from '../context/bean';
...
-            }
-        });
-
-        return res;
+        this.rowGroupColsService.extractCols(source, oldProvidedCols);
+        this.pivotColsService.extractCols(source, oldProvidedCols);
+        this.valueColsService.extractCols(source, oldProvidedCols);
     }
 
     public generateColumnStateForRowGroupAndPivotIndexes(
```
AndrewGlazier: Can you move this one into the `dataTypesService`, seems like the only place that uses it then we could delete the `funcColsService` entirely

packages/ag-grid-community/src/columns/funcColsService.ts
```diff
@@ -2,564 +2,96 @@ import type { NamedBean } from '../context/bean';
...
+        return this.pivotColsService.columns;
+    }
 
-        const column = this.columnModel.getColDefCol(sourceColumnId);
-        return column ? [column] : null;
+    public getSourceColumnsForGroupColumn(groupCol: AgColumn): AgColumn[] | null {
+        return this.rowGroupColsService.getSourceColumnsForGroupColumn(groupCol);
     }
 
     public sortRowGroupColumns(compareFn?: (a: AgColumn, b: AgColumn) => number): void {
```
AndrewGlazier: these sort funcs seem to only be called once, probably fine to delete and do in place

packages/ag-grid-enterprise/src/aggregation/aggColumnNameService.ts
```diff
@@ -2,22 +2,26 @@ import type {
...
 
-    private funcColsService: FuncColsService;
+    private rowGroupColsService: RowGroupColsService;
+    private valueColsService: ValueColsService;
     private columnModel: ColumnModel;
 
     public wireBeans(beans: BeanCollection) {
-        this.funcColsService = beans.funcColsService;
         this.columnModel = beans.columnModel;
+        this.rowGroupColsService = beans.rowGroupColsService as RowGroupColsService;
```
AndrewGlazier: Are these casts because the service is optional? We should leave them as optional types and treat them defensively as it's possible the services will be dropped, it's odd but not impossible to have aggregation with row grouping (grandTotalRows)

packages/ag-grid-community/src/columns/baseColsService.ts
AndrewGlazier: Don't think the _ is necessary, but there does seem to be precedent for the Base prefix (BaseSelectionService)

Personally would prefer the Abstract convention, but as the Base one seems more prevalent we should probably stick with it

packages/ag-grid-community/src/interfaces/iColsService.ts
AndrewGlazier: @peterjrreynolds doesn't like the i prefix
He also doesn't want to fix the iprefix
But he doesn't allow new i-prefixes

Might be worth asking him about his desired convention here

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -48,7 +50,8 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
         this.columnModel = beans.columnModel;
         this.columnNameService = beans.columnNameService;
         this.columnStateService = beans.columnStateService;
-        this.funcColsService = beans.funcColsService;
+        this.rowGroupColsService = beans.rowGroupColsService as RowGroupColsService;
```
AndrewGlazier: Treat as optional + defensively

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -173,7 +176,7 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         this.gos.get('functionsReadOnly') ||
                         column?.isRowGroupActive() ||
                         !column?.getColDef().enableRowGroup,
-                    action: () => this.funcColsService.addRowGroupColumns([column], source),
```
AndrewGlazier, line 176: It might be worth returning this as a disabled action if the relevant service is omitted

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -187,10 +190,10 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         disabled:
                             this.gos.get('functionsReadOnly') ||
                             lockedGroups === -1 ||
-                            lockedGroups >= this.funcColsService.rowGroupCols.length,
+                            lockedGroups >= this.rowGroupColsService.columns.length,
                         action: () =>
-                            this.funcColsService.setRowGroupColumns(
-                                this.funcColsService.rowGroupCols.slice(0, lockedGroups),
+                            this.rowGroupColsService.setColumns(
```
AndrewGlazier: If no service, disabled: true

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -208,7 +211,7 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         disabled:
                             this.gos.get('functionsReadOnly') ||
                             (underlyingColumn != null && this.columnModel.isColGroupLocked(underlyingColumn)),
-                        action: () => this.funcColsService.removeRowGroupColumns([showRowGroup], source),
+                        action: () => this.rowGroupColsService.removeColumns([showRowGroup], source),
```
AndrewGlazier: If no service, disabled: true

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -223,7 +226,7 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         !column?.isRowGroupActive() ||
                         !column?.getColDef().enableRowGroup ||
                         this.columnModel.isColGroupLocked(column),
-                    action: () => this.funcColsService.removeRowGroupColumns([column], source),
+                    action: () => this.rowGroupColsService.removeColumns([column], source),
```
AndrewGlazier: If no service, disabled: true

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -419,8 +422,8 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                 result.push({
                     name: localeTextFunc(funcName, aggFuncService.getDefaultFuncLabel(funcName)),
                     action: () => {
-                        this.funcColsService.setColumnAggFunc(columnToUse, funcName, 'contextMenu');
-                        this.funcColsService.addValueColumns([columnToUse!], 'contextMenu');
+                        this.valueColsService.setColumnAggFunc(columnToUse, funcName, 'contextMenu');
+                        this.rowGroupColsService.addColumns([columnToUse!], 'contextMenu');
```
AndrewGlazier: This should be valueColsService

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -409,8 +412,8 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
             result.push({
                 name: localeTextFunc('noAggregation', 'None'),
                 action: () => {
-                    this.funcColsService.removeValueColumns([columnToUse!], 'contextMenu');
-                    this.funcColsService.setColumnAggFunc(columnToUse, undefined, 'contextMenu');
+                    this.rowGroupColsService.removeColumns([columnToUse!], 'contextMenu');
```
AndrewGlazier: This should be valueColsService

AndrewGlazier: Re: parentFunc createAggregationSubMenu - this currently checks for row grouping module, we should change this to just check if the valueColsService is present (maybe? Thoughts? Can you take a dig into implications?)

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/agGridHeaderDropZones.ts
```diff
@@ -1,16 +1,17 @@
...
 
 export class AgGridHeaderDropZones extends Component {
     private columnModel: ColumnModel;
-    private funcColsService: FuncColsService;
+    private rowGroupColsService: RowGroupColsService;
 
     public wireBeans(beans: BeanCollection) {
         this.columnModel = beans.columnModel;
-        this.funcColsService = beans.funcColsService;
+        this.rowGroupColsService = beans.rowGroupColsService as RowGroupColsService;
```
AndrewGlazier: Defensive check, not cast

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/dropZoneColumnComp.ts
```diff
@@ -32,7 +32,7 @@ export class DropZoneColumnComp extends PillDragComp<AgColumn> {
         this.sortController = beans.sortController;
         this.columnModel = beans.columnModel;
         this.columnNameService = beans.columnNameService;
-        this.funcColsService = beans.funcColsService;
+        this.valueColsService = beans.valueColsService as ValueColsService;
```
AndrewGlazier: defensive check, not cast

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/dropZoneColumnComp.ts
```diff
@@ -331,7 +331,7 @@ export class DropZoneColumnComp extends PillDragComp<AgColumn> {
     private createAggSelect(hidePopup: () => void, value: any): Component {
         const itemSelected = () => {
             hidePopup();
-            this.funcColsService.setColumnAggFunc(this.column, value, 'toolPanelDragAndDrop');
+            this.valueColsService.setColumnAggFunc(this.column, value, 'toolPanelDragAndDrop');
```
AndrewGlazier: I think createAggSelect probably needs some defensive logic, so we don't do whatever is calling this func if no valueService

packages/ag-grid-community/src/interfaces/iColsService.ts
peterjrreynolds: `I` prefix should only be used going forward for things that have a  class implementation. E.g. services with interface in community and implementation in enterprise.

Params and things that are only ever going to be simple objects should not start with `I`

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/dropZoneColumnComp.ts
AndrewGlazier: Looking at this class wondering if it's worth extracting the agg logic, or breaking it down into  different classes which extend BaseDropZoneColumn etc, thoughts?

packages/ag-grid-enterprise/src/rowGrouping/rowGroupingApi.ts
AndrewGlazier: Agree with not casting, but in general can you make this a valueColsApiModule? We could use one for pivoting too then this func should live in there, at that point it should be safe to `beans.valueColsService!` as value cols service will be required by valueColsApi

packages/ag-grid-community/src/columns/baseColsService.ts
AndrewGlazier: We don't tend to, but it's inconsistent - but also not keen on `doSetColumns`

Looked at the funcs and it seems this just gets called by `setColumns` in each extending class, which is just passing up a callback for setting the col, I'd personally just pull the implementation of setColumns up to this level, and abstract the impl for setting a single column

packages/ag-grid-community/src/columns/baseColsService.ts
SalvatorePreviti: Let's use Base then, as it is also minimally shorter for bundle size :) will change the tree stuff to Base.

So, no underscore, yes Base

Consider that we were discussing with Pete ways to split internal stuff from public exports in a nicer way, in the next big release or the future

packages/ag-grid-community/src/columns/funcColsService.ts
StephM-W: Refactored

packages/ag-grid-community/src/columns/funcColsService.ts
StephM-W: Moved

packages/ag-grid-community/src/columns/baseColsService.ts
StephM-W: Pulled up implementations

StephM-W:
Rebuilt these changes on a clean branch, https://github.com/ag-grid/ag-grid/pull/9103

PR #8946, opened 2024-10-08
AG-8628 CTRL+A selects all rows

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-8628

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
@@ -4,11 +4,13 @@ import { BeanStub } from '../../context/beanStub';
 import type { BeanCollection } from '../../context/context';
 import type { AgColumn } from '../../entities/agColumn';
 import type { FocusService } from '../../focusService';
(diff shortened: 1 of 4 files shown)


packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
```diff
@@ -297,7 +301,21 @@ export class RowContainerEventsFeature extends BeanStub {
                 columnStart: allDisplayedColumns[0],
                 columnEnd: _last(allDisplayedColumns),
             });
+        } else if (selectionService) {
+            const justFiltered = _getSelectAllFiltered(gos);
+            const justCurrentPage = _getSelectAllCurrentPage(gos);
+            const source = justCurrentPage
+                ? 'keyboardSelectAllCurrentPage'
```
AndrewGlazier: I wonder if we need all three event sources - is keyboardSelectAll not sufficient? Users will be unlikely to ever get the scenario where they need multiple of these - and if they do they can check their own grid options

Greglar4:
How would I suppress this functionality without completely suppressing any ctrl+A keyboard event?

PR #8954, opened 2024-10-09
AG-12861 Add support for SelectionColumnDef to state service

eliasmalik:
See:
- https://ag-grid.atlassian.net/browse/AG-12861
- https://ag-grid.atlassian.net/browse/AG-13101
- https://ag-grid.atlassian.net/browse/AG-13102


packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -3,14 +3,14 @@ import { BeanStub } from '../context/beanStub';
 import { AgColumn } from '../entities/agColumn';
 import type { ColDef } from '../entities/colDef';
 import { _getCheckboxes, _getHeaderCheckbox } from '../gridOptionsUtils';
-import { isColumnControlsCol } from './columnUtils';
+import { isColumnSelectionCol } from './columnUtils';
 
-export const CONTROLS_COLUMN_ID_PREFIX = 'ag-Grid-ControlsColumn' as const;
```
AndrewGlazier, line 8: We can't change this at this point without breaking state or apps that may have hardcoded this ID - it might not be an issue because we're so early on but we need to agree this as a breaking change if we're going to do it

packages/ag-grid-community/src/columns/columnStateService.ts
```diff
@@ -201,6 +207,24 @@ export class ColumnStateService extends BeanStub implements NamedBean {
             // autogroup cols with nothing else, apply the default
             autoColsCopy.forEach(applyDefaultsFunc);
 
+            // sync selection cols with ColumnState
+            const selectionCols = this.columnModel.getAutoCols()?.slice() ?? [];
```
AndrewGlazier: Is the selection col in the auto cols?

eliasmalik: 🤦🏽

packages/ag-grid-community/src/columns/columnStateService.ts
```diff
@@ -201,6 +207,24 @@ export class ColumnStateService extends BeanStub implements NamedBean {
...
+                    params.defaultState,
+                    null,
+                    null,
+                    true,
+                    source,
+                    callbacks
+                );
+            });
+            selectionCols.forEach(applyDefaultsFunc);
+
```
peterjrreynolds, line 226: Can we combine these two into a common function (e.g. `syncColStates = (colStates, getColFuncName, getColsFuncName) => void`)?

eliasmalik: done

peterjrreynolds (approved):

PR #8955, opened 2024-10-09
AG-12272 - step 11 - simplify master detail logic

SalvatorePreviti:
Simplify and correct the logic to handle masterDetail in the CSRM
- In one of the previous PR I wrongly removed the setMasterForRow call in transaction executeUpdate - fixed here
- use a field change detection logic to understand when setMasterForRow should be executed in all rows

Other refactoring
- rename rootNode to rootRow, as we will move in the next PR the TreeNodeManager into the AbstractClientSideTreeNodeManager
- remove unused expanded function


packages/ag-grid-community/src/clientSideRowModel/abstractClientSideNodeManager.ts
```diff
@@ -460,6 +463,10 @@ export abstract class AbstractClientSideNodeManager<TData = any>
 
         this.nextId++;
 
+        if (this.isMasterDetail()) {
```
AndrewGlazier: Maybe to skip this tree data check we could provide an impl in the clientSideNodeManager (the one for normal grouping) step for createRowNode, which calls to `super.createRowNode` and does this step afterwards?

Just noticed we were doing this, and now aren't - can you remind me why?

SalvatorePreviti: implementation changed and currently the check is in the masterdetail service. We will remove this check when we allow treeData + masterDetail

packages/ag-grid-community/src/clientSideRowModel/abstractClientSideNodeManager.ts
```diff
@@ -45,7 +45,7 @@ export abstract class AbstractClientSideNodeManager<TData = any>
     private nextId = 0;
     protected allNodesMap: { [id: string]: RowNode } = {};
 
-    public rootNode: AbstractClientSideNodeManager.RootNode<TData>;
+    public rootRow: AbstractClientSideNodeManager.RootNode<TData>;
```
AndrewGlazier: I note you're renaming these to row - this might make sense for tree data because of the tree manager, but tree data is still less common than row grouping, and there is conventional differences in the grid for node vs row

I think we should keep calling these nodes, not rows, at the very least where it is impacting grouping/CSRM etc (which this abstraction does)

See
https://www.ag-grid.com/javascript-data-grid/grid-api/#reference-selection-getSelectedNodes
https://www.ag-grid.com/javascript-data-grid/grid-api/#reference-selection-getSelectedRows

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -274,8 +274,14 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             if (rowDataChanged) {
                 const rowData = this.gos.get('rowData');
                 if (rowData) {
-                    if (this.isImmutableRowDataActive()) {
-                        this.setImmutableRowData(rowData, masterDetailChanged);
+                    const getRowIdProvided =
```
AndrewGlazier, line 275: This feels like a node manager concern now, maybe the CSRM shouldn't care about how we set data?

Bit messy as we have to generate a transaction for one case in immutable data, but we could return the generated transaction where available, and the generated changed path (which could help simplify the pattern for new tree data) 

We could then move the rowDataUpdated event into the nodeManagers too (which I think also feels correct?)

I think this would then clarify a lot of the paths for the CSRM, what do you think?

This would also remove the `rowNodeOrderChanged` concern from the normal transactions approach, as this could just then live in the node manager

packages/ag-grid-enterprise/src/masterDetail/detailGridApiService.ts
```diff
@@ -3,48 +3,67 @@ import type {
...
+    private rowModel: IRowModel;
     private funcColsService: FuncColsService;
 
     public wireBeans(beans: BeanCollection): void {
         this.funcColsService = beans.funcColsService;
+        this.rowModel = beans.rowModel;
     }
 
-    private detailGridInfoMap: { [id: string]: DetailGridInfo | undefined } = {};
+    private isMasterDetailEnabled() {
```
AndrewGlazier: This needs to live in a new masterDetail service, as the API Service isn't a core module, this behaviour will be core to using master detail (with CSRM, maybe make a CSRM Master Detail Module?)

packages/ag-grid-enterprise/src/masterDetail/detailGridApiService.ts
```diff
@@ -3,48 +3,67 @@ import type {
...
-            for (let i = 0, len = rows.length; i < len; i++) {
-                const rowNode = rows[i];
-                const data = rowNode.data;
-                if (data) {
-                    this.setMasterForRow(rowNode, data, true);
+    public postConstruct(): void {
+        let enabled = this.isMasterDetailEnabled();
+        this.addManagedPropertyListeners(['treeData', 'masterDetail'], (params) => {
+            const properties = params.changeSet?.properties;
+            if (properties && _isClientSideRowModel(this.gos)) {
```
AndrewGlazier: As row model isn't reactive, we should probably only add this listener when in CSRM, rather than checking when they change

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
```diff
@@ -70,27 +70,13 @@ export class ClientSideChildrenTreeNodeManager<TData>
         treeNodeManager.commitTree();
     }
 
-    public setMasterForAllRows(rows: RowNode<TData>[] | null | undefined, shouldSetExpanded: boolean): void {
-        if (!this.gos.get('treeData')) {
-            this.beans.detailGridApiService?.setMasterForAllRows(rows, shouldSetExpanded);
-        }
-    }
-
     public onTreeDataChanged() {
```
AndrewGlazier, line 73: I'm still a bit unsure about why we do this - would it not be better to recreate the manager when something is impacting which node manager is being used?

packages/ag-grid-community/src/clientSideRowModel/abstractClientSideNodeManager.ts
SalvatorePreviti: agreed

packages/ag-grid-enterprise/src/masterDetail/detailGridApiService.ts
SalvatorePreviti: agreed

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: It should be indeed moved and the way changedPath and transactions for immutable row data are passed improved - but that's not for this PR, it was planned for the next and this is just an initial cleanup. It was easy in this case to inline the method

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
SalvatorePreviti: Ah no this is not being called when node manager changes but only when the property "treeData" or (flatten if you want to change the API) is toggled and there was not a setNewRowData.
In case of tree with children node manager we do not switch the node manager when treeData changes but instead we invalivalidate all the tree nodes and flatten/unflatted by calling a commit tree

packages/ag-grid-enterprise/src/masterDetail/detailGridApiService.ts
SalvatorePreviti: Added a new ClientSideDetailService

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
AndrewGlazier: Does the tree data prop changing not change the node manager? Flatten will be flattening the existing data as a new step, not changing the node manager right? So I assume flatten wouldn't even need to impact any internals of node manager

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
AndrewGlazier: OK happy to do in another PR - I'd say this should be done before we consider merging to latest though.

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
SalvatorePreviti: I don't think we need an extra step, is simpler to just invalidate all tree nodes and call a tree commit whenever the flag we use for flattening will change - the tree node manager checks the treeData flag (or if we want to rename the flatten flag) when it builds or updates the tree - if the flag is off it does not build sublevels but just put everything in the first level. This was part of the previous PR. In this way we also do not recreate rows and we just update them

PR #8961, opened 2024-10-09
AG-12915 - Redirect to archive version of website if version of the error is different to the deployed website grid version

taktran:
From the next major release v33.0.0 onwards, an error will link to the production website, and if it isn't the same version as production, it will redirect to the archive version of the website.


markdurrant (approved):
As much as I understand this, LGTM 👍

PR #8968, opened 2024-10-10
AG-7440 batch 6

StephenCooper:


packages/ag-grid-community/src/validation/rules/colDefValidations.ts
```diff
@@ -125,6 +125,39 @@ const COLUMN_DEFINITION_VALIDATIONS: Validations<ColDef | ColGroupDef> = {
         },
     },
 
+    filter: {
```
StephenCooper: Might need to put these back until the col def validation changes are made.

packages/ag-grid-community/src/widgets/component.ts
```diff
@@ -177,7 +177,7 @@ export class Component<TLocalEvent extends string = ComponentEvent>
                     // 3. The property is on a child component and not available on the parent during construction.
                     //    In which case you may need to pass the template via setTemplate() instead of in the super constructor.
                     // 4. The data-ref is not used by the component and should be removed from the template.
-                    _warnOnce(`Issue with data-ref: ${elementRef} on ${this.constructor.name} with ${current}`);
+                    _errorOnce(`Issue with data-ref: ${elementRef} on ${this.constructor.name} with ${current}`);
```
peterjrreynolds: I know these errors are for us, but shouldn't we extract them anyway?

documentation/ag-grid-docs/src/features/example-runner/framework-templates/lib/Extras.tsx
```diff
@@ -47,8 +47,8 @@ export const Extras = ({ extras }: Props) => {
     extras.forEach((extra) => {
         requiredExtras.add(extra);
 
-        if (extra === 'bootstrap' || extra === 'jqueryui') {
-            // bootstrap and jQuery UI require jQuery
```
StephenCooper, line 51: We removed bootstrap a while ago as an extra

packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -304,18 +312,185 @@ export const AG_GRID_ERRORS = {
...
+    136: () =>
+        `Unable to update chart as the 'type' is missing. It must be either 'rangeChartUpdate', 'pivotChartUpdate', or 'crossFilterChartUpdate'.` as const,
+    137: ({ type, currentChartType }: { type: string; currentChartType: string }) =>
+        `Unable to update chart as a '${type}' update type is not permitted on a ${currentChartType}.` as const,
+    138: ({ chartType }: { chartType: string }) => `invalid chart type supplied: ${chartType}` as const,
+    139: ({ customThemeName }: { customThemeName: string }) =>
+        `a custom chart theme with the name ${customThemeName} has been supplied but not added to the 'chartThemes' list` as const,
+    140: ({ name }: { name: string }) =>
+        `no stock theme exists with the name '${name}' and no custom chart theme with that name was supplied to 'customChartThemes'` as const,
+    141: () => 'crossing filtering with row grouping is not supported.' as const,
```
peterjrreynolds: Can fix the typo here and below - `cross filtering`

packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -304,18 +312,185 @@ export const AG_GRID_ERRORS = {
...
+    142: () => 'crossing filtering is only supported in the client side row model.' as const,
+    143: ({ panel }: { panel: string | undefined }) => `'${panel}' is not a valid Chart Tool Panel name` as const,
+    144: ({ type }: { type: string }) => `Invalid charts data panel group name supplied: '${type}'` as const,
+    145: ({ group }: { group: string }) =>
+        `As of v32, only one charts customize panel group can be expanded at a time. '${group}' will not be expanded.` as const,
```
peterjrreynolds, line 329: Maybe we should add a v32 deprecation comment here so we can remove this later

packages/ag-grid-community/src/widgets/component.ts
AndrewGlazier: If they're just for us I wonder if we should be more aggressive and just throw an error

StephenCooper: Yeah I think I will change to throw an error as if this is happening something is definitely wrong.

PR #8971, opened 2024-10-10
Ag 12236 docs nav improvements

markdurrant:
- Replaces docs nav component with newer component from Charts 
- Replaces old "menus" json with specific data per menu 
- Lazy "sync" of website-shared code


documentation/ag-grid-docs/src/utils/apiMenuPath.ts
```diff
@@ -1,20 +1,22 @@
-import { type MenuData } from '@ag-grid-types';
-
 function findSectionPaths(sections) {
+    // TODO: Understand why this is necessary
```
taktran: Did you figure this out?

taktran (approved):
Looks good 👍🏼 . Just a comment

documentation/ag-grid-docs/src/utils/apiMenuPath.ts
markdurrant: No longer required. (So no I never figured it out).

PR #8972, opened 2024-10-10
AG-12847 - row cell loops

peterjrreynolds:


packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -1813,7 +1816,7 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
     public getCellCtrl(column: AgColumn): CellCtrl | null {
         // first up, check for cell directly linked to this column
         let res: CellCtrl | null = null;
-        this.getAllCellCtrls().forEach((cellCtrl) => {
+        this.someCellCtrls((cellCtrl) => {
             if (cellCtrl.getColumn() == column) {
                 res = cellCtrl;
```
peterjrreynolds, line 1821: Can we return `true` here and break the loop?

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -1828,7 +1831,7 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
         // more expensive, as spanning cols is a
         // infrequently used feature so we don't need to do this most
         // of the time
-        this.getAllCellCtrls().forEach((cellCtrl) => {
+        this.someCellCtrls((cellCtrl) => {
             if (cellCtrl.getColSpanningList().indexOf(column) >= 0) {
                 res = cellCtrl;
```
peterjrreynolds, line 1836: Can we return `true` here and break the loop?

packages/ag-grid-community/src/edit/editService.ts
```diff
@@ -137,9 +137,7 @@ export class EditService extends BeanStub implements NamedBean {
     }
 
     public stopAllEditing(cancel: boolean = false): void {
-        this.rowRenderer.getAllRowCtrls().forEach((rowCtrl) => {
-            rowCtrl.stopEditing(cancel);
-        });
+        this.rowRenderer.someRowCtrls((rowCtrl) => rowCtrl.stopEditing(cancel));
```
AndrewGlazier, line 140: Will this not return true when editing is stopped in one cell preventing other cells from being stopped? Or is that not a valid scenario

packages/ag-grid-community/src/edit/rowEditService.ts
```diff
@@ -14,19 +14,16 @@ export class RowEditService extends BeanStub implements NamedBean {
         event: KeyboardEvent | null = null
     ): boolean {
         let preventDefault = true;
-        const atLeastOneEditing = rowCtrl.getAllCellCtrls().reduce((prev: boolean, cellCtrl: CellCtrl) => {
```
AndrewGlazier: maybe you're not here yet, but this service looks redundant? Can we unwrap these funcs?

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
StephenCooper: Can a column match more than one cellCtrl? Current implementation would get the last one if multiple matches whereas with a return we would get the first. 

On paper seems that this would be safe to return the first.

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -775,10 +777,10 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
     private addListenersForCellComps(): void {
         this.addManagedListeners(this.rowNode, {
             rowIndexChanged: () => {
-                this.getAllCellCtrls().forEach((cellCtrl) => cellCtrl.onRowIndexChanged());
+                this.someCellCtrls((cellCtrl) => cellCtrl.onRowIndexChanged());
             },
             cellChanged: (event) => {
-                this.getAllCellCtrls().forEach((cellCtrl) => cellCtrl.onCellChanged(event));
+                this.someCellCtrls((cellCtrl) => cellCtrl.onCellChanged(event));
```
AndrewGlazier, line 783: Might be able to gain some perf here by returning true if this hits the right cell

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
AndrewGlazier: Should be yeah

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
AndrewGlazier: Yes... though this looks broken anyway, as our impl lets multiple cells span the same col, we need to fix spanning but should be fine to make this change anyway

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -771,16 +769,16 @@ export class RowRenderer extends BeanStub implements NamedBean {
             cellFocused = this.getCellToRestoreFocusToAfterRefresh() || null;
         }
 
-        for (const rowCtrl of this.getRowCtrls(rowNodes)) {
+        this.someMatchingRowCtrls(rowNodes, (rowCtrl) => {
             if (!rowCtrl.isFullWidth()) {
-                continue;
+                return;
```
StephenCooper, line 774: I think this should have a comment about how it's not actually returning but continuing / skipping non full width rows.

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -751,10 +750,9 @@ export class RowRenderer extends BeanStub implements NamedBean {
             newData: false,
             suppressFlash: params.suppressFlash,
         };
-
-        for (const cellCtrl of this.getCellCtrls(params.rowNodes, params.columns as AgColumn[])) {
-            cellCtrl.refreshOrDestroyCell(refreshCellParams);
-        }
+        this.someMatchingCellCtrls(params.rowNodes, params.columns as AgColumn[], (cellCtrl) =>
+            cellCtrl.refreshOrDestroyCell(refreshCellParams)
```
AndrewGlazier, line 754: I don't have time to finish reviewing right now, but note to self this is warning bells

Previously we were iterating a copy, and any modifications to the original wouldn't be an issue - if we're doing refresh or destroy is this going to impact the list we're currently iterating?

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -771,16 +769,16 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
-        for (const rowCtrl of this.getRowCtrls(rowNodes)) {
+        this.someMatchingRowCtrls(rowNodes, (rowCtrl) => {
             if (!rowCtrl.isFullWidth()) {
-                continue;
+                return;
             }
 
             const refreshed = rowCtrl.refreshFullWidth();
             if (!refreshed) {
                 this.redrawRow(rowCtrl.getRowNode(), true);
```
AndrewGlazier, line 779: Same as previous warning bells, note to self revisit later

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -1073,7 +1075,7 @@ export class RowRenderer extends BeanStub implements NamedBean {
         // embedded, as what appears in each section depends on whether we are pinned or not
         const rowsToRemove: string[] = [];
 
-        this.getFullWidthRowCtrls().forEach((fullWidthCtrl) => {
+        this.someFullWidthRowCtrls((fullWidthCtrl) => {
             const rowIndex = fullWidthCtrl.getRowNode().rowIndex;
             rowsToRemove.push(rowIndex!.toString());
```
AndrewGlazier, line 1080: warning bells

packages/ag-grid-community/src/edit/rowEditService.ts
peterjrreynolds: This service is specific to full row editing only (not used for normal single cell editing)

packages/ag-grid-community/src/edit/editService.ts
peterjrreynolds: No, `stopEditing` is a `void` method

AndrewGlazier: Ah got confused with cellCtrl, fair enough

packages/ag-grid-community/src/edit/rowEditService.ts
AndrewGlazier: I... don't see how that changes things. I'm taking about the `rowEditService` - it contains no state, we can remove this as a service and export stand-alone funcs

AndrewGlazier:
Not sure if you want to extend this to the `.allRowCtrls` property - it's a little extra work due to having to having to change how the row container works, but we could gain a small chunk of perf by not having to `updateAllRowCtrls` anymore

packages/ag-grid-community/src/edit/rowEditService.ts
peterjrreynolds: We can't. Full row editing is a separate module (`FullRowEditModule`). For us to tree-shake the code it needs to be tied to something which can be optionally referenced elsewhere (in this case, inside `RowCtrl`).  Therefore it needs to live in a service.

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
yukhartz2625: Does breaking the loop early by returning true significantly improve performance, especially with large datasets?

PR #8976, opened 2024-10-11
AG-12847 - row/cell iterators

peterjrreynolds:


packages/ag-grid-community/src/utils/array.ts
```diff
@@ -131,3 +131,21 @@ export function _forEachReverse<T>(list: T[], action: (value: T, index: number)
         action(list[i], i);
     }
 }
+
+export function _some<T>(iterator: Iterable<T>, predicate: (item: T) => boolean): boolean {
```
AndrewGlazier, line 135: We could reduce code here by leaning on _find and searching for not undefined, except it would break that odd use case you're searching for an undefined 😂

peterjrreynolds: Probably better to stay clean and keep separate

packages/ag-grid-community/src/edit/editApi.ts
```diff
@@ -22,26 +22,26 @@ export function getCellEditorInstances<TData = any>(
...
 
     return res;
 }
 
 export function getEditingCells(beans: BeanCollection): CellPosition[] {
     const res: CellPosition[] = [];
 
-    beans.rowRenderer.getAllCellCtrls().forEach((cellCtrl) => {
+    for (const cellCtrl of beans.rowRenderer.getAllCellCtrls()) {
         if (cellCtrl.isEditing()) {
```
gportela85, line 40: Not a big deal, but I would go for `if (!cellCtrl.isEditing()) { continue; }`

PR #8985, opened 2024-10-14
AG-12847 - more bundle size changes

peterjrreynolds:


packages/ag-grid-community/src/columnMove/internalColumnMoveUtils.ts
```diff
@@ -257,9 +257,9 @@ function calculateValidMoves(params: {
     // so the result we return has to be and index location for this list
     const allGridCols = columnModel.getCols();
 
-    const movingDisplayedCols = allDisplayedCols.filter((col) => _includes(movingCols, col));
-    const otherDisplayedCols = allDisplayedCols.filter((col) => !_includes(movingCols, col));
-    const otherGridCols = allGridCols.filter((col) => !_includes(movingCols, col));
+    const movingDisplayedCols = allDisplayedCols.filter((col) => movingCols.includes(col));
```
StephenCooper, line 260: Do we need to also handle the undefined case which the old _includes method did in case any of these arrays can be undefined?

packages/ag-grid-community/src/dragAndDrop/dragAndDropService.ts
```diff
@@ -28,6 +26,14 @@ export enum DragSourceType {
     AdvancedFilterBuilder,
 }
 
+function _getBodyWidth(): number {
```
StephenCooper, line 29: Does this really benefit us moving these in here? I thought previously we would just leave these utils in place to reduce the risk of duplication across features. They are tree shaken if not used anyway.

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -561,7 +560,7 @@ export class ColumnFilterService extends BeanStub implements NamedBean {
     public createFilterParams(column: AgColumn, colDef: ColDef): IFilterParams {
         const params: IFilterParams = this.gos.addGridCommonParams({
             column,
-            colDef: _cloneObject(colDef),
+            colDef,
```
StephenCooper, line 563: Was this clone in place for a reason? Not to do with something like if a colDef is coming from a store or something?

If not blame available or link to a jira understand this might be hard to validate.

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
```diff
@@ -21,16 +21,48 @@ import type { UndoRedoService } from '../../undoRedo/undoRedoService';
...
 import type { MouseEventService } from './../mouseEventService';
 
+const A_KEYCODE = 65;
+const C_KEYCODE = 67;
+const V_KEYCODE = 86;
+const D_KEYCODE = 68;
+const Z_KEYCODE = 90;
+const Y_KEYCODE = 89;
+
+function _normaliseQwertyAzerty(keyboardEvent: KeyboardEvent): string {
```
StephenCooper, line 36: Again is this really better here?

packages/ag-grid-community/src/rendering/cell/cellComp.ts
```diff
@@ -573,7 +572,7 @@ export class CellComp extends Component implements TooltipParentComp {
 
         // if focus is inside the cell, we move focus to the cell itself
         // before removing it's contents, otherwise errors could be thrown.
-        if (eGui.contains(_getActiveDomElement(this.beans.gos)) && _browserSupportsPreventScroll()) {
+        if (eGui.contains(_getActiveDomElement(this.beans.gos))) {
```
StephenCooper, line 575: Why has _browserSupportsPreventScroll been removed from the check?

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
```diff
@@ -4,10 +4,20 @@ import type { BeanCollection } from '../../context/context';
 import type { AgColumn } from '../../entities/agColumn';
 import type { RowNode } from '../../entities/rowNode';
 import { _getGroupSelection, _isCellSelectionEnabled, _isRowSelection } from '../../gridOptionsUtils';
-import { _isDeleteKey } from '../../utils/keyboard';
+import { _isMacOsUserAgent } from '../../utils/browser';
 import type { RowCtrl } from '../row/rowCtrl';
 import type { CellCtrl } from './cellCtrl';
 
+function _isDeleteKey(key: string, alwaysReturnFalseOnBackspace = false) {
```
StephenCooper, line 11: Some query as other utils.

packages/ag-grid-enterprise/src/multiFilter/multiFilter.ts
```diff
@@ -43,6 +42,16 @@ export function getMultiFilterDefs(params: MultiFilterParams): IMultiFilterDef[]
         : [{ filter: 'agTextColumnFilter' }, { filter: 'agSetColumnFilter' }];
 }
 
+function _forEachReverse<T>(list: T[] | null | undefined, action: (value: T, index: number) => void): void {
```
StephenCooper, line 45: Same

StephenCooper: Although I guess by moving this one we don't have to export it from community, which is a cleaner thing.

packages/ag-grid-community/src/clientSideRowModel/immutableService.ts
```diff
@@ -105,11 +104,11 @@ export class ImmutableService extends BeanStub implements NamedBean {
         }
 
         // at this point, all rows that are left, should be removed
-        _iterateObject(existingNodesMap, (id, rowNode) => {
+        for (const rowNode of Object.values(existingNodesMap)) {
```
AndrewGlazier, line 107: Probs let Salvo know if you're impacting stuff again

packages/ag-grid-community/src/dragAndDrop/dragAndDropService.ts
peterjrreynolds: If something is used in a single place, it should not be a global util. Our historical approach is a bit of an anti-pattern.

I have moved these all back so we can clearly see where they are being used.

packages/ag-grid-community/src/filter/columnFilterService.ts
peterjrreynolds: It has been this way forever. I could not see a good reason for it though, and it is inconsistent with all other user components.

packages/ag-grid-community/src/rendering/cell/cellComp.ts
peterjrreynolds: This check was only in place for an old safari version that is no longer supported.

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
peterjrreynolds: See previous

packages/ag-grid-community/src/clientSideRowModel/immutableService.ts
peterjrreynolds: No change in behaviour here

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
peterjrreynolds: See previous

packages/ag-grid-enterprise/src/multiFilter/multiFilter.ts
peterjrreynolds: See previous

packages/ag-grid-community/src/columnMove/internalColumnMoveUtils.ts
peterjrreynolds: In this particular case, it can't be undefined.

Anywhere else, it will help us to expose any incorrect typings.

packages/ag-grid-community/src/filter/provided/date/defaultDateComponent.ts
```diff
@@ -179,6 +179,6 @@ export class DefaultDateComponent extends Component implements IDateComp {
             return params.filterParams.browserDatePicker;
         }
 
-        return _isBrowserChrome() || _isBrowserFirefox() || (_isBrowserSafari() && _getSafariVersion() >= 14.1);
+        return true;
```
AndrewGlazier: return params?.filterParams?.browserDatePicker ?? true

peterjrreynolds: Done

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -130,7 +129,7 @@ export class HeaderRowCtrl extends BeanStub {
         if (!this.headerCellCtrls) {
             return;
         }
-        return _values(this.headerCellCtrls).find((cellCtrl) => cellCtrl.getColumnGroupChild() === column);
+        return Array.from(this.headerCellCtrls.values()).find((cellCtrl) => cellCtrl.getColumnGroupChild() === column);
```
AndrewGlazier: This would benefit from being unwrappred from the array, we can just manually find with the iterator

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -621,7 +619,7 @@ export class CellCtrl extends BeanStub {
             const flashCell = !suppressFlash && !processingFilterChange && colDef.enableCellChangeFlash;
 
             if (flashCell) {
-                this.flashCell();
+                this.beans.flashCellService?.flashCell(this);
```
AndrewGlazier: feels like this name should be `cellFlashService`

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -960,9 +960,12 @@ export class RowRenderer extends BeanStub implements NamedBean {
 
     private calculateIndexesToDraw(rowsToRecycle?: { [key: string]: RowCtrl } | null): number[] {
         // all in all indexes in the viewport
-        const indexesToDraw = _createArrayOfNumbers(this.firstRenderedRow, this.lastRenderedRow);
+        const indexesToDraw: number[] = [];
+        for (let i = this.firstRenderedRow; i <= this.lastRenderedRow; i++) {
```
AndrewGlazier, line 964: i know you're maintaining logic, but this func is a bit of a mess doing a lot of redundant stuff - and we call it a lot, would you mind tidying it up?

We don't need to fill `indexesToDraw`, we just need to iterate these indexes later in the func - omitting these and having a "indices before" and "indices after" array filled by `checkRowToDraw` means we can sort these as two distinct arrays and improve perf

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -1150,37 +1159,39 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
+        }
         if (animate) {
             // this ensures we fire displayedRowsChanged AFTER all the 'executeInAWhileFuncs' get
             // executed, as we added it to the end of the list.
             executeInAWhileFuncs.push(() => {
                 this.updateAllRowCtrls();
                 this.dispatchDisplayedRowsChanged();
             });
-            _executeInAWhile(executeInAWhileFuncs);
+            window.setTimeout(() => executeInAWhileFuncs.forEach((func) => func()), 400);
```
AndrewGlazier: Is the number 400 important for something to do with animation?, might be worth checking if we want this as a global variable

packages/ag-grid-enterprise/src/widgets/pillDropZonePanel.ts
```diff
@@ -38,6 +35,18 @@ export interface PillDropZonePanelParams {
...
 
+function _insertArrayIntoArray<T>(dest: T[], src: T[], toIndex: number) {
+    if (dest == null || src == null) {
+        return;
+    }
+
+    // put items in backwards, otherwise inserted items end up in reverse order
+    for (let i = src.length - 1; i >= 0; i--) {
+        const item = src[i];
+        dest.splice(toIndex, 0, item);
```
AndrewGlazier: dest.splice(toIndex, 0, ...src)

packages/ag-grid-community/src/clientSideRowModel/immutableService.ts
AndrewGlazier: I know, but it's for when he has to handle conflicts - good-to-be-nice :)

packages/ag-grid-community/src/rendering/rowRenderer.ts
peterjrreynolds: @gportela85 do you know the answer to this? `_executeInAWhile` had 400 set as the timeout, but the function was only used here.

packages/ag-grid-community/src/rendering/rowRenderer.ts
peterjrreynolds: Not going to do this as part of this ticket.

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
peterjrreynolds: Done

packages/ag-grid-enterprise/src/widgets/pillDropZonePanel.ts
peterjrreynolds: Done

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
peterjrreynolds: Done

packages/ag-grid-community/src/rendering/rowRenderer.ts
AndrewGlazier: Ok - can you make a ticket to do so?

peterjrreynolds: Done

packages/ag-grid-community/src/rendering/rowRenderer.ts
gportela85: It's a fair assumption that this is related to animation since it's inside the `if (animated)` condition.
Ideally here, we need to find out what calls this method with `animate: true`, but I would guess that this is related to the following CSS selectors:

```
// for row animations.
.ag-row-animation .ag-row {
    transition:
        transform 0.4s,
        top 0.4s,
        opacity 0.2s;
}
// for rows older than one second, we also animate the height. we don't include the height
// initially so we are not animating auto-height rows on initial render.
.ag-row-animation .ag-row.ag-after-created {
    transition:
        transform 0.4s,
        top 0.4s,
        height 0.4s,
        opacity 0.2s;
}
```

peterjrreynolds: Have extracted to a const

PR #8986, opened 2024-10-14
AG-13128 Angular min to v17 for AG Grid v33

StephenCooper:
Bump Angular version to 17 as by the time v33 is released it will be outside of LTS.

- Bump TS to 5.2


packages/ag-grid-angular/projects/ag-grid-angular/tsconfig.lib.json
```diff
@@ -6,13 +6,10 @@
     "declarationMap": true,
     "declaration": true,
     "inlineSources": true,
-    "types": [],
-    "lib": ["dom", "es2018"]
+    "types": []
   },
-  "angularCompilerOptions": {
```
StephenCooper, line 12: Removed duplication from higher level

packages/ag-grid-angular/tsconfig.json
```diff
@@ -7,19 +7,27 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "strictPropertyInitialization": false,
```
StephenCooper, line 9: Matched values setup for a new v17 lib where we have not specifically added flags.

PR #8989, opened 2024-10-14
AG-12884 Prevent all row selection operations unless enabled via grid options

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12884

### DIFF
--- a/packages/ag-grid-community/src/validation/errorMessages/errorText.ts
+++ b/packages/ag-grid-community/src/validation/errorMessages/errorText.ts
@@ -309,7 +309,7 @@ export const AG_GRID_ERRORS = {
         `${feature} is only available if using 'clientSide' or 'serverSide' rowModelType, you are using ${rowModel}.` as const,
     130: () => 'cannot multi select unless selection mode is "multiRow"' as const,
     131: () => 'cannot range select while selecting multiple rows' as const,
-    132: () => '' as const,
+    132: () => 'Row selection features are not available unless `rowSelection` is enabled.' as const,
     133: () => 'iconRenderer should return back a string or a dom object' as const,
     134: ({ iconName }: { iconName: string }) => `Did not find icon ${iconName}` as const,
     135: () => `Data type of the new value does not match the cell data type of the column` as const,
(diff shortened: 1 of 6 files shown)


packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -309,7 +309,7 @@ export const AG_GRID_ERRORS = {
         `${feature} is only available if using 'clientSide' or 'serverSide' rowModelType, you are using ${rowModel}.` as const,
     130: () => 'cannot multi select unless selection mode is "multiRow"' as const,
     131: () => 'cannot range select while selecting multiple rows' as const,
-    132: () => `cannot multi select unless selection mode is 'multiRow'` as const,
+    132: () => '' as const,
```
SalvatorePreviti: since we didn't go live yet I think we can reuse this number without adding an empty here
Removing warnings I think makes sense once we first go live once with it

moving 241 here maybe?

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideSelectionService.ts
```diff
@@ -243,11 +255,7 @@ export class ServerSideSelectionService extends BaseSelectionService implements
      *  - after grouping / treeData
      */
     protected override updateSelectable(): void {
-        const { gos } = this;
-
-        const isRowSelecting = _isRowSelection(gos);
-
-        if (!isRowSelecting) {
+        if (!this.enabled) {
```
SalvatorePreviti: ohhh there is a bug I found with the order of events for property changes by multiple services - see https://ag-grid.atlassian.net/browse/AG-13089 

I would just keep reading gos here to avoid this issue, that is also very hard to debug, until that problem is solved

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideSelectionService.ts
SalvatorePreviti: The problem is that enabled might contain an old value when this is executed if multiple properties are changed at once

PR #8990, opened 2024-10-14
AG-4489 Prevent clearing cell selection when row dragging

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-4489

### DIFF
--- a/packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
+++ b/packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
@@ -300,10 +300,22 @@ export class RowDragFeature extends BeanStub implements DropTarget {
     }
 
     private moveRows(rowNodes: RowNode[], pixel: number, increment: number = 0): void {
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -304,7 +304,6 @@ export class RowDragFeature extends BeanStub implements DropTarget {
 
         if (rowWasMoved) {
             this.focusService.clearFocusedCell();
```
gportela85: We probably need to add some logic to refocus the cell that just got moved here.

eliasmalik: Not sure if there's a simpler way to accomplish this, but it seems to behave as we would want: https://github.com/ag-grid/ag-grid/pull/8990/commits/238ebdeccc1b1a4151f52b8f48c69f06a50e40c5

PR #8994, opened 2024-10-14
AG-13078 - Update JS/TS html with appropriate dark mode class

taktran:
Framework examples get dark mode class by injecting dark mode JS, but JS/TS can't use that because dark mode is determined by the class in the html


StephenCooper (approved):
LGTM

PR #8995, opened 2024-10-14
AG-12272 - step 11 - master detail logic rewrite

SalvatorePreviti:
- add the optional transactions to the rowDataUpdated event, this is required to see which rows were updated or added during a transaction to selectively call isRowMaster and the initial expanded state only to updated added rows
- remove completely reference to masterDetail in CSRM and node managers
- add handlers for rowDataUpdated and 'treeData', 'masterDetail' property changes in MasterDetailService (note: this service was added in latest, and just modified it)
- move the get/create of the detail node from the flatten stage to MasterDetailService so it can be tree shaken away


packages/ag-grid-community/src/clientSideRowModel/flattenStage.ts
```diff
@@ -174,6 +175,10 @@ export class FlattenStage extends BeanStub implements IRowNodeStage, NamedBean {
                 const detailNode = this.createDetailNode(rowNode);
                 this.addRowNodeToRowsToDisplay(details, detailNode, result, uiLevel);
             }
+
+            if (rowNode.master === undefined && rowNode.data) {
```
AndrewGlazier: I think this is breaking, setting rowNode.master without an event wasn't done before, this should be using rowNode.setMaster, but also why do we need to do this in the flatten stage? We should be setting this on `createNode` and `update` transactions right? If this is for tree data this logic should live in the tree node manager or master detail service, not here

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public postConstruct(): void {
+        if (_isClientSideRowModel(this.gos)) {
+            this.addManagedPropertyListeners(['treeData', 'masterDetail'], (params) => {
+                const properties = params.changeSet?.properties;
+                if (properties) {
+                    const masterDetail = this.gos.get('masterDetail');
+                    const masterDetailChanged = properties.includes('masterDetail');
+                    let needUpdate = false;
```
AndrewGlazier: This is a bit messy, can we just do
```
const treeDataChanged = masterDetail && properties.includes('treeData')
const masterDetailOrTreeDataChanged = masterDetailChanged || treeDataChanged
```

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+
+                    if (masterDetail && properties.includes('treeData')) {
+                        needUpdate = true; // When treeData and masterDetail are enabled, we need to check if leafs/groups changed
+                    } else if (masterDetailChanged) {
+                        needUpdate = true; // When masterDetail changes, we need to update all rows
+                    }
+
+                    if (needUpdate) {
+                        let changed = false;
+                        (this.rowModel as IClientSideRowModel).forEachLeafNode!((node) => {
```
AndrewGlazier: rather than casting several times here, move the row model cast outside the listeners and cast just once

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+                            if (this.updateMaster(node, masterDetailChanged, true)) {
+                                changed = true;
+                            }
+                        });
+
+                        if (changed) {
+                            // We need to refresh the model to ensure all row components are rerendered
```
AndrewGlazier: better explanation is to ensure all detail nodes are available in the flattened row structure, rendered implies `rowRenderer` should be called, not map stage

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+                rowDataUpdated: ({ transactions }) => {
+                    if (transactions && transactions.length) {
+                        // Row data transactional update
+                        const updatedNodes = new Set<RowNode>();
+                        for (const transaction of transactions) {
+                            for (const rowNode of transaction.update) {
+                                updatedNodes.add(rowNode as RowNode);
+                            }
+                        }
+                        (this.rowModel as IClientSideRowModel).forEachLeafNode!((node) => {
```
AndrewGlazier: if we ignore tree data scenario for now we don't need to do this, as all rows will be leaf children, so we can clean this up and just iterate the transaction nodes

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+            row.expanded = false;
+        }
+
+        if (oldMaster === master) {
+            return false; // No changes
+        }
+
+        row.master = master;
+        if (!newRow) {
+            row.dispatchRowEvent('masterChanged');
```
AndrewGlazier: Why do we not use row.setMaster? There's a fair bit of duplicate code in here doing what's done there

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private updateMaster<TData = any>(row: RowNode<TData>, overrideExpanded: boolean, updated: boolean): boolean {
+        const gos = this.gos;
+        const oldMaster = row.master;
+        const newRow = oldMaster === undefined;
+
+        let master = gos.get('masterDetail');
+
+        const treeGroup = row.group && !!row.treeNode;
```
AndrewGlazier: As discussed, let's simplify and not add tree data master-detail as part of this FR

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+                    // Note that CSRM will invoke refreshModel after this event
+                },
+            });
+        }
+    }
+
+    private updateMaster<TData = any>(row: RowNode<TData>, overrideExpanded: boolean, updated: boolean): boolean {
+        const gos = this.gos;
+        const oldMaster = row.master;
+        const newRow = oldMaster === undefined;
```
AndrewGlazier: this looks like a crutch that I don't think is very clear, we don't have a precedent for relying on undefined as "not-set-up-yet" vars, and as discussed previously we should probably change these in the future to defaults, so using this here blocks us from setting that pattern later

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+                            for (const rowNode of transaction.update) {
+                                updatedNodes.add(rowNode as RowNode);
+                            }
+                        }
+                        (this.rowModel as IClientSideRowModel).forEachLeafNode!((node) => {
+                            this.updateMaster(node, false, updatedNodes.has(node));
+                        });
+                    } else {
+                        // New row data
+                        (this.rowModel as IClientSideRowModel).forEachLeafNode!((node) => {
```
AndrewGlazier: I might be missing something, but looks like we now update on update transaction and row data change, but not on add transaction?

Probably safer to move this side of the statement back into ref from the node manager when it creates nodes

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+                    // if we are doing master detail, then the default is that everything can be a Master Row.
+                    const isRowMasterFunc = gos.get('isRowMaster');
+                    master = !isRowMasterFunc || !!isRowMasterFunc(row.data);
+                }
+            } else {
+                master = oldMaster;
+            }
+        }
+
+        if ((overrideExpanded || newRow) && !treeGroup) {
```
AndrewGlazier: Should master-detail be re-expanding rows? I think historically changes in master-detail haven't done this

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    private rowModel: IRowModel;
+    private funcColsService: FuncColsService;
+
+    public wireBeans(beans: BeanCollection): void {
+        this.funcColsService = beans.funcColsService;
+        this.rowModel = beans.rowModel;
+    }
+
+    public postConstruct(): void {
```
AndrewGlazier: postConstruct is normally kept quite tidy, can we move the listener logic for these into funcs

packages/ag-grid-community/src/clientSideRowModel/flattenStage.ts
SalvatorePreviti: I am using here master === undefined to indicate this is a new row and master was neither set before to true or false. This means the row was not processed before but we just set it to false if it was undefined during the flattening stage to be as similar as possible to the previous implementation. We were not raising the event for a new row with this value to false so that behaviour is the same, we do not raise event if we switch from undefined to true or false. However we can see if there is a better way

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: sure - but as we said this logic will now change as we are disabling tree + masterDetail for now, will simplify it

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: We removed row.setMaster in this pr: https://ag-grid.atlassian.net/browse/AG-12975 and we agreed to move the logic outside of the rowNode into the masterdetail module so there is no repetition and this code will be tree shaken for who does not use masterdetail

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: it is a normal way to use undefined or null to do lazy loading or deferred computation, in many javascript projects.
Also if we did not use it before much I am sure this is gonna be used more in the future and I don't see a problem with it.
Besides the break in traditions, does this poses another issue?

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: We do not need to check for add transaction as we check for the master to be undefined to see that we need to recompute the row master flag. If master is undefined we need to set it either to true or false, and this happens for new rows

packages/ag-grid-community/src/clientSideRowModel/flattenStage.ts
SalvatorePreviti: this piece of code makes sense when MasterDetail module is not loaded and we just want to set the flag to false at least for the flattened rows to maintain the previous weird behaviour where this flag was sometime undefined and sometime false, and this follows that logic.
When MasterDetail module is loaded this flag will always be true or false

If it is fine with you we can just leave it undefined also if is a slight change in behaviour. It would be simpler and more correct. Would be much better to change the IRowNode interface to specify that master can be true false or undefined (if the MasterDetail module is not loaded).
When the MasterDetail module is loaded, this value will always be true or false

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: I think this is following the existing behaviour and when masterDetail changes this was happening.
I can check again maybe I misunderstood or I wrote wrongly this logic here

SalvatorePreviti: Oh yes, it was happening as when changing masterDetail we were rebuilding the whole row nodes from rowData, so every row was new, so every expanded state was recomputed fresh. It was on the list of reset properties in CSRM property change listener

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: That's not so simple. Transactions may contain rows updated, added and then deleted and we do not want to touch those, we want to see what's the final result and just know which rows were updated skipping the deleted ones
I think we still need to iterate all rows

Think about a transaction [ 1: { add 1}, 2: {update 2}, 3: {delete 1, delete 2}]

packages/ag-grid-community/src/clientSideRowModel/flattenStage.ts
AndrewGlazier: I'm concerned because the event _was_ fired before, see:
``` 
       if (isTreeData) {
            rowNode.setMaster(false);
            if (setExpanded) {
                rowNode.expanded = false;
            }
        } else {
```
If you want to make this change to not fire events anymore, but you need to do due dilligence and ensure it doesn't break anything else in the grid

When we set master previously the event was _always_ fired (by rowNode.setMaster) as far as I can see?

I know why you're using undefined here, but as my message says it's about blocking the future precedent where we can set defaults, if you want to do this you won't be able to also set defaults on the rownode in future, because of inconsistent patterns - I'd rather we didn't use undefined to mean unconfigured as it's also not clear to someone who is unfamiliar with the code base - I also think it's unnecessary as the flow of calling setMaster in the node manager when a row was updated/added was fine, and simplifies the logic flow a lot more than trying to lean on implicit variable meanings

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
AndrewGlazier: This is such a minor performance improvement - it's actually more of a detriment - why do we need to skip nodes which may have been added + deleted?

This scenario is handling what is really an edge case at the detriment to the normal case, while adding unnecessary complexity

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
AndrewGlazier: It's inconsistent - it's not about tradition

Some properties rely on undefined values (rowNode.selected)
You've stated an interest in setting default rowNode values
This is a third inconsistency, and seems unnecessary when we should simplify the flow anyway and just set this on new row nodes

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
AndrewGlazier: I misunderstood this - but it does still seem excessive to me that we aren't just calling this through from create/update node like before - the logic here is much harder to follow

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: To be sure to not invoke user provided isRowMaster on deleted nodes - to avoid existing codebases eventually break as this is the current behaviour but let me think around it

packages/ag-grid-community/src/clientSideRowModel/flattenStage.ts
SalvatorePreviti: so: if the row is new and was never used before there is nothing that register to this event, not us nor the user. The only place where masterChanged is registered for is in the groupCellRenderer that happens way after we set the master field for the first time. But, I can simplify it and just always raise it

SalvatorePreviti: Changed it to always been raised

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: changed and simplified the logic, masterDetail will be considered disabled if treeData is active

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: simplified the code

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: Changed as you discussed and now we go for all updated and added nodes - however - is ok to call it on deleted rows in this edge case?

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: Forced to always be initialized to false in the rownode, and fixed the logic so it works with master detail.
Verified that nothing else sets master or depends on master being undefined

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: changed the implementation and the logic so this should be outdated

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: Simplified the logic but still maintaining the previous behaviour of reloading the expanded state and the master by calling the user provided function isRowMaster if there is a change in masterDetail+treeData flags

packages/ag-grid-enterprise/src/masterDetail/clientSideDetailService.ts
SalvatorePreviti: done

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -1,15 +1,163 @@
...
+        const setMaster = (row: RowNode, created: boolean, updated: boolean) => {
+            const oldMaster = row.master;
+
+            let newMaster = enabled;
+
+            if (enabled) {
+                if (created || updated) {
+                    if (isRowMaster) {
+                        const data = row.data;
+                        newMaster = !!data && !!isRowMaster(data);
```
AndrewGlazier, line 83: PR looks great, only I'm not sure this can be used to determine if the row is new - new rows have data before 'rowDataUpdated' is set I think?

PR #9001, opened 2024-10-15
AG-13164 Fix getBestCostNodeSelection

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-13164

### DIFF
--- a/packages/ag-grid-community/src/selection/selectionService.ts
+++ b/packages/ag-grid-community/src/selection/selectionService.ts
@@ -425,9 +425,8 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
                 } else {
                     // if not selected, then if it's a group, and the group
                     // has children, continue to search for selections
-                    const maybeGroup = node as any;
-                    if (maybeGroup.group && maybeGroup.children) {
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModelApi.ts
```diff
@@ -69,7 +69,3 @@ export function applyTransactionAsync<TData = any>(
 export function flushAsyncTransactions(beans: BeanCollection): void {
     beans.frameworkOverrides.wrapIncoming(() => _getClientSideRowModel(beans)?.flushAsyncTransactions());
 }
-
-export function getBestCostNodeSelection<TData = any>(beans: BeanCollection): IRowNode<TData>[] | undefined {
```
eliasmalik: @peterjrreynolds any issues with moving this into the row selection module? Seems a better fit?

eliasmalik: reverted, method only makes sense in CSRM, is stubbed out in serverSideSelectionService

PR #9012, opened 2024-10-16
AG-11845 Update property coercions

StephenCooper:
- Stopped property coercion of grid options props.
  - Move boolean coercion into Angular / Vue code  
- Remove PropertyKeys static class, expose raw arrays
- Remove ComponentUtil static class, expose raw array / methods
- Move initial prop check just to validations.


packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
```diff
@@ -245,20 +242,32 @@ export class AgGridAngular<TData = any, TColDef extends ColDef<TData> = ColDef<a
...
+                this._viewContainerRef,
+                this._angularFrameworkOverrides
+            );
+
+            const excludeAngularCompProps = [
+                'gridOptions',
+                'modules',
+                ...Object.entries(this)
+                    .filter(([key, value]) => key.startsWith('_') || value instanceof EventEmitter)
+                    .map(([key]) => key),
```
peterjrreynolds: Can just loop the entries and then push into the map rather than filtering then mapping then spreading.

PR #9014, opened 2024-10-16
Update fa-IR.ts

amir1704:
export in this situation does not mean صادرات. it is best to use خروجی


sina-ss:
I'm Iranian too and I can say 'خروجی' is much better

PR #9018, opened 2024-10-16
AG-12272 hierarchical tree data initial support

SalvatorePreviti:
Merge to latest the hierarchical tree data initial support.

Todos after merged to latest for tree data with children (hierarchical):
- Immutable rowData is still not implemented/not working.
- decide what to do with transactions for hierarchical tree data and/or how to disable them

Breaking changes in this PR:
- TreeDataModule is now required for tree data, it was separated from RowGroupingModule 
- Tree data structure is built by the client side node manager, so, it happens before grouping, and calling the api refreshClientSideModel with "group" or "everything" will not call again getDataPath or process the children. To reload the tree setRowData must be called

Internal changes:
- Removed the ImmutableService, now its responsibility is moved to the ClientSideNodeManager, and triggered via the ClientSideRowModel
- Moved all the tree data related things into the TreeDataModule
- Moved GroupStrategy to GroupStage, there is no grouping stage with tree data anymore
- Added a new "transactions" property to the rowDataUpdated event
- Moved all the masterDetail logic and event handling outside of CSRM into MasterDetail module
- There are multiple ClientSideNodeManagers now, all singleton beans that gets selected by ClientSideRowModel depending on the grid options


packages/ag-grid-community/src/interfaces/iRowModel.ts
```diff
@@ -9,6 +10,11 @@ export interface RowBounds {
 export type RowModelType = 'infinite' | 'viewport' | 'clientSide' | 'serverSide';
 
 export interface IRowModel {
+    readonly rootNode?: RowNode;
+
+    /** Available only in ClientSideRowModel. */
+    refreshModel?(paramsOrStep: RefreshModelParams | ClientSideRowModelStep | undefined): void;
```
peterjrreynolds: I'm not keen on putting stuff like this in the shared interface. `IRowModel` shouldn't have any dependencies on CSRM stuff. Can't the caller just check the row model type?

Can use the helper function `_getClientSideRowModel`

PR #9025, opened 2024-10-17
AG-13189 - fix memory leaks and fix csrm deallocation

SalvatorePreviti:
Fix setting CSRM fields to null on destruction breaking other services.

- rootNode and treeRoot are now nullable to be sure we do not access them after the bean is destroyed
- _debounce now requires a bean or component so we do not risk to execute beans code after destroyed
- fix some setTimeout in the application not checking for isAlive
- exposing the new readonly _EmptyArray that was previously used only in trees and that can now be used to clear arrays without using extra memory and make them readonly (it will throw if someone tries to write on it)


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -734,7 +743,11 @@ export class StateService extends BeanStub implements NamedBean {
     }
 
     private getRowGroupExpansionState(): RowGroupExpansionState | undefined {
-        if (!this.expansionService) {
+        if (
+            !this.expansionService ||
+            // Check if alive is required as this method might be debounced
+            !this.isAlive()
```
peterjrreynolds: The is alive check should live in the async caller rather than in the sync method here

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -620,10 +633,6 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
     }
 
     public getRowBounds(index: number): RowBounds | null {
-        if (_missing(this.rowsToDisplay)) {
```
StephenCooper, line 623: Why don't we need this anymore?

StephenCooper: I see it is always defined as an array now

packages/ag-grid-community/src/gridBodyComp/fakeHScrollComp.ts
```diff
@@ -143,9 +149,7 @@ export class FakeHScrollComp extends AbstractFakeScrollComp {
         if (!hScrollShowing) {
             apply();
         } else {
-            this.setScrollVisibleDebounce = window.setTimeout(() => {
-                apply();
-            }, 100);
+            this.setScrollVisibleDebounce = window.setTimeout(apply, 100);
```
StephenCooper, line 152: Does this not need `bind(this)`?

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -232,10 +251,8 @@ export class StateService extends BeanStub implements NamedBean {
             columnRowGroupChanged: updateRowGroupExpansionState,
             rowDataUpdated: () => {
                 if (this.gos.get('groupDefaultExpanded') !== 0) {
-                    setTimeout(() => {
-                        // once rows are loaded, they may be expanded
-                        updateRowGroupExpansionState();
-                    });
+                    // once rows are loaded, they may be expanded, start the timer only once
+                    this.updateRowGroupExpansionStateTimer ||= setTimeout(updateRowGroupExpansionState);
```
StephenCooper, line 257: Do we need to bind this?

packages/ag-grid-community/src/gridBodyComp/fakeHScrollComp.ts
SalvatorePreviti: no because apply is defined as a arrow function, and arrow functions sets this and that this cannot be changed - arrow functions do an implicit binding, this will be the same as the current this the moment they get created

packages/ag-grid-community/src/misc/state/stateService.ts
SalvatorePreviti: no because updateRowGroupExpansionState is defined as a arrow function, and arrow functions sets this and that this cannot be overwritten - arrow functions do an implicit binding

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Yeah nothing is setting it to null or undefined anymore - and is initialized on construction - i think is safe to not check for it

PR #9026, opened 2024-10-17
AG-12847 - extract logic out of row node

peterjrreynolds:


packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -371,71 +362,25 @@ export class RowNode<TData = any> implements IEventEmitter<RowNodeEventType>, IR
...
+        const { selectionService } = this.beans;
+        const oldNode = selectionService?.createDaemonNode?.(this);
         const oldData = this.data;
 
         this.data = data;
         this.updateDataOnDetailNode();
         this.setId(id);
-        this.checkRowSelectable();
-        this.beans.selectionService?.syncInRowNode(this, oldNode);
+        if (selectionService) {
```
StephenCooper, line 375: Don't need the if block

StephenCooper (approved):

packages/ag-grid-community/src/entities/rowNode.ts
peterjrreynolds: If block means we only need to check existence once rather than twice

PR #9034, opened 2024-10-18
AG-13199 fix logic if tree data module is not loaded

SalvatorePreviti:
- Fix the logic of the CSRM to handle the case treeData is true but no TreeDataModule is present. Note that this is a temporary fix that I'd like to add because it shows what need to be changed in the next PR where I am working out on simplifying, make smaller and cleaning up the updateRowData+refresh flow to support immutable tree data with children updates
- Add a new test around it
- Improves some test code


packages/ag-grid-enterprise/src/main-umd-ce-noStyles.ts
```diff
@@ -27,6 +27,7 @@ import {
     SideBarModule,
     SparklinesModule,
     StatusBarModule,
+    TreeDataModule,
```
peterjrreynolds: This isn't needed. For backwards compatibility, `RowGroupingModule` bundles `TreeDataModule`. There's a separate row grouping module that doesn't include tree data.

StephenCooper (approved):

packages/ag-grid-enterprise/src/main-umd-ce-noStyles.ts
SalvatorePreviti: ohhh removing it then now

PR #9044, opened 2024-10-21
RTI-2366 - Fix framework selector in docs

taktran:
Broke on license install page, because it wasn't passing in the menu items. Not required there since the page is generated for all frameworks

---

Cherry picked from https://github.com/ag-grid/ag-grid/commit/99276c4993500d6b9bef709627fac27c01ece821


markdurrant (approved):
LGTM 👍

PR #9051, opened 2024-10-21
RTI-2371 Fix reactivity of header checkbox

eliasmalik:
See https://ag-grid.atlassian.net/browse/RTI-2371

### DIFF
--- a/community-modules/core/src/headerRendering/cells/column/selectAllFeature.ts
+++ b/community-modules/core/src/headerRendering/cells/column/selectAllFeature.ts
@@ -2,7 +2,6 @@ import { isColumnControlsCol } from '../../../columns/columnUtils';
 import { BeanStub } from '../../../context/beanStub';
 import type { BeanCollection } from '../../../context/context';
 import type { AgColumn } from '../../../entities/agColumn';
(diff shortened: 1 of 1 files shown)


community-modules/core/src/headerRendering/cells/column/selectAllFeature.ts
```diff
@@ -40,17 +38,6 @@ export class SelectAllFeature extends BeanStub {
...
 
-    public postConstruct() {
-        const setRowSelectionOptions = (rowSelection: GridOptions['rowSelection']) => {
-            if (rowSelection && typeof rowSelection !== 'string') {
-                this.selectionOptions = rowSelection;
-            }
-        };
-
-        setRowSelectionOptions(this.gos.get('rowSelection'));
-        this.addManagedPropertyListener('rowSelection', (e) => setRowSelectionOptions(e.currentValue));
```
eliasmalik, line 51: This only gets called after `isCheckboxSelection` which is part of the codepath that gets run when `setComp` is called. So just grabbing `rowSelection` from the `gos` directly instead to simplify (for now).

Issue #200, opened 2015-06-09
suppress third click to sort, so doesn't go back into 'not sorted'


ceolter:
http://www.angulargrid.com/forum/showthread.php?tid=2295

PR #9104, opened 2024-10-22
AG-12847 - module validations

peterjrreynolds:


packages/ag-grid-community/src/validation/validationService.ts
```diff
@@ -62,6 +65,28 @@ export class ValidationService extends BeanStub implements NamedBean {
...
+        propertyName: string,
+        componentName: string,
+        agGridDefaults: { [key in UserComponentName]?: any },
+        jsComps: { [key: string]: any }
+    ): void {
+        const moduleForComponent = USER_COMP_MODULES[componentName as UserComponentName];
+        if (moduleForComponent) {
+            this.gos.assertModuleRegistered(
+                moduleForComponent,
+                `AG Grid '${propertyName}' component: ${componentName}`
```
StephenCooper, line 78: Can you check if this results in `AG Grid AG Grid`?

StephenCooper (approved):

packages/ag-grid-community/src/validation/validationService.ts
peterjrreynolds: No, the message looks like this `AG Grid: error #200 Unable to use AG Grid 'filter' component: agMultiColumnFilter as MultiFilterModule is not registered...`

PR #9111, opened 2024-10-22
AG-12847 - add beans to bean stub

peterjrreynolds:


packages/ag-grid-community/src/gridOptionsService.ts
```diff
@@ -108,7 +108,7 @@ export class GridOptionsService extends BeanStub implements NamedBean {
         this.eventService.addGlobalListener(this.globalEventHandlerFactory(true).bind(this), false);
 
         // Ensure the propertyEventService has framework overrides set so that it can fire events outside of angular
-        this.propertyEventService.setFrameworkOverrides(this.frameworkOverrides);
+        this.propertyEventService.setFrameworkOverrides(this.beans.frameworkOverrides);
```
StephenCooper, line 111: Should this be this.getFrameworkOverrides() or should that be removed from beanStub?

peterjrreynolds: `getFrameworkOverrides()` is actually no longer needed as everything is using `this.beans` (e.g. `RowCtrl`). Have removed

PR #9117, opened 2024-10-23
AG-13233 - Fix Integrated Charts in shadow-DOM.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13233


enterprise-modules/charts/src/charts/chartComp/chartProxies/chartProxy.ts
```diff
@@ -194,6 +194,7 @@ export abstract class ChartProxy<
         } as const;
         newOptions.theme = theme;
         newOptions.container = this.chartProxyParams.parentElement;
+        (newOptions as any).styleContainer = this.chartProxyParams.parentElement;
```
alantreadway, line 197: This is the actual change, everything else is noise from the Charts version bump.

PR #9128, opened 2024-10-25
AG-13239 modularise Theming API CSS for tree shaking

BernieSumption:
1. Allow CSS to be attached to modules (it is injected by the `_injectCoreAndModuleCSS` function)
2. Allow CSS to be attached to components (it is registered via `_registerComponentCSS` then when it's clear that the theming API is being used, injected with `_injectGlobalCSS`)
3. Update the build to process .css files in both community and enterprise
4. Move lots of CSS files from the Theming API folder to the appropriate module


packages/ag-grid-community/src/widgets/component.ts
```diff
@@ -119,7 +122,10 @@ export class Component<TLocalEvent extends string = ComponentEvent>
                     // 3. The property is on a child component and not available on the parent during construction.
                     //    In which case you may need to pass the template via setTemplate() instead of in the super constructor.
                     // 4. The data-ref is not used by the component and should be removed from the template.
-                    throw new Error(`Issue with data-ref: ${elementRef} on ${this.constructor.name} with ${current}`);
+                    console.error('current:', current);
```
peterjrreynolds: Remove?

peterjrreynolds (approved):

PR #9147, opened 2024-10-29
AG-12272 csrm immutable row data refactor

SalvatorePreviti:
To ensure the possibility to give to the node manager the ability to set immutable data without using transactions, as required by the new tree data with children node manager, we need to change the flow that CSRM uses to do refresh.

- remove the property `keepEditingRows` from RefreshModelParams as it was set but never read
- add more properties to RefreshModelParams so we can move some of the processing logic in different methods and in the client side node manager
- make setImmutableRowData able to update the RefreshModelParams if it has transactions to execute, and move some execution in the refresh method
- Split the parsing of refresh model step and the refreshModel method in two different methods (refreshModelApi and refreshModel)
- Simplify and linearize the logic to update rowData and refresh
- Added a new private event beforeRefreshModel, currently used only by MasterDetail. This simplifies the logic of master detail service

### DIFF
--- a/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
+++ b/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
@@ -19,7 +19,6 @@ import {
 import type { IClientSideNodeManager } from '../interfaces/iClientSideNodeManager';
 import type {
     ClientSideRowModelStage,
-    ClientSideRowModelStep,
     IClientSideRowModel,
     RefreshModelParams,
 } from '../interfaces/iClientSideRowModel';
@@ -737,17 +736,6 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         // return true if we are only doing update transactions
         const transactionsContainUpdatesOnly = !transWithAddsOrDeletes;
         return transactionsContainUpdatesOnly;
-    }
-
-    public refreshModelApi(step: ClientSideRowModelStep | undefined): void {
-        if (!step || step === 'everything') {
-            step = 'group';
-        }
-        this.refreshModel({
-            step,
-            keepRenderedRows: true,
-            animate: !this.gos.get('suppressAnimationFrame'),
-        });
     }
 
     public refreshModel(params: RefreshModelParams): void {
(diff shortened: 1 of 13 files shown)


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -718,34 +738,20 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         return transactionsContainUpdatesOnly;
     }
 
-    private buildRefreshModelParams(step: ClientSideRowModelStage = 'group'): RefreshModelParams | undefined {
-        return {
+    public refreshModelApi(step: ClientSideRowModelStep | undefined): void {
```
peterjrreynolds: If this is only used in the API function, we can remove this method and move all the code into the API function so that it can be tree shaken

packages/ag-grid-enterprise/src/treeData/clientSidePathTreeNodeManager.ts
```diff
@@ -24,34 +24,33 @@ export class ClientSidePathTreeNodeManager<TData>
...
 
-    public commitTransactions(
-        transactions: RowNodeTransaction<TData>[],
-        changedPath: ChangedPath | undefined,
-        rowNodesOrderChanged: boolean
-    ): void {
-        this.treeRoot?.setRow(this.rootNode);
+    public refreshModel(params: RefreshModelParams<TData>): void {
+        const transactions = params.rowNodeTransactions;
+        if (transactions?.length) {
```
peterjrreynolds, line 29: Would be neater if this was an early return

packages/ag-grid-enterprise/src/treeData/abstractClientSideTreeNodeManager.ts
```diff
@@ -546,4 +517,33 @@ export abstract class AbstractClientSideTreeNodeManager<TData> extends AbstractC
             }
         }
     }
+
+    public refreshModel(params: RefreshModelParams<TData>): void {
+        if (params.afterColumnsChanged) {
```
peterjrreynolds: Could be an early return

peterjrreynolds (approved):

PR #9155, opened 2024-10-30
AG-12272 - support immutable row data with hierarchical tree data

SalvatorePreviti:
Implements ClientSideChildrenTreeNodeManager.setImmutableRowData to properly handle immutable data.

The deviation from the other node managers is that this does not use transactions.

Also, improves the previous implementation of setImmutableRowData by skipping any execution if there is no change

### DIFF
--- a/packages/ag-grid-enterprise/src/treeData/treeNode.ts
+++ b/packages/ag-grid-enterprise/src/treeData/treeNode.ts
@@ -123,9 +123,11 @@ export class TreeNode implements ITreeNode {
     /**
      * Gets a node a key in the given parent. If the node does not exists, creates a filler node, with null row.
      * We cast to string just to be sure the user passed a string correctly and not a number or something else.
+     * @param key - The key of the node to get.
+     * @param append - If true, the node will be moved to the end of the children list.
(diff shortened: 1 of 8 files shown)


packages/ag-grid-enterprise/src/treeData/treeNode.ts
```diff
@@ -123,16 +123,22 @@ export class TreeNode implements ITreeNode {
...
+    public upsertKey(key: string | number, append: boolean = false): TreeNode {
         if (typeof key !== 'string') {
             key = String(key);
         }
         let node = this.children?.get(key);
         if (!node) {
             node = new TreeNode(this, key, this.level + 1);
             (this.children ??= new Map())?.set(node.key, node); // Add to the map
+        } else if (append) {
+            const children = this.children!;
```
peterjrreynolds, line 139: Could move this up

peterjrreynolds (approved):

packages/ag-grid-enterprise/src/treeData/treeNode.ts
SalvatorePreviti: I refactored in the next PR, by declaring two different methods

PR #9174, opened 2024-11-04
fix: provide using prototype chain in vue3, fix #9173

### LINKED ISSUE #9173: When using ag-grid-vue3, I can't use different provides in different components, it will always take the first component
```
```

Current behaviour.
https://github.com/hcg1023/ag-grid-vue-bug

**Current behavior**
When using ag-grid-vue3, I can't use different provides in different components, it will always take the first component

**Expected behavior**
Correct use of provide

**Please tell us about your environment:all
* **AG Grid version:32.3.2

* **Browser:all

* **Language:all

hcg1023:
With this PR, it will be closer to the provides implementation of vue itself, fixing #9173

### DIFF
--- a/packages/ag-grid-vue3/src/AgGridVue.ts
+++ b/packages/ag-grid-vue3/src/AgGridVue.ts
@@ -140,17 +140,7 @@ export const AgGridVue = defineComponent({
         },
         getProvides() {
             let instance = getCurrentInstance() as any;
-            let provides = {};
-
-            while (instance) {
-                if (instance && instance.provides) {
-                    provides = { ...provides, ...instance.provides };
-                }
-
-                instance = instance.parent;
-            }
-
-            return provides;
+            return Object.create(instance.provides);
         },
         /*
          * Prevents an infinite loop when using v-model for the rowData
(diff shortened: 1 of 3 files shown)


seanlandsman:
Thanks @hcg1023 - we're in the process of a vue 3 rewrite so I will ensure this PR is incorporated into the next version.

hcg1023:
Awesome, I found something else. If I use v-model, `push()`, `splice()` will not cause the grid to redraw. If I change the reference of the array, it will be correct. Hope the new version can fix it.

hcg1023:
I would like to know when the new version will be availabl?

seanlandsman:
We'll be releasing a new version at the end of this month

hcg1023:
I added a new commit to fix the problem that the vue script setup syntax sugar has inconsistent effects in dev and prod. In dev, we can get setupState, but in prod, we cannot get setupState. Users should always be advised to expose functions such as getValue.

@seanlandsman

hcg1023:
https://github.com/vuejs/core/blob/main/packages/runtime-core/src/componentPublicInstance.ts#L691

setupState dev only in here

seanlandsman:
Awesome!

seanlandsman:
@hcg1023 Thank you very much for this great contribution - I've taken your changes on board with some slight changes due to a larger refactoring and rewriting, but they're in there!

Your changes have been enormously helpful - thanks again!

PR #9182, opened 2024-11-04
AG-12847 - module metadata plus tweaks

peterjrreynolds:


StephenCooper:
I like the pattern but I am worried about the manual nature of these comments. It seems like the `@gridOptions` and `@colDef` are going to be the same information that we have as part of the validations module. I suspect we might be able to do something automated off that going forward but that shouldn't block this.

PR #9183, opened 2024-11-04
AG-12915 - Truncate log url if parameters are too long

taktran:
Update tsconfig.base.json to fix typescript error for `params.entries()`

### DIFF
--- a/packages/ag-grid-community/src/validation/logging.ts
+++ b/packages/ag-grid-community/src/validation/logging.ts
@@ -4,6 +4,10 @@ import { VERSION } from '../version';
 import type { ErrorId, ErrorMap, GetErrorParams } from './errorMessages/errorText';
 import type { ValidationService } from './validationService';
 
+const MAX_URL_LENGTH = 2000;
+const MIN_PARAM_LENGTH = 100;
+const VERSION_PARAM_NAME = '_version_';
+
 let validation: ValidationService | null = null;
 let suppressAllLogging = false;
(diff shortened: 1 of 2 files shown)


tsconfig.base.json
```diff
@@ -11,7 +11,7 @@
     "esModuleInterop": true,
     "target": "ES2020",
     "module": "commonjs",
-    "lib": ["ES2022", "DOM"],
+    "lib": ["ES2022", "DOM", "dom.iterable"],
```
StephenCooper, line 14: Do we really need to make this change?

packages/ag-grid-community/src/validation/logging.ts
```diff
@@ -65,15 +69,53 @@ export function toStringWithNullUndefined(str: string | null | undefined) {
     return str === undefined ? 'undefined' : str === null ? 'null' : str;
 }
 
+function getParamsUrl(baseUrl: string, params: URLSearchParams) {
+    return `${baseUrl}?${params.toString()}`;
+}
+
+function truncateUrl(baseUrl: string, params: URLSearchParams, maxLength: number) {
```
StephenCooper, line 76: Good idea to only shorten the longest of the params.

StephenCooper (approved):

tsconfig.base.json
taktran: Without it, there are VSCode errors for `.entries`

Screenshot 2024-11-04 at 4 34 44 pm

PR #9188, opened 2024-11-05
AG-13280 - use AgCharts sparklines

StephM-W:


packages/ag-grid-enterprise/src/sparkline/sparklineCellRenderer.ts
```diff
@@ -1,20 +1,13 @@
-import type { BeanCollection, ICellRenderer, ISparklineCellRendererParams } from 'ag-grid-community';
-import { Component, RefPlaceholder, _observeResize } from 'ag-grid-community';
+import { AgCharts, _Util } from 'ag-charts-community';
+import type { AgChartInstance, AgSparklineOptions } from 'ag-charts-community';
```
StephenCooper: Update these to come from `ag-charts-types`

StephM-W: Updated

PR #9203, opened 2024-11-07
AG-12272 changed row nodes, delta sorting and master detail service without transactions

SalvatorePreviti:
- Add a new class ChangedRowNodes, and an interface IChangedRowNodes that holds information of what changed during a transactional update or a set immutable row data in a better way
- Changes CSRM transactions and immutable data handling to write in this new data structure
- Change the delta sorting algorithm to read this new data structure
- Change the master detail service to read this new data structure
- Change the clientSidePathTreeNodeManager to read this new data structure

This is required because the new client side node manager with children does NOT generate transactions for set immutable data, instead, it directly rebuild the structure correctly and now uses this class to notify of what's changed

### DIFF
--- a/packages/ag-grid-community/src/interfaces/iRowNodeStage.ts
+++ b/packages/ag-grid-community/src/interfaces/iRowNodeStage.ts
@@ -1,15 +1,18 @@
 import type { GridOptions } from '../entities/gridOptions';
 import type { RowNode } from '../entities/rowNode';
 import type { ChangedPath } from '../utils/changedPath';
-import type { ClientSideRowModelStage } from './iClientSideRowModel';
+import type { ClientSideRowModelStage, IChangedRowNodes } from './iClientSideRowModel';
 import type { RowNodeTransaction } from './rowNodeTransaction';
 
 export interface StageExecuteParams<TData = any> {
     rowNode: RowNode<TData>;
-    // used in group stage, as group stage does one transaction at a time
-    rowNodeTransaction?: RowNodeTransaction<TData> | null;
+
     // used in sort stage, as sort stage looks at all transactions in one go
+    changedRowNodes?: IChangedRowNodes<TData>;
+
+    // used in group stage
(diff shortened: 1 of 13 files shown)


packages/ag-grid-community/src/interfaces/iRowNodeStage.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export interface StageExecuteParams<TData = any> {
     rowNode: RowNode<TData>;
-    // used in group stage, as group stage does one transaction at a time
-    rowNodeTransaction?: RowNodeTransaction<TData> | null;
+
     // used in sort stage, as sort stage looks at all transactions in one go
+    changedRowNodes?: IChangedRowNodes<TData>;
+
+    // used in group stage
```
peterjrreynolds, line 13: Would be nice to switch these to JSDoc

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
```diff
@@ -96,6 +105,8 @@ export class ClientSideChildrenTreeNodeManager<TData>
         const changedPath = new ChangedPath(false, rootNode);
         params.changedPath = changedPath;
 
+        const changedRowNodes = params.changedRowNodes!;
```
peterjrreynolds, line 99: Is the `!` needed here?

SalvatorePreviti: unfortunately yes ... but alternatively I can pass it as parameter to be sure is not null

SalvatorePreviti: I will however plan to do other passes of refactoring in the near future, so if is fine, we can merge it for now as it is and we review later

PR #9209, opened 2024-11-08
AG-11763 - module validation improvements

peterjrreynolds:
- Fix duplicate validations.
- Add specific menu item validations. Display correctly for user-provided lists versus grid-provided.
- Add a type for menu items (AG-13374).
- Optimise sidebar validations.
- Optimise menu code.


packages/ag-grid-enterprise/src/menu/chartMenuItemMapper.ts
```diff
@@ -101,7 +101,7 @@ export class ChartMenuItemMapper extends BeanStub implements NamedBean {
 
             if (chartConfigGroup == undefined) {
                 _warn(173, { group });
-                return undefined;
+                return;
```
StephenCooper, line 104: Is this meant to be `null` as well?

peterjrreynolds: This is inside a `forEach`

PR #9230, opened 2024-11-14
AG-12850 Theme part installation performance refactor

BernieSumption:
This change improves the performance of theme installation by 100x or so

### Summary of changes

Previously, a Theming API theme compiled like so:

1. Generate a big blob of CSS combining all active parts (e.g. icon sets, input styles) and default parameter value
2. Generate a unique theme class name e.g. `ag-theme-1` and wrap the generated CSS in `.ag-theme-1 { ... generated css ...}`
3. Inject the wrapped CSS into the document head
4. Apply the `ag-theme-1` class to the wrapper of any grid using the theme
5. If all grids on the page stop using the theme, remove the generated CSS from the head to prevent memory leaks when lots of themes are in use

This PR changes it to work as follows:

1. For each part (e.g. icon sets, input styles) generate a unique class name e.g. `ag-theme-part-1` and wrap the part's CSS in `.ag-theme-part-1 { ... part css ...}`.
2. The first time a part is used, insert its wrapped CSS into the document head. Never remove it.
3. When a theme is used by a grid, get the classes for all the active parts and set them all on the grid wrapper element. For example if a theme uses 3 parts, the grid's wrapper might have `class="ag-theme-part-1 ag-theme-part-2 ag-theme-part-3"`
4. For theme parameters, compile them as CSS and insert them into a `<style>` element in the grid DOM itself, so that it will be removed automatically when the grid is destroyed. This prevents memory leaks when lots of themes are in use without the need to track how many grids are using a theme.

The net effect is that when rapidly regenerating themes, e.g. when dragging a color picker slider in the theme builder (which generates a new theme for each color change), much less work is done. Dragging a color slider in the theme builder in 32.3 was noticeably very laggy. Now no lag is noticeable.


documentation/ag-grid-docs/src/components/theme-builder/components/component-utils.ts
```diff
@@ -1,5 +1,8 @@
 import { type ReactNode, memo, useEffect, useRef } from 'react';
 
+import type { Theme } from 'ag-grid-community';
+import { _themeBuilder as themes } from 'ag-grid-community';
```
gportela85: It looks like we always want this to be `themes`, so why not export `themes` instead of `_themeBuilder`.
If this is will never be imported by the end-user, then maybe export `_themes`.

documentation/ag-grid-docs/src/content/docs/theming-distribution/_examples/creating-parts/main.ts
```diff
@@ -3,29 +3,38 @@ import {
...
         .ag-checkbox-input-wrapper {
             border-radius: 4px;
-            /* Parts' CSS can use new parameters - define support
-               for them using withAdditionalParams below */
-            box-shadow: 0 0 5px 4px var(--ag-checkbox-glow-color);
+            box-shadow: 0 0 5px 4px red;
             width: 16px;
             height: 16px;
-
+
```
gportela85, line 35: empty space here

documentation/ag-grid-docs/src/content/docs/theming-distribution/_examples/creating-parts/main.ts
```diff
@@ -37,25 +46,24 @@ const myCheckboxStyle = createPart('checkboxStyle')
                 }
             }
         }
-
+
```
gportela85: empty space here

PR #9240, opened 2024-11-18
AG-13351 - Abstract homepage section into website shared folder

taktran:


markdurrant (changes requested):
All code looks great 👍 

But can I request some name changes...

Can we change the `homepage` dir to `landing-pages` and the component name from `Section` to `LandingPageSection`? 

Hopefully we'll use these components for more than just homepages. And I think it might be confusing to have a react component with the same name as a default HTML element.

taktran:
> Can we change the homepage dir to landing-pages and the component name from Section to LandingPageSection?

Yep, sounds good. Have updated.

markdurrant (approved):
Everything great now with new names 👍

PR #9255, opened 2024-11-19
AG-13423 - updates and fixes to modules

peterjrreynolds:


packages/ag-grid-enterprise/src/allEnterpriseModule.ts
```diff
@@ -48,3 +55,18 @@ export const AllEnterpriseModule: _ModuleWithoutApi = {
         TreeDataModule,
     ],
 };
+
+/**
+ * @feature All Enterprise and Community features
+ */
+export const AllEnterpriseModule: AllEnterpriseModuleType = {
```
StephenCooper, line 62: Yes, that does make sense as a missing option for setting up charts.

StephenCooper (approved):

PR #9272, opened 2024-11-21
Ag 13351 react table shared components

markdurrant:
Refactor react-table page to use shared landing page components


taktran (approved):
LGTM 👍🏼

PR #9279, opened 2024-11-22
AG-12272 rework csrm state handling and no transactions for immutable set row data

SalvatorePreviti:
In this PR we remove the use and dependency on transactions in the CSRM and the group stages.
Transactions are only a way to produce a change and will not be anymore passed around to stages.

To do so, we introduce a new class called RefreshModelState that brings around all the state informations needed during a model refresh, and, allows to keep the set of changed row nodes that can be later consumed by stages and deselection
This also make refreshModel correctly handle the start state and eventual stacked refreshModel calls, as we keep and update the state between multiple invocations of refreshModel.

This reduces state handling complexity and allow to implement setImmutableRowData that does not need to generate transactions, but, can just update the data in place.
This allows the grouping stage to be able to handle immutable data from hierarchical treeDataChildrenField when grouping is enabled and treeData is false, as it was not generating transactions.


testing/behavioural/src/tree-data/hierarchical/hierarchical-immutable-tree-data.test.ts
```diff
@@ -282,6 +282,8 @@ describe('ag-grid hierarchical immutable tree data', () => {
             · · └── K LEAF id:K ag-Grid-AutoColumn:"K" v:110
         `);
 
+        return;
```
StephenCooper: Is this return meant to be here?

SalvatorePreviti: sorry a leftover! thank you

PR #9289, opened 2024-11-25
AG-13485 Avoid running doesFilterPass inside ngZone

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-13485

### DIFF
--- a/packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkComponentWrapper.ts
+++ b/packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkComponentWrapper.ts
@@ -47,9 +47,13 @@ export class AngularFrameworkComponentWrapper
 
             callMethod(name: string, args: IArguments): void {
                 const componentRef = this.getFrameworkComponentInstance();
-                return angularFrameworkOverrides.runInsideAngular(() =>
-                    wrapper.getFrameworkComponentInstance()[name].apply(componentRef, args)
-                );
+                const methodCall = componentRef[name];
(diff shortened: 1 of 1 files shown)


packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkComponentWrapper.ts
```diff
@@ -47,9 +47,13 @@ export class AngularFrameworkComponentWrapper
 
             callMethod(name: string, args: IArguments): void {
                 const componentRef = this.getFrameworkComponentInstance();
-                return angularFrameworkOverrides.runInsideAngular(() =>
-                    wrapper.getFrameworkComponentInstance()[name].apply(componentRef, args)
-                );
+                const methodCall = componentRef[name];
+                // Special case for `doesFilterPass` as it's called very often and current implementation has
+                // this filter logic as part of the component when really it is just part of the filter model.
+                if (name === 'doesFilterPass') {
```
seanlandsman, line 53: Do we want to make this a bit more general? ie have a list of special cases that needs special processing?

seanlandsman (approved):
lgtm

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkComponentWrapper.ts
StephenCooper: I looked through the other component methods and this is really a stand out case due to the way it is called based on the number of rows in the grid. Most other methods on custom components are called once or twice when they are used.

PR #9290, opened 2024-11-25
AG-13115 Fallback to selection column if not grouping

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-13115

### DIFF
--- a/packages/ag-grid-community/src/columns/selectionColService.ts
+++ b/packages/ag-grid-community/src/columns/selectionColService.ts
@@ -13,6 +13,7 @@ import {
     _convertColumnEventSourceType,
     _destroyColumnTree,
     _updateColsMap,
+    isColumnGroupAutoCol,
     isColumnSelectionCol,
 } from './columnUtils';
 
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/selection/selectAllFeature.ts
```diff
@@ -203,7 +203,10 @@ export class SelectAllFeature extends BeanStub {
             const isAutoCol = isColumnGroupAutoCol(column);
             // default to displaying header checkbox in the selection column
             const location = _getCheckboxLocation(rowSelection);
-            if ((location === 'autoGroupColumn' && isAutoCol) || (location === 'selectionColumn' && isSelectionCol)) {
+            if (
+                (location === 'autoGroupColumn' && isAutoCol) ||
+                (isSelectionCol && beans.selectionColSvc?.isSelectionColumnEnabled())
```
AndrewGlazier, line 208: Is this selection col check needed? Wouldn't the selection col be hidden if location==='autoGroupColumn'?

packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -87,14 +88,17 @@ export class SelectionColService extends BeanStub implements NamedBean {
...
+        const { gos, beans } = this;
         const rowSelection = gos.get('rowSelection');
         if (typeof rowSelection !== 'object') {
             return false;
         }
 
-        if (rowSelection.checkboxLocation === 'autoGroupColumn') {
+        if (
+            rowSelection.checkboxLocation === 'autoGroupColumn' &&
+            beans.visibleCols.allCols.some(isColumnGroupAutoCol)
```
AndrewGlazier: Not certain, but autoColsSvc might be a better check here as it has _just_ the auto cols, so probably a faster look up

eliasmalik: I think we can use `autoColsSvc` assuming there's no way to hide a group column?

packages/ag-grid-community/src/selection/selectAllFeature.ts
eliasmalik: We're using the selection col check in the case where `location != 'autoGroupColumn'`

packages/ag-grid-community/src/columns/selectionColService.ts
AndrewGlazier: We don't currently allow it so should be 👍

PR #9294, opened 2024-11-26
Ag 13351 code block improvements

kylerphillips:


documentation/ag-grid-docs/src/components/theme-builder-homepage/ThemeBuilderHomepage.tsx
```diff
@@ -174,7 +174,7 @@ import { ${themeSelection} } from 'ag-grid-community';
                         </div>
                     </ShadowDom>
                 </div>
-                <div className={styles.codeBlockWrapper}>
+                <div className={`${styles.codeBlockWrapper} code-block-homepage`}>
```
taktran, line 177: It feels like `code-block-homepage` is specific only to the homepage, so why not put it in `ThemeBuilderExample.module.scss`?

PR #9295, opened 2024-11-26
Ag 13468 modules config design updates

markdurrant:
Modules configurator design improvements

- Enterprise Ⓔ in cells vertical alignment
- Style module names as <code> without breaking single click copying
- Style (xx) number of modules display. De-emphasise compared to feature name? 
- Make grid height better match row height. Row is aligned so the middle is on the bottom of the grid (cut off the bottom half of the row).

### DIFF
--- a/documentation/ag-grid-docs/src/components/module-mappings/ModuleConfiguration.tsx
+++ b/documentation/ag-grid-docs/src/components/module-mappings/ModuleConfiguration.tsx
@@ -21,7 +21,7 @@ export const ModuleConfiguration: FunctionComponent<Props> = ({ moduleConfig }) 
     return (
         <div className={styles.configuration}>
             <div className={styles.rowModel}>
-                <span>Row Model:</span>
+                <span className={styles.label}>Row Model:</span>
                 <div>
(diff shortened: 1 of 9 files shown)


documentation/ag-grid-docs/src/components/module-mappings/ModuleConfiguration.tsx
```diff
@@ -21,79 +21,87 @@ export const ModuleConfiguration: FunctionComponent<Props> = ({ moduleConfig })
...
+                                    checked={rowModelOption === moduleName}
+                                    onChange={() => {
+                                        updateRowModelOption(moduleName);
+                                    }}
+                                />{' '}
+                                {name}
+                                {isEnterprise && (
+                                    <>
+                                        {' '}
+                                        <EnterpriseIcon />
```
taktran: Does this one need to be changed to `<Icon name="enterprise" />` too?

taktran (approved):
Code looks good. Just a question in there.

documentation/ag-grid-docs/src/components/module-mappings/ModuleConfiguration.tsx
markdurrant: Good spot. Now fixed.

PR #9300, opened 2024-11-27
AG-13353 - Homepage hero updates

taktran:


documentation/ag-grid-docs/src/pages-styles/homepage.module.scss
```diff
@@ -111,6 +112,17 @@ body {
     }
 }
 
+.heroLinks {
+    display: flex;
+    position: absolute;
+    bottom: -72px;
```
taktran: Not sure if this is the best way to position the hero links below the grid, but it works.

markdurrant: I don't think you need the `position`, `bottom`, or `right` properties here. The flex stuff is doing all the work.

markdurrant (approved):
Answered one question. Everything else LGTM 👍

kylerphillips (approved):

PR #9307, opened 2024-11-28
AG-13353 - Make darkmode header light

taktran:
* Create a `data-is-homepage` attribute for homepage only
* Use `#{$selector-darkmode}:not(:has([data-is-homepage])) &` selector to only target homepage header darkmode
* Override header CSS vars to only use light mode colors on the homepage
* Update `--color-bg-site-header` CSS var to be the site brand color


external/ag-website-shared/src/components/site-header/SiteHeader.module.scss
```diff
@@ -12,10 +12,120 @@
         z-index: 6; // needed in order to prevent grid z-indexes overlapping on small height
     }
 
-    #{$selector-darkmode} & {
+    #{$selector-darkmode}:not(:has([data-is-homepage])) & {
         background: var(--color-bg-primary);
         border-bottom: 1px solid var(--color-border-primary);
     }
+
+    // Reset homepage colors to light mode
```
taktran: There is probably a better way to do this, but I didn't want to make too many changes to the design system - but maybe we should make those changes, if we want this change.

kylerphillips (approved):

markdurrant (approved):
LGTM 👍

kylerphillips (approved):

PR #9311, opened 2024-11-29
AG-7238 - added aria configuration to IMenuItemComp

gportela85:
https://ag-grid.atlassian.net/browse/AG-7238


packages/ag-grid-enterprise/src/widgets/agMenuItemComponent.ts
```diff
@@ -125,6 +125,7 @@ export class AgMenuItemComponent extends BeanStub<AgMenuItemComponentEvent> {
                 if (configureDefaults) {
                     this.configureDefaults(configureDefaults === true ? undefined : configureDefaults);
                 }
+                comp.setupAriaAttributes?.(this.setAriaAttribute.bind(this));
```
peterjrreynolds: `setupAriaAttributes` needs adding to `optionalMethods` of `MenuItemComponent` above

PR #9330, opened 2024-12-04
AG-10823 Replace individual icons in Theming API

BernieSumption:
This PR adds a new part, `iconOverrides`, that can be used to replace individual icons with icon font characters, SVGs, or foreground color with SVG masks.

Documentation on branch build: https://testing.ag-grid.com/AG-10823/javascript-data-grid/custom-icons/


packages/ag-grid-community/src/theming/parts/icon-set/overrides/icon-overrides.ts
```diff
@@ -0,0 +1,61 @@
...
+    cssImports?: string[];
+    icons: {
+        [key: string]: string;
+    };
+};
+
+type IconSetOverridesArgs = IconSetOverridesImage | IconSetOverridesFont;
+
+export const iconOverrides = (args: IconSetOverridesArgs) => {
+    const cssParts = [sharedIconStylesCSS];
```
gportela85, line 29: should probably deconstruct args, so we don't keep reading its value over and over again.
```
const { mask, type, cssImports } = args;
if (type === 'image') {
    for (const [key, value] of Object.entries(args.icons)) {
        const imageCssValue = imageValueToCss(value);
        let targetImage: string = 'mask';
        let styleSuffix: string = '';

        if (!mask) {
            targetImage = 'background';
            styleSuffix = unsetMaskIcon;
        }
        cssParts.push(`.ag-icon-${key}::before { ${targetImage}-image: ${imageCssValue}; ${styleSuffix} }`)
    }
}
if (type === 'font') {
    const { family, weight, color, icons } = args;
   //...
```

PR #9338, opened 2024-12-05
AG-13333 - module docs updates

peterjrreynolds:


documentation/ag-grid-docs/src/content/docs/modules/index.mdoc
```diff
@@ -4,211 +4,45 @@ title: "AG Grid Modules"
...
+`ModuleRegistry.registerModules()` can be called multiple times to register additional modules, so `ModuleRegistry.registerModules([ValidationModule])` can be run when in development.
 
 ## Registering AG Grid Modules
 
 When including AG Grid in your application via modules it is your responsibility to register the required modules with the grid before it is instantiated. You can either register them globally or pass them individually to each grid instance.
 
 ### Providing Modules Globally
 
-You can import and register modules globally, but you need to ensure that this is done before **_any_** Grids are instantiated. Any modules registered globally will be available to all Grids.
+You can import and register modules globally, but you need to ensure that this is done before **_any_** grids are instantiated. Any modules registered globally will be available to all grids.
```
StephenCooper, line 32: We may want a note for React SSR to remind people that modules need to be registered on the client and not the server.

StephenCooper (approved):
Just  a small update

documentation/ag-grid-docs/src/content/docs/modules/index.mdoc
peterjrreynolds: Done

PR #9352, opened 2024-12-06
AG-13333 - remove docs modules references

peterjrreynolds:


documentation/ag-grid-docs/src/content/docs/theming-v32-upgrading-to-v28-css/index.mdoc
```diff
@@ -30,24 +28,6 @@ There are many ways to import CSS, but however you are doing this in your app yo
   href="https://cdn.jsdelivr.net/npm/ag-grid-community@{% $agGridVersion %}/styles/ag-grid.css" />
 ```
 
-### Modules
```
StephenCooper, line 33: Is this valid though for the upgrade that is being talked about? Or is this actually an upgrade to lastet?

StephenCooper (approved):

documentation/ag-grid-docs/src/content/docs/theming-v32-upgrading-to-v28-css/index.mdoc
peterjrreynolds: It's for an upgrade from <v28 to v28+. However the code snippets default to the latest version, and that's not going to exist for v33, so they will be wrong.

Our policy is for the docs to be for the latest version only, so we should be treating this as an upgrade to v33. If people want to do an upgrade to an older version, they can use the archive version of the page.

PR #9392, opened 2024-12-09
RTI-2387 fix example aggregation

StephM-W:


documentation/ag-grid-docs/src/content/docs/excel-export-rows/_examples/excel-export-pinned-rows/main.ts
```diff
@@ -82,9 +82,9 @@ const gridOptions: GridOptions<IOlympicData> = {
             date: '01/08/2030',
             sport: 'Track & Field',
             gold: 222,
-            silver: '005',
+            silver: 5,
             bronze: 244,
-            total: 255,
+            total: 571,
```
eliasmalik: isn't the total 471?

PR #9459, opened 2024-12-10
RTI-2508 - update label.placement documentation

StephM-W:


adam-wang-uk (changes requested):
1. The link for AG Charts label placement option is missing a /charts/
https://www.ag-grid.com/charts/javascript/bar-series/#reference-AgBarSeriesOptions-label-placement

2. Metatag included should be "sparklineOptions.label.placement" this is so when people search this option it comes up in migration guide

adam-wang-uk (approved):

PR #9494, opened 2024-12-11
Fix angular sparklines dark mode / package.json

StephenCooper:


StephM-W (approved):

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
```diff
@@ -680,7 +680,21 @@ const darkModeTs = `
         };
         
         // update chart themes when example first loads
-        updateChartThemes(isInitialModeDark);
+        let initialSet = false;
+        const maxTries = 5;
+        let tries = 0;
+        const trySetInitial = (delay) => {
+            if(params.api){
```
taktran, line 687: It feels like this would fail the format check

taktran (approved):
Looks good. I think it has to run through the formatter though.

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
StephenCooper: thats just a string

PR #9500, opened 2024-12-11
AG-13215 - Fix sticky row horizontal position

AndrewGlazier:


gportela85:
I think this fix is good, even if there is a 1px border that would be visible due to overflow, `ag-invisible` will hide it and we won't get an invisible gap because the height is 0.
Looks good to me.

PR #9522, opened 2024-12-12
AG-13496 - date filter handle non dates

peterjrreynolds:


packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -300,7 +304,7 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
     // The default comparator assumes that the cellValue is a date
     const cellAsDate = cellValue as Date;
 
-    if (cellValue == null || cellAsDate < filterDate) {
```
peterjrreynolds, line 303: comparator never gets called with null values

StephM-W (approved):

PR #9523, opened 2024-12-12
AG-12210 - don't autogenerate chartTheme related imports

StephM-W:


plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/grid-vanilla-src-parser.ts
```diff
@@ -381,7 +382,7 @@ function internalParser(
         return colDefs || '';
     };
 
-    PROPERTIES.forEach((propertyName) => {
+    PROPERTIES.filter((property) => !EXCLUDED_PROPERTIES.includes(property)).forEach((propertyName) => {
```
StephenCooper, line 385: Does this not break the generated examples? As in if the property is defined in the main.ts example it won't be included in the framework ones?

StephM-W: Will double-check and manually updated any affected examples

StephM-W: No existing examples use these types currently, so I think we should either:
* continue with the PR as-is, or
* add them wherever chartThemeOverride is used (`... as AgChartThemeOverride`)

PR #9541, opened 2024-12-16
Fix Spelling mistake

YazaHealth:
This was annoying me slightly


StephenCooper:
Thanks for reporting! I have made the change for our next release.

PR #9551, opened 2024-12-16
AG-13702 Correct Regex

StephenCooper:


plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
```diff
@@ -60,11 +60,12 @@ export function getFunctionName(code: string): string {
 
 export const convertFunctionToProperty = (code: string) =>
     code.replace(/function\s+([^(\s]+)\s*\(([^)]*)\)/, '$1 = ($2) =>');
-export const convertFunctionToPropertyTs = (code: string) => {
-    return code.replace(/function\s+([^(\s]+)\s*\(([^)]*)\):(\s+[^{]*)/, '$1: ($2) => $3 = ($2) =>');
+
```
StephenCooper, line 63: This is actually reverting to previous state of the file with no changes before my previous commit.

PR #9569, opened 2024-12-17
AG-13328 - fix grid serializer value getter to handle properly tree groups and add excel tests

SalvatorePreviti:
When we changed the tree data logic, we fixed a bug where nodes were not correctly marked as groups, and group nodes were marked as group nodes only in some circumstances. Now they are correctly marked as groups after tree data rewrite.
In turns, this activate another pre-existing bug in the excel value serializer, that caused the value getter to not be invoked if group is true, and this is a feature used by grouping but should not be used by tree data

Changes:

- One line fix to properly handle value getter in tree groups by checking if the node is a tree data node
- Add a unit test and the utilities able to test excel files in behavioural unit tests


packages/ag-grid-community/src/export/baseGridSerializingSession.ts
```diff
@@ -94,8 +94,8 @@ export abstract class BaseGridSerializingSession<T> implements GridSerializingSe
     }
 
     private shouldRenderGroupSummaryCell(node: RowNode, column: AgColumn, currentColumnIndex: number): boolean {
-        const isGroupNode = node && node.group;
-        // only on group rows
+        // only on group rows when grouping, and not for tree data group nodes
+        const isGroupNode = node.group && !this.gos.get('treeData');
```
StephenCooper, line 98: Are we sure the node is defined all the time given the previous code?

testing/behavioural/src/tree-data/hierarchical/hierarchical-tree-excel-export.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            rowData,
+            treeData: true,
+            groupDefaultExpanded: -1,
+            ['treeDataChildrenField' as any]: 'children',
+            getRowId: (params) => params.data.uiId,
+        });
+
+        api.exportDataAsExcel({ fileName: 'test.xlsx' });
+
+        expect(await getExcelJsonData(await objectUrls.pullBlob())).toEqual([
```
StephenCooper, line 63: These could be so valuable for other excel export features.

packages/ag-grid-community/src/export/baseGridSerializingSession.ts
SalvatorePreviti: I am going to check again but it seems it cannot be null so it seems it was a typo

SalvatorePreviti: yes i am pretty sure or the code that call that method would fail as well, and that method is private and used only there. 
image

StephenCooper (approved):

testing/behavioural/src/tree-data/hierarchical/hierarchical-tree-excel-export.test.ts
SalvatorePreviti: "left as an exercise to the reader"™

packages/ag-grid-community/src/export/baseGridSerializingSession.ts
gportela85: It looks like `node` used to be optional many years ago: https://github.com/ag-grid/ag-grid/blob/8496a5d5b3c232fd2dbbfe853a9d18ec456f2bf0/packages/ag-grid-community/src/ts/exporter/gridSerializer.ts#L137

Then at some point the `node && node.group` check was added, then `node` was made mandatory but the failsafe was never removed. It should be ok to be removed.

PR #9590, opened 2024-12-19
Ag 13658 fw animation update

markdurrant:
- Refactor so component only includes animated text
- Add `prefixText` and `suffixText` props
- Sync `ag-website-shared`


kylerphillips (approved):
LGTM 🚀

PR #9619, opened 2025-01-02
Ag 13709 clean changed row nodes

AndrewGlazier:


packages/ag-grid-community/src/clientSideRowModel/changedRowNodes.ts
```diff
@@ -3,27 +3,26 @@ import type { IRowNode } from '../interfaces/iRowNode';
...
+    public readonly adds = new Set<RowNode<TData>>();
 
     /** Marks a row as removed. Order of operations is: remove, update, add */
     public remove(node: IRowNode<TData>): void {
-        const { updates, removals } = this;
-        removals.add(node as RowNode<TData>);
-        updates.delete(node as RowNode<TData>);
+        this.removals.add(node as RowNode<TData>);
+        if (!this.updates.delete(node as RowNode<TData>)) {
+            this.adds.delete(node as RowNode<TData>);
```
AndrewGlazier: if node was just added, does it need to be in removals list still? @SalvatorePreviti

SalvatorePreviti: changedRowNodes.remove(node) = add to the set of removed nodes, remove it from the sets of add or updates nodes, this is needed for example when doing batched transactions - we need to be sure that if transactions are ({add:2}, {update:[1, 2]}, {remove:[1, 2]}) we respect the rule that the 3 sets are mutually exclusive, so one node can be only in one of the 3 sets or none.
We can change in such a way if the node was just added and then removed in the same batch we don't add it to the removals set and remove from the adds set, resulting in only node 1 in the removals set and updates and adds empty, in the example above. If this is the question. However currently either way this don't break and all works as expected, currently the consumers are fine with processing removed nodes that weren't added before. - let me know what you think

changedRowNodes.update(node) = if the node was not added before, add it to the update set. And we don't need to remove it from the removals set, as nodes are never reused, once a node is delete we create a new one in any case. Also this makes sense in batched transactions, as [{add:1}, {update:1}] we still want node 1 to be just in the set of added nodes and not in the set of updated nodes

changedRowNodes.add(node) = just add the node to the add set, and we are sure we call this only when the node is freshly created so we don't need to check or update the updates or removed nodes.

SalvatorePreviti: Tried to do this in this draft PR, please have a look @AndrewGlazier  - https://github.com/ag-grid/ag-grid/pull/9622 - the problem in doing this is that there is a minor change in behaviour and modelUpdate is not executed when there is a transaction batch [{add:1}{remove:1}] while before it was called. From a certain point of view seems more correct that modelUpdated is not called as the transaction batch in this case would result in no changes, but is a different behaviour from previous versions and is a different behaviour if transactions were to be executed one by one that can be a bit confusing.

AndrewGlazier: I think I'm of the opinion that if we're making no changes to the nodes (i.e row is added and deleted in the same op) then the row model hasn't actually updated - and us then running the row model update is wrong, because there's been no change

It's fine that we do update this if it's two different ops as the grid is synchronous and the row model updates twice in this instance

I'd be happy to include the draft PR personally @SalvatorePreviti

PR #9631, opened 2025-01-06
AG-13729 Separate public private event handlers

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-13729


packages/ag-grid-community/src/context/beanStub.ts
```diff
@@ -161,6 +174,13 @@ export abstract class BeanStub<TEventType extends string = BeanStubEvent>
         };
     }
 
+    // type guard for IAgEventEmitter
+    private isAgEventEmitter<TEvent extends string>(
```
peterjrreynolds: This could be a standalone function

peterjrreynolds (approved):

PR #9635, opened 2025-01-06
AG-13719 Adjusted footer design

kylerphillips:
Make footer no longer full width, fits with rest of homepage.

Looks slightly odd on docs page but I think it's ok?

image

Screenshot 2025-01-06 at 15 20 12

### DIFF
--- a/external/ag-website-shared/src/components/footer/Footer.module.scss
+++ b/external/ag-website-shared/src/components/footer/Footer.module.scss
@@ -1,6 +1,7 @@
 @use 'design-system' as *;
 
 .footer {
+    width: 100%;
     border-top: 1px solid var(--color-border-secondary);
     padding-top: $spacing-size-10;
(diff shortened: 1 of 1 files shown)


taktran (approved):

external/ag-website-shared/src/components/footer/Footer.module.scss
```diff
@@ -33,6 +34,10 @@
 }
 
 .footerColumns {
+    max-width: calc(var(--layout-max-width-small) + var(--layout-horizontal-margins) * 2);
+    margin-right: auto;
+    margin-left: auto;
+    padding: $spacing-size-2;
     @media screen and (min-width: $breakpoint-site-footer-five-column) {
```
markdurrant, line 42: Just need a space between padding & @media here 👍

markdurrant (changes requested):
One minor nitpick, otherwise LGTM 👍

PR #9642, opened 2025-01-07
AG-12904 - remove Math.random usage

StephM-W:


documentation/ag-grid-docs/src/content/docs/change-cell-renderers/_examples/animated-flashing-cells/main.ts
```diff
@@ -9,6 +9,8 @@ import {
     createGrid,
 } from 'ag-grid-community';
 
+import './random';
```
seanlandsman, line 13: I think we need a comment here that explains why this import is here - and that its only needed for example/test data purposes

seanlandsman (changes requested):

documentation/ag-grid-docs/src/content/docs/change-cell-renderers/_examples/animated-flashing-cells/main.ts
StephM-W: Updated with additional comment

seanlandsman (approved):
lgtm

PR #9643, opened 2025-01-07
AG-13797 Make selectionColDef reactive

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-13797


packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -53,15 +61,14 @@ export class SelectionColService extends BeanStub implements NamedBean {
...
 
-        // the new tree dept will equal the current tree dept of cols
         const newTreeDepth = cols.treeDepth;
         const oldTreeDepth = this.selectionCols?.treeDepth ?? -1;
-        const treeDeptSame = oldTreeDepth == newTreeDepth;
+        const treeDepthSame = oldTreeDepth == newTreeDepth;
 
         const list = this.generateSelectionCols();
-        const areSame = _areColIdsEqual(list, this.selectionCols?.list ?? []);
+        const areSame = _isDeepEqual(list, this.selectionCols?.list ?? []);
```
AndrewGlazier: Little uncomfortable with this as I imagine this is comparing a lot of column properties we would expect to change, ie col width or other stateful properties

Have you looked at the approach we use for updating `autoGroupColumnDef`? We apply a listener in the `autoColService` and just forcefully reset the existing column state/update the existing columns colDefs, this means we can avoid doing a full `colModel.refresh` too which can be quite cumbersome - it also means users are less impacted when not memoising this obj, though they should anyway

PR #9647, opened 2025-01-08
AG-13809 - fix api functions tied to row models

peterjrreynolds:


packages/ag-grid-community/src/interfaces/iModule.ts
```diff
@@ -47,7 +47,7 @@ export interface Module {
 }
 
 /** Used to define a module that contains api functions. */
-export type _ModuleWithApi<TGridApi extends Readonly<Partial<GridApi>>> = Module & {
+export type _ModuleWithApi<TGridApi extends Readonly<Partial<GridApi>>> = Omit<Module, 'rowModels'> & {
```
peterjrreynolds, line 50: This will prevent us from accidentally tying API functions to row models in future.

PR #9652, opened 2025-01-09
AG-13828 - Add gradient background to lightmode hero grid

markdurrant:
Add gradient background to lightmode hero grid to increase contrast. 

Left: Current, Right: with new gradient 
Screenshot 2025-01-09 at 11 26 28


kylerphillips:
Added a small shadow, otherwise think this is a good improvement. Dig it!

kylerphillips (approved):

PR #9657, opened 2025-01-12
AG-10167 - allowed rich select values to have a different structure as the grid column

gportela85:
https://ag-grid.atlassian.net/browse/AG-10167

### DIFF
--- a/packages/ag-grid-enterprise/src/richSelect/richSelectCellEditor.ts
+++ b/packages/ag-grid-enterprise/src/richSelect/richSelectCellEditor.ts
@@ -128,25 +128,35 @@ export class RichSelectCellEditor<TData = any, TValue = any> extends PopupCompon
     }
 
     private getSearchStringCallback(values: TValue[]): ((values: TValue[]) => string[]) | undefined {
-        const params = this.params;
-        const { colDef } = params;
-
-        if (typeof values[0] !== 'object' || !colDef.keyCreator) {
+        if (typeof values[0] !== 'object') {
             return;
         }
 
-        const { column, node, data } = params;
-        return (values: TValue[]) =>
-            values.map((value: TValue) => {
-                const keyParams: KeyCreatorParams = this.gos.addGridCommonParams({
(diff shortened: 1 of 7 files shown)


packages/ag-grid-enterprise/src/richSelect/richSelectCellEditor.ts
```diff
@@ -128,25 +128,35 @@ export class RichSelectCellEditor<TData = any, TValue = any> extends PopupCompon
...
+        }
+
         const params = this.params;
-        const { colDef } = params;
+        const { colDef, formatValue } = params;
 
-        if (typeof values[0] !== 'object' || !colDef.keyCreator) {
-            return;
+        if (colDef.cellEditorParams?.formatValue) {
+            return (values: TValue[]) => values.map(formatValue!);
```
peterjrreynolds, line 139: Are `colDef.cellEditorParams?.formatValue` and `formatValue` the same here?

packages/ag-grid-enterprise/src/richSelect/richSelectCellEditor.ts
```diff
@@ -128,25 +128,35 @@ export class RichSelectCellEditor<TData = any, TValue = any> extends PopupCompon
...
+                        node,
+                        data,
+                    });
+                    return keyCreator(keyParams);
                 });
-                return colDef.keyCreator!(keyParams);
-            });
+        }
+
+        return (values: TValue[]) => values.map(formatValue!);
```
peterjrreynolds, line 159: Would `formatValue` be defined here?

packages/ag-grid-enterprise/src/richSelect/richSelectCellEditor.ts
gportela85: No, `formatValue` is created automatically by `cellEditService.createCellEditorParams` if none is provided. So i need to check if one was actually provided for the Rich Select Editor by the user.

PR #9661, opened 2025-01-13
AG-13850 initializing columns after rowdata does not trigger refreshModel

SalvatorePreviti:
A regression caused by the new way refreshModel and init service and column service interact causes the refreshModel to not be called if columns are set after rowData in a grid that has nor rowData or columns.

changes:

- Add a unit test for this case
- Fix the CSRM initialization logic to properly invoke refresh


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -209,19 +209,15 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
 
     public start(): void {
         this.started = true;
         if (this.shouldSkipSettingDataOnStart) {
-            this.refreshModel({
-                step: 'group',
-                newData: true,
-                rowDataUpdated: true,
-            });
+            this.refreshModel({ step: 'group' });
```
StephenCooper: Is this change required for this fix?

SalvatorePreviti: all right - reverted this and added tests and improved logic to be correct with raising events

PR #9665, opened 2025-01-13
Ag 13591 module callout component

kylerphillips:
Screenshot 2025-01-13 at 15 37 51

### DIFF
--- a/documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.astro
+++ b/documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.astro
@@ -1,5 +1,11 @@
 ---
 import { ModuleCallout } from '@components/module-callout/components/ModuleCallout';
+
+interface Props {
+    moduleName: string;
+}
+
+const { moduleName } = Astro.props as Props;
 ---
 
-<ModuleCallout client:only="react" />
+<ModuleCallout moduleName={moduleName} client:only="react" />
--- a/documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.astro
+++ b/documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.astro
@@ -1,0 +1,5 @@
+---
+import { ModuleCallout } from '@components/module-callout/components/ModuleCallout';
+---
+
+<ModuleCallout client:only="react" />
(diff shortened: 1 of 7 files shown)


documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.tsx
```diff
@@ -0,0 +1,30 @@
+import { Alert } from '@ag-website-shared/components/alert/Alert';
+import { getFrameworkFromPath } from '@components/docs/utils/urlPaths';
+import { urlWithPrefix } from '@utils/urlWithPrefix';
+
+interface Props {
+    moduleName: string;
+}
+
+export const ModuleCallout = ({ moduleName }: Props) => {
+    const framework = getFrameworkFromPath(window.location.pathname);
```
taktran: You would pass this in from the Astro layer, so that this component is simpler eg, 

https://github.com/ag-grid/ag-grid/blob/latest/documentation/ag-grid-docs/src/components/license-setup/components/LicenseSetup.astro#L7

documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.tsx
```diff
@@ -0,0 +1,30 @@
...
+    return (
+        <>
+            <Alert type="info">
```
taktran, line 13: How come you have a `module` type, but don't use it here? It feels like you don't need the code in `Alert.tsx`?

documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.astro
```diff
@@ -0,0 +1,11 @@
...
+import { ModuleCallout } from '@components/module-callout/components/ModuleCallout';
+
+interface Props {
+    moduleName: string;
+}
+
+const { moduleName } = Astro.props as Props;
+---
+
+<ModuleCallout moduleName={moduleName} client:only="react" />
```
taktran: It seems like static content so could probably use `client:load`:

```suggestion
<ModuleCallout moduleName={moduleName} client:load />
```

documentation/ag-grid-docs/src/components/module-callout/components/ModuleCallout.tsx
kylerphillips: Removed unused module type in latest commit

external/ag-website-shared/src/components/alert/Alert.module.scss
```diff
@@ -75,3 +75,26 @@
         fill: var(--color-success);
     }
 }
+
+.module {
```
taktran: You might be able to remove these styles too?

taktran (approved):
Looks good 👍🏼

PR #9671, opened 2025-01-15
AG-13841 Fix flex, pinning and width column updates

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-13841

### DIFF
--- a/packages/ag-grid-community/src/columns/columnFactoryUtils.ts
+++ b/packages/ag-grid-community/src/columns/columnFactoryUtils.ts
@@ -184,13 +184,12 @@ export function updateSomeColumnState(
 
     // pinned - anything but undefined, thus null or empty string will remove pinned
     if (pinned !== undefined && pinned !== column.getPinned()) {
-        pinnedCols?.setColsPinned([column], pinned, source);
+        pinnedCols?.setColPinned(column, pinned);
     }
 
     // flex
     if (flex !== undefined && flex !== column.getFlex()) {
         colFlex?.setColFlex(column, flex);
-        // colFlex?.refreshFlexedColumns();
     }
 }
 
@@ -227,9 +226,6 @@ export function _updateColumnState(
         const widthBeforeUpdate = column.getActualWidth();
         column.setActualWidth(widthBeforeUpdate, source);
     }
-    // // Do a minimal refresh here: widths changing imply left positions and body widths have also changed.
-    // beans.visibleCols.setLeftValues(source);
-    // beans.visibleCols.updateBodyWidths();
 }
 
 function findExistingColumn(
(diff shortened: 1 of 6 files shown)


packages/ag-grid-community/src/columns/columnFactoryUtils.ts
```diff
@@ -183,13 +183,14 @@ export function updateSomeColumnState(
     }
 
     // pinned - anything but undefined, thus null or empty string will remove pinned
-    if (pinned !== undefined) {
-        pinnedCols?.setColPinned(column, pinned);
+    if (pinned !== undefined && pinned !== column.getPinned()) {
+        pinnedCols?.setColsPinned([column], pinned, source);
```
eliasmalik: Using `setColsPinned` because it refreshes visible cols and dispatches pinned event, instead of having to do it manually.

packages/ag-grid-community/src/columns/columnFactoryUtils.ts
```diff
@@ -226,6 +227,9 @@ export function _updateColumnState(
         const widthBeforeUpdate = column.getActualWidth();
         column.setActualWidth(widthBeforeUpdate, source);
     }
+    // Do a minimal refresh here: widths changing imply left positions and body widths have also changed.
+    beans.visibleCols.setLeftValues(source);
+    beans.visibleCols.updateBodyWidths();
```
eliasmalik: alternatively could just do a full `visibleCols.refresh` but it's slightly more work

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -69,14 +69,13 @@ export class HeaderRowCtrl extends BeanStub {
      * @param comp Proxy to the actual component
      * @param initCompState Should the component be initialised with the current state of the controller. Default: true
      */
-    public setComp(comp: IHeaderRowComp, compBean: BeanStub | undefined, initCompState: boolean = true): void {
+    public setComp(comp: IHeaderRowComp): void {
```
eliasmalik: Simplifying signature, was only ever called in one place with `undefined` as second argument.

packages/ag-grid-community/src/columns/columnFactoryUtils.ts
AndrewGlazier: I think this should be being handled by `_compareColumnStatesAndDispatchEvents` which contains this code:

```
const pinnedChangePredicate = (cs: ColumnState, c: AgColumn) => cs.pinned != c.getPinned();
dispatchColumnPinnedEvent(eventSvc, getChangedColumns(pinnedChangePredicate), source);
```

Adding this in here is adding another layer of event firing, I assume the flex below is the same problem

AndrewGlazier:
Having looked into this a bit I think what needs to be done is `_updateColumnState` shouldn't be explicitly called (my original mistake) 

This is also done in `columnFactoryUtils.createColumn` but as this is a fresh column the lack of events here is not important

I think instead changing the `_updateColumnState` functions to use `_applyColumnState` instead will add the required event handling/checking here

AndrewGlazier (changes requested):

PR #9673, opened 2025-01-15
AG-13242 - Update Nx plugins to use ag-shared/plugin-utils.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13242


taktran (approved):
LGTM 👍🏼

PR #9674, opened 2025-01-15
AG-13242 - Upgrade nx to nx@20.3.1

taktran:


nx.json
```diff
@@ -155,5 +171,6 @@
       "inputs": ["default", "^production"]
     }
   },
-  "defaultProject": "all"
+  "defaultProject": "all",
+  "useLegacyCache": true
```
alantreadway: Should we try `false` here?

alantreadway (commented):
LGTM

nx.json
taktran: Ah yes, I'll try that

PR #9682, opened 2025-01-16
AG-9010 - Initial commit for row span

AndrewGlazier:


packages/ag-grid-community/src/allCommunityModule.ts
```diff
@@ -27,6 +27,7 @@ import {
 } from './filter/filterModule';
 import { InfiniteRowModelModule } from './infiniteRowModel/infiniteRowModelModule';
 import type { _ModuleWithoutApi } from './interfaces/iModule';
+import { CellSpanModule } from './main';
```
gportela85: No imports from 'main', right?

packages/ag-grid-community/src/rendering/spanning/rowSpanService.ts
```diff
@@ -0,0 +1,114 @@
+import type { NamedBean } from '../../context/bean';
+import { BeanStub } from '../../context/beanStub';
+import type { AgColumn } from '../../entities/agColumn';
+import type { RowNode } from '../../entities/rowNode';
+import type { CellPosition } from '../../interfaces/iCellPosition';
+import { _debounce } from '../../main';
```
gportela85: Import from `main`.

packages/ag-grid-community/src/theming/core/css/_grid-layout.css
```diff
@@ -78,6 +78,33 @@
     );
 }
 
+/* Cells not contained by rows are spanned without the parent row - they need some basic styles rows would normally apply */
```
gportela85: don't forget to add these style for customer using legacy themes, so we need these changes added to:

`/community-modules/styles/src/interna/base/parts/_grid-layout.scss`

AndrewGlazier: I've done it for the file suggested, does the other css change require similar? I couldn't find an equivalent? (_general.css)

packages/ag-grid-community/src/entities/colDef.ts
```diff
@@ -698,6 +698,18 @@ export interface ColDef<TData = any, TValue = any> extends AbstractColDef<TData,
...
+    /**
+     * Set to `true` to automatically merge cells in this column with equal values. Provide a callback to specify custom merging logic.
+     */
+    spanRows?:
+        | boolean
+        | ((
+              valueA: TValue | null | undefined,
+              valueB: TValue | null | undefined,
+              nodeA: IRowNode<TData>,
+              nodeB: IRowNode<TData>
```
peterjrreynolds: Should make this a params object

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
```diff
@@ -327,6 +360,19 @@ export class RowContainerCtrl extends BeanStub implements ScrollPartner {
 
         this.onDisplayedColumnsChanged();
         this.onDisplayedRowsChanged();
+
+        if (this.beans.spannedCellRenderer && this.options.getSpannedRowCtrls) {
+            this.addManagedListeners(this.beans.spannedCellRenderer, {
+                spannedRowsUpdated: () => {
+                    const spannedCtrls = this.options.getSpannedRowCtrls!(this.beans.spannedCellRenderer!);
```
peterjrreynolds: bundle size (`this.beans.spannedCellRenderer`)

packages/ag-grid-community/src/interfaces/iServerSideRowModel.ts
```diff
@@ -7,7 +7,6 @@ import type { ServerSideTransaction, ServerSideTransactionResult } from './serve
 export interface IServerSideRowModel<TData = any> extends IRowModel {
     refreshStore(params?: RefreshServerSideParams): void;
     onRowHeightChanged(): void;
-    onRowHeightChangedDebounced(): void;
```
peterjrreynolds, line 10: I think this is technically a breaking change. Although we removed the model from the API, you can still get it in custom filter params - https://ag-grid.com/javascript-data-grid/component-filter/#reference-IFilterParams-rowModel.

packages/ag-grid-community/src/rendering/row/rowAutoHeightService.ts
```diff
@@ -17,107 +17,136 @@ export class RowAutoHeightService extends BeanStub implements NamedBean {
...
+                if (spannedCell) {
+                    // only last row gets additional auto height of spanned cell
+                    if (spannedCell.getLastNode() !== row) {
+                        continue;
                     }
-                    if (activeColsForRow.includes(col)) {
-                        // Column is present in the row, i.e not spanned over, but no auto height was provided so we cannot calculate the row height
-                        notAllPresent = true;
+
+                    cellHeight = this.beans.rowSpanSvc?.getCellSpan(col, row)?.getLastNodeAutoHeight();
```
peterjrreynolds: bundle size

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -711,6 +711,9 @@ export class RowRenderer extends BeanStub implements NamedBean {
             ...this.bottomRowCtrls,
             ...stickyTopRowCtrls,
             ...stickyBottomRowCtrls,
+            ...(this.beans.spannedCellRenderer?.getCtrls('top') ?? []),
+            ...(this.beans.spannedCellRenderer?.getCtrls('bottom') ?? []),
+            ...(this.beans.spannedCellRenderer?.getCtrls('center') ?? []),
```
peterjrreynolds: bundle size

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-auto-height/main.ts
```diff
@@ -0,0 +1,54 @@
...
+    columnDefs: columnDefs,
+    defaultColDef: {
+        flex: 1,
+    },
+};
+
+// setup the grid after the page has finished loading
+document.addEventListener('DOMContentLoaded', function () {
+    const gridDiv = document.querySelector<HTMLElement>('#myGrid')!;
+    const registry = ModuleRegistry;
```
StephenCooper: Why is this here in the example as well as the registration at the top?

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-custom/main.ts
```diff
@@ -0,0 +1,43 @@
...
+    columnDefs: columnDefs,
+    defaultColDef: {
+        flex: 1,
+    },
+};
+
+// setup the grid after the page has finished loading
+document.addEventListener('DOMContentLoaded', function () {
+    const gridDiv = document.querySelector<HTMLElement>('#myGrid')!;
+    const registry = ModuleRegistry;
```
StephenCooper: Duplicated?

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-simple/main.ts
```diff
@@ -48,16 +23,16 @@ let gridApi: GridApi<IOlympicData>;
...
-        width: 170,
+        flex: 1,
     },
-    suppressRowTransform: true,
 };
 
 // setup the grid after the page has finished loading
 document.addEventListener('DOMContentLoaded', function () {
     const gridDiv = document.querySelector<HTMLElement>('#myGrid')!;
+    const registry = ModuleRegistry;
```
StephenCooper: Duplicate

packages/ag-grid-community/src/context/context.ts
```diff
@@ -301,6 +303,8 @@ export interface CoreBeanCollection {
     rowChildrenSvc?: IRowChildrenService;
     footerSvc?: IFooterService;
     touchSvc?: TouchService;
+    rowSpanSvc?: RowSpanService;
+    spannedCellRenderer?: SpannedCellRenderer;
```
StephenCooper: I was under the impression that only services should be beans. I don't think we have any other renderers that are beans. 

Maybe just needs a different name.

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -149,11 +155,16 @@ export class AgColumn<TValue = any>
         userProvidedColDef: ColDef<any, TValue> | null,
         source: ColumnEventType
     ): void {
+        const colSpanChanged = colDef.spanRows !== this.colDef.spanRows;
```
StephenCooper, line 158: Nit Pick: Can this be colocated next to the if.

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        const newRowCtrls = new Map<RowNode, RowCtrl>();
+
+        let hasNewSpans = false;
+        for (const ctrl of rowCtrls) {
+            // skip cells that are animating out.
+            if (!ctrl.isAlive()) {
+                continue;
+            }
+            this.beans.rowSpanSvc?.forEachSpannedColumn(ctrl.rowNode, (col, cellSpan) => {
```
StephenCooper: Nit: Destructure beans to reduce repetition of this.beans.

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    public getCellByPosition(cellPosition: CellPosition): CellCtrl | undefined {
+        const cellSpan = this.beans.rowSpanSvc?.getCellSpanByPosition(cellPosition);
+        if (!cellSpan) {
+            return undefined;
+        }
+
+        const ctrlsName = `${_normalisePinnedValue(cellPosition.rowPinned)}Ctrls` as const;
```
StephenCooper: I would be worried about the performance of this when compared to a switch statement.

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        if (!ctrl) {
+            return undefined;
+        }
+
+        return ctrl.getAllCellCtrls().find((cellCtrl) => cellCtrl.column === cellPosition.column);
+    }
+
+    public getCtrls(container: 'top' | 'bottom' | 'center'): RowCtrl[] {
+        const ctrlsName = `${container}Ctrls` as const;
```
StephenCooper: I am not a fan of the string concatenation approach to select from the controllers.

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
```diff
@@ -0,0 +1,137 @@
+import type { NamedBean } from '../../context/bean';
+import { BeanStub } from '../../context/beanStub';
+import type { RowNode } from '../../entities/rowNode';
+import type { CellPosition } from '../../interfaces/iCellPosition';
+import type { RowPinnedType } from '../../interfaces/iRowNode';
+import type { CellCtrl } from '../cell/cellCtrl';
+import type { RowCtrl } from '../row/rowCtrl';
+import { SpannedRowCtrl } from './spannedRowCtrl';
+
+export class SpannedCellRenderer extends BeanStub<'spannedRowsUpdated'> implements NamedBean {
```
StephenCooper: I think this name needs changing to avoid confusion with our existing Components.

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -39,19 +50,31 @@ export class RowContainerComp extends Component {
...
+            setSpannedRowCtrls: (rowCtrls: RowCtrl[]) => this.setRowCtrls(rowCtrls, true),
             setDomOrder: (domOrder) => {
                 this.domOrder = domOrder;
             },
-            setContainerWidth: (width) => (this.eContainer.style.width = width),
-            setOffsetTop: (offset) => (this.eContainer.style.transform = `translateY(${offset})`),
+            setContainerWidth: (width) => {
+                this.eContainer.style.width = width;
+                if (this.eSpannedContainer) {
+                    this.eSpannedContainer.style.width = width;
```
StephenCooper, line 69: Nit: use optional chaining to reduce size.

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -89,33 +116,35 @@ export class RowContainerComp extends Component {
                 }
                 rowComp = new RowComp(rowCtrl, beans, options.type);
             }
-            this.rowComps[instanceId] = rowComp;
+            this[rowCompsName][instanceId] = rowComp;
```
StephenCooper: Request: Can we ensure this way of accessing is not a performance regression if this is on a hot path.

packages/ag-grid-react/src/reactUi/rows/rowContainerComp.tsx
```diff
@@ -1,48 +1,61 @@
...
 
     const eViewport = useRef<HTMLDivElement | null>(null);
     const eContainer = useRef<HTMLDivElement | null>(null);
-
+    const eSpanContainer = useRef<HTMLDivElement | null>(null);
     const rowCtrlsRef = useRef<RowCtrl[]>([]);
     const prevRowCtrlsRef = useRef<RowCtrl[]>([]);
     const [rowCtrlsOrdered, setRowCtrlsOrdered] = useState<RowCtrl[]>(() => []);
+
+    const isSpanning = useMemo(
```
StephenCooper: no need to useMemo here as its just a simple boolean value.

packages/ag-grid-community/src/theming/core/css/_general.css
```diff
@@ -92,6 +92,15 @@
     flex: 1 1 auto;
 }
 
+.ag-viewport {
+    position: relative;
```
StephenCooper, line 96: What is the reason for this change?

gportela85: @AndrewGlazier I see that this `ag-viewport` is a new container that was added to `rowContainerComp` but I don't see this reference in the React component. Do we have a different structure for React?

You also need to add these styles to `_common-structural.scss` or they won't work with the legacy themes.

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-custom/main.ts
AndrewGlazier: This was a cheap hack seeing as it's only temp (until the module is in AllCommunity...), it's here to prevent the example generator extracting the module logic for JS so the examples still get the module registered, it's stripped by all the examples _other than_ typescript

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-auto-height/main.ts
AndrewGlazier: This was a cheap hack seeing as it's only temp (until the module is in AllCommunity...), it's here to prevent the example generator extracting the module logic for JS so the examples still get the module registered, it's only duplicated in the final typescript example result

Open to another idea, or can add a proper exclusion in to the generator, but this seemed like a reasonable short term work-around for little cost

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-custom/main.ts
AndrewGlazier: See above

documentation/ag-grid-docs/src/content/docs/row-spanning/_examples/row-spanning-simple/main.ts
AndrewGlazier: See above

packages/ag-grid-community/src/context/context.ts
AndrewGlazier: This is performing a similar role as the `rowRenderer` so it shares that convention - I'll rename this to `spannedRowRenderer` which is more directly applicable

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
```diff
@@ -0,0 +1,215 @@
...
+import type { RowNode } from '../../entities/rowNode';
+import type { CellPosition } from '../../interfaces/iCellPosition';
+import type { IRowModel } from '../../interfaces/iRowModel';
+import type { RowPinnedType } from '../../interfaces/iRowNode';
+import type { ValueService } from '../../valueService/valueService';
+import { _normalisePinnedValue } from './spannedCellRenderer';
+
+export const _doesColumnSpan = (column: AgColumn) => {
+    // todo replace with new row spanning
+    return column.getColDef().rowSpan;
```
peterjrreynolds: `colDef` is now public on `AgColumn`

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
```diff
@@ -0,0 +1,215 @@
...
+    public readonly cellSpan = true;
+
+    private spannedNodes: Set<RowNode>;
+    private lastNode: RowNode;
+
+    constructor(
+        public readonly col: AgColumn,
+        public readonly firstNode: RowNode
+    ) {
+        this.firstNode = firstNode;
```
peterjrreynolds: unneeded

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
```diff
@@ -0,0 +1,215 @@
...
+    public getSpannedNodes() {
+        return this.spannedNodes;
+    }
+
+    public getFirstNode(): RowNode {
+        return this.firstNode;
+    }
+
+    public getLastNode(): RowNode {
+        return this.lastNode;
```
peterjrreynolds, line 36: For the sake of bundle size, can we make all these public instead?

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * Belongs to a column, when cells are to be rendered they call back to this service with the values.
+ * This service determines if the cell should instead be replaced with a spanning cell, in which case the cell is
+ * stretched over multiple rows.
+ *
+ * Only create if spanning is enabled for this column.
+ */
+export class RowSpanCache extends BeanStub {
+    private readonly column: AgColumn;
```
peterjrreynolds: move to constructor

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    // cannot use getAllRowCtrls as it returns this services row ctrls.
+    private getAllRelevantRowControls(ctrlsKey: 'top' | 'bottom' | 'center'): RowCtrl[] {
+        switch (ctrlsKey) {
+            case 'top':
+                return this.beans.rowRenderer.topRowCtrls;
+            case 'bottom':
+                return this.beans.rowRenderer.bottomRowCtrls;
+            case 'center':
+                return this.beans.rowRenderer.allRowCtrls;
```
peterjrreynolds: bundle size

packages/ag-grid-community/src/entities/agColumn.ts
AndrewGlazier: No, the next line changes the contents of `this.colDef`

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
AndrewGlazier: Done

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
AndrewGlazier: This isn't a hot path - can change this but it'll be at the cost of higher bundle size I imagine

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
AndrewGlazier: For what reason? If it's performance based, then this is again not a hot path, and this is the more bundle-size efficient approach

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
AndrewGlazier: Address this in the higher up comment, I've renamed it to `spannedRowRenderer`, I can't think of a more applicable name without also renaming `rowRenderer`

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
AndrewGlazier: No optional chaining of left side of an assignment operator

packages/ag-grid-react/src/reactUi/rows/rowContainerComp.tsx
AndrewGlazier: Done

packages/ag-grid-community/src/theming/core/css/_general.css
AndrewGlazier: @StephenCooper explicitly assigning position: relative creates a context for descendant elements with position: absolute, this ensures that the spanning container is positioned relative to the viewport

@gportela85 I've added this to the react component and _common-structural thanks

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
AndrewGlazier: done

packages/ag-grid-community/src/interfaces/iServerSideRowModel.ts
AndrewGlazier: It's not in our public interface, there's a lot of stuff on rowModel that isn't in our public interface - I don't think this is breaking?

(Note this isn't in IRowModel, but IServerSideRowModel/IClientSideRowModel)

packages/ag-grid-community/src/rendering/row/rowAutoHeightService.ts
AndrewGlazier: done

packages/ag-grid-community/src/rendering/rowRenderer.ts
AndrewGlazier: done

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
AndrewGlazier: done

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
AndrewGlazier: done

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
AndrewGlazier: I'd rather keep this.lastNode private on the cache to avoid incorrectly changing this without appending to the list, but have done so for `firstNode` and `spannedNodes`

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
AndrewGlazier: done

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
AndrewGlazier: done

packages/ag-grid-community/src/entities/colDef.ts
AndrewGlazier: done

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
AndrewGlazier: I don't believe this is a hot path, but rather than testing it I've made some changes to remove concern

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -39,19 +50,31 @@ export class RowContainerComp extends Component {
...
-            setContainerWidth: (width) => (this.eContainer.style.width = width),
-            setOffsetTop: (offset) => (this.eContainer.style.transform = `translateY(${offset})`),
+            setContainerWidth: (width) => {
+                this.eContainer.style.width = width;
+                if (this.eSpannedContainer) {
+                    this.eSpannedContainer.style.width = width;
+                }
+            },
+            setOffsetTop: (offset) => {
+                this.eContainer.style.transform = `translateY(${offset})`;
```
StephenCooper: Might save a few bytes. 

```
const {eContainer, eSpannedContainer} = this;
const top = `translateY(${offset})`;
eContainer.style.transform = top;
if (eSpannedContainer) {
  eSpannedContainer.style.transform = top;
}
```

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
```diff
@@ -0,0 +1,210 @@
...
+    public wireBeans(beans: BeanCollection) {
+        this.rowModel = beans.rowModel;
+        this.valueService = beans.valueSvc;
+    }
+
+    public buildCache(pinned: 'top' | 'center' | 'bottom'): void {
+        const newMap = new Map();
+
+        const isFullWidthCellFunc = this.beans.gos.getCallback('isFullWidthRow');
+        const equalsFnc = this.column.getColDef().equals;
```
StephenCooper: Could we deconstruct from `this` within this function, also don't duplicate getColDef call.

packages/ag-grid-community/src/rendering/spanning/spannedCellRenderer.ts
StephenCooper: Is this not a hot path when scrolling the grid with spanned cells?

AndrewGlazier: It is on the scroll path, but this happening three times per scroll event is incredibly negligible still. If this was happening inside of a sort, or even potentially large iterator I'd consider the impact more carefully

packages/ag-grid-community/src/rendering/spanning/rowSpanCache.ts
AndrewGlazier: done

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
AndrewGlazier: done

packages/ag-grid-community/src/interfaces/iServerSideRowModel.ts
AndrewGlazier: recovered & deprecated

packages/ag-grid-community/src/context/context.ts
AndrewGlazier: done

packages/ag-grid-community/src/allCommunityModule.ts
```diff
@@ -90,5 +91,6 @@ export const AllCommunityModule: _ModuleWithoutApi = {
         RowAutoHeightModule,
         DragAndDropModule,
         ClientSideRowModelApiModule,
+        // CellSpanModule,
```
peterjrreynolds: This should be added now

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -1021,6 +1021,12 @@ export interface GridOptions<TData = any> {
      * @initial
      */
     ensureDomOrder?: boolean;
+    /**
+     * When `true`, enables the cell span feature allowing for the use of the `colDef.spanRows` property.
+     * @default false
+     * @initial
+     */
+    enableCellSpan?: boolean;
```
peterjrreynolds, line 1029: Can also add a grid option validation for this to check the module is present

StephenCooper: Do we warn them if they have set rowSpan on a column but not set the grid option?

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
```diff
@@ -274,7 +307,7 @@ export class RowContainerCtrl extends BeanStub implements ScrollPartner {
         this.eContainer = eContainer;
         this.eViewport = eViewport;
 
-        this.createManagedBean(new RowContainerEventsFeature(this.eContainer));
+        this.createManagedBean(new RowContainerEventsFeature(this.eViewport ?? this.eContainer));
```
StephenCooper, line 310: Should the type of eViewport reflect that it can be undefined? Am I right in assuming this is undefined when not doing row spanning?

StephenCooper (approved):
I don't have any more feedback

packages/ag-grid-community/src/entities/gridOptions.ts
AndrewGlazier: @StephenCooper yes this is already in colDefValidations

done

packages/ag-grid-community/src/allCommunityModule.ts
AndrewGlazier: done

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
AndrewGlazier: Not part of the scope of this ticket, the typing behaviour around eViewport has not changed - you're welcome to pick up this type fixing work however I don't have scope to pick this up

peterjrreynolds (approved):

PR #9694, opened 2025-01-20
AG-13896 - Hide error pages from search engines

taktran:


documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
```diff
@@ -108,10 +115,19 @@ const migrationUrl = pathJoin(
                     </Warning>
                 )
             }
+
             <div class={styles.pageSections}>
                 <h2 id={fullErrorTextHeading.slug}>{fullErrorTextHeading.text}</h2>
-                <pre id="errorCodeText" style="white-space: pre-wrap;">{defaultErrorText}</pre>
-                <Note>To show this error text in the dev console, import the <code>ValidationModule</code>.</Note>
+                <div class="missing-params-message">
```
Comment (StephenCooper) [inline documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro:121]: Is this always shown? Should it not be optional?

Comment (taktran) [reply documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro:121]: It's hidden by default using css: https://github.com/ag-grid/ag-grid/pull/9694/files#diff-1b01847949957cfaf0a22048342287dd8d5a4aef87a28457dada87b31aaab6d9R165-R167

I did it this way because it's an astro file, and it allows it to not have to go through a react layer

StephenCooper (approved):
LGTM

peterjrreynolds (approved):

documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
```diff
@@ -141,6 +157,16 @@ const migrationUrl = pathJoin(
...
+<style>
+    #errorCodeText {
+        white-space: pre-wrap;
+    }
+
+    .missing-params-message {
+        display: none;
+    }
+</style>
+
```
Comment (markdurrant) [inline documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro]: Nitpick: I think I prefer having all styles in `*.module.scss` files. Makes things like searching for classes a bit more consistent. This file is already importing `docs.module.scss` so maybe add styles there?

markdurrant (approved):
Generally looks good. One Nitpick on `<style>` tags.

PR #9695, opened 2025-01-20
AG-13898 Just sanitize once and store on column

StephenCooper:


packages/ag-grid-community/src/rendering/row/rowComp.ts
```diff
@@ -92,9 +92,7 @@ export class RowComp extends Component {
             }
         });
 
-        const cellCompsToRemove = Object.values(cellsToRemove).filter((cellComp) => cellComp != null);
```
StephenCooper, line 95: null value handling is already done in the destroyCells method so no need to filter the nulls out before passing the array.

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -545,6 +549,9 @@ export class AgColumn<TValue = any>
     public getId(): string {
         return this.colId;
     }
+    public getColIdSanitised(): string {
+        return this.colIdSanitised;
+    }
```
peterjrreynolds: can remove this and update to `public readonly colIdSanitised: string;`

PR #9706, opened 2025-01-22
AG-13778 - fixed issues while scrolling with ensureDomOrder

gportela85:
https://ag-grid.atlassian.net/browse/AG-13778

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
@@ -64,6 +64,7 @@ export class RowContainerComp extends Component {
         // destroys all row comps
         this.setRowCtrls([]);
         super.destroy();
+        this.lastPlacedElement = null;
     }
 
     private setRowCtrls(rowCtrls: RowCtrl[]): void {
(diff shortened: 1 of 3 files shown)


packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -110,8 +125,11 @@ export class RowContainerComp extends Component {
         this.lastPlacedElement = element;
     }
 
-    private ensureDomOrder(eRow: HTMLElement, rowCtrl: RowCtrl): void {
+    private ensureDomOrder(eRow: HTMLElement, isNew: boolean, rowCtrl: RowCtrl): void {
```
peterjrreynolds: Should rename this as it doesn't just do stuff when ensuring dom order anymore

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -72,32 +73,46 @@ export class RowContainerComp extends Component {
...
         };
 
-        rowCtrls.forEach(processRow);
+        const orderedRows: [eGui: HTMLElement, rowCtrl: RowCtrl, isNew: boolean][] = [];
+
+        for (const rowCtrl of rowCtrls) {
+            const [eGui, isNew = false] = processRow(rowCtrl) || [];
+
+            if (eGui) {
+                orderedRows.push([eGui, rowCtrl, isNew]);
```
AndrewGlazier: `orderedRows` looks like it is assuming that the `rowCtrls` array is ordered - I'm not sure this is always true?

Haven't dug all the way through but one example is in the `stickyRowFeature` we reverse the list depending on top/bottom, also in the `rowRenderer` we merge different lists of ctrls with no adherence to order (`this.allRowCtrls = [...liveList, ...zombieList, ...cachedList];`)

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -72,32 +73,46 @@ export class RowContainerComp extends Component {
 
         this.lastPlacedElement = null;
 
-        const processRow = (rowCon: RowCtrl) => {
+        const processRow = (rowCon: RowCtrl): [HTMLElement, boolean] | undefined => {
```
AndrewGlazier: I'm not fussy about this, but I'm surprised you haven't flattened this down now you've updated the loop to for(of)

packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerComp.ts
```diff
@@ -72,32 +73,46 @@ export class RowContainerComp extends Component {
...
             const instanceId = rowCon.instanceId;
             const existingRowComp = oldRows[instanceId];
 
             if (existingRowComp) {
                 this.rowComps[instanceId] = existingRowComp;
                 delete oldRows[instanceId];
-                this.ensureDomOrder(existingRowComp.getGui(), rowCon);
```
AndrewGlazier, line 82: I'm not sure I understand how the feature works, it looks like we always add the row after the previous row... but that wouldn't work right? Surely we need to be doing a lookup for the comp in this.rowComps[] with the preceding row index and inserting after that? Not just basing it on `lastInsertedNode`?

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -477,12 +477,17 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
             map: {},
         };
 
-        const addCell = (colInstanceId: ColumnInstanceId, cellCtrl: CellCtrl) => {
-            res.list.push(cellCtrl);
+        const addCell = (colInstanceId: ColumnInstanceId, cellCtrl: CellCtrl, index?: number) => {
+            if (index != null) {
+                res.list.splice(index, 0, cellCtrl);
```
AndrewGlazier, line 482: Splice isn't exactly fast, are we ok with this in scrolling?

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -491,24 +496,34 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
...
+                continue;
             }
 
             prevCellCtrl.destroy();
-        });
+        }
+
+        if (colsFromPrev.length) {
+            for (const [colInstanceId, cellCtrl] of colsFromPrev) {
+                const index = res.list.findIndex((ctrl) => ctrl.column.getLeft()! > cellCtrl.column.getLeft()!);
```
AndrewGlazier, line 522: Wondering if we might be better just sorting the list afterwards before returning instead of so many findIndex/splicing while building the array - is this only for ensureDomOrder or do we always need to do this?

PR #9729, opened 2025-01-27
AG-13144 - Update migration page to be more consistent

taktran:


external/ag-website-shared/src/components/expanding-section/ExpandingSection.astro
```diff
@@ -121,8 +121,18 @@ const { headerText, isOpen = false } = Astro.props as Props;
...
+            const isLargeContent = this.#contentHeight > distanceFromBottom;
+            const contentHeight = isLargeContent ? distanceFromBottom : this.#contentHeight;
+
+            if (isLargeContent) {
+                // Set content to be less than the content viewport
+                this.#setCSSProp('--content-transition-timing', `0s`);
+                this.#setCSSProp('--content-height', `${contentHeight + 2}px`);
+            }
+
+            const transitionTiming = clamp(this.#contentHeight / 2500, 0.25, 1);
```
markdurrant: This we also want to set `transitionTiming` based on `isLargeContent`. If `isLargeContent` then the timing should be based on `distanceFromBottom`. Currently we have different animation speeds for expanding and collapsing.

markdurrant (changes requested):

markdurrant (approved):
Everything looking great now. Animation timing is 👌!

PR #9742, opened 2025-01-31
Enable nx batching by default.

alantreadway:
Apply Charts changes from https://github.com/ag-grid/ag-charts/pull/3424 to Grid Nx build process.

Before:
```
$ nx clean && nx reset
$ time nx dev
...
79.81s user 6.89s system 190% cpu 45.402 total
```

After:
```
$ nx clean && nx reset
$ time nx dev
...
59.84s user 4.84s system 189% cpu 34.188 total
```


packages/ag-grid-enterprise/tsconfig.lib.json
alantreadway, line 1: Will revert

packages/ag-grid-enterprise/tsconfig.lib.json
```diff
@@ -3,8 +3,8 @@
   "compilerOptions": {
     "types": [],
     "baseUrl": "src",
-    "module": "node16",
-    "moduleResolution": "node16"
+    "module": "es2020",
+    "moduleResolution": "bundler"
```
seanlandsman: We should apply the same thing to community - community and enterprise should always be in lockstep

What are the output implications - if any - to switching to these values?

patches/@nx+jest+20.3.1.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/@nx/jest/executors.json b/node_modules/@nx/jest/executors.json
+index 136c230..c0ea260 100644
+--- a/node_modules/@nx/jest/executors.json
++++ b/node_modules/@nx/jest/executors.json
+@@ -2,7 +2,6 @@
+   "executors": {
+     "jest": {
+       "implementation": "./src/executors/jest/jest.impl",
+-      "batchImplementation": "./src/executors/jest/jest.impl#batchJest",
```
seanlandsman, line 9: Do we still need to patch? Even after the upgrade?

utilities/all/project.json
```diff
@@ -44,31 +47,35 @@
...
         "production": {}
       }
     },
     "pack:extract": {
       "executor": "nx:noop",
+      "dependsOn": ["^pack:extract"],
       "inputs": [],
       "outputs": [],
-      "dependsOn": ["^pack:extract"],
+      "cache": false,
```
seanlandsman, line 75: Why was this necessary (just curious)?

seanlandsman (changes requested):
Changes to community tsconfig?

taktran (approved):
NIce 👍🏼 LGTM

utilities/all/project.json
alantreadway: It's not strictly needed, but I was just errring on the side of paranoia, `nx:noop` tasks should never be cached or need caching since they don't do anything.

seanlandsman (approved):

packages/ag-grid-enterprise/tsconfig.lib.json
alantreadway: Discussed offline, reverted for now.

patches/@nx+jest+20.3.1.patch
alantreadway: Yes, until Nx allow selective batching, this is the simplest thing for us to do unfortunately (force off batching support in the task types that don't work).
https://github.com/nrwl/nx/discussions/29729

PR #9748, opened 2025-02-03
AG-13855 disable transactions for tree data with children

SalvatorePreviti:
Display a warning when executing a transaction against tree data with "children" hierarchical data

Tree data with hierarchical data cannot support transactions.


packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -609,6 +609,7 @@ export const AG_GRID_ERRORS = {
         'As of v33.1, using "keyCreator" with the Rich Select Editor has been deprecated. It now requires the "formatValue" callback to convert complex data to strings.' as const,
     267: () =>
         'Detail grids can not use a different theme to the master grid, the `theme` detail grid option will be ignored.' as const,
+    268: () => "Transactions aren't supported with tree data when providing hierarchical data" as const,
```
AndrewGlazier: Can we change this to 
`Transactions aren't supported with tree data when using treeDataChildrenField` (all TreeData is hierarchical)

PR #9750, opened 2025-02-03
AG-13970 - fix multi filter menu issue

peterjrreynolds:


packages/ag-grid-community/src/widgets/tabGuardFeature.ts
```diff
@@ -126,6 +126,13 @@ export class TabGuardFeature extends BeanStub {
         this.addTabGuards(...tabGuards);
     }
 
+    public removeTabGuards(): void {
+        const { comp, eTopGuard, eBottomGuard } = this;
+        const parent = comp.getFocusableElement();
+        parent.removeChild(eTopGuard);
+        parent.removeChild(eBottomGuard);
+    }
+
```
gportela85: Shouldn't we call `removeTabGuards` when destroying this class? This way, we wouldn't have to call it manually in the `agMenuPanel` class and this method could be changed to private.

PR #9759, opened 2025-02-04
AG-5211 Initial master detail selection

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-5211


packages/ag-grid-community/src/api/gridApi.ts
```diff
@@ -175,6 +175,15 @@ export interface _RowSelectionGridApi<TData = any> {
 
     /** Returns an unsorted list of selected rows (i.e. row data that you provided). */
     getSelectedRows(): TData[];
+
+    /**
+     * Returns true if all rows have been selected, null if some rows have been selected, and false otherwise.
+     *
+     * Intended for internal use only. This is not a supported API and may change at any time.
+     *
+     * @internal
```
AndrewGlazier: I'm not sure if we should prefix `getSelectAllState` with `_` as it's internal, not sure if we have convention here but it might prevent auto completes offering it quite as often? @StephenCooper probaby knows better if we have a convention/preference here

packages/ag-grid-community/src/gridOptionsUtils.ts
```diff
@@ -553,8 +554,13 @@ export function _getIsRowSelectable(gos: GridOptionsService): IsRowSelectable |
...
 }
 
-export function _getRowSelectionMode(gos: GridOptionsService): RowSelectionMode | undefined {
-    const selection = gos.get('rowSelection');
+export function _getRowSelectionMode(gridOptions: GridOptions): RowSelectionMode | undefined;
+export function _getRowSelectionMode(gos: GridOptionsService): RowSelectionMode | undefined;
+export function _getRowSelectionMode(arg: object): RowSelectionMode | undefined {
+    const selection =
+        'beanName' in arg && arg.beanName === 'gos'
+            ? (arg as GridOptionsService).get('rowSelection')
```
AndrewGlazier, line 562: Minor thing, do we need the casting if the ternary is correctly inferring?

packages/ag-grid-community/src/api/gridApi.ts
StephenCooper: This method should not be on the gridApi at all. This should just contain public methods. 

So I assume this will require some changes elsewhere to avoid depending on the actual gridApi

eliasmalik: Removed

PR #9762, opened 2025-02-04
AG-13938 Perfomance of params handling

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-13938


packages/ag-grid-community/src/gridOptionsService.ts
```diff
@@ -247,13 +247,7 @@ export class GridOptionsService extends BeanStub implements NamedBean {
         return this.domDataKey;
     }
 
-    public getGridCommonParams<TData = any, TContext = any>(): AgGridCommon<TData, TContext> {
```
StephenCooper, line 250: Not required. Can call addGridCommonParams with an {} instead.

PR #9779, opened 2025-02-06
AG-12914 getting started fixes

JamesSwinton:
Content improvements & build fixes


documentation/ag-grid-docs/src/content/docs/license-install/index.mdoc
```diff
@@ -32,3 +29,11 @@ Validate your licence key and configure your application. See sample code and ex
 
 {% metaTag tags=["LicenseManager"] /%}
 {% licenseSetup /%}
+
+{% if isFramework("react") %}
+
+## Next.js
+
+If you're using Next.js, we have a [blog](https://blog.ag-grid.com/using-ag-grid-with-next-js-to-build-a-react-table/#how-to-set-up-a-license-key-for-enterprise) that details how to get you up and running, as well as how to set the license key.
```
taktran: That link doesn't go to any anchor link. I'm thinking it should be:

The this hash

```suggestion
If you're using Next.js, we have a [blog](https://blog.ag-grid.com/using-ag-grid-with-next-js-to-build-a-react-table/#how-to-set-up-a-license-key-for-enterprise) that details how to get you up and running, as well as how to set the license key.
```

or the root page

```suggestion
If you're using Next.js, we have a [blog](https://blog.ag-grid.com/using-ag-grid-with-next-js-to-build-a-react-table) that details how to get you up and running, as well as how to set the license key.
```

documentation/ag-grid-docs/src/content/docs-nav/docsNav.json
```diff
@@ -24,8 +24,8 @@
                     "children": [
                         {
                             "type": "item",
-                            "title": "NPM Installation",
-                            "path": "npm-installation"
+                            "title": "Installation",
+                            "path": "installation"
```
taktran, line 28: You might want to add a redirect for `npm-installation` pages to go to `installation`

taktran (approved):
Generally looks good, just some minor comments

documentation/ag-grid-docs/src/content/docs-nav/docsNav.json
JamesSwinton: I don't think this is needed as I created the `npm-installation` page, but now reverted to just `installation` so there shouldn't be any links requiring a redirect.

PR #9792, opened 2025-02-07
AG-9010 - Split cells when data changes

AndrewGlazier:


packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -316,6 +316,7 @@ export class RowNode<TData = any>
         const event: DataChangedEvent<TData> = this.createDataChangedEvent(data, oldData, update);
 
         this.__localEventService?.dispatchEvent(event);
+        this.beans.rowSpanSvc?.onRowDataUpdated(this);
```
gportela85: I think we should do a 
`const { rowSpanSvc, valueCache, selectionSvc } = beans;`

here

AndrewGlazier: Addressed

PR #9848, opened 2025-02-10
RTI-2533 - fix rich select async with null value

peterjrreynolds:


packages/ag-grid-enterprise/src/widgets/agRichSelect.ts
```diff
@@ -329,8 +329,8 @@ export class AgRichSelect<TValue = any> extends AgPickerField<
         }
 
         let idx = null;
+        listComponent.selectValue(this.value);
```
peterjrreynolds, line 332: This already contains handling for `null` value, but also runs the logic to setup loading

PR #9870, opened 2025-02-11
RTI-2538 memoize tree data callback getDataPath, and add autoSizeStrategy rowNumbers as useMemo in snippet gen

SalvatorePreviti:
- memoize tree data callback getDataPath
- add autoSizeStrategy and rowNumbers as useMemo in snippet gen
- add a new simple detector to skip the use of useMemo for JS literals that don't need to be memoized


StephenCooper (changes requested):
Can you updated the tests to have examples of these?

StephenCooper (approved):

PR #9880, opened 2025-02-11
AG-11318 - Column hover highlight reactivity

AndrewGlazier:


packages/ag-grid-community/src/columns/columnHover/hoverFeature.ts
```diff
@@ -9,13 +9,32 @@ export class HoverFeature extends BeanStub {
...
+    }
+
+    private enableFeature = (enabled: boolean) => {
+        if (enabled) {
+            this.element.addEventListener('mouseover', this.onMouseOver);
+            this.element.addEventListener('mouseout', this.onMouseOut);
+            return;
         }
+        this.element.removeEventListener('mouseover', this.onMouseOver);
+        this.element.removeEventListener('mouseout', this.onMouseOut);
```
peterjrreynolds: bundle size - `this.element`

gportela85:
I'd have the HoverFeature class be something like the following, so we can leverage the functions returned from addManagedEventLister.

```
export class HoverFeature extends BeanStub {
    constructor(
        private readonly columns: AgColumn[],
        private readonly element: HTMLElement
    ) {
        super();
    }

    private listeners: (() => void)[] = [];

        this.addManagedPropertyListener('columnHoverHighlight', ({ currentValue }) => {
            this.enableFeature(currentValue);
        });
        this.enableFeature();
    }

    private enableFeature = (enabled?: boolean) => {
        const { beans, gos, element, columns } = this;
        const colHover = beans.colHover!
        const active = enabled ?? !!gos.get('columnHoverHighlight');

        if (active) {
            this.listeners = this.addManagedElementListeners(element, {
                mouseover: colHover.setMouseOver.bind(colHover, columns),
                mouseout: colHover.setMouseOver.bind(colHover, columns)
            });
        } else {
            this.listeners.forEach(fn => fn());
            this.listeners = [];
        }
    };

    public override destroy() {
        super.destroy();
        (this.listeners as any) = null;
    }
}
```

packages/ag-grid-community/src/columns/columnHover/hoverFeature.ts
AndrewGlazier: no longer relevant

PR #9881, opened 2025-02-11
AG-8812 - IRM onFirstDataRendered fix

AndrewGlazier:


documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -159,8 +159,6 @@ This can also be performed on demand via the following API methods:
 
 {% apiDocumentation source="grid-api/api.json" section="columnSizing" names=["autoSizeColumns", "autoSizeAllColumns"] /%}
 
-Note that using `autoSizeStrategy` to fit cell contents only works for the Client-Side Row Model and Server-Side Row Model, but the API methods work for all row models.
```
peterjrreynolds, line 162: Still doesn't work for viewport

PR #9884, opened 2025-02-11
AG-14040 - Fix range overwriting pinned border

AndrewGlazier:


packages/ag-grid-enterprise/src/rangeSelection/cellRangeFeature.ts
```diff
@@ -146,6 +146,14 @@ export class CellRangeFeature implements ICellRangeFeature {
...
         left: boolean;
     } {
+        if (this.rangeCount === 0) {
+            return {
+                top: false,
+                right: false,
+                bottom: false,
+                left: false,
+            };
+        }
```
gportela85: This fix doesn't seem quite right, we are checking only if no ranges are present but if there is a range that doesn't select this cell, the columns would still get the wrong style.

The correct fix would be to move the following block:
```
const ranges = rangeSvc.getCellRanges().filter((range) => rangeSvc.isCellInSpecificRange(cellPosition, range));
```
to be below:
```
const thisCol = cellPosition.column as AgColumn;
```

and then check
```
if (!ranges.length) { 
    return { top, right, bottom, left }; 
}
```

PR #9888, opened 2025-02-11
AG-14090 Swap innerHtml for textContent

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-14090


packages/ag-grid-enterprise/src/statusBar/providedPanels/agNameValue.ts
```diff
@@ -20,8 +20,8 @@ export class AgNameValue extends Component {
         this.eLabel.innerHTML = localeTextFunc(key, defaultValue);
     }
 
-    public setValue(value: any): void {
-        this.eValue.innerHTML = value;
+    public setValue(value: string): void {
```
StephenCooper, line 23: This is only every called with a string.

PR #9895, opened 2025-02-12
feat: add named KeyCreatorFunc type

### LINKED ISSUE #9894: feat: add named type for KeyCreator
I think it would be nice to add a named type for the KeyCreator function. The current type is `(params: KeyCreatorParams<TData, TValue>) => string`. Several other similar functions in col def, such as valueFormatter, valueGetter, valueSetter, valueParser, cellEditorSelector have a named type (`ValueFormatterFunc` and so on).

See relevant type definitions here: https://github.com/ag-grid/ag-grid/blob/3f15d639a5f496a788b9178ab74ad70cc258298b/packages/ag-grid-community/src/entities/colDef.ts#L943-L955

braaar:
fixes #9894


StephenCooper:
Will be completed as part of wider type improvements as part of AG-12960

PR #9896, opened 2025-02-12
AG-5211 Avoid calling methods on destroyed grids

eliasmalik:


packages/ag-grid-enterprise/src/masterDetail/detailCellRendererCtrl.ts
```diff
@@ -153,24 +153,25 @@ export class DetailCellRendererCtrl extends BeanStub implements IDetailCellRende
...
             // the gridInfo can be stale if a refresh happens and
             // a new row is created before the old one is destroyed.
             if (rowNode.detailGridInfo !== gridInfo) {
                 return;
             }
             if (!masterGridApi.isDestroyed()) {
+                masterGridApi.removeEventListener('rowSelected', onMasterRowSelected);
                 masterGridApi.removeDetailGridInfo(rowId); // unregister from api
             }
+            !api.isDestroyed() && api.removeEventListener('selectionChanged', onDetailSelectionChanged);
```
SalvatorePreviti: I'd prefer an if, minor stylistic comment - but also, I think that isDestroyed might be always true here?

packages/ag-grid-enterprise/src/masterDetail/detailCellRendererCtrl.ts
```diff
@@ -153,24 +153,25 @@ export class DetailCellRendererCtrl extends BeanStub implements IDetailCellRende
...
-            api.removeEventListener('selectionChanged', onDetailSelectionChanged);
-            masterGridApi.removeEventListener('rowSelected', onMasterRowSelected);
-
             // the gridInfo can be stale if a refresh happens and
             // a new row is created before the old one is destroyed.
             if (rowNode.detailGridInfo !== gridInfo) {
                 return;
             }
             if (!masterGridApi.isDestroyed()) {
+                masterGridApi.removeEventListener('rowSelected', onMasterRowSelected);
```
SalvatorePreviti: does it need to be done? is not automatically executed by the bean base class on destruction?

packages/ag-grid-enterprise/src/masterDetail/detailCellRendererCtrl.ts
SalvatorePreviti: does it need to be done? is not automatically executed by the bean base class on destruction?

packages/ag-grid-enterprise/src/masterDetail/detailCellRendererCtrl.ts
eliasmalik: Ah I didn't realise they were also managed. I've removed these lines

SalvatorePreviti (approved):

PR #9898, opened 2025-02-12
AG-14113 - Fix server side/client side render of react components in darkmode

taktran:
## Changes

* Use [suppressHydrationWarning](https://18.react.dev/reference/react-dom/client/hydrateRoot#suppressing-unavoidable-hydration-mismatch-errors) for `DarkModeToggle.tsx` - it's only text that changes on darkmode change, and the text is not visible
* Delete grid `imageCaption` component and use shared one
* Update shared `imageCaption` as Astro component to speed up rendering of darkmode images


markdurrant (approved):
LGTM 👍

PR #9907, opened 2025-02-13
AG-14024 - agSkeletonCellRenderer should not be used with RowNumbers column

gportela85:
https://ag-grid.atlassian.net/browse/AG-14024

### DIFF
--- a/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
+++ b/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
@@ -269,13 +269,18 @@ export class CellCtrl extends BeanStub {
 
         // if node is stub, and no group data for this node (groupSelectsChildren can populate group data)
         const isSsrmLoading = rowNode.stub && rowNode.groupData?.[column.getId()] == null;
-        if (isSsrmLoading) {
+        const colDef = column.getColDef();
+        const isCellRenderer = this.isCellRenderer();
+
+        if (isSsrmLoading || isCellRenderer) {
             const params = this.createCellRendererParams();
-            compDetails = _getLoadingCellRendererDetails(userCompFactory, column.getColDef(), params);
-        } else if (this.isCellRenderer()) {
-            const params = this.createCellRendererParams();
-            compDetails = _getCellRendererDetails(userCompFactory, column.getColDef(), params);
+            if (!isSsrmLoading || isRowNumberCol(column)) {
+                compDetails = _getCellRendererDetails(userCompFactory, colDef, params);
+            } else {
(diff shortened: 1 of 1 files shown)


packages/ag-grid-enterprise/src/cellRenderers/loadingCellRenderer.ts
```diff
@@ -13,7 +18,47 @@ export class LoadingCellRenderer extends Component implements ILoadingCellRender
...
+        if (isEmbedFullWidthRows && (!pinned || isLeftPinned === isRtl)) {
+            return;
+        }
+
+        const eGui = this.getGui();
+        const propSuffix = gos.get('enableRtl') ? 'right' : 'left';
+        eGui.style.setProperty(`padding-${propSuffix}`, '0');
+        const cell = rowNumbersSvc?.getPlaceholderCellForNode(node);
+
+        if (cell) {
```
peterjrreynolds: Don't need this check or the `rowNumbersSvc?` above, as `rowNumbersSvc` must be defined at this point

packages/ag-grid-enterprise/src/cellRenderers/loadingCellRenderer.ts
```diff
@@ -13,7 +18,47 @@ export class LoadingCellRenderer extends Component implements ILoadingCellRender
...
-        params.node.failedLoad ? this.setupFailed() : this.setupLoading();
+        const { node, pinned } = params;
+        node.failedLoad ? this.setupFailed() : this.setupLoading();
+
+        if (this.gos.get('rowNumbers')) {
+            this.injectRowNumbersAtStart(node, pinned);
+        }
+    }
+
+    private injectRowNumbersAtStart(node: IRowNode, pinned?: ColumnPinnedType): void {
```
peterjrreynolds: SSRM is more likely to be used than row numbers, so ideally the majority of the contents of this method would live in the row numbers service so as to offer better bundle size optimisation.

gportela85: The majority of the contents have been isolated to `const cell = rowNumbersSvc.getPlaceholderCellForNode(node);`

The part that was left here, is related to the Cell Renderer itself (eg. how to embed the extra content when using RTL, or embedFullWidthRows, removing padding to the eGui of the renderer)
Also, the listener to change the width of the Row Number place holder should be removed from the element once the Loading Cell Renderer is destroyed.
I left everything else in the `rowNumbersSvc`.

peterjrreynolds (approved):

PR #9912, opened 2025-02-14
AG-14096 - Fix integrated charts automated example on mobile

taktran:
* Remove `pointer-events: none` that was preventing the script from interacting with the example from https://github.com/ag-grid/ag-charts/pull/2415
* Add `ignoreError` action property to keep going if there is an error. Needed for when the context menu or dropdown menu is open and user interacts with the page


markdurrant (approved):
LGTM 👍

PR #9922, opened 2025-02-17
AG-13752 - fall back to data-order when pivotComparator returns 0

StephM-W:
https://ag-grid.atlassian.net/browse/AG-13752

### DIFF
--- a/packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
+++ b/packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
@@ -49,7 +49,7 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
         });
     }
 
-    public createPivotColumnDefs(uniqueValues: any): PivotColDefServiceResult {
+    public createPivotColumnDefs(uniqueValues: Map<string, any>): PivotColDefServiceResult {
         // this is passed to the colModel, to configure the columns and groups we show
 
         const pivotColumnGroupDefs: (ColDef | ColGroupDef)[] = this.createPivotColumnsFromUniqueValues(uniqueValues);
@@ -86,7 +86,7 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
         };
     }
 
-    private createPivotColumnsFromUniqueValues(uniqueValues: any): (ColDef | ColGroupDef)[] {
+    private createPivotColumnsFromUniqueValues(uniqueValues: Map<string, any>): (ColDef | ColGroupDef)[] {
         const pivotColumns = this.pivotColsSvc?.columns ?? [];
         const maxDepth = pivotColumns.length;
 
@@ -102,7 +102,7 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
 
     private recursivelyBuildGroup(
         index: number,
-        uniqueValue: any,
+        uniqueValue: Map<string, any>,
         pivotKeys: string[],
         maxDepth: number,
         primaryPivotColumns: AgColumn[]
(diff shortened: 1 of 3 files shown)


packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -252,24 +260,27 @@ export class PivotStage extends BeanStub implements NamedBean, IRowNodeStage {
...
-            const result: Record<string, any> = {};
+        }
 
-            for (const key of Object.keys(mappedChildren)) {
-                result[key] = this.bucketChildren(mappedChildren[key], pivotColumns, pivotIndex + 1, uniqueValues[key]);
-            }
+        const result: Map<string, any> = new Map();
 
-            return result;
+        for (const key of Array.from(mappedChildren.keys())) {
```
AndrewGlazier: Array.from here is probably unnecessary and constructing this array will slow things down a little I think?

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -160,6 +170,23 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
         return groups;
     }
 
+    private addOrderedComparator(keys: string[], pivotComparator: (valueA: string, valueB: string) => number) {
```
AndrewGlazier: Is this wrapping necessary? As js Map is first out/first in I'd assume returning 0 would retain the order anyway so this is probably just slowing things down?

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -115,7 +115,17 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
 
         // sort by either user provided comparator, or our own one
         const primaryPivotColumnDefs = primaryPivotColumns[index].getColDef();
-        const comparator = this.headerNameComparator.bind(this, primaryPivotColumnDefs.pivotComparator);
+
+        let pivotComparator = primaryPivotColumnDefs.pivotComparator;
+
+        const keys = Array.from(uniqueValue.keys());
```
AndrewGlazier: I think this is just being used to iterate, would be better to just iterate the iterator as making the array would either do nothing at best (if the engine is smart enough to optimise) or slow things down at worst

AndrewGlazier (changes requested):

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
StephM-W: We can't easily tell whether the users are returning 0 _all the time_ to trigger data-order, or just some of the time to get  their bespoke ordering. Chaining seems to support both cases with low impact?

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
StephM-W: Agreed - in this function it's just iterated over, I'm moving this conversion to the comparator where it's also used for indexing the keys (instead of using indexOf), so it's only done when needed.

StephM-W: And removed the keys variable altogether to avoid confusion when the iterator's already been used

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
StephM-W: Agreed, removed

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
AndrewGlazier: I'm not entirely sure I understand;

Now that we batch data into a Map we can be sure that the underlying data isn't being sorted - originally it was being sorted because we were using objects, which sort numeric keys

Now we use the map, we can be sure that the leaf cols are being created in the order they appeared in the data as uniqueValue is no longer sorted due to being an object: (prev code ahead)
```
for (const key of Object.keys(uniqueValue)) {
    const newPivotKeys = [...pivotKeys, key];
    const colDef = this.createColDef(measureColumns[0], key, newPivotKeys);
    colDef.columnGroupShow = 'open';
    leafCols.push(colDef);
}
```

as we can now be assured that `uniqueValue` is in the correct order then returning "0" in a sort comparator means "this value is in the correct location already" and don't make extra steps to move it

This is why I don't think we need to wrap the comparator; as just returning 0 should be fine

FYI - your test case would need to provide a pivotComparator to catch the performance impact of this wrapper func

StephM-W: Even though we use a Map, the default non-pivotComparator codepath still sorts (by headerName) as that's the existing behaviour we don't want to change (see a couple of lines down where we `leafCols.sort(comparator)`).

viknesh-ai (approved):

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
StephM-W: As discussed, I've simplified the code and confirmed the behaviour remains as required

PR #9924, opened 2025-02-17
AG-119 - find

peterjrreynolds:


packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -318,6 +318,9 @@ export class RowNode<TData = any>
 
         this.__localEventService?.dispatchEvent(event);
         rowSpanSvc?.onRowDataUpdated(this);
+        // if you want to add new feature stuff here, consider whether it can listen
```
AndrewGlazier: Odd comment seems redundant? Would be better to just change the `rowSpanSvc` and establish the pattern properly, anyone looking to implement a feature now would just use the established event

packages/ag-grid-community/src/interfaces/iRowModel.ts
```diff
@@ -20,6 +20,7 @@ export interface IRowModel {
 
     getTopLevelRowCount(): number;
     getTopLevelRowDisplayedIndex(topLevelIndex: number): number;
+    getTopLevelIndexFromDisplayedIndex?(displayedIndex: number): number;
```
AndrewGlazier, line 23: As this has only been introduced to CSRM would be better on IClientSideRowModel

packages/ag-grid-community/src/pagination/paginationService.ts
```diff
@@ -83,31 +81,24 @@ export class PaginationService extends BeanStub implements NamedBean {
...
+        let adjustedIndex = index;
+        if (!this.paginateChildRows) {
+            adjustedIndex = this.beans.rowModel.getTopLevelIndexFromDisplayedIndex?.(index) ?? index;
+        }
+
+        this.goToPage(Math.floor(adjustedIndex / this.pageSize));
     }
 
-    public isRowInPage(row: RowPosition): boolean {
+    public isRowInPage(rowIndex: number): boolean {
```
AndrewGlazier, line 97: The existing places that used `isRowInPage` could potentially have pinned rows, while the original func doesn't work correctly for it would be safer to fix this func to return true for pinned, rather than eradicating any potential support for pinned from this type

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -275,6 +275,13 @@ export class CellCtrl extends BeanStub {
         } else if (this.isCellRenderer()) {
             const params = this.createCellRendererParams();
             compDetails = _getCellRendererDetails(userCompFactory, column.getColDef(), params);
+        } else if (beans.find?.isMatch(rowNode, column)) {
```
AndrewGlazier: find seems like a really obscure/abstract name, can we add Svc to it or something? I appreciate you're probably limited due to other features, but just `find` as a svc isn't clear

packages/ag-grid-enterprise/src/find/findApi.ts
```diff
@@ -0,0 +1,30 @@
+import type { BeanCollection, FindCellParams, FindCellValueParams, FindMatch, FindPart } from 'ag-grid-community';
+
+export function findNext(beans: BeanCollection): void {
```
AndrewGlazier, line 3: bundle size, all of these beans could be destructured inline in params

packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -0,0 +1,581 @@
+import type {
```
AndrewGlazier, line 1: There's a lot going on in this service that would benefit from some comments making it easier to navigate - I feel like our new services should be made as maintainable as possible

packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
```diff
@@ -466,6 +464,14 @@ export class GroupCellRendererCtrl extends BeanStub implements IGroupCellRendere
                 params
             );
         }
+
+        if (this.beans.find?.isMatch(params.node, params.column!)) {
+            return _getCellRendererDetails(
+                this.userCompFactory,
+                { ...params.colDef, cellRenderer: 'agFindCellRenderer' },
```
AndrewGlazier: Are you sure this should be colDef and not relatedColDef?

AndrewGlazier:
Does this feature not support spanned cells? Doesn't look like they've been accounted for so should add a warning if we're not supporting them

AndrewGlazier (changes requested):

packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -0,0 +1,581 @@
...
+
+        const refreshAndWipeActive = this.refresh.bind(this, false);
+        const refreshAndKeepActive = this.refresh.bind(this, true);
+        this.addManagedPropertyListeners(['findSearchValue', 'findOptions'], refreshAndWipeActive);
+        this.addManagedEventListeners({
+            modelUpdated: refreshAndKeepActive,
+            displayedColumnsChanged: refreshAndKeepActive,
+            pinnedRowDataChanged: refreshAndKeepActive,
+            cellValueChanged: refreshAndKeepActive,
+            rowNodeDataChanged: refreshAndKeepActive,
```
AndrewGlazier: Just realised this is being called here (rowNodeDataChanged)

This event will fire _a lot_ and looks like it does a lot too, note to solve the same problem in the rowSpanService I introduced a special debounce for just this case as batch data changes/transactions/immutable data changes will fire this event _loads_

packages/ag-grid-community/src/entities/rowNode.ts
peterjrreynolds: Updated as discussed.

packages/ag-grid-community/src/interfaces/iRowModel.ts
peterjrreynolds: We will need to implement it for SSRM for AG-14152. I haven't done it yet as find doesn't work for SSRM.

packages/ag-grid-community/src/pagination/paginationService.ts
peterjrreynolds: Discussed offline, will leave as is

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
peterjrreynolds: Changed to `findSvc`

packages/ag-grid-enterprise/src/find/findApi.ts
peterjrreynolds: Discussed offline, will leave as is

packages/ag-grid-enterprise/src/find/findService.ts
peterjrreynolds: Done

packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
peterjrreynolds: Should be `relatedColDef` - updated

packages/ag-grid-enterprise/src/find/findService.ts
peterjrreynolds: Added a debounce

PR #9928, opened 2025-02-18
AG-14147 update and reorder installation docs

JamesSwinton:
Remove `ag-grid-enterprise` import from JS Installation docs

Reorder Installation, Getting Started and README docs so that registering modules comes before importing.


StephenCooper (approved):
LGTM

PR #9931, opened 2025-02-18
AG-14062 - add date component to col def

peterjrreynolds:


packages/ag-grid-community/src/components/framework/userCompUtils.ts
```diff
@@ -250,9 +250,10 @@ export function _getFilterDetails(
 
 export function _getDateCompDetails(
     userCompFactory: UserComponentFactory,
+    def: ColDef,
     params: IDateParams
 ): UserCompDetails<IDateComp> | undefined {
-    return userCompFactory.getCompDetailsFromGridOptions(DateComponent, 'agDateInput', params, true);
+    return userCompFactory.getCompDetails(def, DateComponent, 'agDateInput', params, true);
```
StephenCooper, line 256: Will this still work with the previous approach?

peterjrreynolds: Yes

StephenCooper (approved):

PR #9970, opened 2025-02-26
AG-12794 - Tree data parent id implementation

SalvatorePreviti:
Implementation of treeDataParentIdField grouping strategy for tree data.

As we discussed, the other implementations and the selection logic for strategy and node manager can be refactored after this.

The GroupStage now selects between grouping or tree data parent id strategy, and every time there is a switch, we reset all the row nodes states.
In the future refactoring work we can move the getDataPath and part of the treeDataChildrenField implementation as grouping strategies as well.

The algorithmic approach used in the parent id tree data strategy is to keep the logic as simple and fast as possible reducing the complexity by not handling partial updates in a way too different as we handle full reload, with the same code and the same loop over all the nodes. The diff logic should be fast enough for most circumstances.
We have two main stages, first loop through all nodes and check if any child changed, and then a tree traversal to finalise the arrays and all the changes. This should be fast in the general case also for single updates, keeping implementation simple.
We handle the case of circular references and writing the warning for it separately and only in case there is a cycle.


packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        let newStrategy = oldStrategy;
+        const treeDataManagedByNodeManager = params.nodeManager!.treeData;
+        if (treeDataManagedByNodeManager) {
+            newStrategy = undefined;
+        } else {
+            const { changedRowNodes, changedProps } = params;
+            if (
+                !changedRowNodes ||
+                (changedProps &&
```
AndrewGlazier: This looks like it could be simplified & doesn't really need changedProps nor nodeManager passed down so we could keep the IRowNodeStage model clean, and not need nodeManager or changedParams to be added for this one use case i.e:

gridOptionsUtils: (Suggesting this goes in gridOptionsUtils so it can be used in the existing places we do this determination, as this logic is in a few places now
```
function _getTreeDataApproach(gos: GridOptionsService) {
   if (!gos.get('treeData')) {
      return null;
   } 

   if (gos.get('getDataPath')) {
      return 'dataPaths';
   }

   // etc
}
```

```
getStrategy() {
   const approach = _getTreeDataApproach(gos);
   if (approach === 'childrenField') {
         return treeDataParentIdStrategy;
   }

  return groupStrategy;
```

  The changedPath seems unnecessary here as you do a determination to see if the strategy changes anyway - which seems sufficient enough

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1053,29 +1054,30 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
-            } else {
-                const sibling: ClientSideRowModelRootNode = rootNode.sibling;
-                rootNode.childrenAfterGroup = rootNode.allLeafChildren;
-                if (sibling) {
-                    sibling.childrenAfterGroup = rootNode.childrenAfterGroup;
-                }
-                rootNode.updateHasChildren();
+
+        if (
+            !this.groupStage?.execute({
```
AndrewGlazier: Inlining this complicates the flow as it implies execute can return/have side effects; code is easier to read/follow with this not inline

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -137,7 +137,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
     private getNewNodeManager(): IClientSideNodeManager<any> {
         const { gos, beans } = this;
         let nodeManager: IClientSideNodeManager<any> | undefined;
-        if (gos.get('treeData')) {
+        if (gos.get('treeData') && !gos.get('treeDataParentIdField' as any)) {
```
AndrewGlazier: Need to add this to the interface so we don't need to use casting

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
```diff
@@ -0,0 +1,351 @@
...
+        // This ensures a simpler code and less complexity, and also that enough speed for the vast majority of cases.
+        // Consider that trying other approaches might be more complex and potentially not as fast,
+        // as the user can always move an entire subtree by changing a single parentId.
+        //
+        // To further reduce allocations, we use treeNodeFlags to store both temporary flags,
+        // the expanded initialized state and the future children count between the first loop and the tree traversal.
+        // This avoid the needs to create complex data structures to store temporary data or add more fields to the row nodes.
+
+        const rootNode: TreeRow<TData> = params.rowNode;
+        const nodeManager = params.nodeManager!;
```
AndrewGlazier: I don't think the stages should have access to the node manager - as this is just done for a node lookup we can still do this with `rootNode.allLeafChildren`

AndrewGlazier (changes requested):

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
```diff
@@ -0,0 +1,349 @@
...
+        sibling.allLeafChildren = row.allLeafChildren;
+        sibling.childrenAfterGroup = row.childrenAfterGroup;
+        sibling.childrenAfterAggFilter = row.childrenAfterAggFilter;
+        sibling.childrenAfterFilter = row.childrenAfterFilter;
+        sibling.childrenAfterSort = row.childrenAfterSort;
+    }
+};
+
+const updateAllLeafChildren = <TData>(
+    row: TreeRow<TData>,
```
gportela85, line 245: I haven't tested this, but couldn't this be written as

```
const updateAllLeafChildren = <TData>(
    row: TreeRow<TData>,
    allLeafChildren: TreeRow<TData>[],
    allLeafChildrenLen: number
): boolean => {
    let changed = allLeafChildren.length !== allLeafChildrenLen;

    if (changed) {
        row.allLeafChildren = allLeafChildrenLen
            ? (allLeafChildren === _EmptyArray ? new Array(allLeafChildrenLen) : allLeafChildren.slice(0, allLeafChildrenLen))
            : _EmptyArray;
    }

    let writeIdx = 0;
    for (const child of row.childrenAfterGroup!) {
        for (const leaf of child.allLeafChildren!) {
            if (changed || allLeafChildren[writeIdx] !== leaf) {
                allLeafChildren[writeIdx] = leaf;
                changed = true;
            }
            writeIdx++;
        }
    }

    return changed;
};
```

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: this would have big performance implications as we need to find the row by parentId, and we have already a way to find a node by id in the nodeManager in O(1) as lookup - maybe we can discuss other ways to handle this more elegantly without the need to loop through allLeafChildren in O(n) (that would become O(n*n) in that loop) and without the need to build another temporary map that would be heavy for single updates?

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: If you are ok with making it public straight away, sure!

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: Threee possible options I could think of:
1) define a different interface for each stage, instead of using a common interface, so we can pass just the fields we need and avoid using ? and ! where not needed
2) define a new interface IRowNodeLookup<TData> { getRowNode(id:string): RowNode<TData> | undefined } and pass this instead - better with option 1
3) use the getRow from ClientSideRowModel (also if that has also a check for group rows that is a bit heavy, is not too bad)

packages/ag-grid-community/src/context/context.ts
```diff
@@ -283,6 +283,8 @@ export interface CoreBeanCollection {
     sortStage?: IRowNodeStage;
     flattenStage?: IRowNodeStage;
     groupStage?: IRowNodeStage;
+    groupStrategy?: IRowGroupingStrategy;
```
AndrewGlazier: Do these need to be named beans? I'm pretty certain they're only used by the groupStage and don't need to be available anywhere else

SalvatorePreviti: So - the problem is where they are exposed and used - we need the GroupStage to be available both for tree data module and grouping module - and the shared things are in the RowHierarchy module.
If we put the strategies all inside the RowHierarchy module to keep them hidden this is going to make the split between tree data and grouping not possible, if someone wants to load just tree data OR grouping in the future if we fully spit them - at least this is what i discussed with @peterjrreynolds - so he might provide a better set of options here?

The named bean there gives the level of indirection needed if someone can load group module without tree data module or viceversa.

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: I currently opted for the option to define IClientSideNodeLookup interface, and IClientSideNodeManager inherit this interface. So we pass the node manager, but just a view with just one method, getRowNode(id: string): RowNode<TData> | undefined; - that should make things clearer and simpler for now and the future

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: I would not centralise this at the moment - I sugges to do this work of generalization, centralization and cleanup in the future refactoring PRs and not in this one

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
AndrewGlazier: This is not a hot path; we don't need to over optimise in this scenario - it's better to follow the established patterns and simply rebuild the map in this case - the performance implications are negligible

We optimise where optimisation is needed, and unless there's an evidenced scenario where this performance of this is unacceptable, we should not be passing the node manager through

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
AndrewGlazier: Why not? This change is introducing a repeat to this logic, it's good practice to not introduce repeated logic

packages/ag-grid-community/src/context/context.ts
AndrewGlazier: This is possible by using the dynamicBeans pattern, and this stops us polluting the global `beans` namespace

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Agreed, could you have another look? This will be further simplified in the next refactoring PRs when we remove most of the logic from the tree node managers

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: I rewrote the logic and should be simpler now overall - I suggest to move the deduplication (is not the only piece of code that is duplicated, there is duplicated logic as well for expanded state, group flag, haschildren and so on) on a subsequent refactoring to not touch too much in this feature PR

We have already AG-13709 in backlog

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: Passing the row node lookup looks clean enough to me now, pushed the changes. Could you have another look?
I do consider in this context this an hot path - this gets executed every time there is an update, including transactions and immutable data, even for a single row update - we should try to avoid O(n*n) or unnecessary allocations of big lookup maps, when we already have them available, during updates

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: Thank you, I rewrote it, and should be much simpler now, can you have another look?

packages/ag-grid-community/src/context/context.ts
SalvatorePreviti: It isn't possible to tree shake dynamic beans - if we put grouping strategy and tree data strategies inside the row hierarchy module, whenever the grouping stage is loaded it will include both tree data and grouping strategies in the build. Maybe we can have a quick chat with Pete?

peterjrreynolds: Dynamic beans would work here and allow tree-shaking. The dynamic beans would be defined on the tree data and grouping modules. The shared module would just refer to the dynamic bean by (string) name.

SalvatorePreviti: Spoke with pete - done as a named dynamic bean exposed by the modules, could you check again?

packages/ag-grid-community/src/context/context.ts
```diff
@@ -427,6 +429,8 @@ export type BeanName =
     | 'gridOptionsWrapper'
     | 'gridSerializer'
     | 'groupStage'
+    | 'groupStrategy'
```
AndrewGlazier: These strategies don't need to be part of this list anymore as they're dynamic beans - not named beans

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
AndrewGlazier: This is not a hot path - a hot path, for example is rendering and scrolling.

The impact benefit of this is negligible but it has a negative impact on the established patterns in the grid.

It is important to prioritise code patterns and maintainability, not to early optimise every line of code. If a user reports an issue with performance that we can trace back to this specific issue, we can revisit this in the future, for now please make the requested change

Note this doesn't need to be O(n^2), you can rebuild the map in scope here, but the stages should not have access to the nodeManager

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
AndrewGlazier: 13709 was an exception that was made, it is not the standard approach. PRs with needed changes should include the changes before they are merged

We should dedupe as we write code - please make the suggested change, you do not need to fix all other instances of duplication - but the one that is being introduced here should be fixed as it is introduced

packages/ag-grid-community/src/interfaces/iRowNodeStage.ts
```diff
@@ -16,6 +30,11 @@ export interface StageExecuteParams<TData = any> {
     afterColumnsChanged?: boolean;
 }
 
+export interface IRowGroupingStrategy<TData = any> extends Bean {
+    execute(params: StageExecuteParams<TData>): void;
+    deactivate?(): void;
```
AndrewGlazier: deactivate probably not needed anymore, if we are destroying beans properly we can use BeanStub.destroy func

packages/ag-grid-community/src/validation/rules/gridOptionsValidations.ts
```diff
@@ -412,12 +412,25 @@ const GRID_OPTION_VALIDATIONS: () => Validations<GridOptions> = () => {
                 const rowModel = options.rowModelType ?? 'clientSide';
                 switch (rowModel) {
                     case 'clientSide': {
-                        const { treeDataChildrenField, getDataPath } = options;
-                        if (!treeDataChildrenField && !getDataPath) {
-                            return "treeData requires either 'treeDataChildrenField' or 'getDataPath' in the clientSide row model.";
+                        const { treeDataChildrenField, treeDataParentIdField, getDataPath, getRowId } = options;
+                        if (!treeDataChildrenField && !treeDataParentIdField && !getDataPath) {
+                            return "treeData requires either 'treeDataChildrenField' or 'treeDataParentIdField' or or 'getDataPath' in the clientSide row model.";
```
AndrewGlazier: or or in message

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -56,17 +54,8 @@ interface GroupingDetails {
...
-
-    public refreshProps: Set<keyof GridOptions<any>> = new Set([
-        'groupDefaultExpanded',
-        'groupAllowUnbalanced',
-        'initialGroupOrderComparator',
-        'groupHideOpenParents',
-        'groupDisplayType',
-    ]);
-    public step: ClientSideRowModelStage = 'group';
+export class GroupStrategy extends BeanStub implements NamedBean, IRowGroupingStrategy {
```
AndrewGlazier: These aren't namedBeans anymore - they're dynamicBeans (no interface, just BeanStub), also means no need for beanName

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
```diff
@@ -0,0 +1,142 @@
...
+            }
+            if (gos.get('getDataPath')) {
+                return undefined; // managed by node manager
+            }
+        }
+
+        return (this.groupStrategy ??= this.createStrategy('GroupStrategy'));
+    }
+
+    private treeDataManagedByNodeManager(): boolean {
```
AndrewGlazier: this func is identical to `getNewStrategy() === undefined`

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
```diff
@@ -0,0 +1,142 @@
...
+        super.destroy();
+        this.parentIdTreeStrategy = context.destroyBean(this.parentIdTreeStrategy);
+        this.groupStrategy = context.destroyBean(this.groupStrategy);
+    }
+
+    private createStrategy(name: 'GroupStrategy' | 'TreeParentIdStrategy'): IRowGroupingStrategy<TData> | undefined {
+        const beans = this.beans;
+        const result = beans.registry.createDynamicBean<IRowGroupingStrategy<TData>>(name, false);
+        if (result) {
+            beans.context.createBean(result);
```
AndrewGlazier: this.createdManagedBean to avoid memory leak

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
```diff
@@ -0,0 +1,142 @@
...
+    public execute(params: StageExecuteParams<TData>): boolean {
+        const { strategy: oldStrategy } = this;
+
+        const newStrategy = this.getNewStrategy();
+
+        const strategyChanged = oldStrategy !== newStrategy;
+        if (strategyChanged) {
+            this.strategy = newStrategy;
+            if (oldStrategy) {
+                oldStrategy.deactivate?.();
```
AndrewGlazier: this.destroyBean(oldStrategy), then use destroy in strategy beans to do needed cleanup

packages/ag-grid-community/src/context/context.ts
SalvatorePreviti: thank you, was a leftover!

packages/ag-grid-community/src/validation/rules/gridOptionsValidations.ts
SalvatorePreviti: thank you!

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
SalvatorePreviti: thank you! was a leftover

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: solved

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: not really as the TreeData or Grouping module might not be loaded - in that case getNewStrategy would return undefined too - let me think/verify what would happen

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: Introduced _getGroupingApproach in grid option utils, used it in CSRM as well. Could you have another look?

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: As discussed and agreed we are using CSRM getRowNode(id)

packages/ag-grid-community/src/interfaces/iRowNodeStage.ts
SalvatorePreviti: Removed as now we use fully dynamic beans that gets created or destroyed

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: Removed this as we use fully dynamic modules and _getGroupingApproach

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: using destroyBean

packages/ag-grid-community/src/context/context.ts
```diff
@@ -124,7 +124,9 @@ export type DynamicBeanName =
     | 'headerFilterCellCtrl'
     | 'headerGroupCellCtrl'
     | 'rangeHandle'
-    | 'tooltipFeature';
+    | 'tooltipFeature'
+    | 'GroupStrategy'
```
gportela85: This should be `groupStrategy`

packages/ag-grid-community/src/context/context.ts
```diff
@@ -124,7 +124,9 @@ export type DynamicBeanName =
...
+    | 'TreeParentIdStrategy';
```
gportela85: This should be `treeParentIdStrategy`

PR #9972, opened 2025-02-26
Ag 14256 plausible tracking sign up license

kylerphillips:
Adds success and error tracking to plausible when trial license form gets submitted


taktran (approved):
LGTM 👍🏼 Tested locally, and it works well

PR #9975, opened 2025-02-26
AG-119 - fix find with group rendering override params

peterjrreynolds:


packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -378,12 +411,16 @@ export class FindService extends BeanStub implements NamedBean, IFindService {
...
                     continue;
                 }
-                const cellSpan = rowSpanSvc?.getCellSpan(column, node as RowNode);
+                const cellSpan = rowSpanSvc?.getCellSpan(column, node);
                 if (cellSpan && cellSpan.firstNode !== node) {
                     // only match on first row of span
                     continue;
                 }
-                const value = valueSvc.getValueForDisplay(column, node);
+                // we don't want to call `getValueForDisplay` if the node is not actually displayed,
```
AndrewGlazier, line 419: Do we actually get here for groupHideOpenParents? I thought the early return added above would prevent this being necessary?

peterjrreynolds: Discussed offline (yes)

PR #9978, opened 2025-02-26
AG-12174 Manual row pinning

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12174


packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -1284,7 +1284,21 @@ export interface GridOptions<TData = any> {
      * Data to be displayed as pinned bottom rows in the grid.
      */
     pinnedBottomRowData?: any[];
-
+    /**
+     * Determines whether manual row pinning is enabled via the row context menu.
+     *
+     * Set to `true` to allow pinning rows to both top and bottom pinned row containers.
```
AndrewGlazier: nit, "row containers" are more of an internal concept, I think just leaving it as "to both the top and bottom" potentially suffix with "of the viewport" would be sufficient

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -1284,7 +1284,21 @@ export interface GridOptions<TData = any> {
...
+     * Set to `'bottom'` to allow pinning rows to the bottom pinned row container only.
+     */
+    enableRowPinning?: boolean | 'top' | 'bottom';
+    /**
+     * Return `true` if the grid should allow the row to be manually pinned.
+     * Return `false` if the grid should prevent the row from being pinned
+     *
+     * When not defined, all rows default to pinnable.
+     */
+    isRowPinnable?: IsRowPinnable<TData>;
```
AndrewGlazier, line 1301: Would also suggest "isRowPinnedByDefault"

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -2448,6 +2462,10 @@ export interface IsRowSelectable<TData = any> {
     (node: IRowNode<TData>): boolean;
 }
 
+export interface IsRowPinnable<TData = any> {
+    (node: IRowNode<TData>): boolean;
```
AndrewGlazier, line 2473: Might be more than needed; but would we ever want to restrict certain rows to be pinned in specific directions? I assume not; but worth considering

packages/ag-grid-community/src/export/gridSerializer.ts
```diff
@@ -210,7 +210,7 @@ export class GridSerializer extends BeanStub implements NamedBean {
                     .sort((a, b) => a.rowIndex - b.rowIndex)
                     .map((position) => this.pinnedRowModel?.getPinnedTopRow(position.rowIndex))
                     .forEach(processRow);
-            } else {
+            } else if (!this.pinnedRowModel?.isManual()) {
```
AndrewGlazier, line 213: Did you check what happens with copying pinned row values? Just checking as copy/paste uses serializer

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -241,6 +244,9 @@ export class StateService extends BeanStub implements NamedBean {
         if (scrollState) {
             this.setScrollState(scrollState);
         }
+        if (rowPinning) {
+            this.setRowPinningState(rowPinning);
```
AndrewGlazier, line 247: This approach won't work for SSRM, as first data rendered may only be one block of the data - maybe worth asking Pete/looking at how he extracts row group state for SSRM case (if we do)

packages/ag-grid-community/src/rendering/cell/cellPositionFeature.ts
```diff
@@ -62,10 +62,14 @@ export class CellPositionFeature extends BeanStub {
...
+            const refreshSpanHeight = this.refreshSpanHeight.bind(this, cellSpan);
+            refreshSpanHeight();
             this.addManagedListeners(this.beans.eventSvc, {
-                paginationChanged: this.refreshSpanHeight.bind(this, cellSpan),
-                recalculateRowBounds: this.refreshSpanHeight.bind(this, cellSpan),
+                paginationChanged: refreshSpanHeight,
+                recalculateRowBounds: refreshSpanHeight,
+                // XXX: This might be too eager but there are cases where pinning spanned
+                // rows doesn't fire 'recalculateRowBounds'
+                displayedRowsChanged: refreshSpanHeight,
```
AndrewGlazier: I think it'd be better to remove this and try to debug an alternative - do you have details on the scenario where this breaks?

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -728,6 +728,14 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
             rowHighlightChanged: this.onRowNodeHighlightChanged.bind(this),
             draggingChanged: this.postProcessRowDragging.bind(this),
             uiLevelChanged: this.onUiLevelChanged.bind(this),
+            rowPinned: () => {
```
AndrewGlazier: nit: can you move this into a class function for consistency?

packages/ag-grid-community/src/validation/enterpriseModuleNames.ts
```diff
@@ -23,6 +25,7 @@ export const ENTERPRISE_MODULE_NAMES: Record<EnterpriseModuleName, 1> = {
     RowNumbers: 1,
     RowGrouping: 1,
     RowGroupingPanel: 1,
+    ServerSideManualPinnedRow: 1,
```
AndrewGlazier: I'm not sure this module is required - as long as SSRM is updated it shouldn't need an entirely different approach

packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowModel.ts
```diff
@@ -0,0 +1,384 @@
...
+}
+
+export class ManualPinnedRowModel extends BeanStub implements IPinnedRowModel {
+    private top = new OrderedSet();
+    private bottom = new OrderedSet();
+
+    public postConstruct(): void {
+        const filterManager = this.beans.filterManager;
+        const hideFilteredNodes = (node: RowNode) =>
+            filterManager ? !filterManager.doesRowPassFilter({ rowNode: node }) : false;
```
AndrewGlazier: This approach to filtering might not work, considering we have aggregate filters which can cause the children to pass

packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowModel.ts
```diff
@@ -0,0 +1,384 @@
...
+    let rowTop = 0;
+    container.forEach((node, index) => {
+        node.setRowTop(rowTop);
+        node.setRowHeight(_getRowHeightForNode(beans, node).height);
+        node.setRowIndex(index);
+        rowTop += node.rowHeight!;
+    });
+}
+
+function validatePinningOptions(gos: GridOptionsService): void {
```
AndrewGlazier: This should live inside of the `gridOptionsValidations`

packages/ag-grid-enterprise/src/menu/contextMenu.ts
```diff
@@ -93,6 +93,15 @@ export class ContextMenuService extends BeanStub implements NamedBean, IContextM
             if (anyExport) {
                 defaultMenuOptions.push('export');
             }
+
+            const enableRowPinning = gos.get('enableRowPinning');
+            const isRowPinnable = gos.get('isRowPinnable');
+            if (enableRowPinning) {
+                const pinnable = isRowPinnable?.(node) ?? true;
```
AndrewGlazier, line 101: Odd use case, but what if pinned row returns false to isRowPinnable? Should pinned rows always be unpinnable? Not for grand total etc, but wondering if this edge case is a concern

packages/ag-grid-enterprise/src/menu/contextMenu.ts
```diff
@@ -93,6 +93,15 @@ export class ContextMenuService extends BeanStub implements NamedBean, IContextM
...
+                if (pinnable) {
+                    defaultMenuOptions.push(node.rowPinned ? 'unpinRow' : 'pinRowSubMenu');
```
AndrewGlazier: Any option to directly move pinned row from top to bottom? Do we not want this?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -109,6 +109,12 @@ export class RowNode<TData = any>
      */
     public rowPinned: RowPinnedType;
 
+    /** `true` if node represents a manually pinned row. This row is similar in behaviour to a footer. */
+    public manualPinned?: boolean;
```
AndrewGlazier: Is this important? Can we not just use the existing .rowPinned property?

packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowModel.ts
```diff
@@ -0,0 +1,384 @@
...
+    private visible = new Set<RowNode>();
+    /** Ordering of nodes in the pinned area */
+    private order: RowNode[] = [];
+    /**
+     * We cache the row index of nodes to handle the case where they become not displayed (e.g
+     * collapsing a group). Their `rowIndex` then becomes `null` and pinned row order becomes
+     * unstable in a way that looks odd for users. In this case we fall back to their `rowIndex`
+     * when they were last displayed.
+     */
+    private indexCache = new Map<RowNode, number>();
```
AndrewGlazier: Not sure if this will be enough as if sorts change the rows will be in wrong position - we might want to call the sortService to determine the rowIndices just for these pinned rows - instead of trying to determine this from the cache

packages/ag-grid-community/src/clientSideRowModel/abstractClientSideNodeManager.ts
```diff
@@ -350,6 +350,11 @@ export abstract class AbstractClientSideNodeManager<TData = any>
                 nodesToUnselect.push(rowNode);
             }
 
+            // If a row has been manually pinned, ensure its sibling is also removed
+            if (rowNode.pinnedSibling) {
```
AndrewGlazier, line 357: Not sure this will catch the use case for immutable data changes in tree data nested records mode, as that doesn't support transactions

packages/ag-grid-community/src/export/gridSerializer.ts
eliasmalik: Copying cell ranges works as expected. I haven't implemented row selection in this PR but since we expect pinned rows to mirror the selection state of their source rows, I think we'd want those pinned rows to be excluded from the copy in the same way that as for the export.

packages/ag-grid-community/src/entities/gridOptions.ts
eliasmalik: Is there an advantage to this over using initial state?

AndrewGlazier: Yes; we don't want users to mutate initial state, it's a recovery device but shouldn't really be mutated, meaning isRowPinnedByDefault would be used for the first time the grid loads

packages/ag-grid-community/src/gridOptionsUtils.ts
```diff
@@ -335,7 +335,7 @@ export function _getGroupAggFiltering(
     return undefined;
 }
 
-export function _getGrandTotalRow(gos: GridOptionsService): 'top' | 'bottom' | undefined {
+export function _getGrandTotalRow(gos: GridOptionsService): GridOptions['grandTotalRow'] {
```
AndrewGlazier, line 338: I know this wasn't you (and may have even been me)... but any idea what value we gain from this function? Seems odd

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -307,17 +308,13 @@ export class CellCtrl extends BeanStub {
...
     }
 
     // eslint-disable-next-line @typescript-eslint/ban-types
-    private isIncludeControl(value: boolean | Function | undefined): boolean {
-        const rowNodePinned = this.rowNode.rowPinned != null;
-        const isFunc = typeof value === 'function';
-        const res = rowNodePinned ? false : isFunc || value === true;
-
-        return res;
+    private isIncludeControl(value: boolean | Function | undefined, allowManuallyPinned = false): boolean {
```
AndrewGlazier, line 321: Can you improve this func name or give it a jsdoc comment, "is include control" is quite a vague name and I'm still not entirely sure what it's doing

packages/ag-grid-community/src/rendering/spanning/rowSpanService.ts
```diff
@@ -15,9 +15,11 @@ export class RowSpanService extends BeanStub<'spannedCellsUpdated'> implements N
 
     public postConstruct(): void {
         const onRowDataUpdated = this.onRowDataUpdated.bind(this);
+        const buildPinnedCaches = this.buildPinnedCaches.bind(this);
         this.addManagedEventListeners({
             paginationChanged: this.buildModelCaches.bind(this),
-            pinnedRowDataChanged: this.buildPinnedCaches.bind(this),
+            pinnedRowDataChanged: buildPinnedCaches,
+            rowPinnedChanged: buildPinnedCaches,
```
AndrewGlazier: Personally if we're ever going to make this public I think pinnedRowsChanged would be a better event name

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -513,45 +514,41 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
         }
 
         const nodes: RowNode[] = [];
+        const push = (node: RowNode) => nodes.push(node);
```
AndrewGlazier: Same as `const push = nodes.push` - though might be valuable to rename this to addToResult or something?

eliasmalik: Defining it point-free gives us type errors when passing to `forEachNode` and `forEachNodeAfterFilter` because they accept callbacks that take an index as a second parameter and `.push` is variadic.

PR #9981, opened 2025-02-27
AG-7887 Angular batch framework comps

StephenCooper:
Main part of the change is in the animationFrameService where for Angular we batch the cell renderer component creation within a single `ngZone.run()` so that we only trigger zone.onStable after a full batch of components instead of after every one is created.

In testing this has lead to significant performance improvements for grids with Angular custom cell renderers. This fix only applies when `suppressAnimationFrame=false` (the default value). Users should be pushed away from using this flag if possible especially when users may not have high powered machines.


documentation/ag-grid-docs/src/utils/sitemap.ts
```diff
@@ -12,6 +12,13 @@ const isDebugPage = (page: string) => {
     return page.includes('/debug/');
 };
 
+/*
+ * Test pages for testing
+ */
+export const isTestPage = (page: string) => {
```
StephenCooper, line 18: Copies the pattern used in AG Charts docs

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkComponentWrapper.ts
```diff
@@ -38,7 +38,7 @@ export class AngularFrameworkComponentWrapper
             }
 
             protected createComponent(): ComponentRef<AgFrameworkComponent<any>> {
-                return angularFrameworkOverrides.runInsideAngular(() => that.createComponent(OriginalConstructor));
+                return that.createComponent(OriginalConstructor);
```
StephenCooper, line 41: This is already called within a higher runInsideAngular call so no need to wrap again.

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
```diff
@@ -91,13 +93,16 @@ export class AngularFrameworkOverrides extends VanillaFrameworkOverrides {
...
         const prototype = comp.prototype;
-        const isAngularComp = prototype && 'agInit' in prototype;
-        return isAngularComp;
+        return prototype && 'agInit' in prototype;
     }
 
     runInsideAngular<T>(callback: () => T): T {
+        if (!this._ngZone || NgZone.isInAngularZone()) {
+            return callback();
+        }
```
StephenCooper, line 102: Found it slightly quicker to check if already in the zone as opposed to blindly calling into run again.

packages/ag-grid-community/src/misc/animationFrameService.ts
```diff
@@ -50,7 +52,6 @@ export class AnimationFrameService extends BeanStub implements NamedBean {
     private lastScrollTop = 0;
 
     private taskCount = 0;
-    private cancelledTasks = new Set();
```
StephenCooper, line 53: We never cancel tasks so removing the dead code

packages/ag-grid-community/src/gridBodyComp/gridBodyScrollFeature.ts
```diff
@@ -263,10 +263,10 @@ export class GridBodyScrollFeature extends BeanStub {
         // the `scrollGridIfNeeded` will recalculate the rows to be rendered by the grid
         // so it should only be called after `eBodyViewport` has been scrolled to the correct
         // position, otherwise the `first` and `last` row could be miscalculated.
-        if (!animationFrameSvc || this.gos.get('suppressAnimationFrame')) {
-            this.scrollGridIfNeeded(true);
-        } else {
+        if (animationFrameSvc?.active) {
             animationFrameSvc.schedule();
+        } else {
+            this.scrollGridIfNeeded(true);
```
StephenCooper, line 269: active property already records the value of `gos.get('suppressAnimationFrame')`

documentation/ag-grid-docs/src/utils/sitemap.ts
peterjrreynolds: This may also need adding to the robots files check

packages/ag-grid-community/src/misc/animationFrameService.ts
```diff
@@ -139,32 +137,56 @@ export class AnimationFrameService extends BeanStub implements NamedBean {
...
+                    this.sortTaskList(p1);
                     task = p1Tasks.pop()!.task;
                 } else if (p2Tasks.length) {
-                    this.sortTaskList(p2TaskList);
+                    this.sortTaskList(p2);
                     task = p2Tasks.pop()!.task;
+                } else if (f1Tasks.length) {
+                    // Assuming that framework tasks do not schedule p1 or p2 tasks so that it is safe
+                    // to loop through all framework tasks for as long as we have time left
+                    this.beans.frameworkOverrides.wrapOutgoing(() => {
```
peterjrreynolds: extract to variable outside the loop

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/angularFrameworkOverrides.ts
```diff
@@ -11,6 +11,8 @@ import { AngularFrameworkEventListenerService } from './angularFrameworkEventLis
 
 @Injectable()
 export class AngularFrameworkOverrides extends VanillaFrameworkOverrides {
+    public override batchFrameworkComponents: boolean = true;
```
peterjrreynolds: readonly?

packages/ag-grid-community/src/vanillaFrameworkOverrides.ts
```diff
@@ -7,6 +7,7 @@ import { setValidationDocLink } from './validation/logging';
 /** The base frameworks, eg React & Angular, override this bean with implementations specific to their requirement. */
 export class VanillaFrameworkOverrides implements IFrameworkOverrides {
     public renderingEngine: 'vanilla' | 'react' = 'vanilla';
+    public batchFrameworkComponents = false;
```
peterjrreynolds: readonly?

documentation/ag-grid-docs/src/utils/sitemap.ts
StephenCooper: Added a hardcoded exclude to the robots page and will follow up with Tak.

peterjrreynolds (approved):

gportela85 (approved):

PR #9992, opened 2025-02-28
AG-14268 - improved code to refocus rows that were removed

gportela85:
https://ag-grid.atlassian.net/browse/AG-14268

### DIFF
--- a/packages/ag-grid-community/src/entities/positionUtils.ts
+++ b/packages/ag-grid-community/src/entities/positionUtils.ts
@@ -130,6 +130,10 @@ export function _getRowAbove(beans: BeanCollection, rowPosition: RowPosition): R
     const { rowIndex: index, rowPinned: pinned } = rowPosition;
     const { pageBounds, pinnedRowModel, rowModel } = beans;
 
+    if (pinned === 'top' && index === 0) {
+        return null;
+    }
+
     if (pinned == null && index === pageBounds.getFirstRow()) {
         return pinnedRowModel?.isRowsToRender('top')
             ? { rowIndex: pinnedRowModel.getPinnedTopRowCount() - 1, rowPinned: 'top' }
--- a/packages/ag-grid-community/src/entities/positionUtils.ts
+++ b/packages/ag-grid-community/src/entities/positionUtils.ts
@@ -189,21 +189,19 @@ function getNextStickyPosition(beans: BeanCollection, rowNode?: RowNode, up?: bo
 }
 
 function isLastRowInContainer(beans: BeanCollection, rowPosition: RowPosition): boolean {
-    const pinned = rowPosition.rowPinned;
-    const index = rowPosition.rowIndex;
-
+    const { rowPinned, rowIndex } = rowPosition;
     const { pinnedRowModel, pageBounds } = beans;
 
-    if (pinned === 'top') {
+    if (rowPinned === 'top') {
         const lastTopIndex = pinnedRowModel?.getPinnedTopRowCount() ?? 0 - 1;
-        return lastTopIndex <= index;
+        return lastTopIndex <= rowIndex;
     }
 
-    if (pinned === 'bottom') {
+    if (rowPinned === 'bottom') {
         const lastBottomIndex = pinnedRowModel?.getPinnedBottomRowCount() ?? 0 - 1;
-        return lastBottomIndex <= index;
+        return lastBottomIndex <= rowIndex;
     }
 
     const lastBodyIndex = pageBounds.getLastRow();
-    return lastBodyIndex <= index;
+    return lastBodyIndex <= rowIndex;
 }
--- a/packages/ag-grid-community/src/entities/positionUtils.ts
+++ b/packages/ag-grid-community/src/entities/positionUtils.ts
@@ -127,131 +127,61 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
 }
 
 export function _getRowAbove(beans: BeanCollection, rowPosition: RowPosition): RowPosition | null {
-    // if already on top row, do nothing
-    const index = rowPosition.rowIndex;
-    const pinned = rowPosition.rowPinned;
+    const { rowIndex: index, rowPinned: pinned } = rowPosition;
     const { pageBounds, pinnedRowModel, rowModel } = beans;
-    const isFirstRow = pinned ? index === 0 : index === pageBounds.getFirstRow();
-    let ignoreSticky = false;
 
-    const getLastFloatingTopRow = (): RowPosition => {
-        const lastFloatingRow = pinnedRowModel?.getPinnedTopRowCount() ?? 0 - 1;
-
-        return { rowIndex: lastFloatingRow, rowPinned: 'top' } as RowPosition;
-    };
-
-    // if already on top row, do nothing
-    if (isFirstRow) {
-        if (pinned === 'top') {
-            return null;
-        }
-
-        if (!pinned) {
-            if (pinnedRowModel?.isRowsToRender('top')) {
-                return getLastFloatingTopRow();
-            }
-            return null;
-        }
-
-        // last floating bottom
-        if (rowModel.isRowsToRender()) {
-            const lastBodyRow = pageBounds.getLastRow();
-            return { rowIndex: lastBodyRow, rowPinned: null } as RowPosition;
-        }
-
-        if (pinnedRowModel?.isRowsToRender('top')) {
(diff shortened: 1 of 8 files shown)


packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,81 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
     return rowCtrl.getCellCtrl(cellPosition.column as AgColumn);
 }
+
+export function _getRowAbove(beans: BeanCollection, rowPosition: RowPosition): RowPosition | null {
+    const { rowIndex: index, rowPinned: pinned } = rowPosition;
+    const { pageBounds, pinnedRowModel, rowModel } = beans;
+
+    if (pinned === 'top' || (pinned === null && index === pageBounds.getFirstRow())) {
+        return pinnedRowModel?.isRowsToRender('top')
+            ? { rowIndex: pinnedRowModel.getPinnedTopRowCount() - 1, rowPinned: 'top' }
```
AndrewGlazier, line 139: If pinned === 'top' don't you need the pinned rowIndex - 1? Am I misunderstanding this bit?

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,81 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+            : null;
+    }
+
+    if (pinned === 'bottom' && rowModel.isRowsToRender()) {
```
AndrewGlazier: Same query as above; surely you could be in a pinned row that isn't the top pinned bottom row?

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,81 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+    }
+
+    const rowNode = rowModel.getRow(index);
+    return getNextStickyPosition(beans, rowNode) ?? { rowIndex: index + 1, rowPinned: pinned };
+}
+
+function getNextStickyPosition(beans: BeanCollection, rowNode?: RowNode, up?: boolean): RowPosition | undefined {
+    const { gos, rowRenderer } = beans;
+
+    if (!rowNode?.sticky || !_isGroupRowsSticky(gos)) {
```
AndrewGlazier, line 174: mildest performance point, and not concerned about requesting this change as it's not a hot path; but we could probably check isGroupRowsSticky + that there's sticky nodes before doing `getRow` if it was done inside this func

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,81 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+
+function isLastRowInContainer(beans: BeanCollection, rowPosition: RowPosition): boolean {
+    const pinned = rowPosition.rowPinned;
+    const index = rowPosition.rowIndex;
+
+    const { pinnedRowModel, pageBounds } = beans;
+
+    if (pinned === 'top') {
+        const lastTopIndex = pinnedRowModel?.getPinnedTopRowCount() ?? 0 - 1;
+        return lastTopIndex <= index;
```
AndrewGlazier: I'm curious; is there a scenario where the last index could be less than the index? Or is this just additional caution

If there's a scenario it'd be good to add a comment here outlining it

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -679,21 +681,51 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
-        });
+                },
+            });
+        } else {
+            const params = {
+                ...cellToFocus,
+                preventScrollOnBrowserFocus: true,
+                forceBrowserFocus: true,
+            };
+            if (cellPosition.rowIndex !== cellToFocus.rowIndex || cellPosition.rowPinned != cellPosition.rowPinned) {
```
AndrewGlazier: cellPosition.rowPinned != cellPosition.rowPinned

bug

gportela85: :(

packages/ag-grid-community/src/entities/positionUtils.ts
gportela85: Added an extra condition at the top to handle top and index 0.

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -679,21 +681,51 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
+                this.focusSvc.setFocusedCell(params);
```
AndrewGlazier, line 700: This is going to fire a public event again - not sure this is right

packages/ag-grid-community/src/entities/positionUtils.ts
gportela85: Actually, this was the only bit of code that I just moved from the `cellNavigationService` without refactoring it.
I'm not sure why this is there as it is, but I assumed there was a reason so I kept it.

packages/ag-grid-community/src/entities/positionUtils.ts
gportela85: The reason why this was kept here is exactly that.. it isn't a hot path, so I kept it here for readability / separation.

packages/ag-grid-community/src/rendering/rowRenderer.ts
gportela85: We are actually focusing a different row/cell in this case, so I think the event should be fired to let users know.

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,85 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+        return pinnedRowModel?.isRowsToRender('top')
+            ? { rowIndex: pinnedRowModel.getPinnedTopRowCount() - 1, rowPinned: 'top' }
+            : null;
+    }
+
+    if (pinned === 'bottom' && index === 0 && rowModel.isRowsToRender()) {
+        return { rowIndex: pageBounds.getLastRow(), rowPinned: null };
+    }
+
+    const rowNode = rowModel.getRow(index);
```
peterjrreynolds: If we're in a pinned section, this is getting a random row node.

The sticky logic should only be called here if `pinned == null`

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,85 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+        }
+
+        if (pinned === 'top' && rowModel.isRowsToRender()) {
+            return { rowIndex: pageBounds.getFirstRow(), rowPinned: null };
+        }
+
+        return pinnedRowModel?.isRowsToRender('bottom') ? { rowIndex: 0, rowPinned: 'bottom' } : null;
+    }
+
+    const rowNode = rowModel.getRow(index);
```
peterjrreynolds: Same as above

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -123,3 +125,85 @@ export function _getCellByPosition(beans: BeanCollection, cellPosition: CellPosi
...
+            break;
+        }
+    }
+
+    return nextCtrl ? { rowIndex: nextCtrl.rowNode.rowIndex!, rowPinned: null } : undefined;
+}
+
+function isLastRowInContainer(beans: BeanCollection, rowPosition: RowPosition): boolean {
+    const pinned = rowPosition.rowPinned;
+    const index = rowPosition.rowIndex;
```
peterjrreynolds: Can destructure these

packages/ag-grid-community/src/entities/positionUtils.ts
gportela85: This shouldn't really matter because `getNextStickyPosition` always checks for `rowNode.sticky` which will always be false for pinned nodes.

packages/ag-grid-community/src/entities/positionUtils.ts
gportela85: done

peterjrreynolds (approved):

PR #9993, opened 2025-02-28
AG-13810 - Prevent animations on col changes when grouping is impacted

AndrewGlazier:


peterjrreynolds (approved):

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -102,18 +102,29 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        let skipNextColsAnimation = false;
         this.addManagedEventListeners({
-            newColumnsLoaded: refreshEverythingAfterColsChangedFunc,
-            columnRowGroupChanged: refreshEverythingFunc,
+            newColumnsLoaded: () => {
+                this.refreshModel({
+                    step: 'group', // after cols change, row grouping (the first stage) could of changed
+                    afterColumnsChanged: true,
+                    keepRenderedRows: true,
+                    // we want animations if sorting or filtering could be impacted, but not if grouping has changed
```
gportela85, line 117: I see this comment there but there isn't a lot of detail explaining when this happens? When exactly is it important for us to animate the group step during the  `newColumnsLoaded`.

I can see the commit that added this, but it wasn't fixing a JIRA in particular.
https://github.com/ag-grid/ag-grid/commit/f5870aade8438e12f413b15578acee0401575d40 

Basically, if we set animate to `false` here, what breaks? And if there is something that actually breaks, couldn't we pass a flag through the `newColumnsLoaded` event?

StephenCooper:
Fix applied via https://github.com/ag-grid/ag-grid/pull/12605

PR #9994, opened 2025-02-28
Fix small typo in ValueGetter docs

iwburns:
Convert a few words from possessive to plural


StephenCooper:
These have been fixed as part of a wider grammar check. Thanks!

PR #10020, opened 2025-03-05
AG-14121 - Cell focus changes

AndrewGlazier:


packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -793,6 +787,12 @@ export class CellCtrl extends BeanStub {
     public override destroy(): void {
         this.onCompAttachedFuncs = [];
         this.onEditorAttachedFuncs = [];
+
+        // if this was focused; focus will need recovered
+        if (this.eGui === _getActiveDomElement(this.beans)) {
+            this.beans.focusSvc.needsFocusRestored = true;
```
peterjrreynolds, line 826: `this.beans` - bundle size

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -208,15 +208,29 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
...
         // pinned rows render before the main grid body in the SSRM, only fire the event after the main body has rendered.
-        if (!isSsrmLoadingRow && !isIrmLoadingRow && !this.rowNode.rowPinned) {
+        if (!isSsrmLoadingRow && !isIrmLoadingRow && !rowNode.rowPinned) {
             // this is fired within setComp as we know that the component renderer is now trying to render.
             // linked with the fact the function implementation queues behind requestAnimationFrame should allow
             // us to be certain that all rendering is done by the time the event fires.
             this.beans.rowRenderer.dispatchFirstDataRenderedEvent();
         }
+
+        const focusSvc = this.beans.focusSvc;
```
peterjrreynolds: `this.beans` - bundle size

packages/ag-grid-community/src/focusService.ts
```diff
@@ -75,6 +73,30 @@ export class FocusService extends BeanStub implements NamedBean {
         this.addDestroyFunc(_registerKeyboardFocusEvents(this.beans));
     }
 
+    /**
+     * Specifies whether to take focus, as grid either already has focus, or lost it due
+     * to a destroyed cell
+     * @returns true if the grid should re-take focus, otherwise false
+     */
+    public shouldTakeFocus(): boolean {
+        if (this.gos.get('suppressFocusAfterRefresh')) {
```
StephenCooper, line 81: If this is an initial property then I think we should store it in a class variable seeing as this method is called as part of the cell ctrl

packages/ag-grid-community/src/focusService.ts
```diff
@@ -75,6 +73,30 @@ export class FocusService extends BeanStub implements NamedBean {
...
+            return false;
+        }
+
+        if (this.needsFocusRestored) {
+            this.needsFocusRestored = false;
+            return true;
+        }
+
+        const activeElement = _getActiveDomElement(this.beans);
+        const cellDomData = _getDomData(this.gos, activeElement, DOM_DATA_KEY_CELL_CTRL);
```
StephenCooper: Why are we checking domData as part of this check?

packages/ag-grid-community/src/focusService.ts
AndrewGlazier: It isn't initial. We do lots of gos checks in cell ctrl init phase and I don't believe they have been proven to present any significant performance issue

packages/ag-grid-community/src/focusService.ts
```diff
@@ -75,6 +73,30 @@ export class FocusService extends BeanStub implements NamedBean {
...
+        const rowDomData = _getDomData(this.gos, activeElement, DOM_DATA_KEY_ROW_CTRL);
+
+        const gridElementFocused = cellDomData || rowDomData;
+        return gridElementFocused;
```
gportela85: I don't think this is correct, cellDomData is CellCtrl and rowDomData is a rowCtrl but this method should return a boolean.

Maybe we should update `_getDomData` to return `CellCtrl` or `RowCtrl` instead of `any`.
Also, this looks very similar to `rowRenderer.getCellToRestoreFocusToAfterRefresh()`, so we should refactor them and avoid duplication.

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -681,34 +680,25 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
+        const cellToFocus = this.findPositionToFocus(cellPosition);
         if (!cellToFocus) {
-            this.focusSvc.focusHeaderPosition({
+            focusSvc.needsFocusRestored = true;
+            focusSvc.focusHeaderPosition({
                 headerPosition: {
                     headerRowIndex: getFocusHeaderRowCount(this.beans) - 1,
                     column: cellPosition.column,
                 },
             });
```
AndrewGlazier, line 685: self note, return early and remove else

packages/ag-grid-community/src/focusService.ts
AndrewGlazier: To ensure focus is captured in the grid (and so not stealing focus from elsewhere), this is the way it's previously been done and I cannot think of a more performant way to do so

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
AndrewGlazier: Only one instance of this.beans

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
AndrewGlazier: done

packages/ag-grid-community/src/focusService.ts
AndrewGlazier: Haven't fixed the typing on those functions, but I no longer directly call these/have also changed rowRenderer to no longer directly call these

packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -220,7 +211,28 @@ export class RowRenderer extends BeanStub implements NamedBean {
...
     }
 
+    private isCellRendered(rowIndex: number, column?: AgColumn): boolean {
+        const rowCtrl = this.rowCtrlsByRowIndex[rowIndex!];
+        if (!rowCtrl) {
+            return false;
+        }
+
+        if (!column) {
+            return true;
```
StephenCooper, line 222: Add a comment.

AndrewGlazier: done

packages/ag-grid-community/src/entities/positionUtils.ts
```diff
@@ -144,6 +144,10 @@ export function _getRowAbove(beans: BeanCollection, rowPosition: RowPosition): R
         return { rowIndex: pageBounds.getLastRow(), rowPinned: null };
     }
 
+    if (index === 0) {
```
peterjrreynolds: There's another case missing here. If index is 0, pinned bottom, no (center) rows to render, and pinned top exists, then should go to pinned top

peterjrreynolds: Looks like we also need similar handling for `_getRowBelow`

AndrewGlazier: done

packages/ag-grid-community/src/gridOptionsUtils.ts
```diff
@@ -202,6 +202,21 @@ export function _getActiveDomElement(beans: BeanCollection): Element | null {
     return _getRootNode(beans).activeElement;
 }
 
+export function _doesPopupHaveFocus(beans: BeanCollection, popupGui?: HTMLElement): boolean {
```
peterjrreynolds: Can we move this to `focus.ts` as this is not related to grid options

AndrewGlazier: I thought this too, but  _getActiveDomElement was here and is relevant looks like gos needs a bigger refactor as a bunch of random stuff has been thrown in here?

Regardless this is a non-issue as I'm removing this piece of code, after review I don't believe the changes in this ticket are impacted by the nested-popup logic at all...

PR #10058, opened 2025-03-11
AG-14358 Manually build dom elements

StephenCooper:


packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -179,21 +207,25 @@ export class HeaderComp extends Component implements IHeaderComp {
...
     }
 
-    private workOutTemplate(): string | null | undefined {
-        const { params, beans } = this;
-        const template: string | null | undefined = params.template ?? getHeaderCompTemplate(!!beans.sortSvc);
-
-        // take account of any newlines & whitespace before/after the actual template
-        return template?.trim ? template.trim() : template;
+    private setupTemplate(paramsTemplate?: string, sortSvc?: SortService) {
+        const template = paramsTemplate;
```
AndrewGlazier: This seems oddly redundant

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -179,21 +207,25 @@ export class HeaderComp extends Component implements IHeaderComp {
...
+        const sortComp = sortSvc ? [sortSvc.getSortIndicatorSelector()] : undefined;
+        if (template) {
+            // take account of any newlines & whitespace before/after the actual template
+            this.currentParamsTemplate = template;
+            this.setTemplate(template?.trim(), sortComp);
+        } else {
+            this.setTemplate(getHeaderCompElementParams(!!sortComp), sortComp);
```
AndrewGlazier: currentParamsTemplate needs set back to null, in case refresh is called I think?

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -303,8 +293,8 @@ export class HeaderRowCtrl extends BeanStub {
                     break;
             }
         }
-
-        this.headerCellCtrls.set(idOfChild, headerCtrl);
+        // Method only called when this.headerCellCtrls is defined
```
AndrewGlazier: Can we add a proper jsdoc header to the func disclaiming this to hopefully prevent it being called incorrectly in the future, as we're removing some defensive checks

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -341,20 +331,8 @@ export class HeaderRowCtrl extends BeanStub {
     public findHeaderCellCtrl(
```
AndrewGlazier: This function feels redundant & the interface over complicated, it's only used in two places (one using func, one using ref), can you delete in favour of just inlining allCtrls.find?

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
```diff
@@ -370,10 +348,11 @@ export class HeaderRowCtrl extends BeanStub {
     }
 
     public override destroy(): void {
-        this.headerCellCtrls?.forEach((ctrl) => {
+        for (const ctrl of this.allCtrls) {
```
AndrewGlazier: Can we use .destroyBeans here?

packages/ag-grid-community/src/headerRendering/rowContainer/headerRowContainerCtrl.ts
```diff
@@ -147,15 +147,15 @@ export class HeaderRowContainerCtrl extends BeanStub implements ScrollPartner {
 
     public getHeaderCtrlForColumn(column: AgColumn | AgColumnGroup): AbstractHeaderCellCtrl | undefined {
```
AndrewGlazier, line 148: This func looks slow, would it benefit from an indexed lookup?

packages/ag-grid-community/src/headerRendering/rowContainer/headerRowContainerCtrl.ts
```diff
@@ -147,15 +147,15 @@ export class HeaderRowContainerCtrl extends BeanStub implements ScrollPartner {
...
+            return this.columnsRowCtrl?.findHeaderCellCtrl(column);
         }
 
         if (this.groupsRowCtrls.length === 0) {
             return;
         }
 
         for (let i = 0; i < this.groupsRowCtrls.length; i++) {
-            const ctrl = this.groupsRowCtrls[i].getHeaderCellCtrl(column);
+            const ctrl = this.groupsRowCtrls[i].findHeaderCellCtrl(column);
```
AndrewGlazier: This logic is a little dumb; might not be a major improvement but we could speed this up by working out which header row the group is in based on the column group depth

Appreciate it's scope creep so feel free to ignore

packages/ag-grid-community/src/sort/sortIndicatorComp.ts
```diff
@@ -1,22 +1,33 @@
...
+import type { ElementParams } from '../utils/dom';
 import { _clearElement, _setDisplayed } from '../utils/dom';
 import type { IconName } from '../utils/icon';
 import { _createIconNoSpan } from '../utils/icon';
 import type { ComponentSelector } from '../widgets/component';
 import { Component, RefPlaceholder } from '../widgets/component';
 
-function makeSpan(dataRefSuffix: string, classSuffix: string) {
-    return /* html */ `<span data-ref="eSort${dataRefSuffix}" class="ag-sort-indicator-icon ag-sort-${classSuffix} ag-hidden" aria-hidden="true"></span>`;
+function buildTemplate(): ElementParams {
```
AndrewGlazier: Do all these templates need to be funcs? They're mostly static objects (with a few exceptions) These could simply be class variables

packages/ag-grid-community/src/utils/array.ts
```diff
@@ -28,8 +28,17 @@ export function _areEqual<T>(
     );
 }
 
-export function _sortNumerically(array: number[]): number[] {
-    return array.sort((a, b) => a - b);
+/**
+ * Utility that uses the fastest looping approach to apply a callback to each element of the array
+ * https://jsperf.app/for-for-of-for-in-foreach-comparison
+ */
+export function _forAll<T>(array: T[] | undefined, callback: (value: T) => void) {
```
AndrewGlazier, line 35: Why do we need this util? Isn't the fact that we're wrapping a callback slower than just inlining content in for ( ... of ....)?

packages/ag-grid-community/src/utils/dom.ts
```diff
@@ -424,17 +425,13 @@ export function _setElementHeight(element: HTMLElement, height: string | number)
...
+    element.style.minHeight = height;
 }
 
 export function _formatSize(size: number | string) {
-    if (typeof size === 'number') {
-        return `${size}px`;
-    }
-
-    return size;
+    return typeof size === 'number' ? `${size}px` : size.toString();
```
AndrewGlazier: I appreciate you probably didn't want to change the existing logic, but can we not just trust our types that size is a string and not call toString on it again?

packages/ag-grid-community/src/widgets/component.ts
```diff
@@ -219,11 +221,17 @@ export class Component<TLocalEvent extends string = ComponentEvent>
...
+        templateOfParams: ElementParams | string | null | undefined,
         componentSelectors?: ComponentSelector[],
         paramsMap?: { [key: string]: any }
     ): void {
-        const eGui = _loadTemplate(template as string);
+        let eGui: HTMLElement;
+        if (typeof templateOfParams === 'object' && templateOfParams != null) {
+            eGui = _createElement(templateOfParams);
+        } else {
+            eGui = _loadTemplate(templateOfParams as string);
```
AndrewGlazier: _loadTemplate typing is wrong if we want it to handle null/undefined

packages/ag-grid-community/src/utils/dom.ts
```diff
@@ -573,3 +570,58 @@ export function _requestAnimationFrame(beans: BeanCollection, callback: any) {
...
+    /** The type of input field */
+    inputType?: string;
+    tabindex?: string;
+    children?: (ElementParams | null)[]; // nulls are allowed to allow for optional children
+};
+
+export function _createElement(params: ElementParams) {
+    const { ariaHidden, children, class: className, inputType, ref: dataRef, role, tabindex, tag } = params;
+
+    const element = document.createElement(tag);
```
AndrewGlazier, line 559: Probably minor; but there's bundle size to be saved by assuming tag is 'div' if omitted

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: If the paramsTemplate has changed then refresh returns false to re-create the component. See existing comment on Line 185. 

So we don't need to set back to null due to that behaviour.

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
StephenCooper: Good spot. I think I will pass that into the function so it can be typed as required. Also the check around this call is not required because we always define it as a new map before calling this function.

packages/ag-grid-community/src/headerRendering/row/headerRowCtrl.ts
StephenCooper: It is used outside of the class but agree it is not really adding much value now that we expose the array of ctrls.

packages/ag-grid-community/src/headerRendering/rowContainer/headerRowContainerCtrl.ts
StephenCooper: I don't know this enough to fully understand the change so will ignore for now.

packages/ag-grid-community/src/utils/array.ts
StephenCooper: I have done some tests, both with benchmark and on our code and there is no noticable difference doing this one extra level of wrapping. Probably because where this is being used the callback has already been created. 

It does make that code a lot cleaner and smaller where this is currently used.

packages/ag-grid-community/src/utils/dom.ts
StephenCooper: I would rather not do this, as I feel this does reduce the readability and won't pay off in terms of bundle size.

StephenCooper: Have introduced helper functions for `div` and `span` to reduce duplication of those tag names which is most elements.

AndrewGlazier (approved):

packages/ag-grid-community/src/utils/dom.ts
StephenCooper: Turns out the helper functions do more damage then good due to the fact that bundlers then have to know they are pure functions. 

Even when they do the right thing there is actually no significant difference in bundle size with the helper functions and just specify the tag each time. Hence I will just specify the tag and this is also easiest for all bundlers to reason about and tree shaking away when the element objects are defined outside of the class.

PR #10078, opened 2025-03-17
AG-14425 Fix selection column column reset

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14425

### DIFF
--- a/packages/ag-grid-community/src/columns/columnStateUtils.ts
+++ b/packages/ag-grid-community/src/columns/columnStateUtils.ts
@@ -306,10 +306,7 @@ export function _resetColumnState(beans: BeanCollection, source: ColumnEventType
 
     autoColSvc?.getColumns()?.forEach(addColState);
     selectionColSvc?.getColumns()?.forEach(addColState);
-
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/columns/columnStateUtils.ts
```diff
@@ -306,9 +304,8 @@ export function _resetColumnState(beans: BeanCollection, source: ColumnEventType
...
     };
 
-    if (groupAutoCols) {
-        groupAutoCols.forEach(addColState);
-    }
+    autoColSvc?.getColumns()?.forEach(addColState);
+    selectionColSvc?.getColumns()?.forEach(addColState);
 
     if (primaryColumns) {
         primaryCols.forEach(addColState);
```
eliasmalik: Is there are reason we're checking `primaryColumns` but acting on `primaryCols`?

AndrewGlazier: I trusted auto-complete too much, should be primaryColumns (nice catch)

PR #10088, opened 2025-03-18
AG-14059 - Add console log component

taktran:
## Changes

* Add console log component based off https://github.com/ag-grid/ag-charts/pull/3759
  * Add support for `supportedFrameworks` (grid only)
  * Move `safeStringify` to the `postMessage` side and only allow simple objects/arrays to be logged
  * Increase buffer size to `20`
  * Update styles of log component
* Remove all `window.alert`s in examples

### DIFF
--- a/plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
+++ b/plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
@@ -19,6 +19,7 @@ import {
     convertTsxToJsx,
     getBoilerPlateFiles,
     getEntryFileName,
+    getHasExampleConsoleLog,
     getIsEnterprise,
     getIsLocale,
     getMainFileName,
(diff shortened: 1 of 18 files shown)


plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
```diff
@@ -92,6 +95,9 @@ async function getProvidedFiles(folderPath: string) {
 export async function generateFiles(options: ExecutorOptions, gridOptionsTypes: Record<string, GridOptionsType>) {
     const isDev = options.mode === 'dev';
     const folderPath = options.examplePath;
+    const folders = folderPath.split('/');
+    const pageName = folders[folders.length - 3];
+    const exampleName = folders[folders.length - 1];
```
peterjrreynolds: Might be worth pulling this out to somewhere sensible, so if we change the structure in the future this doesn't suddenly break.

peterjrreynolds (approved):

kylerphillips (approved):

kylerphillips:
LGTM!

markdurrant (commented):
👍

markdurrant (approved):

plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
taktran: Putting the logic inside website docs (where the other `fileUtils` live), which is outside of the `plugins` folder, is not that straight forward because the example generation is dependant on the website docs, and would cause a circular dependency. It's also only grid specific.

Instead I've pulled it out as a function within plugins and throw an error if it's empty. Hopefully, it fails loud enough if the structure changes, so we know to change it in future

PR #10091, opened 2025-03-18
AG-14412 fix all leaf children computation for tree data

SalvatorePreviti:
Part 1:

- Fix the computation of allLeaf children so that includes all nodes that have data, like it was in version 31
- Instead of using _EmptyArray, uses null as it was in version 31
- make the documentation about allLeafChildren field clearer


packages/ag-grid-community/src/interfaces/iRowNode.ts
```diff
@@ -187,7 +187,13 @@ interface GroupRowNode<TData = any> {
 
     /** `true` if this node is a group and the group is the bottom level in the tree. */
     leafGroup: boolean | undefined;
-    /** All lowest level nodes beneath this node, no groups. */
+    /**
+     * All the nodes that have user provided data below this node. Can be null if empty.
+     * This excludes:
+     * - filler nodes when using treeData and getDataPath
+     * - group nodes when using grouping
+     * - header and footer nodes
```
AndrewGlazier: I'm not sure what a header node is in this context; not sure if this was an intentional add?

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
```diff
@@ -246,31 +247,37 @@ const updateRowArrays = <TData>(row: TreeRow<TData>, childrenAfterGroup: TreeRow
 
 const updateAllLeafChildren = <TData>(
     row: TreeRow<TData>,
-    allLeafChildren: TreeRow<TData>[],
+    allLeafChildren: TreeRow<TData>[] | null,
     newAllLeafChildrenLen: number
 ): boolean => {
-    let changed = allLeafChildren.length !== newAllLeafChildrenLen;
-    if (changed) {
+    let changed = true;
```
AndrewGlazier: Can you move this down as it's only used after the else block now, and remove the else block

packages/ag-grid-community/src/interfaces/iRowNode.ts
SalvatorePreviti: fixed

packages/ag-grid-enterprise/src/treeData/treeParentIdStrategy.ts
SalvatorePreviti: simplified the logic

PR #10102, opened 2025-03-19
AG-14357 unmanaged tree data drag n drop doc example

SalvatorePreviti:
Better example for tree data row dragging, that supports repositioning rows as well as moving parents.


documentation/ag-grid-docs/src/content/docs/row-dragging/_examples/dragging-with-tree-data/fileCellRenderer.component_angular.ts
```diff
@@ -0,0 +1,37 @@
...
+    imports: [CommonModule],
+    template: `
+        <span class="filename">
+            <i [ngClass]="fileIconClass"></i>
+            {{ value }}
+        </span>
+    `,
+})
+export class FileCellRenderer implements ICellRendererAngularComp {
+    public value: string = '';
```
StephenCooper: These would both be better as a signal.

StephenCooper (approved):
Just a small change required in the Angular example.

documentation/ag-grid-docs/src/content/docs/row-dragging/_examples/dragging-with-tree-data/fileCellRenderer.component_angular.ts
SalvatorePreviti: done

PR #10106, opened 2025-03-19
AG-14325 Correctly render checkboxes when checkboxLocation = 'autoGroupColumn'

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14325

### DIFF
--- a/packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
+++ b/packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
@@ -22,7 +22,6 @@ import {
     _setAriaExpanded,
     _stopPropagationForAgGrid,
     _warn,
-    isColumnGroupAutoCol,
 } from 'ag-grid-community';
 
 import { _getGroupValue, _isHiddenParent } from '../rowHierarchyUtils';
@@ -559,8 +558,7 @@ export class GroupCellRendererCtrl extends BeanStub implements IGroupCellRendere
 
         const { node, column } = this.params;
         const rowSelection = this.gos.get('rowSelection');
-        const isAutoGroupColumn =
-            column?.getColDef().showRowGroup != null || (column && isColumnGroupAutoCol(column as AgColumn));
+        const isAutoGroupColumn = column?.getColDef().showRowGroup != null;
         const isCheckboxLocationHere = isAutoGroupColumn && _getCheckboxLocation(rowSelection) === 'autoGroupColumn';
         const checkboxes =
             typeof rowSelection === 'object'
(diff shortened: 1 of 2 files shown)


packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
```diff
@@ -556,12 +557,13 @@ export class GroupCellRendererCtrl extends BeanStub implements IGroupCellRendere
             return;
         }
 
-        const node = this.params.node as RowNode;
+        const { node, column } = this.params;
         const rowSelection = this.gos.get('rowSelection');
-        const checkboxLocation = _getCheckboxLocation(rowSelection);
+        const isAutoGroupColumn = column && isColumnGroupAutoCol(column as AgColumn);
```
AndrewGlazier: `isColumnGroupAutoCol` won't work with `groupDisplayType='custom'`

Might be better to check colDef.showRowGroup != null if we need it to work with this, though as the value is literally `autoGroupColumn` maybe this is correct?

packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
```diff
@@ -581,8 +583,8 @@ export class GroupCellRendererCtrl extends BeanStub implements IGroupCellRendere
             this.createBean(cbSelectionComponent);
 
             cbSelectionComponent.init({
-                rowNode: node, // when groupHideOpenParents = true and group expanded, we want the checkbox to refer to leaf node state (not group node state)
-                column: this.params.column as AgColumn,
+                rowNode: node as RowNode, // when groupHideOpenParents = true and group expanded, we want the checkbox to refer to leaf node state (not group node state)
```
AndrewGlazier, line 585: Was this needed? I'm not seeing the point of this cast

eliasmalik: Yeah, `GroupCellRenderParams.node` is typed as `IRowNode` but the `CheckboxSelectionComponent.init` method expects a `RowNode` (was being cast on line 559 before)

AndrewGlazier: Got it, missed that 👍

packages/ag-grid-enterprise/src/rowHierarchy/rendering/groupCellRendererCtrl.ts
```diff
@@ -556,12 +557,14 @@ export class GroupCellRendererCtrl extends BeanStub implements IGroupCellRendere
             return;
         }
 
-        const node = this.params.node as RowNode;
+        const { node, column } = this.params;
         const rowSelection = this.gos.get('rowSelection');
-        const checkboxLocation = _getCheckboxLocation(rowSelection);
+        const isAutoGroupColumn =
+            column?.getColDef().showRowGroup != null || (column && isColumnGroupAutoCol(column as AgColumn));
```
AndrewGlazier: column.getColDef().showRowGroup works for both auto and custom columns, we don't need to also check isColumnGroupAutoCol

eliasmalik: Removed the `isColumnGroupAutoCol` check

PR #10116, opened 2025-03-20
AG-14480 Fix example generator with empty interface files

SalvatorePreviti:
fix example generator to ignore undefined outputs and remove unused interfaces.ts files in two examples


taktran (approved):
LGTM

StephenCooper (approved):

PR #10118, opened 2025-03-20
AG-3390 - Return success value of generateFiles

taktran:
If `generateFiles` is not successful, it should pass that through the batch processing.

NOTE: `latest` with fail CI until https://github.com/ag-grid/ag-grid/pull/10116 is merged


alantreadway (approved):
LGTM

PR #10132, opened 2025-03-21
RTI-2619 - Wrap logger lines

taktran:


external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
```diff
@@ -35,6 +35,8 @@
     > div {
         padding: $spacing-size-2;
         border-bottom: 1px solid var(--color-border-secondary);
+        white-space: normal;
```
markdurrant: I think this should be pre-wrap. 

```suggestion
        white-space: pre-wrap;
```

markdurrant (changes requested):
One nitpick

external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
markdurrant: Just in case people add extra spaces indentation ect

taktran: Updated

markdurrant (approved):
👍

PR #10170, opened 2025-03-24
RTI-2633 Fix Framework Example styles

StephenCooper:


documentation/ag-grid-docs/src/content/docs/component-floating-filter/_examples/custom-complex-filter-and-floating-filter/index.html
```diff
@@ -1,7 +1 @@
-<style>
```
StephenCooper, line 1: Style is only present in the complex jquery example.

PR #10174, opened 2025-03-24
RTI-2650 - Update missing darkmode image

taktran:


taktran:
Should fix it in the component

PR #10183, opened 2025-03-24
AG-14520 handle split range on chart create

StephM-W:
https://ag-grid.atlassian.net/browse/AG-14520


packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -320,7 +321,33 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
 
     private getSelectedRange(): PartialCellRange {
         const ranges = this.rangeSvc?.getCellRanges() ?? [];
-        return ranges.length > 0 ? ranges[0] : { columns: [] };
+        if (ranges.length === 0) {
+            return { columns: [] };
+        }
+
+        const columns = ranges.reduce((cols, range) => cols.concat(range.columns), [] as Column[]);
```
peterjrreynolds: Can have multiple ranges with the same columns - need to dedupe

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -320,7 +321,33 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
...
+        let startRow = Number.MAX_VALUE;
+        let endRow = Number.MIN_VALUE;
+
+        ranges.forEach((range) => {
+            if (range.startRow && range.endRow) {
+                startRow = Math.min(startRow, range.startRow.rowIndex);
+                endRow = Math.max(endRow, range.endRow.rowIndex);
```
peterjrreynolds: This doesn't handle ranges that cross pinned rows, e.g.
- Range entirely within pinned top
- Range entirely within pinned bottom
- Range that goes from middle of pinned top into centre
- Etc.

StephM-W: Pinned rows aren't currently supported on chart ranges - will update code accordingly

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -319,8 +321,59 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
...
 
     private getSelectedRange(): PartialCellRange {
-        const ranges = this.rangeSvc?.getCellRanges() ?? [];
-        return ranges.length > 0 ? ranges[0] : { columns: [] };
+        const ranges = _uniqBy(this.rangeSvc?.getCellRanges() ?? [], 'id');
+        if (ranges.length === 0) {
+            return { columns: [] };
+        }
+
+        const columns = _uniq(ranges.reduce((cols, range) => cols.concat(range.columns), [] as Column[]));
```
peterjrreynolds: This is potentially creating multiple arrays to then convert to a set, then back to an array again.

`ranges` is already being iterated below. Much more efficient to just create a set, then iterate the columns and add to the column set in the ranges loop below.

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -319,8 +321,59 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
...
+            // if range crosses pinned rows, adjust the start/end row indexes to exclude pinned rows
+            // pinned rows aren't part of the main row model and:
+            //   * aren't easily accessible during chart data extraction
+            //   * aren't included in aggregation functions
+            //   * can have completely bespoke data shapes
+            //
+            if (range.startRow.rowPinned === 'top' && range.endRow.rowPinned !== 'top') {
+                // range crosses pinned top boundary, so start at first row in the row model
+                sIndex = 0;
+            }
```
peterjrreynolds, line 357: This is missing the case where the range is entirely within top and it should be ignore:
```
if (range.startRow.rowPinned === 'top') {
  if (range.endRow.rowPinned === 'top') {
    // range entirely within top - ignore
    continue;
  } else {
    // range crosses pinned top boundary, so start at first row in the row model
    sIndex = 0;
  }
}
```

Also need the same for bottom

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -319,8 +321,59 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
...
+        if (ranges.length === 0) {
+            return { columns: [] };
+        }
+
+        const columns = _uniq(ranges.reduce((cols, range) => cols.concat(range.columns), [] as Column[]));
+        const startRow: RowPosition = { rowIndex: Number.MAX_VALUE, rowPinned: undefined };
+        const endRow: RowPosition = { rowIndex: Number.MIN_VALUE, rowPinned: undefined };
+
+        ranges.forEach((range) => {
+            if (!(range.startRow && range.endRow)) {
```
peterjrreynolds: Bundle size - deconstruct `startRow` and `endRow` here

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -319,8 +321,59 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
     }
 
     private getSelectedRange(): PartialCellRange {
-        const ranges = this.rangeSvc?.getCellRanges() ?? [];
-        return ranges.length > 0 ? ranges[0] : { columns: [] };
+        const ranges = _uniqBy(this.rangeSvc?.getCellRanges() ?? [], 'id');
```
peterjrreynolds: In what situation do we have duplicate IDs?

IDs can also be undefined, won't we lose ranges here in that situation?

packages/ag-grid-enterprise/src/charts/chartService.ts
```diff
@@ -319,8 +321,59 @@ export class ChartService extends BeanStub implements NamedBean, IChartService {
...
+            }
+        });
+
+        if (startRow.rowIndex === Number.MAX_VALUE || endRow.rowIndex === Number.MIN_VALUE) {
+            // if we didn't find any valid ranges, return an empty range
+            return { columns: [] };
+        }
+
+        return {
+            id: this.generateId(),
```
peterjrreynolds: In `ChartDataModel` we set the ID to be the chart ID. Does that matter here?

packages/ag-grid-community/src/utils/array.ts
```diff
@@ -67,3 +67,19 @@ export function _flatten<T>(arrays: Array<T[]>): T[] {
     // Currently the fastest way to flatten an array according to https://jsbench.me/adlib26t2y/2
     return ([] as T[]).concat.apply([], arrays);
 }
+
+export function _uniq(array: any[]): any[] {
+    return [...new Set(array)];
+}
+
+export function _uniqBy<T>(array: T[], iteratee: ((value: T) => any) | string): T[] {
```
peterjrreynolds: We're trying to avoid creating more generic util functions (unless/until actually needed by multiple modules). Can move this into chartservice

packages/ag-grid-community/src/utils/array.ts
```diff
@@ -67,3 +67,19 @@ export function _flatten<T>(arrays: Array<T[]>): T[] {
     // Currently the fastest way to flatten an array according to https://jsbench.me/adlib26t2y/2
     return ([] as T[]).concat.apply([], arrays);
 }
+
+export function _uniq(array: any[]): any[] {
+    return [...new Set(array)];
```
peterjrreynolds: Same as below. In this case, not worth the util

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: I'll revisit this - I had tested this condition and the range had correctly been ignored

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: It makes more sense for it to be chartId, agreed - which we don't yet know at this stage, but which chartModel will.

I'll remove this here

packages/ag-grid-community/src/utils/array.ts
StephM-W: Removed

packages/ag-grid-community/src/utils/array.ts
StephM-W: Removed

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: Removed

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: Removed uniqueness handling - we really just want to capture the superset of non-pinned rows and columns

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: Deconstructed

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: Updated, this occurs for instance when selecting the non-first pinned row

packages/ag-grid-enterprise/src/charts/chartService.ts
StephM-W: Removed

peterjrreynolds (approved):

PR #10211, opened 2025-03-27
AG-14535 - drive suppressFieldDotNotation for chart from grid options

StephM-W:
https://ag-grid.atlassian.net/browse/AG-14535


packages/ag-grid-enterprise/src/charts/chartComp/chartProxies/chartTheme.ts
```diff
@@ -217,7 +217,6 @@ function inbuiltStockThemeOverrides(params: ChartProxyParams, isEnterprise: bool
                 bottom: extraPadding.includes('bottom') ? 40 : 20,
                 left: extraPadding.includes('left') ? 30 : 20,
             },
-            suppressFieldDotNotation: true,
```
StephenCooper, line 220: Is this removed as replaced with the file below?

StephM-W: That's correct, this default via theme overrides is removed, and instead is replaced by propagating the gridOptions settings of the same name if present

StephenCooper (approved):

PR #10216, opened 2025-03-28
AG-14477 Reduce use of innerHTML

StephenCooper:


gportela85 (approved):

packages/ag-grid-enterprise/src/widgets/agRichSelect.ts
```diff
@@ -866,6 +865,7 @@ export function _bindCellRendererToHtmlElement(
             if (typeof gui === 'object') {
                 eTarget.appendChild(gui);
             } else {
+                // eslint-disable-next-line no-restricted-properties -- no other way to parse custom HTML strings from the user
```
peterjrreynolds: This can be removed completely. This is legacy code (https://github.com/ag-grid/ag-grid/commit/8ea2207a3d06d5c20c2a5b6baf8102e60d5a206b#diff-60e953661995e1222876f4447976d570a2027998516284901c432ab373059cc5R1978)

Cell renderer `getGui()` has only been able to return `HTMLElement` since AngularJS support was removed.

peterjrreynolds (approved):

PR #10218, opened 2025-03-28
AG-12174 Row Pinning Documentation

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12174


documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
```diff
@@ -1412,14 +1412,32 @@
...
+        },
+        "enableRowPinning": {
+            "more": {
+                "name": "Manual Row Pinning",
+                "url": "./row-pinning-manual/#enabling-manual-row-pinning"
+            }
+        },
+        "isRowPinnable": {
+            "more": {
+                "name": "Manual Row Pinning",
```
AndrewGlazier: Can we remove 'Manual' - as this is the new preference let's just go with Row Pinned

documentation/ag-grid-docs/src/content/docs-nav/nav.json
```diff
@@ -596,7 +596,20 @@
...
-                            "path": "row-pinning"
+                            "path": "row-pinning",
+                            "children": [
+                                {
+                                    "type": "item",
+                                    "title": "Static Row Pinning",
+                                    "path": "row-pinning-static"
+                                },
+                                {
+                                    "type": "item",
```
AndrewGlazier: Let's remove Manual and bump this above Static

documentation/ag-grid-docs/src/content/docs-nav/nav.json
```diff
@@ -596,7 +596,20 @@
...
+                                    "title": "Manual Row Pinning",
+                                    "path": "row-pinning-manual",
```
AndrewGlazier: let's repurpose the old row-pinning path so we can avoid issues with needing to add a redirect

documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
```diff
@@ -1410,16 +1410,34 @@
             },
             "description": "Row data uses the [generic type](./typescript-generics) `TData` if provided via `GridOptions&lt;TData&gt;`. Otherwise `TData` will default to `any`."
         },
+        "enableRowPinning": {
```
AndrewGlazier, line 1413: Just to check as it's not here, should the new pinned grand total row option be in this PR?

packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowUtils.ts
```diff
@@ -67,6 +71,18 @@ export class PinnedRows {
         // pre-sort by existing row-index otherwise we'll fall back to order in which rows are pinned
         this.order.sort((a, b) => (a.pinnedSibling?.rowIndex ?? 0) - (b.pinnedSibling?.rowIndex ?? 0));
         this.order = rowNodeSorter?.doFullSort(this.order, sortOptions) ?? this.order;
+        // post-sort to move the grand total rows to the correct place
+        const grandTotalRow = this.beans.gos.get('grandTotalRow');
```
AndrewGlazier, line 78: Nit: rather than running another sort; might be better to just remove the grand total row from the order and add it at head/tail

PR #10223, opened 2025-03-28
AG-14529 - Add grid logger

taktran:
## Changes

* Add React JSON View to display complex objects in logger component + style the same as code viewer
* Create example logger test page
* Extract code color css vars into root design system file
* Update docs example page page to show all examples, with example properties
* Move example runner id to outer container, so it can be scrolled to before the example loads
* Check for console log in all script files including provided examples
* Convert all `alert`s to `console.log`
* Ignore example links in links checker
* Remove `ENABLE_GENERATE_DEBUG_PAGES` flag, as it's no longer needed


markdurrant (approved):
All LGTM 👍

PR #10224, opened 2025-03-28
AG-14564 - support value objects in integrated charts

StephM-W:
https://ag-grid.atlassian.net/browse/AG-14564

### DIFF
--- a/packages/ag-grid-enterprise/src/charts/chartComp/datasource/chartDatasource.ts
+++ b/packages/ag-grid-enterprise/src/charts/chartComp/datasource/chartDatasource.ts
@@ -216,11 +216,15 @@ export class ChartDatasource extends BeanStub {
 
                     // add data value to value column
                     const value = this.valueSvc.getValue(col, rowNode);
-                    let actualValue = value != null && typeof value.toNumber === 'function' ? value.toNumber() : value;
+                    let actualValue = value;
 
-                    // unwrap value object if present
-                    if (actualValue && typeof actualValue.value === 'number') {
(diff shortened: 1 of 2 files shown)


packages/ag-grid-enterprise/src/charts/chartComp/datasource/chartDatasource.ts
```diff
@@ -216,8 +216,12 @@ export class ChartDatasource extends BeanStub {
 
                     // add data value to value column
                     const value = this.valueSvc.getValue(col, rowNode);
-                    const actualValue =
-                        value != null && typeof value.toNumber === 'function' ? value.toNumber() : value;
+                    let actualValue = value != null && typeof value.toNumber === 'function' ? value.toNumber() : value;
+
+                    // unwrap value object if present
+                    if (actualValue && typeof actualValue.value === 'number') {
```
peterjrreynolds: Can we have a situation where `toNumber` returns an object with a `value` property?

If not, can make these checks exclusive (same in chart column service)

StephM-W: It's a user-provided object, so... users could potentially have such toNumber implementation.

peterjrreynolds (approved):

packages/ag-grid-enterprise/src/charts/chartComp/datasource/chartDatasource.ts
StephM-W: Consolidated the codepaths

PR #10250, opened 2025-04-02
AG-7027 API method to iterate pinned rows 

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-7027

### DIFF
--- a/packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowApi.ts
+++ b/packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowApi.ts
@@ -1,4 +1,4 @@
-import type { BeanCollection, IRowNode } from 'ag-grid-community';
+import type { BeanCollection, IRowNode, RowPinnedType } from 'ag-grid-community';
 
 export function getPinnedTopRowCount(beans: BeanCollection): number {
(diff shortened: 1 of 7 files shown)


packages/ag-grid-enterprise/src/manualPinnedRows/manualPinnedRowApi.ts
```diff
@@ -15,3 +15,11 @@ export function getPinnedTopRow(beans: BeanCollection, index: number): IRowNode
 export function getPinnedBottomRow(beans: BeanCollection, index: number): IRowNode | undefined {
     return beans.pinnedRowModel?.getPinnedBottomRow(index);
 }
+
+export function forEachPinnedRow(
```
AndrewGlazier, line 19: Can we also make this work for static pinned rows?

The users requesting this feature are more likely needing this for that scenario, so if we close the FR without capturing that scenario they may not be pleased

eliasmalik: I added it for the existing pinned row model too: https://github.com/ag-grid/ag-grid/pull/10250/files#diff-cca04775b4899a57bfc9bc3feea50c859a78efb2227b4f825c04a02529e91df2

PR #10251, opened 2025-04-02
AG-14630 - fix set filter open from floating filter

peterjrreynolds:


BernieSumption (changes requested):
Use `:where(input)[disabled]` to add conditions to a selector without increasing specificity. Then the stylelint-disable will not be required

PR #10264, opened 2025-04-03
AG-14578 refresh visibility when columns change

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14578?focusedCommentId=81283

### DIFF
--- a/packages/ag-grid-community/src/columns/selectionColService.ts
+++ b/packages/ag-grid-community/src/columns/selectionColService.ts
@@ -35,10 +35,6 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
         });
 
         this.addManagedPropertyListener('selectionColumnDef', this.updateColumns.bind(this));
-
-        this.addManagedEventListeners({
-            displayedColumnsChanged: (event) => this.refreshVisibility(event.source),
-        });
     }
 
     public addColumns(cols: ColumnCollections): void {
@@ -96,12 +92,11 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
 
     public updateColumns(event: PropertyValueChangedEvent<'selectionColumnDef'>): void {
         const source = _convertColumnEventSourceType(event.source);
-        const current = event.currentValue;
 
         this.columns?.list.forEach((col) => {
-            const newColDef = this.createSelectionColDef(current);
+            const newColDef = this.createSelectionColDef(event.currentValue);
             col.setColDef(newColDef, null, source);
-            _applyColumnState(this.beans, { state: [{ colId: col.getColId(), ...newColDef }] }, source);
+            _applyColumnState(this.beans, { state: [{ ...newColDef, colId: col.getColId() }] }, source);
         });
     }
 
@@ -210,7 +205,7 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
         }
 
         const beans = this.beans;
-        const visibleColumns = beans.visibleCols.getAllTrees() ?? [];
+        const visibleColumns = beans.visibleCols.allCols;
 
         if (visibleColumns.length === 0) {
             return;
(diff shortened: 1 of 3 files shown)


packages/ag-grid-community/src/columns/visibleColsService.ts
```diff
@@ -60,28 +60,37 @@ export class VisibleColsService extends BeanStub implements NamedBean {
...
+            this.joinCols();
+        };
+
+        pickCols();
+
+        // Run `pickCols` first because SelectionColService needs the full list of columns first.
+        // If we need to act we just re-run `pickCols` (cheap)
+        const shouldHideCol = selectionColSvc?.shouldHideCol() ?? false;
+        if (shouldHideCol) {
+            pickCols(shouldHideCol);
```
AndrewGlazier: Not keen on this, can you look into if we can do another way please

packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -200,63 +200,60 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
...
-                        state: [{ colId: existingState.colId, hide: !existingState.hide }],
-                    },
-                    source
-                );
-            }
-        }
+        const selectionColLeft = leftCols.find(isColumnSelectionCol);
+        const selectionColCenter = centerCols.find(isColumnSelectionCol);
+        const selectionColRight = rightCols.find(isColumnSelectionCol);
+        const selectionColPresent = selectionColLeft || selectionColCenter || selectionColRight;
```
AndrewGlazier: Nit:
The service uses the shared interface for cols, but can only have 1 col - this is fine

We're now using find here, again leaning into this "1 col" pattern (as if there was more than 1 col this would be a bug right?)

Can we add a comment or two around this svc clarifying that the service only manages 1 col?

The actual nit here, is because we're assuming 1 selection col - we can just use the pinned state on that one col to find whether to search left/right/center, and not need to check all - same applies for row numbers below I think?

On the same note; can we also short-circuit this if the col is !visible in its state? I assume if selection col is !visible, then all of this logic doesn't need run

Won't impact most scenarios - but in the future we may want to support up to a million columns to support some niche features, so good to keep cols fast where possible

packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -200,63 +200,60 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
...
+     * @param rightCols Visible columns in the right-pinned container
+     */
+    public refreshVisibility(leftCols: AgColumn[], centerCols: AgColumn[], rightCols: AgColumn[]): void {
         if (!this.isSelectionColumnEnabled()) {
             return;
         }
 
-        const beans = this.beans;
-        const visibleColumns = beans.visibleCols.getAllTrees() ?? [];
+        const numVisibleCols = leftCols.length + centerCols.length + rightCols.length;
```
AndrewGlazier, line 221: Nit aimed at line 216 (this.isSelectionColumnEnabled()): Would we better/faster just checking if we've already created some selection cols? E.g this.columns.list.length > 0

This might save us from any issues with order of events later on, as we won't need to care about whether they _should_ be enabled here, just that we have some cols so we have to handle them

packages/ag-grid-community/src/columns/selectionColService.ts
```diff
@@ -213,45 +213,47 @@ export class SelectionColService extends BeanStub implements NamedBean, IColumnC
...
+            cols && _removeFromArray(cols, column);
         };
 
+        const rowNumbersCol = this.beans.rowNumbersSvc?.getColumn(ROW_NUMBERS_COLUMN_ID);
+
         // two conditions for which we hide selection column:
         //   1. Only selection column and row numbers column are visible
         //   2. Only selection column is visible
-        if (rowNumbersColPresent && selectionColPresent && numVisibleCols === 2) {
+        if (rowNumbersCol && numVisibleCols === 2) {
```
AndrewGlazier: Nit: Personally think this might be more readable:
Either:
const expectedNumCols = rowNumbersCol ? 2 : 1;
if (numVisibleCols === expectedNumCols) {
 hide

or this if we want it to be potentially more extendable...
let expectedNumCols = 1;
if (rowNumbersCol) {expectedNumCols += 1;}
if (numVisibleCols === expectedNumCols) {
 hide

PR #10273, opened 2025-04-04
Ag 14625 add a migration guide page for v30.0.0 and v29.0.0

sabahang:
QA please test
1- framework specific sections of migration guide
2- algolia search (specially mor common words like "test" etc that appear in migration guides)
3- content indentations and whether each section is appearing under the correct section


documentation/ag-grid-docs/src/content/docs/upgrading-to-ag-grid-29/index.mdoc
```diff
@@ -0,0 +1,350 @@
...
+
+**Typescript Upgrade**
+
+Minimum version of Typescript for AG Grid v29 is v4.0.8.
+
+{% if isFramework("angular") %}
+**Angular Upgrade**
+{% /if %}
+
+{% if isFramework("angular") %}
```
taktran: I think you can just use 1 `{% if %}` code block if the condition is the same and next to each other eg, 

```
{% if isFramework("angular") %}
**Angular Upgrade**

* The minimum Angular dependency supported by AG Grid v29 is now Angular v10 via legacy packages. Previous versions of Angular aren't supported by AG Grid v29. See [Angular Compatibility](https://www.ag-grid.com/archive/29.0.0/angular-data-grid/angular-compatibility/) for more information.
{% /if %}

```

taktran (approved):
Looks good, just a minor improvement comment

documentation/ag-grid-docs/src/content/docs/upgrading-to-ag-grid-29/index.mdoc
sabahang: thanks. resolved and pushed

taktran (approved):

PR #10275, opened 2025-04-04
AG-14636 Move modules validations to core

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-14636


packages/ag-grid-community/src/eventTypes.ts
```diff
@@ -155,7 +155,7 @@ export const _INTERNAL_EVENTS = [
     'pinnedRowsChanged',
 ] as const;
 
-export const _ALL_EVENTS = [..._PUBLIC_EVENTS, ..._INTERNAL_EVENTS] as const;
+export const _ALL_EVENTS = () => [..._PUBLIC_EVENTS, ..._INTERNAL_EVENTS] as const;
```
StephenCooper: This helps esbuild not include all the events which is a couple of kbs.

packages/ag-grid-community/src/validation/rules/colDefValidations.ts
```diff
@@ -25,237 +25,228 @@ const COLUMN_DEFINITION_DEPRECATIONS: () => Deprecations<ColDef | ColGroupDef> =
...
-const COLUMN_DEFINITION_VALIDATIONS: () => Validations<ColDef | ColGroupDef> = () => ({
-    aggFunc: { module: 'SharedAggregation' },
-    autoHeight: {
-        supportedRowModels: ['clientSide', 'serverSide'],
-        module: 'RowAutoHeight',
-    },
-    cellClass: { module: 'CellStyle' },
-    cellClassRules: { module: 'CellStyle' },
-    cellEditor: ({ cellEditor, editable }) => {
+export const COLUMN_DEFINITION_MOD_VALIDATIONS: ModuleValidation<ColDef | ColGroupDef> = {
```
StephenCooper, line 28: Pulled the module validations out from the other forms of validation.

packages/ag-grid-community/src/validation/rules/gridOptionsValidations.ts
```diff
@@ -91,41 +91,76 @@ const GRID_OPTION_DEPRECATIONS = (): Deprecations<GridOptions> => ({
...
+function toConstrainedNum(key: keyof GridOptions, value: any, min: number): string | null {
     if (typeof value === 'number' || value == null) {
         if (value == null) {
             return null;
         }
-
-        if (value >= min && value <= max) {
-            return null;
-        }
-        if (max === Number.MAX_VALUE) {
```
StephenCooper, line 108: Never use the max value so removing.

packages/ag-grid-community/src/validation/validationService.ts
```diff
@@ -33,10 +31,6 @@ export class ValidationService extends BeanStub implements NamedBean {
         provideValidationServiceLogger(this);
     }
 
-    public postConstruct(): void {
-        this.processGridOptions(this.gridOptions);
-    }
```
StephenCooper, line 38: Now called from GridOptionsService which will call through if the validationService is present.

packages/ag-grid-community/src/validation/validationService.ts
```diff
@@ -136,36 +122,6 @@ export class ValidationService extends BeanStub implements NamedBean {
...
-            if (!rulesOrGetter) {
-                return;
-            } else if (typeof rulesOrGetter === 'function') {
-                const fromGetter = rulesOrGetter(options, this.gridOptions, this.beans);
-                if (!fromGetter) {
-                    return;
-                }
-
-                // this is a sub validator.
-                if ('objectName' in fromGetter) {
```
StephenCooper, line 150: Due to the way colDefs are created we run validation on each when they are created. This means we do not need the recursive checking here.

packages/ag-grid-community/src/eventTypes.ts
peterjrreynolds: Could rename this to be a getter so that it's less confusing

StephenCooper: Done

peterjrreynolds (approved):

PR #10284, opened 2025-04-07
AG-14661 Update select all checkbox when rowSelection changes

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14661


packages/ag-grid-community/src/selection/selectAllFeature.ts
```diff
@@ -66,6 +68,7 @@ export class SelectAllFeature extends BeanStub {
             if (getSelectAll(currentValue) !== getSelectAll(previousValue)) {
                 this.showOrHideSelectAll();
             }
+            this.updateStateOfCheckbox();
```
eliasmalik, line 71: This is the actual fix, the rest is tidying.

PR #10285, opened 2025-04-07
AG-14634 add isModuleRegistered method

StephenCooper:


documentation/ag-grid-docs/src/content/docs/modules/_examples/individual-registration/main.ts
```diff
@@ -16,13 +16,17 @@ import {
...
     ColumnMenuModule,
     ContextMenuModule,
     ...(process.env.NODE_ENV !== 'production' ? [ValidationModule] : []),
-]);
+];
+const leftModules = [ClipboardModule, CsvExportModule, SetFilterModule];
+const rightModules = [CsvExportModule, ExcelExportModule, NumberFilterModule, TextFilterModule];
+
+// Register shared Modules globally
+ModuleRegistry.registerModules(sharedModules);
```
StephenCooper, line 29: Updated all versions of the example to log registered modules per grid.

packages/ag-grid-community/src/interfaces/iModule.ts
```diff
@@ -188,7 +188,10 @@ export type EnterpriseModuleName =
     | 'TreeData'
     | 'ViewportRowModel';
 
-export type ModuleName = InternalModuleName | CommunityModuleName | EnterpriseModuleName;
+/** The names of all publicly available AG Grid modules */
+export type AgModuleName = CommunityModuleName | EnterpriseModuleName;
+/** INTERNAL: All public and internal module names */
+export type ModuleName = InternalModuleName | AgModuleName;
```
StephenCooper: Had to add a new type as otherwise users might run into type issues if they try to pass `ModuleName` into the `isRegistered` method. 

This was also motivated by only allowing users to check for publicly registered modules to avoid exposing internal module names which we may change in the future.

plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
```diff
@@ -202,7 +202,7 @@ export async function generateFiles(options: ExecutorOptions, gridOptionsTypes:
             let transformedEntryFile = entryFile;
 
             if (hasExampleConsoleLog) {
-                transformedEntryFile = transformedEntryFile + '\n' + getConsoleLogSnippet({ pageName, exampleName });
+                transformedEntryFile = getConsoleLogSnippet({ pageName, exampleName }) + '\n' + transformedEntryFile;
```
StephenCooper, line 205: This fixes the logger component so that it patches `console.log` before the example code runs so that any top level logs will be directed to the logger component correctly.

PR #10295, opened 2025-04-07
AG-14662 Move setGridApi to an internal interface only

StephenCooper:


packages/ag-grid-react/src/shared/interfaces.ts
```diff
@@ -9,8 +9,12 @@ export interface AgGridReactProps<TData = any> extends GridOptions<TData> {
     modules?: Module[];
     containerStyle?: any;
     className?: string;
-    setGridApi?: (gridApi: GridApi<TData>) => void;
```
StephenCooper, line 12: It is never called on the top level and it is not documented.

PR #10317, opened 2025-04-08
Add missing safe-navigation operator in getOptionalSelectors

amcbain-globys:
Hi!

We're not sure how we triggered this code in our usages, but we found it explodes and upon inspection line 72 assumes that `beans.registry` can be undefined, but line 70 doesn't.

Since one already assumes it could be undefined, I figure it's safe to add it to the other.

Thank you! 🙂

--------

For others hitting this, we RCA'd this to nested [lazy](https://react.dev/reference/react/lazy) React components. [Nesting lazy components does not appear to currently be supported](https://github.com/facebook/react/issues/29235) in any fashion, even with proper caching.

When we attempted to load a lazy component in a tree with a lazy grandparent that also contained our grid, React unmounted and remounted the AgGridReact component, leaving the grid in a useless partially loaded state.

Additionally, while merging this MR makes the immediate error go away, it doesn't solve the grid state issue. Our workaround was to skip lazy for the nested component and directly await a function returning an `import()` call then caching the result.

**ag-grid devs:** Even though our problem has been solved, this is still a potentially useful fix to have. Thanks!

### DIFF
--- a/packages/ag-grid-community/src/gridComp/gridCtrl.ts
+++ b/packages/ag-grid-community/src/gridComp/gridCtrl.ts
@@ -67,7 +67,7 @@ export class GridCtrl extends BeanStub {
         const beans = this.beans;
         return {
             paginationSelector: beans.pagination?.getPaginationSelector(),
-            gridHeaderDropZonesSelector: beans.registry.getSelector('AG-GRID-HEADER-DROP-ZONES'),
+            gridHeaderDropZonesSelector: beans.registry?.getSelector('AG-GRID-HEADER-DROP-ZONES'),
             sideBarSelector: beans.sideBar?.getSelector(),
             statusBarSelector: beans.registry?.getSelector('AG-STATUS-BAR'),
             watermarkSelector: (beans.licenseManager as IWatermark)?.getWatermarkSelector(),


StephenCooper:
This is another workaround for Lazy Cell Components via a wrapper component that handles the loading and prevents it having to suspend though the rest of the grid. 

There is an existing ticket AG-8609 to support lazy cell componets.

https://plnkr.co/edit/wOnA4ROxVZNGsijQ?open=index.jsx

amcbain-globys:
> This is another workaround for Lazy Cell Components via a wrapper component that handles the loading and prevents it having to suspend though the rest of the grid.
> 
> There is an existing ticket AG-8609 to support lazy cell components.
> 
> https://plnkr.co/edit/wOnA4ROxVZNGsijQ?open=index.jsx

In our case the lazy component wasn't a grid cell renderer or inside the grid at all.

Our structure was effectively this:
```html
<Suspense>
  <LazyLoadedParent>
    <Grid />
    <ComponentThatLoadsALazyChild />
  </LazyLoadedParent>
</Suspense>
```

It didn't matter whether we used suspense with our lazy grandchild or not. From [React's own issue tracker](https://github.com/facebook/react/issues/29235), nesting lazy components seems to be unsupported and thus undefined behavior.

Unfortunately, [the React docs](https://react.dev/reference/react/lazy) don't mention anywhere _not_ to nest lazy components and the dev suggestion in the ticket to module level cache lazy components would only prevent extra renders but can't solve the UB when nesting.

As partially called out in my original post, I suspect that loading the lazy grandchild skips straight to the topmost suspense in the tree, causing it to initiate a rerender. At minimum, this results in the AgGridReact component unmounting and remounting, which it doesn't respond positively to as evidenced by the grid returning to a "loading..." state without any columns or data.

Fixing _that_ would be a separate ticket/PR, but since it's caused by UB it's likely not worth the energy.

PR #10322, opened 2025-04-09
AG-14529 - Updates to example logger

taktran:
* Collapse logged object by default
* Make example logger resizable
* Allow `undefined` to be passed through to example logger
* Sort object keys alphabetically
* Add clear button to example logger

### DIFF
--- a/external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
+++ b/external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
@@ -1,7 +1,10 @@
 @use 'design-system' as *;
 
 .loggerOuter {
+    min-height: 34px;
     height: 154px;
+    max-height: 700px;
+    resize: vertical;
     display: flex;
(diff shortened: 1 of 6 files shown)


external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
```diff
@@ -1,7 +1,10 @@
 @use 'design-system' as *;
 
 .loggerOuter {
+    min-height: 34px;
```
markdurrant: I think min-height should be the same as height so the logger can't be collapsed below its current size.

markdurrant (approved):
Code all looks good 👍  One small design note.

external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
taktran: Updated.

PR #10324, opened 2025-04-10
AG-14705 Method renames

StephenCooper:


packages/ag-grid-community/src/environment.ts
```diff
@@ -117,28 +117,17 @@ export class Environment extends BeanStub implements NamedBean {
         return this.getCSSVariablePixelValue(CELL_HORIZONTAL_PADDING);
     }
 
-    public getDefaultIndentation(): number {
-        return this.getCSSVariablePixelValue(INDENTATION_LEVEL);
```
StephenCooper, line 121: Removed single use functions and just inlined them

packages/ag-grid-community/src/rendering/cssClassManager.ts
```diff
@@ -9,24 +9,7 @@ export class CssClassManager {
         this.getGui = getGui;
     }
 
-    public addCssClass(className: string): void {
```
StephenCooper, line 12: Methods not required.

testing/module-size/moduleValidator.ts
```diff
@@ -28,15 +28,16 @@ function validateSizes() {
         // validate that all results their selfSize is less than the expectedSize + 2%
 
         // Some modules are very small and the expected size is very close to the actual size
-        const bufferSize = (expected) => Math.max(expected * 0.03, 1);
+        // But for large modules cap at a 5kb change.
```
StephenCooper, line 31: Make the buffer stricter for big modules.

packages/ag-grid-community/src/rendering/cssClassManager.ts
```diff
@@ -9,24 +9,7 @@ export class CssClassManager {
...
-        const eGui = this.getGui();
-        if (!eGui) {
-            return false;
-        }
-
-        return eGui.classList.contains(className);
-    }
-
-    public addOrRemoveCssClass(className: string, addOrRemove: boolean): void {
+    public updateCss(className: string, addOrRemove: boolean): void {
```
peterjrreynolds: Can we name this (and the `Component` one) `toggleCss`? Makes more sense as then it matches `classList.toggle`.

peterjrreynolds (approved):

PR #10326, opened 2025-04-10
fix: rtl and zoom issue for chrome in version 26.0.0

shayankarimi2000:
used relative solution for rtl and zoom issue on latest version for version 26.0.0 

hi @gportela85, please merge as we are using version 26.0.0 on vue 2
thank u


package.json
shayankarimi2000, line 1: ignore this file it was added for urgent fix of issue on private repo

StephenCooper:
We do not release patches to old versions of AG Grid. If this is still an issue on the latest version please raise an issue with a reproduction.

PR #10329, opened 2025-04-10
AG-14415 - Add repeat message count to example logger

taktran:
## Changes

* Add repeat message count to example logger
* Remove quotes from keys in objects in the example logger

### DIFF
--- a/external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
+++ b/external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
@@ -41,21 +41,44 @@
         background-color: var(--color-gray-900);
         color: var(--color-text-secondary);
     }
+}
 
-    > div {
-        display: flex;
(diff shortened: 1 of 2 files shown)


external/ag-website-shared/src/components/example-runner/components/ExampleLogger.module.scss
```diff
@@ -41,23 +41,46 @@
...
+
+.count {
+    background-color: var(--color-gray-200);
+    color: var(--color-gray-500);
+    border-radius: 50%;
+    padding: 2px 4px 0;
+    font-size: var(--text-fs-2xs);
+    min-width: 2em;
+    height: fit-content;
+    text-align: center;
```
markdurrant: Think we'd want to have a set absolute border-radius to so shape is a 'pill' not an oval.

markdurrant (approved):
All looks good. One design nitpick.

Issue #10338, opened 2025-04-12
[Bug]: Unable to customize drag handle icon


chunkerchunker:
### Link to reproducible scenario

https://plnkr.co/edit/iqVgN46EYzdfj8fa?open=main.ts

### Describe the bug

Steps to reproduce 
1. See attached plunker, based on [first row dragging example from docs](https://www.ag-grid.com/javascript-data-grid/row-dragging/#managed-dragging).  Added lines 29-30, which should customize the drag grip icon for the column (per the icon customization docs).

Actual behaviour
RowDragComp ignores the column def when constructing the icon.

Expected behaviour
Standard drag icon should be replaced with text "test".

More information
Will create a pull request fix.

### Version

33.2.3

### Does the issue occur for a specific framework only?

All languages

### Is the issue only observable on a specific browser?

All browsers

AG-Zoheil:
Hi @chunkerchunker,

Thanks for getting in touch.

To replace individual icons please follow this approach shown here in our theming documentation:
https://www.ag-grid.com/javascript-data-grid/custom-icons/#replacing-icons-example

Kind regards,
Zoheil

chunkerchunker:
@AG-Zoheil I've updated the reproduction plunker to show that the suggestion of replacing the rowDrag icon via the theme doesn't work either.

Note that this is only for the row drag handle; customizing other icons works as expected.

AG-Zoheil:
See this working here: https://plnkr.co/edit/kV429MzJl1jmi1cX?open=main.ts

Please note:

![Image](https://github.com/user-attachments/assets/91a12230-6857-4c6f-a7ba-d48e84b0fdd1)

chunkerchunker:
Apologies!  My mistake on using the wrong icon name.  Thank you for pointing that out.

(very embarrassed)

PR #10339, opened 2025-04-12
Fix for issue #10338: Unable to customize drag handle icon

chunkerchunker:
Component should include column definition when creating the icon, in case the column def includes any icon customizations.


chunkerchunker:
Invalid bug report.

PR #10345, opened 2025-04-14
AG-14665 - Update docs examples page with AG Grid

taktran:


external/ag-website-shared/src/components/docs-examples/utils/transformData.ts
```diff
@@ -0,0 +1,27 @@
+function allPropertiesAreTruthy(entries: Record<string, object>, property: string) {
+    return entries.every(([_, data]) => {
+        return data[property];
+    });
+}
+
+export function transformData(data) {
```
taktran: Probably should move this to `getDocsExampleContents`, closer to the source of the data. The only downside is that it'll break the json endpoint data, but that shouldn't be used by anything other than internal testing.

markdurrant (approved):
LGTM 👍

PR #10356, opened 2025-04-15
AG-14715 - ensureIndexVisible with autoHeight cells

AndrewGlazier:


packages/ag-grid-community/src/gridBodyComp/gridBodyScrollFeature.ts
```diff
@@ -618,6 +618,12 @@ export class GridBodyScrollFeature extends BeanStub {
 
             // so when we return back to user, the cells have rendered
             this.animationFrameSvc?.flushAllFrames();
+
+            if (!this.beans.rowAutoHeight?.areRowsMeasured()) {
+                setTimeout(() => {
```
gportela85, line 623: I think it could be good to limit the amount of times this can be called in recursion to avoid infinite loops.

PR #10357, opened 2025-04-15
AG-2055 - added row resizer to row numbers

gportela85:
https://ag-grid.atlassian.net/browse/AG-2055

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
@@ -2,6 +2,7 @@ import type { ColumnModel } from '../columns/columnModel';
 import { BeanStub } from '../context/beanStub';
 import type { BeanCollection } from '../context/context';
 import type { CtrlsService } from '../ctrlsService';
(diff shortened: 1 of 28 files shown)


packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
```diff
@@ -147,11 +149,18 @@ export class GridBodyCtrl extends BeanStub {
...
+            rowResizeStarted: toggleRowResizeStyle,
+            rowResizeEnded: toggleRowResizeStyle,
         });
 
         this.addManagedPropertyListener('treeData', setGridRootRole);
     }
 
+    private toggleRowResizeStyles(params: RowResizeStartedEvent | RowResizeEndedEvent) {
+        const isResizingRow = params.type === 'rowResizeStarted';
+        this.eBodyViewport.classList.toggle('ag-prevent-animation', isResizingRow);
```
AndrewGlazier, line 161: Why is this necessary? Is `redrawRows(animate: false)` not sufficient? If not, why? This is hard to reason, so if it's needed it's important to add code comments documenting this too

Also consider ag-prevent-row-move-animation or the likes for names, this doesn't impact all animations (ie cell flash)

PR #10364, opened 2025-04-17
AG-14722 - Sync example logger updates from charts

taktran:
From https://github.com/ag-grid/ag-charts/pull/3992


markdurrant (approved):
LGTM 👍

PR #10371, opened 2025-04-22
AG-14757 issues dragging during multiple selection

SalvatorePreviti:
Fix the drag and drop logic to be fully consistent


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1321,7 +1308,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         super.destroy();
 
         // Forcefully deallocate memory
-        this.clearHighlightedRow();
+        this.lastHighlightedRow = null;
```
AndrewGlazier: This is no longer firing an event it used to (i.e clear highlight)

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,107 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
-    // returns false if row was moved, otherwise true
-    public ensureRowsAtPixel(rowNodes: RowNode[], pixel: number, increment: number = 0): boolean {
-        const indexAtPixelNow = this.getRowIndexAtPixel(pixel);
-        const rowNodeAtPixelNow = this.getRow(indexAtPixelNow);
-        const animate = !this.gos.get('suppressAnimationFrame');
-
-        if (rowNodeAtPixelNow === rowNodes[0]) {
-            return false;
+    /** Drag and drop. Returns false if at least a row was moved, otherwise true */
+    public moveRows(rowNodes: RowNode[], target: RowNode | null | undefined, position: 'Above' | 'Below'): boolean {
```
AndrewGlazier: This is only used by drag-n-drop, can you move it to the module?

AndrewGlazier: We use lower camel case for string variables, i.e 'above', 'below'

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,107 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        const allLeafChildrenLen = allLeafChildren?.length;
+        if (!allLeafChildrenLen) {
+            return false; // Destroyed or empty
         }
 
-        const allLeafChildren = this.rootNode?.allLeafChildren;
-        if (!allLeafChildren) {
-            return false;
+        const rowsSet = new Set<ClientSideRowModelRowNode>();
+        let left = allLeafChildrenLen;
```
AndrewGlazier: What is left/right? Can this be named/documented better?

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,107 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+                left = left < idx ? left : idx;
+                right = right > idx ? right : idx;
+            }
+        }
+        if (left > right) {
+            return false; // Nothing to move
         }
 
-        // TODO: this implementation is currently quite inefficient and it could be optimized to run in O(n) in a single pass
+        let mid = target?.sourceRowIndex ?? allLeafChildrenLen;
```
AndrewGlazier: This chunk is hard to consume, can you document/improve the naming of the following block of assignments?

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: This shouldn't raise an event I believe as this happens during destruction? I am not sure why before we were doing it - double checking with Gil as well

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: This comes because the type RowHighlightPosition - changing the type to RowHighlightPosition that already used upper case - in the future we will have 3 possible position, above, below, inside (for tree data) - this code will change

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Did split the code and renamed the fields and added some docs

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Did split the code and renamed the fields and added some docs

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: So, this is very specific to CSRM and we might want to keep it here for now as this code will change for the tree data, as I have already another PR in WIP. It seems too low level to be moved to row dragging module.
Moved there the logic that instead check the pixel position and handle row dragging specific feature, separating the "move" operation from the drag and drop specific logic.

We also agreed with Gil that we should make a row highlight service and move all the highlight code there, as it will require also to support highlight indentation for trees and it doesn't belong to CSRM - as part of the ticket https://ag-grid.atlassian.net/browse/AG-14482

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,133 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        } else if (position === 'Below') {
+            ++insertIdx;
+        }
+        let minIdx = insertIdx;
+        let maxIdx = Math.min(insertIdx, totalRows - 1);
+        for (const { sourceRowIndex } of rows) {
+            if (sourceRowIndex < minIdx) minIdx = sourceRowIndex;
+            if (sourceRowIndex > maxIdx) maxIdx = sourceRowIndex;
         }
+        return [minIdx, insertIdx, maxIdx] as const;
```
AndrewGlazier: min/max is still quite abstract - something like the following might be clearer/help readability?
[firstMovedNodeIdx, targetPositionIdx, lastMovedNodeIdx]

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
AndrewGlazier: Managed drag and drop is all CSRM isn't it? - and CSRM is a more commonly used file that should be stripped of unnecessary logic. Still looks to me like this belongs in managed drag and drop

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,133 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+    /** Reorders the children of the root node, so that the rows to move are in the correct order.
+     * @param rowsToMoveSet The valid set of rows to move, as returned by getValidRowsToMove
+     * @param minIdx The first index of the rows to move
+     * @param targetIdx The target index, where the rows will be moved
+     * @param maxIdx The last index of the rows to move
+     * @returns True if the order of the rows changed, false otherwise
+     */
+    private reorderLeafChildren(
+        rowsToMoveSet: ReadonlySet<ClientSideRowModelRowNode>,
+        minIdx: number,
```
AndrewGlazier: Same issue with the names here - they'd help if they were a bit clearer (see suggested in other comment)

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,133 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        // First partition. Filter from left to right, so the middle can be overwritten
+        for (let i = minIdx; i < targetIdx; ++i) {
+            const row: ClientSideRowModelRowNode = allLeafChildren[i];
+            if (!rowsToMoveSet.has(row)) {
+                if (row.sourceRowIndex !== minIdx) {
+                    row.sourceRowIndex = minIdx;
+                    allLeafChildren[minIdx] = row;
+                    orderChanged = true;
+                }
+                ++minIdx;
```
AndrewGlazier: Can you use a new variable than minIdx? The first index of rows to move isn't changing, and mutating the param could be confusing/error prone

nextLeafIndex or something along those lines may be more easily followed

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,133 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        // Third partition. Filter from right to left, so the middle can be overwritten
+        for (let i = maxIdx; i >= targetIdx; --i) {
+            const row = allLeafChildren[i];
+            if (!rowsToMoveSet.has(row)) {
+                if (row.sourceRowIndex !== maxIdx) {
+                    row.sourceRowIndex = maxIdx;
+                    allLeafChildren[maxIdx] = row;
+                    orderChanged = true;
+                }
+                --maxIdx;
```
AndrewGlazier: Same opinion with this one; mutation of params is confusing/name not really applicable to what it's being used for

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -437,120 +437,133 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
-        if (this.getRow(hoveredRowNode.rowIndex! + diff) === movingRowNode) {
-            return true;
+        // Second partition. Overwrites the middle between the other two filtered partitions
+        for (const row of rowsToMoveSet) {
+            if (row.sourceRowIndex !== minIdx) {
+                row.sourceRowIndex = minIdx;
+                allLeafChildren[minIdx] = row;
+                orderChanged = true;
+            }
+            ++minIdx;
```
AndrewGlazier: New var/not mutation

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
AndrewGlazier: I assume the event fires even during destruction? Though admittedly not sure - if you've tested and it doesn't normally fire then 👍  otherwise it's a breaking change

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: As discussed with Gil, we will try to move this code later, in another PR and see if there are better ways to handle this outside of CSRM once we have a clearer vision on how it integrates with treeData better

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: ok putting it back as it was just in case

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: moved to row dragging module

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Is not necessarily the first move index but first affected node index, renamed.

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -259,25 +265,68 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
-        const cellCtrl = cellPosition && _getCellByPosition(this.beans, cellPosition);
-
-        const rowWasMoved = this.clientSideRowModel.ensureRowsAtPixel(rowNodes, pixel, increment);
-        if (rowWasMoved) {
-            if (cellCtrl) {
-                cellCtrl.focusCell();
-            } else {
-                focusSvc.clearFocusedCell();
+    private moveRows(rowsMove: RowsMove | null): void {
+        if (rowsMove) {
```
gportela85: better to do `if (!rowsMove) { return; }`

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -64,6 +69,12 @@ export interface RowDropZoneParams extends RowDropZoneEvents {
     getContainer: () => HTMLElement;
 }
 
+interface RowsMove {
+    rows: RowNode[];
+    target: RowNode | null | undefined;
+    position: 'Above' | 'Below';
```
gportela85: type should be `RowHighlightPosition`

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -1321,12 +1222,11 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         super.destroy();
 
         // Forcefully deallocate memory
-        this.clearHighlightedRow();
+        this.highlightRow();
```
gportela85: should pass `null` explicitly here just to conform with "cleanup".

PR #10373, opened 2025-04-22
AG-13198 Redraw new pinned rows when calling api.redrawRows

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-13198

### DIFF
--- a/packages/ag-grid-community/src/rendering/rowRenderer.ts
+++ b/packages/ag-grid-community/src/rendering/rowRenderer.ts
@@ -469,10 +469,10 @@ export class RowRenderer extends BeanStub implements NamedBean {
         return res;
     }
 
-    public refreshFloatingRowComps(): void {
-        this.refreshFloatingRows(this.topRowCtrls, 'top');
+    public refreshFloatingRowComps(recycleRows = true): void {
+        this.refreshFloatingRows(this.topRowCtrls, 'top', recycleRows);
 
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/rendering/rowRenderer.ts
```diff
@@ -469,10 +469,10 @@ export class RowRenderer extends BeanStub implements NamedBean {
         return res;
     }
 
-    public refreshFloatingRowComps(): void {
-        this.refreshFloatingRows(this.topRowCtrls, 'top');
+    public refreshFloatingRowComps(recycleRows = true): void {
+        this.refreshFloatingRows(this.topRowCtrls, 'top', recycleRows);
```
gportela85, line 473: when is this called with false being passed as param?

eliasmalik: If `enableRowPinning` is `true` and `redrawAfterModelUpdate` is called with `{ recycleRows: false | undefined }` or `{ domLayoutChanged: true }`. The former happens when called via the `redrawRows` API call.

eliasmalik: I deliberately applied this change to only the new pinned rows because of your comment in the JIRA about the static pinned rows not being part of the row model. This is true when using `pinned(Top|Bottom)RowData`, but technically the manually pinned rows are also (in a sense) part of the normal row model.

PR #10377, opened 2025-04-23
AG-14644 Account for border width in pinned container height calculation

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14644

### DIFF
--- a/packages/ag-grid-community/src/environment.ts
+++ b/packages/ag-grid-community/src/environment.ts
@@ -54,6 +54,12 @@ const ROW_BORDER_WIDTH: Variable = {
     defaultValue: 1,
     border: true,
 };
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/environment.ts
```diff
@@ -105,6 +112,10 @@ export class Environment extends BeanStub implements NamedBean {
         this.addDestroyFunc(() => this.mutationObserver.disconnect());
     }
 
+    public getDefaultPinnedBorderWidth(): number {
```
BernieSumption: Should be `getPinnedRowBorderWidth` - it's not a default (for the other methods using this naming pattern, CSS supplies a default size that can be overridden in code)

BernieSumption (approved):

PR #10389, opened 2025-04-25
AG-14747 Remove default pinned row styles

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14747


BernieSumption (approved):
A query on these. font weight parameters are rare. The test we use for whether to include a parameter is that is must be at least one of:

1. Used by our in-house themes
2. Very commonly used by customers
3. Hard to apply with CSS alone

Do any of the above apply to `pinnedRowFontWeight` and `pinnedSourceRowFontWeight`?

PR #10397, opened 2025-04-28
AG-14780 - Pivot column generation improvement

AndrewGlazier:


packages/ag-grid-community/src/columns/columnGroups/columnGroupService.ts
```diff
@@ -486,7 +486,7 @@ export class ColumnGroupService extends BeanStub implements NamedBean {
         return tree;
     }
 
-    private createMergedColGroupDef(colGroupDef: ColGroupDef | null, groupId: string): ColGroupDef {
+    public createMergedColGroupDef(colGroupDef: ColGroupDef | null, groupId: string): ColGroupDef {
```
StephenCooper: Can we pull this out into utils and just pass gos into it so that it can become a standalone function?

AndrewGlazier: Done

peterjrreynolds (approved):

PR #10417, opened 2025-04-29
AG-14757 fixes for external grid drag drop and improves prev/next target selection

SalvatorePreviti:
The previous implementation was still using a slightly different path for external grid drag and drop, this PR standardises the path for both and keep all consistent also when dragging and dropping from one grid to another

### DIFF
--- a/packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
+++ b/packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
@@ -13,6 +13,7 @@ import type {
 import { _getNormalisedMousePosition } from '../gridBodyComp/mouseEventUtils';
 import { _addGridCommonParams, _getRowIdCallback, _isClientSideRowModel } from '../gridOptionsUtils';
 import type { IClientSideRowModel } from '../interfaces/iClientSideRowModel';
+import type { IRowNode } from '../interfaces/iRowNode';
 import { _last } from '../utils/array';
 import { ChangedPath } from '../utils/changedPath';
 import { _warn } from '../validation/logging';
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -531,10 +493,36 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
+        return rowsDrop.sameGrid ? this.moveRows(rowsDrop) : this.addRows(rowsDrop);
+    }
+
+    private addRows({ delta, target, rows }: RowsDrop): boolean {
+        const getRowIdFunc = _getRowIdCallback(this.gos);
+        const clientSideRowModel = this.clientSideRowModel;
+        const add = rows
+            .filter(
+                ({ data, rowPinned }) =>
+                    !clientSideRowModel.getRowNode(getRowIdFunc?.({ data, level: 0, rowPinned }) ?? data.id)
```
AndrewGlazier, line 506: I appreciate this wasn't done by you/this PR - but might be worth having a look into this, looks like it might be buggy

I don't know of anywhere else we access the data randomly and look for an id field

AndrewGlazier (approved):

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
SalvatorePreviti: looking into it! thank you

SalvatorePreviti: I will add a comment and ask Gil in the future and as we discussed we leave as is right now

PR #10421, opened 2025-04-29
AG-10936 - set state

peterjrreynolds:


packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -58,15 +58,26 @@ export class ServerSideExpansionService extends BaseExpansionService implements
         }
     }
 
-    public expandRows(rowIds: string[]): void {
-        rowIds.forEach((rowId) => {
+    public expandRows(rowIdsToExpand: string[], rowIdsToCollapse?: string[]): void {
+        for (const rowId of rowIdsToExpand) {
```
StephenCooper: Not sure if there is potential to share code here?

documentation/ag-grid-docs/src/content/docs/grid-state/_examples/set-state/provided/angular/app.component.ts
```diff
@@ -0,0 +1,158 @@
...
+        filter: true,
+        enableRowGroup: true,
+        enablePivot: true,
+        enableValue: true,
+    };
+    public rowSelection: RowSelectionOptions = {
+        mode: 'multiRow',
+    };
+    public rowData?: IOlympicData[];
+    public gridVisible = true;
```
StephenCooper: If we make this a signal I suspect we won't need to do the manual change detection calls?

peterjrreynolds: Done

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
peterjrreynolds: Done

PR #10436, opened 2025-05-01
Ag 9780 edit enhancements

AndrewGlazier:


packages/ag-grid-community/src/editing-validation/editingValidationService.ts
```diff
@@ -0,0 +1,26 @@
...
+    beanName = 'editingValidationSvc' as const;
+
+    isValid(cellCtrl: CellCtrl): boolean {
+        const {
+            column,
+            rowCtrl: { rowId },
+        } = cellCtrl;
+
+        const { comp: cellComp, rowNode } = cellCtrl ?? {};
+        const oldValue = this.beans.valueSvc.getValueForDisplay(column, rowNode);
```
AndrewGlazier, line 15: Careful for using getValueForDisplay - this is probably not what you want to validate, valueSvc.getValue more likely to be what needs validated

packages/ag-grid-community/src/editing/editingApi.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { AgColumn } from '../entities/agColumn';
+import { _getCellByPosition } from '../entities/positionUtils';
+import { _getActiveDomElement } from '../gridOptionsUtils';
+import type { GetCellEditorInstancesParams, ICellEditor } from '../interfaces/iCellEditor';
+import type { CellPosition } from '../interfaces/iCellPosition';
+import { _warn } from '../validation/logging';
+import { _resolveControllers, _resolveRowController } from './utils/controllers';
+
+export function undoCellEditing(beans: BeanCollection): void {
+    beans.undoRedo?.undo('api');
```
AndrewGlazier, line 14: Is this func needed? Are we exporting this? If so this API will have dependency on undoRedo svc (ignore if for backwards compat)

packages/ag-grid-community/src/editing/editingModelService.ts
```diff
@@ -0,0 +1,181 @@
...
+    rowId: string;
+    columnId: string;
+    oldValue?: any;
+    newValue?: any;
+};
+
+export class EditingModelService extends BeanStub implements NamedBean {
+    beanName = 'editingModelSvc' as const;
+
+    private rowModels: Map<string, RowEditingModel> = new Map();
```
AndrewGlazier, line 20: From a maintenance perspective might be better to name this rowEditModels to avoid confusion with the row models

packages/ag-grid-community/src/editing/model/rowEditingModel.ts
```diff
@@ -0,0 +1,47 @@
...
+    public removeEditModel(columnId: string): boolean {
+        if (!this.models.has(columnId)) {
+            return false;
+        }
+
+        this.models.delete(columnId);
+        return true;
+    }
+
+    public getEditModels(): CellEditingModel[] {
```
AndrewGlazier, line 33: Unused?

packages/ag-grid-community/src/editing/editingModelService.ts
```diff
@@ -0,0 +1,181 @@
...
+            models.push(...cellModels);
+        }
+        return models;
+    }
+
+    public getEditingCellIds(): CellIdPositions[] {
+        const ids: { rowId: string; columnId: string; oldValue?: any; newValue?: any }[] = [];
+        this.rowModels.forEach((rowModel, rowId) => {
+            rowModel
+                .getEditModels()
```
AndrewGlazier, line 79: This is just building an array and iterating it - no issue with this pattern in general but be warned we quite often fall into scenarios where simply implementing our own iterators/callback fns would have been a better pattern, as the arrays end up being large it can end up causing memory or performance concerns

packages/ag-grid-community/src/editing/editingModelService.ts
```diff
@@ -0,0 +1,181 @@
...
+        if (rowModel.getEditModelCount() === 0) {
+            this.rowModels.delete(rowId);
+        }
+    }
+
+    public getEditModel(rowId: string, columnId: string): CellEditingModel | undefined {
+        return this.rowModels.get(rowId)?.getEditModel(columnId);
+    }
+
+    public getEditModels(rowId?: string | null, columnId?: string | null): CellEditingModel[] {
```
AndrewGlazier, line 51: Again concern with the pattern of building up lots of arrays, destructuring and adding to new arrays etc, if func isn't needed remove, if it is, consider iterators/callback approach?

packages/ag-grid-community/src/editing/editingModelService.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    public override destroy(): void {
+        super.destroy();
+        this.stopEditing();
+    }
+
+    private createRowModel(rowId: string): RowEditingModel {
+        const rowModel = new RowEditingModel();
+        this.rowModels.set(rowId, rowModel);
```
AndrewGlazier, line 178: Note to self; what happens when the underlying data changes and the node no longer exists?

packages/ag-grid-community/src/editing/model/rowEditingModel.ts
```diff
@@ -0,0 +1,47 @@
+import { CellEditingModel } from './cellEditingModel';
+
+export class RowEditingModel {
+    private models: Map<string, CellEditingModel> = new Map();
```
AndrewGlazier, line 4: What happens when columns change?

packages/ag-grid-community/src/editing/editingModelService.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    public getEditingCellPositions(): CellPosition[] {
+        const positions: CellPosition[] = [];
+        const cellIds = this.getEditingCellIds();
+        cellIds.forEach((cell) => {
+            const rowNode = _getRowById(this.beans, cell.rowId);
+            if (rowNode) {
+                positions.push({
+                    column: this.beans.colModel.getCol(cell.columnId)!,
```
AndrewGlazier, line 92: If we need to do this; preserve col ref rather than colId?

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -165,13 +170,26 @@ export class ValueService extends BeanStub implements NamedBean {
         const colId = column.getColId();
         const data = rowNode.data;
 
+        const { editingSvc, rowRenderer, rowGroupColsSvc } = this.beans;
+
+        if (editingSvc && source === 'ui') {
```
AndrewGlazier, line 175: valueService is a very _very_ hot path, and I'm concerned that filtering for specific row/cell ctrls is going to be slow. Also not entirely understanding the source === ui requirement, if user edits a value should they not be sorting by that new value for some reason? If not, this logic should maybe actually live in getValueForDisplay?

Same concern though - why do we need to verify if the cell is editing? Am I misunderstanding what 'editing' is representing?

PR #10439, opened 2025-05-01
AG-832 - SSRM continue ensureIndexVisible scroll unless interrupted

AndrewGlazier:


packages/ag-grid-community/src/gridBodyComp/gridBodyScrollFeature.ts
```diff
@@ -623,6 +623,38 @@ export class GridBodyScrollFeature extends BeanStub {
                 setTimeout(() => {
                     this.ensureIndexVisible(index, position, retry + 1);
                 }, 0);
+                return;
+            }
+
+            // SSRM - if the node is a stub, give the grid a chance to load the data
+            // when data loads, try again to scroll to the row.
+            // Cancel if any other scroll event occurs.
+            if (rowNode?.stub) {
```
gportela85, line 638:
```
if (!rowNode?.stub) {
   return;
}

gportela85: Also, this code will add multiple listeners every time `ensureIndexVisible` is called and will pile them up until the scrollFeature is destroyed.

PR #10444, opened 2025-05-02
AG-14481 - Prevent flash of nav for what's new page

taktran:
Allow docs nav framework to be undefined, and use redirect link if so.

Prevents flash of empty nav for pages like what's new, 404 etc.


markdurrant (approved):
LGTM 👍

PR #10449, opened 2025-05-02
AG-14371 - Improve performance of loading logo when not on the page

taktran:
## Changes

* Don't animate loading logo when it isn't visible
  * Create new `ag-grid-logomark-not-loading.svg` that does not animate by default
  * Move loading logo logic into `LoadingLogo` component
  * Move files into shared folder


StephM-W (approved):

markdurrant (approved):
LGTM 👍

PR #10464, opened 2025-05-08
Ag 14236 pricing page scroll improvements

markdurrant:
- Make sure buy buttons are visible on pricing page as the user scrolls down


taktran (approved):
👍🏼

PR #10465, opened 2025-05-08
AG-3390 - Add tsx to example mime types

taktran:


documentation/ag-grid-docs/src/utils/htaccess/htaccessRules.ts
```diff
@@ -129,6 +129,7 @@ ErrorDocument 404 /404.html
 
 # add MIME types for serving example files
 AddType text/javascript ts jsx
+AddType application/typescript tsx
```
jacobp100: We already had TS classed as JS

```suggestion
AddType text/javascript ts tsx jsx
```

taktran: From a discussion, it feels more correct to go with `application/typescript` for `ts` as well, since it’s not as standard as javascript, and we should be consistent with `ts` and `tsx`

PR #10469, opened 2025-05-08
AG-11914-propagate-nonce-option

StephM-W:
https://ag-grid.atlassian.net/browse/AG-14840


documentation/ag-grid-docs/src/content/docs/security-test/_examples/combination-chart-csp/head.html
```diff
@@ -0,0 +1,4 @@
+<meta
+    http-equiv="Content-Security-Policy"
+    content="default-src 'self' localhost 'nonce-ABCDEF'; style-src 'self' 'unsafe-inline'"
```
alantreadway, line 3: If a `styleNonce` is in use (seems to be?), we probably want to remove `'unsafe-inline'` here as that allows `<style>` tags without a `nonce`?

documentation/ag-grid-docs/src/content/docs/security-test/_examples/line-sparkline-customisation-csp/head.html
```diff
@@ -0,0 +1,4 @@
+<meta
+    http-equiv="Content-Security-Policy"
+    content="default-src 'self' localhost 'nonce-ABCDEF'; style-src 'self' 'unsafe-inline'"
```
alantreadway, line 3: Ditto

SalvatorePreviti (approved):

documentation/ag-grid-docs/src/content/docs/security-test/_examples/combination-chart-csp/head.html
StephM-W: We need these in the example as the grid docs don't (yet) add the nonce to inlined generated scripts

documentation/ag-grid-docs/src/content/docs/security-test/_examples/line-sparkline-customisation-csp/head.html
StephM-W: As above

PR #10481, opened 2025-05-11
Ag 14859 campaign page update

markdurrant:


documentation/ag-grid-docs/src/pages/campaigns/power-of-ag-charts.astro
seanlandsman, line 1: Please add

```
hidePageFromSearchEngines={true}
```

to the `Layout` component - to prevent google from indexing it

seanlandsman (changes requested):

documentation/ag-grid-docs/src/pages/campaigns/power-of-ag-charts.astro
markdurrant: Prop added.

seanlandsman (approved):

PR #10506, opened 2025-05-12
RTI-2717 Add module tooltip content label

kylerphillips:
Screenshot 2025-05-12 at 14 51 54

### DIFF
--- a/documentation/ag-grid-docs/src/components/reference-documentation/components/PropertyModules.module.scss
+++ b/documentation/ag-grid-docs/src/components/reference-documentation/components/PropertyModules.module.scss
@@ -92,6 +92,7 @@
 .moduleTooltipContent {
     padding: 0;
     max-width: 400px;
+    padding-top: $spacing-size-2;
 
     > a {
         --icon-color: var(--color-link);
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/src/components/reference-documentation/components/PropertyModules.module.scss
```diff
@@ -120,3 +121,15 @@
...
+    text-transform: uppercase;
+    font-family: 'SF Pro Display', sans-serif;
+    font-size: 12px;
+    font-weight: 600;
+    color: var(--color-text-secondary);
+    margin-bottom: $spacing-size-2;
+    padding-left: $spacing-size-2;
+    opacity: 0.7;
+    padding-top: $spacing-size-4;
+}
```
markdurrant: I think we can use the standards/variables in a few places here. 

`font-family`, `font-size`, `font-weight`. 

Maybe use another colour instead of opacity.

markdurrant (changes requested):
Design looks good. Can improve code a bit by using our variables.

StephenCooper (approved):

markdurrant (approved):
LGTM 👍

PR #10515, opened 2025-05-12
Rti 2731 firefox shows video in wrong aspect ratio

SalvatorePreviti:
Setting display-block and width 100% fixes the problem


markdurrant (approved):
LGTM 👍

PR #10518, opened 2025-05-12
RTI-2729-tooltips-at-the-bottom-of-examples-do-not-point-directly-at-their-icons

SalvatorePreviti:
The problem appears to be due to the fact that .cta is by default an inline element, and so it will have no width when it contains a block element.
Setting it to inline-block fix the problem. Checked places where cta is used (link buttons) and seems that all is fine


markdurrant (approved):
LGTM 👍

PR #10532, opened 2025-05-13
[WIP] AG-14345 Conditional rendering of header checkboxes

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14345


eliasmalik:
maybe best to defer until the old header checkbox options are removed from colDef

PR #10542, opened 2025-05-14
AG-304 Feature(ag-grid): dateTime and dateTimeString types

sdwvit:
This pull request introduces support for the 'dateTime' and 'dateTimeString' cell data types in AG-Grid:

### **Column Definitions**
- Updated column definitions to recognize 'dateTime' and 'dateTimeString'.

### **Filtering**
   - Added new filter options for 'dateTime' and 'dateTimeString' data types across components.
   - Enhanced tree list filtering to handle date-time values.

### **Code changes**
   - Added single-source-of-truth methods within the `DataTypeService` for 'dateTime' and 'dateTimeString'.

### **UI Components**
   - Upgraded filters, pickers, and other components to support time-aware data types.

### **Documentation Updates**
   - Added new data types to the documentation.
   - Provided examples demonstrating their usage in column definitions.

image
   
### **Enterprise-Specific Changes**:
   - Updated advanced filters to include new operators and display value models for time-aware data types.

Fixes [AG-14909](https://ag-grid.atlassian.net/browse/AG-14909) and maybe [AG-304](https://ag-grid.atlassian.net/browse/AG-304)

Test: https://plnkr.co/edit/7LbCplSa3miT463H?open=main.ts&preview and set date filter https://plnkr.co/edit/uMQYuVlEWIeP8hPR?open=main.ts&preview


documentation/ag-grid-docs/src/content/docs/cell-data-types/_examples/enable-cell-data-types/main.ts
```diff
@@ -75,14 +79,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
                     return {
                         ...rowData,
-                        date: `${dateParts[2]}-${dateParts[1]}-${dateParts[0]}`,
-                        dateObject: new Date(
-                            parseInt(dateParts[2]),
-                            parseInt(dateParts[1]) - 1,
-                            parseInt(dateParts[0])
-                        ),
+                        date: dateString,
+                        dateObject: dateTimeObj,
```
peterjrreynolds: `dateObject` needs to be populated with a `Date` without any time, otherwise the date filter won't work.

documentation/ag-grid-docs/src/content/docs/cell-data-types/_examples/pre-defined-cell-data-types/main.ts
```diff
@@ -92,14 +96,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
                     return {
                         ...rowData,
-                        date: `${dateParts[2]}-${dateParts[1]}-${dateParts[0]}`,
-                        dateObject: new Date(
-                            parseInt(dateParts[2]),
-                            parseInt(dateParts[1]) - 1,
-                            parseInt(dateParts[0])
-                        ),
+                        date: dateString,
+                        dateObject: dateTimeObj,
```
peterjrreynolds: Same as the other example

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
```diff
@@ -26,26 +26,32 @@ const gridOptions = {
...
 
-The following example demonstrates the six pre-defined cell data types (inferred from the row data):
+The following example demonstrates the pre-defined cell data types (most of which are inferred from the row data):
 
 * The **Athlete** column has a `'text'` data type.
 * The **Age** column has a `'number'` data type.
 * The **Gold** column has a `'boolean'` data type.
 * The **Date** column has a `'date'` data type (cell values are `Date` objects).
 * The **Date (String)** column has a `'dateString'` data type (cell values are `string`s representing dates).
+* The **DateTime** column has a `'dateTime'` data type (cell values are `Date` objects, [and developer has to explicitly set `cellDataType: 'dateTime'`](#enable-cell-data-types), otherwise data type is set to just `'date'`).
```
peterjrreynolds:
```suggestion
* The **DateTime** column has a `'dateTime'` data type (cell values are `Date` objects). This is explicitly set to `cellDataType: 'dateTime'` as `Date` objects are always inferred to be `'date'` data type.
```
The docs assume the reader is a developer, so we wouldn't refer to them directly. Also, link text should be simple.

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
```diff
@@ -26,26 +26,32 @@ const gridOptions = {
...
 
 ## Inferring Data Types
 
-By default the grid will infer cell data types the first time that row data is passed into the grid.
+By default, the grid will infer cell data types the first time that row data is passed into the grid.
 
 For inference to work for a column, it must contain non-null values and have the `field` property set. The resolved column definition (including the default column definition and column types) must also not have the Value Getter, Value Parser or reference data properties set, or be using [Sparklines](./sparklines-overview/). If these conditions are not met, no cell data type will be set (it will need to be defined directly on the column if desired).
 
+{% note %}
+Because `'dateTime'` corresponds to cell values that are `Date` objects, there is no easy way to tell them apart from regular `'date'` columns. If you wish to enable `'dateTime'`'s higher precision fields, please [explicitly specify `cellDataType: 'dateTime'`](#enable-cell-data-types) in the corresponding `columnDefs` entry.
```
peterjrreynolds:
```suggestion
Because `'dateTime'` corresponds to cell values that are `Date` objects, there is no easy way to tell them apart from regular `'date'` columns. If you wish to enable `'dateTime'`'s higher precision fields, please explicitly specify `cellDataType: 'dateTime'` in the corresponding `columnDefs` entry. 
```
Simple link text (don't really need the link here)

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
```diff
@@ -111,7 +117,31 @@ The following properties are set:
...
+* When the [Set Filter is Disabled by Default](./filter-set/#suppress-set-filter-by-default), the [Date Filter](./filter-text/) is used, and `filterParams.comparator` is set to parse the `string` date values.
+* When the [Set Filter is Enabled by Default](./filter-set/#suppress-set-filter-by-default), the [Set Filter Tree List](./filter-set-tree-list/) is enabled, with `filterParams.treeListPathGetter` set to convert the `string` date values into paths, and the [Values are Formatted](./filter-set-tree-list/#formatting-values) by setting `filterParams.treeListFormatter` to convert the months to names and `filterParams.valueFormatter` to format the Floating Filter values using the Value Formatter.
+
+### DateTime
+
+The `'dateTime'` cell data type is used for date and time values that are represented as `Date` objects. Unlike the `'date'` cell data type which only shows the date portion, `'dateTime'` displays both date and time components.
+
+This data type uses the string format `'YYYY-MM-DD hh:mm:ss'`. If you wish to use a different format, you can [Override the Pre-Defined Cell Data Type Definition](#overriding-the-pre-defined-cell-data-type-definitions).
```
peterjrreynolds: Are we not using the ISO format?

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
```diff
@@ -231,8 +261,8 @@ The following example demonstrates overriding pre-defined cell data types:
 
 {% gridExampleRunner title="Overriding Pre-Defined Cell Data Types" name="overriding-pre-defined-cell-data-types" /%}
 
-### Date as String Data Type Definition
+### Specific case for Date and DateTime as String Data Type Definition
```
peterjrreynolds:
```suggestion
### Date and DateTime as String Data Type Definition
```
Keep headings short where possible as the right-hand nav gets messy otherwise

documentation/ag-grid-docs/src/content/docs/filter-advanced/_examples/advanced-filter/main.ts
```diff
@@ -64,14 +68,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
                     return {
                         ...rowData,
-                        date: `${dateParts[2]}-${dateParts[1]}-${dateParts[0]}`,
-                        dateObject: new Date(
-                            parseInt(dateParts[2]),
-                            parseInt(dateParts[1]) - 1,
-                            parseInt(dateParts[0])
-                        ),
+                        date: dateString,
+                        dateObject: dateTimeObj,
```
peterjrreynolds: See previous comment

packages/ag-grid-community/src/columns/dataTypeService.ts
```diff
@@ -510,17 +545,71 @@ export class DataTypeService extends BeanStub implements NamedBean {
...
+        const formatValue = this.formatValueFuncs[cellDataType];
+        const partialColDef = this.columnDefinitionPropsPerDataType[dataTypeDefinition.baseDataType]({
+            colDef,
+            cellDataType,
+            colModel: this.colModel,
+            dataTypeDefinition,
+            colId,
+            formatValue,
+        });
+        Object.assign(colDef, partialColDef); // todo prefer object copy instead of mutation
```
peterjrreynolds: Can remove todo.

We create col defs by applying in turn:
1. default col def
2. data type props
3. column types
4. col def

If we have a lot of columns (e.g. pivoting), and we created a copy each time, this would be expensive.

packages/ag-grid-community/src/entities/dataType.ts
```diff
@@ -109,32 +121,50 @@ export interface BooleanDataTypeDefinition<TData = any> extends BaseDataTypeDefi
 export interface DateDataTypeDefinition<TData = any> extends BaseDataTypeDefinition<'date', TData, Date> {}
 
 /** Represents a `'dateString'` data type (type `string` that represents a date). */
-export interface DateStringDataTypeDefinition<TData = any> extends BaseDataTypeDefinition<'dateString', TData, string> {
+export interface DateStringDataTypeDefinition<TData = any, K extends BaseCellDataType = 'dateString'>
```
peterjrreynolds: I think this interface should be renamed and made private, and then exposed as another interface, e.g.
```
export interface DateStringDataTypeDefinition<TData = any> extends BaseDateStringDataTypeDefinition<TData, 'dateString'> {}

interface BaseDateStringDataTypeDefinition <TData = any, K extends BaseCellDataType = 'dateString'>
...
```

Otherwise we're exposing the second generic type param to users, and they shouldn't need to know anything about it.

packages/ag-grid-community/src/entities/dataType.ts
```diff
@@ -109,32 +121,50 @@ export interface BooleanDataTypeDefinition<TData = any> extends BaseDataTypeDefi
...
+
 /** Configuration options for a cell data type. */
-export type DataTypeDefinition<TData = any> =
-    | TextDataTypeDefinition<TData>
-    | NumberDataTypeDefinition<TData>
-    | BooleanDataTypeDefinition<TData>
-    | DateDataTypeDefinition<TData>
-    | DateStringDataTypeDefinition<TData>
-    | ObjectDataTypeDefinition<TData, any>;
+export type DataTypeDefinition<TData = any, TValue = any> = ValueOf<_DataTypeDefMap<TData, TValue>>;
```
peterjrreynolds: Note that if you expand `dataTypeDefinitions` on the docs API widget, you see all these types.

Stuff done for the purpose of type checking should be private, and we should just expose a simple union.

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -54,14 +57,14 @@ const MONTH_KEYS: (keyof typeof MONTH_LOCALE_TEXT)[] = [
...
-function setFilterNumberComparator(a: string, b: string): number {
+function setFilterNumberComparator<TValue = any>(a: TValue | null, b: TValue | null): number {
     if (a == null) {
         return -1;
     }
     if (b == null) {
         return 1;
     }
-    return parseFloat(a) - parseFloat(b);
+    return parseFloat(String(a)) - parseFloat(String(b));
```
peterjrreynolds: What was the motivator for changing this?

This is only used for number data type, so `a` and `b` should actually only ever be numbers here.

It's in the comparator, so there is a performance impact to this change.

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -456,40 +459,39 @@ export class ColumnFilterService extends BeanStub implements NamedBean {
...
-            if (cellDataType === 'number') {
-                defaultFilter = 'agNumberColumnFilter';
-            } else if (cellDataType === 'date' || cellDataType === 'dateString') {
-                defaultFilter = 'agDateColumnFilter';
-            } else {
-                defaultFilter = 'agTextColumnFilter';
-            }
-        }
-        return defaultFilter;
+    private defaultFilters: Record<BaseCellDataType | 'set', UserComponentName> = {
```
peterjrreynolds: Move properties to the top of the class with the other properties

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -456,40 +459,39 @@ export class ColumnFilterService extends BeanStub implements NamedBean {
...
+        boolean: 'agTextColumnFilter',
+        date: 'agDateColumnFilter',
+        dateString: 'agDateColumnFilter',
+        dateTime: 'agDateColumnFilter',
+        dateTimeString: 'agDateColumnFilter',
+        number: 'agNumberColumnFilter',
+        object: 'agTextColumnFilter',
+        set: 'agSetColumnFilter',
```
peterjrreynolds: Might be neater if you just kept this to data types, and left the set logic in the method

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -886,148 +888,156 @@ export class ColumnFilterService extends BeanStub implements NamedBean {
...
+            formatValue,
+            t: this.getLocaleTextFunc(),
+            valueSvc: this.beans.valueSvc,
+        });
+        if (filterParamsOrNothing) {
+            if (typeof filterParamsOrNothing === 'object') {
+                colDef.filterParams =
+                    typeof colDef.filterParams === 'object'
+                        ? { colDef, ...colDef.filterParams, ...filterParamsOrNothing }
+                        : { colDef, ...filterParamsOrNothing };
```
peterjrreynolds: Shouldn't be setting the `colDef` in the filter params here. This can just mimic the previous logic.

packages/ag-grid-community/src/filter/columnFilterService.ts
```diff
@@ -886,148 +888,156 @@ export class ColumnFilterService extends BeanStub implements NamedBean {
...
+            } else {
+                colDef.filterValueGetter = filterParamsOrNothing;
```
peterjrreynolds: I'm not a fan of this approach

packages/ag-grid-community/src/filter/columnFilterService.ts
peterjrreynolds: The issue here is actually that `DateFloatingFilter.getDateComponentParams` should be passing `filterParams` instead of `column.getColDef().filterParams` as the `filterParams` (the former includes all the grid-provided params like colDef, whereas the latter only includes the user-provided params).

packages/ag-grid-community/src/interfaces/advancedFilterModel.ts
```diff
@@ -98,8 +97,60 @@ export interface ObjectAdvancedFilterModel {
...
+    date: (args: FilterParamsDefArgs) => ISetFilterParams<any, Date> | IDateFilterParams;
+    dateString: (args: FilterParamsDefArgs) => ISetFilterParams | IDateFilterParams;
+    dateTime: (args: FilterParamsDefArgs) => ISetFilterParams<any, Date> | IDateFilterParams;
+    dateTimeString: (args: FilterParamsDefArgs) => ISetFilterParams | IDateFilterParams;
+    object: (args: FilterParamsDefArgs) => ISetFilterParams<any, any> | ValueGetterFunc;
+    text: () => void;
+}>;
+
+/** Represents a single filter condition on a column */
+export type ColumnAdvancedFilterModel = FilterModelDefMap[keyof FilterModelDefMap];
```
peterjrreynolds: Similar issue to the data type definitions

packages/ag-grid-community/src/utils/date.ts
```diff
@@ -29,6 +29,28 @@ export function _serialiseDate(date: Date | null, includeTime = true, separator
...
+
+    const parts = [d.getFullYear(), d.getMonth() + 1, d.getDate(), d.getHours(), d.getMinutes(), d.getSeconds()].map(
+        String
+    );
+
+    if (includeTime) {
+        return parts;
+    }
+
+    return parts.slice(0, 3);
```
peterjrreynolds: Note that dates are more commonly used than datetimes, and creating intermediate arrays can have a performance impact.

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
sdwvit: ISO format is YYYY-MM-DD**T**hh:mm:ss**Z** (2025-05-23T14:26:58Z)

packages/ag-grid-community/src/entities/dataType.ts
sdwvit: Do I understand correctly that only exported stuff is visible in docs?

packages/ag-grid-community/src/filter/columnFilterService.ts
sdwvit: This is done only for correct typings, thank you for spotting the performance implication. Will revert

packages/ag-grid-community/src/filter/columnFilterService.ts
sdwvit: Elaborate please? This is [replicating the previous switch/case logic](https://github.com/ag-grid/ag-grid/pull/10542/files#diff-88ca7f3951a74023a6ecee154d8a469291c97c7c7ef0a6fe7c9291090f40c6e4L1015-L1029). The "not fun" part I can spot is the reliance on typeof filterParams === 'object' to decide strategy: it doesn't look clean to me neither, and enables future maintainer to make a mistake (hard to do, but still). 

If you confirm, I can explore ways to get rid of it.

packages/ag-grid-community/src/utils/date.ts
sdwvit: Gotcha, thank you for spotting this. Will refactor.

documentation/ag-grid-docs/src/content/docs/cell-data-types/_examples/enable-cell-data-types/main.ts
sdwvit: well spotted, mind if I mention it in the docs? with the new date types, I foresee it may be a confusion for users. I can add a line about this behavior in this PR

packages/ag-grid-community/src/columns/dataTypeService.ts
sdwvit: Good insight, thank you

packages/ag-grid-community/src/filter/columnFilterService.ts
sdwvit: These do not depend on class instance. Okay to move them outside of class? see updated code

packages/ag-grid-community/src/entities/dataType.ts
sdwvit: I think this is that case when extra abstraction actually hurts. Refactored using simpler approach

packages/ag-grid-community/src/entities/dataType.ts
```diff
@@ -116,25 +128,47 @@ export interface DateStringDataTypeDefinition<TData = any> extends BaseDataTypeD
...
-    | Omit<TextDataTypeDefinition<TData>, 'extendsDataType'>
-    | Omit<NumberDataTypeDefinition<TData>, 'extendsDataType'>
-    | Omit<BooleanDataTypeDefinition<TData>, 'extendsDataType'>
-    | Omit<DateDataTypeDefinition<TData>, 'extendsDataType'>
-    | Omit<DateStringDataTypeDefinition<TData>, 'extendsDataType'>
-    | Omit<ObjectDataTypeDefinition<TData, any>, 'extendsDataType'>;
+export type CoreDataTypeDefinition<TData = any, TValue = any> = Omit<
+    DataTypeDefinition<TData, TValue>,
+    'extendsDataType'
+>;
```
sdwvit, line 166: @peterjrreynolds let me know if this is fine, or should I also fall back to union

peterjrreynolds: This is only used internally, so fine

peterjrreynolds (approved):

PR #10557, opened 2025-05-16
AG-14671 Autosizing selection column

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14671

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_widgets.scss
+++ b/community-modules/styles/src/internal/base/parts/_widgets.scss
@@ -157,6 +157,18 @@
         color: var(--ag-secondary-foreground-color);
     }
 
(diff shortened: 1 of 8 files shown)


community-modules/styles/src/internal/base/parts/_widgets.scss
```diff
@@ -157,6 +157,18 @@
         color: var(--ag-secondary-foreground-color);
     }
 
+    [col-id='ag-Grid-SelectionColumn'] .ag-selection-checkbox {
```
AndrewGlazier: Probably need one of Gil/Stephen/Bernie to comment on this specifically - not sure we want to be using col ids for styling

PR #10563, opened 2025-05-16
AG-14845 - Reintroduce console logger

taktran:


documentation/ag-grid-docs/src/content/docs/csv-export/_examples/csv-export-header-footer/main.ts
```diff
@@ -99,11 +99,6 @@ function getParams() {
 
 function onBtnExport() {
     const params = getParams();
-    if (params.suppressQuotes || params.columnSeparator) {
-        alert(
-            'NOTE: you are downloading a file with non-standard quotes or separators - it may not render correctly in Excel.'
-        );
-    }
```
markdurrant, line 106: Alert is removed but not replaced by `console.log`?

markdurrant (approved):
One question, otherwise LGTM 👍

documentation/ag-grid-docs/src/content/docs/csv-export/_examples/csv-export-header-footer/main.ts
taktran: Yep, from above, it has

```
        suppressQuotes: undefined,
        columnSeparator: undefined,
```

So this is never called

peterjrreynolds (approved):

PR #10567, opened 2025-05-19
AG-14345 Making SelectAllFeature optional

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14345


packages/ag-grid-community/src/context/genericContext.ts
```diff
@@ -83,13 +83,10 @@ export class GenericContext<TBeanName extends string, TBeanCollection extends {
         return Object.values(this.beans);
     }
 
-    public createBean<T extends GenericBean<TBeanName, TBeanCollection> | null | undefined>(
+    public createBean<T extends GenericBean<TBeanName, TBeanCollection>>(
         bean: T,
         afterPreCreateCallback?: (bean: GenericBean<TBeanName, TBeanCollection>) => void
     ): T {
-        if (!bean) {
```
eliasmalik, line 90: This is an unrelated change that I made in the process of this PR. I think it's safe, but if it's considered risky then I'm happy to revert.

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -62,7 +62,10 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
             }
         });
 
-        this.addManagedEventListeners({ rowSelected: this.onRowSelected.bind(this) });
+        this.addManagedEventListeners({
+            rowSelected: this.onRowSelected.bind(this),
+            rowDataUpdated: () => this.updateSelectable(),
```
StephenCooper: Is this a required part of this change? What is this solving?

packages/ag-grid-community/src/context/genericContext.ts
StephenCooper: Happy with this. It would have been a blocker for all users if this got hit anyway.

StephenCooper (approved):

packages/ag-grid-community/src/selection/selectionService.ts
eliasmalik: Oh strange, this is a change from a different branch, not sure how it snuck in, will remove

PR #10580, opened 2025-05-20
AG-10936 - set state partial

peterjrreynolds:


packages/ag-grid-community/src/interfaces/gridState.ts
```diff
@@ -188,6 +188,9 @@ export interface GridState {
     /**
      * When providing a partial `initialState` with some but not all column state properties, set this to `true`.
      * Not required if passing the whole state object retrieved from the grid.
+     * Not used for `api.setState()`.
```
StephenCooper: Tell user to use ignore params instead as part of api.setState.

StephenCooper (approved):
Feel free to merge once comment updated.

PR #10587, opened 2025-05-21
AG-14395 treedata path drag drop logic

SalvatorePreviti:
Main changes:
- fixes RowDragFeature to handle trees with path (as they can contain filler nodes)
- Removes the old PathClientSideTreeRowManager and the associated TreeNode, treeRow
- changes the CSRM as there are only two node managers now, clientSideChildrenTreeNodeManager to load the children array, or the standard clientSideNodeManager
- Moves all the tree logic for all the 3 input modes (nested, children, path) all in the TreeGroupStrategy stage, integrating path trees with the other two existing modes
- Provide a new O(1) way to get fille nodes in getRowNode by delegating looking for groups by id to the grouping stage. The groupStage stays unchanged, the improvement is only for trees.

This approach allows also to make drag and drop works properly also if there are nodes with the same key dragged around. And this is an allowed operation, as the user might want to swap two rows, or doesn't care about keys once the data is loaded.
The only limitation is that there cannot be empty fillers, so if all nodes are dragged out from a filler node, the filler node will be deleted.

Despite this is a big PR, this in reality affects only trees and row dragging and do not have impacts in any other feature.

Removing PathClientSideTreeRowManager removed approximately 1000 of code, as the logic in the TreeGroupStrategy is more linear. Description of the approach used in the file itself as a comment.
This also adds a pretty noticeable performance step up:

New benchmark in this branch:
image

Old benchmark in latest:
image

hz is the number of rebuild/update tree per second and note that benchmarks here uses JSDOM that is much slower than a real browser


packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
```diff
@@ -23,57 +24,74 @@ export class GroupStage<TData> extends BeanStub implements NamedBean, IRowNodeSt
...
     ]);
+
     public step: ClientSideRowModelStage = 'group';
 
     private approach: GroupingApproach | null = null;
+    private strategyBeanName: string | null = null;
     private strategy: IRowGroupingStrategy<TData> | undefined = undefined;
 
+    /** Gets a filler row by id */
+    public getFiller(id: string): RowNode<TData> | undefined {
```
AndrewGlazier: Nit - these are only "filler"'s in the context of tree data, would be better to just call this `getNode`

AndrewGlazier: Not sure about using RowNodeStages as node managers from a conceptual/design POV

@gportela85 do you think this is fine?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -193,18 +179,21 @@ export class RowNode<TData = any>
...
     public allChildrenCount: number | null;
 
-    /** Children mapped by the pivot columns or group key */
-    public childrenMapped: { [key: string]: any } | null = null;
+    /**
+     * - For PivotStage is a record object, containing the children mapped by the pivot columns or group key
+     * - For GroupStrategy is a Record<string, RowNode<TData>>, containing the children of this group mapped by the group key.
+     * - For TreeStrategy is used to temporarily store the path to the node from the root, when using getDataPath
+     */
+    public childrenMapped: { [key: string]: any } | string | null = null;
```
AndrewGlazier: Not a fan of extending this type for a temporary measure, as it complicates using this field for anything outside of tree data (as already seen by all the casting that's now required) - is there anywhere else you can temporarily store this string when building trees?

Worst case scenario I think a bespoke variable would be better - but as this is just used within tree construction, finding a way to maintain this within that flow/service would be better if possible

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: getGroupNode maybe?

SalvatorePreviti: Another option is to move the management of filler nodes in the node manager, that I would personally prefer, but it would be much many more changes in CSRM of course and we could do it in another stage. Is a bit more risky as a change as there are more changes in CSRM, and it will not be tree shaken also if it would be a very small amount of code

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: Yes I don't want generally to add more fields to the RowNode as every field added adds weight and complexity, I can create a new map but it is going to be a small overhead, I can try to measure what is this overhead by profiling and instead use a new map, let me have a think about it

SalvatorePreviti: I measured that using a map does not add particular overhead so removing the use of this as a temporary storage

packages/ag-grid-enterprise/src/rowHierarchy/groupStage.ts
SalvatorePreviti: renamed to getNode

PR #10603, opened 2025-05-22
AG-14949 allow grand total row pinning without `enableRowPinning`

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-14949


documentation/ag-grid-docs/src/content/docs/aggregation-total-rows/_examples/enabling-grand-total/main.ts
```diff
@@ -29,11 +36,7 @@ const gridOptions: GridOptions<IOlympicData> = {
 
 function onChange() {
     const grandTotalRow = document.querySelector<HTMLInputElement>('#input-property-value')!.value;
-    if (grandTotalRow === 'bottom' || grandTotalRow === 'top') {
-        gridApi.setGridOption('grandTotalRow', grandTotalRow);
-    } else {
-        gridApi.setGridOption('grandTotalRow', undefined);
-    }
+    gridApi.setGridOption('grandTotalRow', grandTotalRow as 'top' | 'bottom' | 'pinnedTop' | 'pinnedBottom');
```
AndrewGlazier: unsetting this is going to set the _string_ undefined into grandTotalRow, not the value right? It might be worth flipping the old if statement for === 'undefined'

documentation/ag-grid-docs/src/content/docs/aggregation-total-rows/index.mdoc
```diff
@@ -18,6 +18,10 @@ const gridOptions = {
 }
 ```
 
+{% note %}
+Using `grandTotalRow='pinnedTop'` or `grandTotalRow='pinnedBottom'` requires the `PinnedRowModule` to be registered
```
AndrewGlazier: Unioning the pinned top/bottom into here causes users to lose control on whether this row is the first or last row inside the pinned container - I thought we were going to do another grid option to allow this still?

PR #10608, opened 2025-05-22
AG-14972 - fixed wheel event being prevented in pinned areas

gportela85:
https://ag-grid.atlassian.net/browse/AG-14972

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
@@ -334,23 +334,31 @@ export class GridBodyCtrl extends BeanStub {
     }
 
     private addBodyViewportListener(): void {
-        const { popupSvc, touchSvc } = this.beans;
+        const {
+            eBodyViewport,
+            eStickyTop,
+            eStickyBottom,
+            eTop,
+            eBottom,
+            beans: { popupSvc, touchSvc },
+        } = this;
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
```diff
@@ -376,24 +384,23 @@ export class GridBodyCtrl extends BeanStub {
...
 
-        // we test for shift key because some devices will
-        // only change deltaY even when scrolling horizontally
-        const target = e.target as HTMLElement;
         if (!isHorizontalScroll) {
-            e.preventDefault();
-            this.scrollVertically(deltaY);
-        } else if (
-            this.eStickyTopFullWidthContainer.contains(target) ||
-            this.eStickyBottomFullWidthContainer.contains(target) ||
```
eliasmalik, line 392: Looks like we're no longer checking the sticky full width container? Was this just unnecessary? I tried in a plunkr  and scrolling full width rows still seems to work fine.

eliasmalik (approved):

packages/ag-grid-community/src/gridBodyComp/gridBodyCtrl.ts
gportela85: this was only necessary because we were sharing the method before. Now we know only the correct containers have the listener for horizontal wheel, so the check is no longer necessary, it's always true.

PR #10611, opened 2025-05-24
AG-3870 Apply min/max when autosizing columns

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-3870

### DIFF
--- a/packages/ag-grid-community/src/columnAutosize/columnAutosizeApi.ts
+++ b/packages/ag-grid-community/src/columnAutosize/columnAutosizeApi.ts
@@ -6,7 +6,6 @@ import type {
     ISizeColumnsToFitParams,
 } from '../interfaces/autoSize';
 import type { Column } from '../interfaces/iColumn';
-import { _warn } from '../validation/logging';
 
 export function sizeColumnsToFit(beans: BeanCollection, paramsOrGridWidth?: ISizeColumnsToFitParams | number) {
     if (typeof paramsOrGridWidth === 'number') {
@@ -24,7 +23,6 @@ export function autoSizeColumns(
     skipHeader?: boolean
 ): void {
     if (Array.isArray(keysOrParams)) {
-        _warn(277, { methodName: 'autoSizeAllColumns' });
         colAutosize?.autoSizeCols({ colKeys: keysOrParams, skipHeader, source: 'api' as const });
     } else {
         colAutosize?.autoSizeCols({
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/columnAutosize/columnAutosizeApi.ts
```diff
@@ -11,10 +16,38 @@ export function sizeColumnsToFit(beans: BeanCollection, paramsOrGridWidth?: ISiz
...
-export function autoSizeAllColumns(beans: BeanCollection, skipHeader?: boolean): void {
-    beans.colAutosize?.autoSizeAllColumns('api', skipHeader);
+export function autoSizeAllColumns(beans: BeanCollection, params: ISizeAllColumnsToContentParams): void;
+export function autoSizeAllColumns(beans: BeanCollection, skipHeader?: boolean): void;
+export function autoSizeAllColumns(
+    beans: BeanCollection,
+    paramsOrSkipHeader?: ISizeColumnsToContentParams | boolean
+): void {
+    if (paramsOrSkipHeader && typeof paramsOrSkipHeader === 'object') {
+        _warn(277, { methodName: 'autoSizeAllColumns' });
```
peterjrreynolds: Shouldn't the warning be on the boolean version?

packages/ag-grid-community/src/interfaces/autoSize.ts
```diff
@@ -55,4 +63,26 @@ export interface SizeColumnsToContentStrategy {
...
+    defaultMinWidth?: number;
+    /** Default maximum width for every column (does not override the column maximum width). */
+    defaultMaxWidth?: number;
+    /** Provide to limit specific column widths when sizing. */
+    columnLimits?: SizeColumnsToContentColumnLimits[];
+}
+
+export interface ISizeColumnsToContentParams extends ISizeAllColumnsToContentParams {
+    /** If not provided will auto-size all columns. Otherwise will size the specified columns. */
+    colIds?: string[];
```
peterjrreynolds, line 87: The old API allowed you to pass `(string | ColDef | Column)[]`. Now only col IDs are allowed. Should we not maintain the existing behaviour?

packages/ag-grid-community/src/columnAutosize/columnAutosizeApi.ts
eliasmalik: 🤦🏽

PR #10629, opened 2025-05-27
AG-14983 Codemod command line updates

adam-wang-uk:


documentation/ag-grid-docs/src/content/docs/upgrading-to-ag-grid-31-1/index.mdoc
```diff
@@ -20,7 +20,7 @@ Follow these steps to upgrade your project's AG Grid version to `31.1.0`:
 1. Run the `migrate` command of version `31.1` of the AG Grid codemod runner, where `$FROM_VERSION` refers to your project's existing AG Grid version:
 
     ```
-    npx @ag-grid-devtools/cli@31.1 migrate --from=$FROM_VERSION
+    npx @ag-grid-devtools/cli@latest migrate --from=31 --to=31.1
```
StephenCooper: I think we should actually leave it as `--from=$FROM_VERSION` as the user could be coming from any version and that would be valid. 

Right to have the `--to=31.1` as that is specific to this page. 

Can we update all the version pages to follow this pattern.

adam-wang-uk:
Updated @StephenCooper

StephenCooper (approved):

PR #10662, opened 2025-06-02
AG-14155 React Rendering 

StephenCooper:


packages/ag-grid-react/src/shared/interfaces.ts
```diff
@@ -27,6 +27,9 @@ export interface AgGridReactProps<TData = any> extends GridOptions<TData> {
     setGridApi?: (gridApi: GridApi<TData>) => void;
     /** @deprecated v33.3 AgGridReact does not accept children so this property will be removed. */
     children?: any;
+
+    /** Enables support for `cellRendererParams.deferRender` */
+    enableCellDeferRender?: boolean;
```
StephenCooper: The name of this could change depending on whether this is the default or not.

packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
```diff
@@ -238,10 +239,14 @@ export const AgGridReactUi = <TData,>(props: InternalAgGridReactProps<TData>) =>
         });
     }, [props]);
 
+    const enableCellDeferRender = !(props.suppressCellDeferRender ?? false) && !!React.useSyncExternalStore;
```
peterjrreynolds: Don't need `?? false` here

packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
```diff
@@ -74,6 +74,7 @@ const reactPropsNotGridOptions: ReactCompProps = {
     className: undefined,
     passGridApi: undefined,
     componentWrappingElement: undefined,
+    suppressCellDeferRender: undefined,
```
peterjrreynolds: `suppressDeferCellRender`?

PR #10672, opened 2025-06-03
Vitest benchmarks treedata

SalvatorePreviti:
Add vitest capability to the project and write a simple benchmark for tree data with path
The tests are currently meant to be executed manually and just for developer benefit in the local machine

The version we currently use for vitest says is an experimental feature, is not experimental anymore in the new vitest, when we upgrade.

Also updates behavioural tests ts version

image


seanlandsman (approved):
lgtm

StephenCooper (approved):

PR #10717, opened 2025-06-09
Ag 3390 add chevrons

kylerphillips:


documentation/ag-grid-docs/src/components/icon/IconsPanel.tsx
```diff
@@ -61,6 +61,10 @@ const ICONS = [
     'tree-open',
     'unlinked',
     'up',
+    'chevron-up',
+    'chevron-down',
+    'chevron-left',
+    'chevron-right',
```
peterjrreynolds: These should be in alphabetical order

peterjrreynolds:
For legacy themes, I believe we also need to add these to `build.js` and `_icon-font-codes.scss`

PR #10719, opened 2025-06-09
Reduce createElement calls

StephenCooper:
Reduce use of fragments to improve DEV performance by reducing the amount of createElement calls.

### DIFF
--- a/packages/ag-grid-react/src/reactUi/cells/cellComp.tsx
+++ b/packages/ag-grid-react/src/reactUi/cells/cellComp.tsx
@@ -61,14 +61,10 @@ const jsxEditor = (
 ) => {
     const newFormat = editDetails.compProxy;
 
-    return (
-        <>
-            {newFormat ? (
-                jsxEditorProxy(editDetails, CellEditorClass, setRef)
-            ) : (
-                <CellEditorClass {...editDetails.compDetails.params} ref={setRef} />
-            )}
-        </>
+    return newFormat ? (
+        jsxEditorProxy(editDetails, CellEditorClass, setRef)
+    ) : (
+        <CellEditorClass {...editDetails.compDetails.params} ref={setRef} />
     );
 };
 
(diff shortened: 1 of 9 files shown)


packages/ag-grid-react/src/reactUi/cells/cellComp.tsx
```diff
@@ -122,38 +107,31 @@ const jsxShowValue = (
...
-    // like the aggregation functions avg and count, which return objects and depend on toString()
-    // getting called.
-    const valueForNoCellRenderer = value?.toString ? value.toString() : value;
+    const bodyJsxFunc = () => {
+        if (!compDetails) {
+            // No Cell Renderer, so just show the value.
+            // if we didn't do this, objects would cause React error. we depend on objects for things
+            // like the aggregation functions avg and count, which return objects and depend on toString()
+            // getting called.
+            return value?.toString?.() ?? value;
```
StephenCooper, line 116: This is the main performance gain. We no longer require a call to `createElement` when just displaying a text value.

packages/ag-grid-react/src/reactUi/cells/cellComp.tsx
```diff
@@ -122,38 +107,31 @@ const jsxShowValue = (
...
+        }
 
-    const bodyJsxFunc = () => (
-        <>
-            {noCellRenderer && <>{valueForNoCellRenderer}</>}
-            {reactCellRenderer && !reactCellRendererStateless && (
-                <CellRendererClass {...compDetails!.params} key={key} ref={cellRendererRef} />
-            )}
-            {reactCellRenderer && reactCellRendererStateless && (
+        if (compDetails?.componentFromFramework) {
```
peterjrreynolds: `compDetails` must be defined here

peterjrreynolds: Similar below

packages/ag-grid-react/src/reactUi/header/headerFilterCellComp.tsx
```diff
@@ -122,25 +122,26 @@ const HeaderFilterCellComp = ({ ctrl }: { ctrl: HeaderFilterCellCtrl }) => {
...
-                        value={{
-                            setMethods: (methods: CustomFloatingFilterCallbacks) =>
-                                floatingFilterCompProxy.current!.setMethods(methods),
-                        }}
-                    >
-                        <UserCompClass {...floatingFilterProps!} />
-                    </CustomContext.Provider>
-                )}
+                {reactUserComp ? (
+                    !reactiveCustomComponents ? (
```
peterjrreynolds: Could flip the conditions here so it's `booleanValue ? a : b` instead of `!booleanValue ? b : a`

PR #10735, opened 2025-06-10
AG-15116 Deprecating advanced filter eval approach. Always use tree based filter

AlpineJosh:
Deprecated the Advanced Filter function eval in favour of always using the tree based approach. This may cause minor perfromance issues for highly nested filters, but simplifies the approach and allows for more function types.

Documentation updated to reflect this. Unsure how to write relase note or mark deprecations in the docs.

### DIFF
--- a/packages/ag-grid-community/src/entities/gridOptions.ts
+++ b/packages/ag-grid-community/src/entities/gridOptions.ts
@@ -690,7 +690,7 @@ export interface GridOptions<TData = any> {
      */
     advancedFilterBuilderParams?: IAdvancedFilterBuilderParams;
     /**
-     * @deprecated As of v34, advanced filter evaluation is deprecated, and this option will default to true.
+     * @deprecated As of v34, advanced filter no longer uses function evaluation, so this option has no effect.
      * @default true
      * @agModule `AdvancedFilterModule`
      */
--- a/packages/ag-grid-community/src/entities/gridOptions.ts
+++ b/packages/ag-grid-community/src/entities/gridOptions.ts
@@ -699,6 +699,7 @@ export interface GridOptions<TData = any> {
      * When using AG Grid Enterprise, the Set Filter is used by default when `filter: true` is set on column definitions.
      * Set to `true` to prevent this and instead use the Text Filter, Number Filter or Date Filter based on the cell data type,
      * the same as when using AG Grid Community.
+     * @default false
      * @initial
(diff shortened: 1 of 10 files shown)


packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -690,18 +690,15 @@ export interface GridOptions<TData = any> {
      */
     advancedFilterBuilderParams?: IAdvancedFilterBuilderParams;
     /**
-     * By default, Advanced Filter sanitises user input and passes it to `new Function()` to provide the best performance.
-     * Set to `true` to prevent this and use defined functions instead.
-     * This will result in slower filtering, but it enables Advanced Filter to work when `unsafe-eval` is disabled.
-     * @default false
+     * @deprecated As of v34, advanced filter evaluation is deprecated, and this option will default to true.
```
peterjrreynolds:
```suggestion
     * @deprecated As of v34, advanced filter no longer uses function evaluation, so this option has no effect.
```

Perhaps something like this instead to clarify that it does nothing

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -690,18 +690,15 @@ export interface GridOptions<TData = any> {
...
+     * @default true
      * @agModule `AdvancedFilterModule`
      */
     suppressAdvancedFilterEval?: boolean;
     /**
      * When using AG Grid Enterprise, the Set Filter is used by default when `filter: true` is set on column definitions.
      * Set to `true` to prevent this and instead use the Text Filter, Number Filter or Date Filter based on the cell data type,
      * the same as when using AG Grid Community.
-     * @default false
```
peterjrreynolds: Was this supposed to be removed?

packages/ag-grid-enterprise/src/advancedFilter/filterExpressionParser.ts
```diff
@@ -40,18 +40,7 @@ export class FilterExpressionParser {
             : this.params.advFilterExpSvc.translate('advancedFilterValidationMessageAtEnd', [message]);
     }
 
-    public getFunctionString(): {
-        functionString: string;
-        params: FilterExpressionFunctionParams;
-    } {
-        const params = this.createFunctionParams();
-        return {
-            functionString: `return ${this.joinExpressionParser.getFunctionString(params)};`,
```
peterjrreynolds, line 49: Should be able to track this through and remove all the `getFunctionString` implementations too.

PR #10759, opened 2025-06-12
AG-5652 Testing IDs

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-5652


documentation/ag-grid-docs/src/content/docs/testing/_examples/hello-world/exampleConfig.json
```diff
@@ -1,3 +1,3 @@
 {
-    "supportedFrameworks": ["typescript"]
+    "supportedFrameworks": ["typescript", "reactFunctional", "reactFunctionalTs", "angular", "vue3"]
```
StephenCooper: This file can be deleted if all frameworks are supported.

packages/ag-grid-community/src/main.ts
```diff
@@ -1298,3 +1299,7 @@ export type {
     ColorSchemeValue,
     WithParamTypes,
 } from './theming/theme-types';
+
+// Testing
+
+export * from './testing/testIdUtils';
```
StephenCooper: I think this should export on an object to reduce the number of things we are exporting from ag-grid-community.

packages/ag-grid-community/src/testing/testIdUtils.ts
```diff
@@ -0,0 +1,201 @@
...
+    const params = Object.entries(attributes)
+        .map(([k, v]) => (v != null ? `${k}=${v}` : null))
+        .filter(Boolean)
+        .join(';');
+    return [name, params].filter((s) => s.length > 0).join(':');
+}
+
+/** Headers */
+
+export function getTestIdForHeaderGroupCell(colId: string | null): string {
```
StephenCooper: I wonder if we can reduce the number of methods here in some way?

Open to discussion on this as maybe individual methods is clearest.

documentation/ag-grid-docs/src/content/docs/testing/_examples/hello-world/exampleConfig.json
eliasmalik: I'm deliberately not inclduing the javascript framework (i.e. UMD) because when I visit `/javascript-data-grid/**` I want the example to load in typescript, so we can validate the modules are correctly configured.

packages/ag-grid-community/src/testing/testIdUtils.ts
eliasmalik: The alternative I guess would be to have a function like `getTestId(name: ElementType, ...args): string`, where we provide overloads for every possible `ElementType` so that users know which arguments to pass?

In general I think we'd need to solve issues around discoverability; If I am a user and want to write a test that involves a particular element, how do I know which of these functions to use? Some naming is obvious (row, cell, header cell, etc), but others less so. Right now the main path for discovering what test IDs are available is bascially using autocomplete in the IDE. That would still work if we used the overload approach, but do we also want to document all of these on the website? We'd probably want that to be generated, rather than maintained manually.

Happy to talk tomorrow

.github/workflows/doc-tests.yml
```diff
@@ -192,6 +197,20 @@ jobs:
             with:
               framework: angular
 
+    test-recipes:
+      if: ${{ github.event.inputs.skip_recipes != 'true' }}
+      name: Test Public Testing Recipes
+      runs-on: ubuntu-latest
+      strategy:
+        fail-fast: false
+        matrix:
```
StephenCooper: You don't need the matrix if the tests are not being sharded

documentation/ag-grid-docs/src/content/docs/testing/_examples/hello-world/exampleConfig.json
StephenCooper: Target these urls for tests as it has all the frameworks

https://grid-staging.ag-grid.com/examples/column-updating-definitions/add-remove-columns/typescript/

packages/ag-grid-community/src/testing/testIdService.ts
```diff
@@ -0,0 +1,348 @@
+import type { NamedBean } from '../context/bean';
+import { BeanStub } from '../context/beanStub';
+import type { BeanName } from '../context/context';
+import { _getRootNode } from '../gridOptionsUtils';
+import type { ITestIdService } from '../interfaces/iTestIdService';
+import * as ids from './testIdUtils';
+
+const TEST_ID_ATTR = 'data-testid';
```
StephenCooper: Can we make this configurable via the Module.

packages/ag-grid-community/src/main.ts
eliasmalik: Exporting as an object seems to break tree shaking

StephenCooper (approved):

PR #10763, opened 2025-06-12
AG-15034 Fix tooltip in example

jacobp100:
https://ag-grid.atlassian.net/browse/AG-15034


jacobp100:
Hold off merging - we're renaming a property here

PR #10780, opened 2025-06-13
AG-15132 total rows should not be draggable

SalvatorePreviti:
Hide the row drag handle for all the footer rows - without indentation, as this was agreed with Mark and Zoheil.
Indentation when the drag icon is hidden can be added in another ticket if we want to make it behave as intended.

Handles also the total row as target, and ignore total rows during multiple selected row dragging

image

image


packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -261,6 +261,10 @@ export class RowDragFeature extends BeanStub implements DropTarget {
         let targetRowIndex = clientSideRowModel.getRowIndexAtPixel(y);
         let target = clientSideRowModel.getRow(targetRowIndex) ?? null;
 
+        if (target?.footer) {
+            return null; // Cannot drop on a footer row
```
AndrewGlazier: Is this expected? There's nothing stating this in the ticket, seems like a behavioural decision for Zoheil

Personally I'd expect this to work, and the row to drop as the first/last row adjacent to the footer?

SalvatorePreviti: Let me give it a try

SalvatorePreviti: discussed with Zoheil and amended new changes, the total row is now a valid target and behaves consistently

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -323,42 +333,50 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
+            return target.parent; // Same parent as the target
+        }
+
+        let nextRow: RowNode | undefined;
+        do {
+            nextRow = clientSideRowModel.getRow(++targetRowIndex);
+        } while (nextRow && nextRow.footer);
+
+        const { childrenAfterAggFilter } = target;
+        if (!nextRow || nextRow.parent !== target || !childrenAfterAggFilter?.length) {
```
AndrewGlazier: Note I've just had to fix a bug similar to this - childrenAfterAggFilter actually isn't the best metric for length as people can mutate the childrenAfterSort array (unfortunately a legacy pattern we used to offer as a work-around, so it has been documented/provided through ZD)

Here's my ticket for context: https://ag-grid.atlassian.net/browse/AG-14741

Unsure if this same bug would be present in your code here, but an FYI

SalvatorePreviti: Changed to childrenAfterGroup, and this includes the test if the row is visible by checking rowIndex is not null, so it excludes filtered rows.

PR #10807, opened 2025-06-16
Maintenance(ag-grid): follow ups on perf workflow

sdwvit:
- Downgraded gh script action to v6 because of v7 regression
 - Add more and better std output in tests
 - Adjust calculation so it doesn't end up running indefinitely
 - Fix invalid blocks error in slack notification code
 - Increased timeout for performance test in setData.spec.ts

### DIFF
--- a/testing/shared/scripts/convert-playwright-output-to-slack-msg.mjs
+++ b/testing/shared/scripts/convert-playwright-output-to-slack-msg.mjs
@@ -41,7 +41,7 @@ const slackMessage = { channel, username, icon_url, blocks };
 fs.writeFileSync(
     commentFileName,
     blocks
-        .map((b) => b.text?.text.replace(/<(.+)\|(.+)>/g, '[$2]($1)') || '---')
+        .map((b) => b.text?.text?.replace?.(/<(.+)\|(.+)>/g, '[$2]($1)') || '---')
         .concat(['---', process.env.IS_SUCCESS ? '' : `Please address the issues before merging.`])
         .join('\n')
 );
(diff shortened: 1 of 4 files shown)


testing/shared/scripts/convert-playwright-output-to-slack-msg.mjs
```diff
@@ -22,28 +22,29 @@ const logFile = path.join(__dirname, '../../../playwright-report/test-results.js
...
-    },
 ];
 
 const slackMessage = { channel, username, icon_url, blocks };
 
-fs.writeFileSync(commentFileName, blocks.map((b) => b.text?.text || '---').join('\n'));
+fs.writeFileSync(
+    commentFileName,
+    blocks
+        .map((b) => b.text?.text || '---')
```
sdwvit:
```suggestion
        .map((b) => b.text?.text.replace(/<(.+)\|(.+)>/g, '[$2]($1)') || '---')
```

testing/shared/scripts/convert-playwright-output-to-slack-msg.mjs
```diff
@@ -22,28 +22,29 @@ const logFile = path.join(__dirname, '../../../playwright-report/test-results.js
...
-    },
 ];
 
 const slackMessage = { channel, username, icon_url, blocks };
 
-fs.writeFileSync(commentFileName, blocks.map((b) => b.text?.text || '---').join('\n'));
+fs.writeFileSync(
+    commentFileName,
+    blocks
+        .map((b) => b.text?.text.replace(/<(.+)\|(.+)>/g, '[$2]($1)') || '---')
```
sdwvit:
```suggestion
        .map((b) => b.text?.text?.replace?.(/<(.+)\|(.+)>/g, '[$2]($1)') || '---')
```

PR #10831, opened 2025-06-17
AG-15152 - Add CSP testing workflow

taktran:


.github/workflows/csp.yml
```diff
@@ -0,0 +1,60 @@
...
+description: Content Security Policy Testing
+
+env:
+    NX_NO_CLOUD: true
+    CI: true
+    DEFAULT_RETENTION_DAYS: 30
+
+on:
+    push:
+        branches: ['AG-15152-csp-testing']
```
taktran: Can be removed after review

.github/workflows/csp.yml
```diff
@@ -0,0 +1,60 @@
...
+            clean_checkout:
+                description: 'Disable all caching'
+                type: 'choice'
+                required: true
+                default: 'false'
+                options:
+                    - 'true'
+                    - 'false'
+    schedule:
+        - cron: '0 0 * * *' # Run daily at midnight UTC
```
seanlandsman: I wonder if we should consider staggering some of these tests. With this we'll have 3/4 tests that hammer staging all a the same time

seanlandsman (changes requested):
Looks fine but I think we should stagger these nightly tests a bit.

With docs test, accessibility and now this we'll be absolutely hammering staging all at midnight on.

Perhaps start these tests at 1:30am?

seanlandsman (approved):
lgtm

PR #10842, opened 2025-06-18
AG-9295 Extend use of TContext Generic

StephenCooper:


packages/ag-grid-community/src/interfaces/iLoadingCellRenderer.ts
```diff
@@ -1,7 +1,8 @@
 import type { ICellRendererParams } from '../rendering/cellRenderers/iCellRenderer';
 import type { IComponent } from './iComponent';
 
-export interface ILoadingCellRendererParams<TData = any, TContext = any> extends ICellRendererParams<TData, TContext> {
```
StephenCooper, line 4: Here TContext was mislabeled and was actually TValue

peterjrreynolds (approved):

PR #10850, opened 2025-06-18
AG-15130 drag drop nesting timer

SalvatorePreviti:
This addresses:
- [AG-15130](https://ag-grid.atlassian.net/browse/AG-15130) Add delay before nesting moved row into new parent
- [AG-14596](https://ag-grid.atlassian.net/browse/AG-14596) Allow hovering groups to expand them when dragging rows
- [border being 1px off left and right](https://ag-grid.atlassian.net/browse/AG-14394?focusedCommentId=84950) (css changes)

AG-15130 and AG-14596 are intimately related and made sense to address both in the same PR, as is the same timer/feature.
This will also close AG-14394 as closes the 1px left-right border issue

In detail changes:

- added proper destructor to the row drag feature, that disposes and frees all the memory and clear the timer
- added a throttle timer with (as currently agreed) a hardcoded 2 seconds timer, that can both expand a collapsed group or create a new group
- changed the above field in RowsDrop to be a 'above', 'inside', 'below' value
- renamed getRowsDrop to managedRowsDrop as it is responsible to start the timer
- renamed determineNewParent to targetShouldBeParent and made it return a boolean value
- fixed the scss and css to remove left and right 1px border
- corrected the logic to insert above, below, inside

Reviewed by Andrew, comment addressed.
Tested by Mark for the UI/UX on his local machine and addressed his comments.


community-modules/styles/src/internal/base/parts/_grid-layout.scss
```diff
@@ -131,7 +131,7 @@
         display: block;
         width: auto;
         height: auto;
-        inset: 0 1px;
+        inset: 0;
```
AndrewGlazier, line 134: I don't think we can change this - this was an intentional design at some point, we shouldn't be making design changes (unless obviously broken) as they're considered breaking, and this was an intentional design (I doubt anyone accidentally added l/r inset)

SalvatorePreviti: This is a CSS class I made in a previous PR that was not existing before in the previous release that is specifically made for tree data row dragging, this won't break any existing layout, was just a not correct value as Mark points out the correct appearance should be 0 pixels top bottom left right

ag-row-highlight-inside class was introduced just few days ago as part of https://ag-grid.atlassian.net/browse/AG-14393

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -321,45 +370,64 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
+            this.makeGroupThrottleTimer = window.setTimeout(this.makeGroupThrottleCallback, 2000);
+        }
+    }
+
+    private makeGroupThrottleCallback = () => {
+        const event = this.lastDraggingEvent;
+        if (event) {
+            this.makeGroupThrottled = true;
+            this.doManagedDrag(event);
+            this.makeGroupExpanded(this.makeGroupThrottleTarget);
```
AndrewGlazier, line 400: What if transactions cause the rows to move? Do we not need to recalculate target & refresh throttle?

Probably needs modelUpdated to do the same lgoic as onScrollCallback

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -321,45 +370,64 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
-            newParent = target; // Always insert as child
+        const position = inside ? 'inside' : above ? 'above' : 'below';
+        return { sameGrid, position, target, newParent, rows };
+    }
+
+    private makeGroupThrottleStart(target: RowNode) {
+        if (this.makeGroupThrottleTimer === null) {
+            this.makeGroupThrottled = false;
+            this.makeGroupThrottleTarget = target;
+            this.makeGroupThrottleTimer = window.setTimeout(this.makeGroupThrottleCallback, 2000);
```
AndrewGlazier: This magic number (2s) probably needs to be configurable as a grid option

SalvatorePreviti: Added `rowDragInsertDelay` grid option as we decided its name with Mark and Kiril

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
SalvatorePreviti: So, is actually fine, since we recompute all from the event every time, if the target row changes the timer will be cleared. If it does not change, there is nothing to worry about really as the result will be still correct. We also ignore deleted source rows during drop - I am confident it won't break

gportela85 (approved):

PR #10873, opened 2025-06-19
AG-14993 auto test example buttons

StephenCooper:


documentation/ag-grid-docs/src/content/docs/server-side-model-tree-data/_examples/transactions-tree-data/main.ts
```diff
@@ -83,7 +83,7 @@ let latestId = 100000;
 function addToSelected() {
     const selected = gridApi!.getSelectedNodes()[0];
     if (!selected) {
-        console.warn('No row was selected.');
```
StephenCooper, line 86: Downgraded to log as then these will not fail the example runner tests when the buttons are clicked without a selection. 

In the future we may be able to automate a selection first.

peterjrreynolds (approved):

PR #10877, opened 2025-06-19
AG-9780-handle-number-editor-prevend-default-behaviour

StephM-W:


documentation/ag-grid-docs/src/content/docs/cell-editing-batch/index.mdoc
```diff
@@ -9,10 +9,32 @@ Batch editing allows you to edit multiple cells in the grid before committing th
 Batch editing is an advanced feature only available via the API to allow you to tailor it to your specific needs.
 {% /note %}
 
+{% warning %}
```
peterjrreynolds: `note` might be better then `warning` just because `warning` is quite jarring visually.

packages/ag-grid-community/src/edit/editService.ts
```diff
@@ -74,13 +77,14 @@ export class EditService extends BeanStub implements NamedBean, IEditService {
...
     private strategy?: BaseEditStrategy;
-    private includeParents: boolean = true;
+    protected cellEditingInvalidCommitType: boolean = false;
 
     public postConstruct(): void {
         const { beans } = this;
         this.model = beans.editModelSvc!;
         this.valueSvc = beans.valueSvc;
         this.rangeSvc = beans.rangeSvc!;
+        this.cellEditingInvalidCommitType = this.gos.get('cellEditingInvalidCommitType') === 'block';
```
peterjrreynolds: We try to avoid caching gos values due to ordering of property updates (e.g. two services listening to property change, one service reads the update first and calls the second which then has the old value).

packages/ag-grid-community/src/edit/utils/editors.ts
```diff
@@ -260,7 +252,22 @@ export function _refreshEditorOnColDefChanged(beans: BeanCollection, cellCtrl: C
     const editorParams = _createEditorParams(beans, cellCtrl, eventKey, cellStartedEdit);
     const colDef = column.getColDef();
     const compDetails = _getCellEditorDetails(beans.userCompFactory, colDef, editorParams);
-    editor.refresh(compDetails!.params);
+
+    editor.refresh(handleNumberEditor(compDetails!.params, eventKey));
+}
+
+function handleNumberEditor(
```
peterjrreynolds: `checkAndPreventDefault`?

Otherwise you might miss the fact that it's preventing default

StephM-W: Renamed

packages/ag-grid-community/src/edit/editService.ts
StephM-W: Removed to direct access

documentation/ag-grid-docs/src/content/docs/cell-editing-batch/index.mdoc
StephM-W: Updated

gportela85 (approved):

PR #10889, opened 2025-06-19
AG-15040 - handle blocking validation for enter

StephM-W:


gportela85 (approved):

gportela85 (changes requested):
Why does this fix the issue? Why hitting `enter` 3 times caused it?

gportela85 (approved):

PR #10894, opened 2025-06-19
AG-9780 - update batch styles

StephM-W:


AlpineJosh (approved):
Colors all converted correctly. Looks good.

PR #10904, opened 2025-06-19
Maintenance(ci): follow up on performance workflow

sdwvit:
- Remove unused frameworks from build
 - Fix Attach snippet action
 - Call page.requestGC() every 50 iterations, so we get a bit more control when it runs
 - Adjust test settings/iterations
 - Call dispose of handler to help memory pressure
 - Add calculateDriftWithMargin function for drift rate analysis
 - Adjust margin of error calculations in performance metrics
 - Rely on performance.mark and events from the grid to measure click-to-render accurately
 - preallocate data for the grid to speed up the tests

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -132,7 +132,7 @@ jobs:
         if: ${{ steps.setup.outcome == 'success' || steps.setup.outcome == 'skipped' }}
         run: |
           echo "Building UMD packages..."
-          npx nx run-many -t build --projects=ag-grid-{enterprise,community,angular,react,vue3}
+          npx nx run-many -t build --projects=ag-grid-{enterprise,community}
 
       - name: Notify Progress
         uses: actions/github-script@v7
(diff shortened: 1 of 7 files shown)


.github/workflows/benchmark.yml
```diff
@@ -132,7 +132,7 @@ jobs:
         if: ${{ steps.setup.outcome == 'success' || steps.setup.outcome == 'skipped' }}
         run: |
           echo "Building UMD packages..."
-          npx nx run-many -t build --projects=ag-grid-{enterprise,community,angular,react,vue3}
+          npx nx run-many -t build --projects=ag-grid-{enterprise,community}
```
StephenCooper, line 135: I suspect we will want to bring these back when you run the test for each framework?

StephenCooper (approved):

.github/workflows/benchmark.yml
sdwvit: cron job doesn't rely on local build, it grabs already built umds from [web](https://github.com/ag-grid/ag-grid/blob/latest/testing/performance/benchmarking.ts#L110-L111)

PR #10916, opened 2025-06-20
AG-14993 Fixes for examples

StephenCooper:


documentation/ag-grid-docs/src/content/docs/drag-and-drop/_examples/two-grids/main.ts
```diff
@@ -153,10 +153,10 @@ function gridDrop(event: any, grid: string) {
         return;
     }
 
-    const gridApi = grid == 'left' ? leftApi! : rightApi!;
+    const api = grid == 'left' ? leftApi! : rightApi!;
```
StephenCooper, line 156: This is a workaround for the teardown logic thinking this meant there was an instance of the gridApi. 

Tactical fix, should come up with better matching logic is this keeps being an issue.

peterjrreynolds (approved):

PR #10923, opened 2025-06-20
AG-15015 Re-enable Angular 17 support

StephenCooper:


packages/ag-grid-community/src/rendering/cell/cellComp.ts
```diff
@@ -381,7 +381,7 @@ export class CellComp extends Component {
             if (loadingComp) {
                 // Immediately render the loading component and then schedule the task
                 createCellRendererFunc(loadingComp)();
-                onReady.then(() => createTask(true));
+                onReady.then(() => createTask!(true));
```
StephenCooper, line 384: Typescript downgrade required this change

seanlandsman (approved):
lgtm but we'll need to do the same with charts

PR #10930, opened 2025-06-20
AG-15174-batch-edit-docs

StephM-W:


gportela85 (approved):

documentation/ag-grid-docs/src/content/docs/cell-editing-batch-test/_examples/batch-editing/data.ts
bryanakitchen, line 1: Why are all of the `Male`s `Happy` and the `Female`s `Sad`?

PR #10943, opened 2025-06-20
AG-9780 - TC121 - cleanup editors when batch editing

StephM-W:


packages/ag-grid-community/src/edit/utils/controllers.ts
```diff
@@ -85,11 +86,22 @@ export function _getCtrls(beans: BeanCollection, inputs: ResolveControllerType =
     };
 }
 
+function _stopEditing(beans: BeanCollection): void {
+    const { editSvc } = beans;
+    if (editSvc?.isBatchEditing()) {
+        _destroyEditors(beans);
+    } else {
+        editSvc?.stopEditing(undefined, { source: 'api' });
```
StephenCooper, line 94: Is this source of api correct if it is being triggered by a click outside of the grid?

StephM-W: I've been thinking about renaming the sources... really 'api'='force' and 'ui'='try-your-best', but yes, 'api' here is correct

StephenCooper (approved):

PR #10951, opened 2025-06-21
AG-15174 - TC8 - remove redundant info

StephM-W:


gportela85 (changes requested):
This PR looks the same as https://github.com/ag-grid/ag-grid/pull/10950

PR #10972, opened 2025-06-23
Fix invalid template literal in AgProvidedColumnGroup ID

charmingjae:
While reviewing the code that generates column group IDs, I noticed a small typo in the template literal:

```ts
FAKE_PATH_${col.getId()}}_${i}
```

While working with the Selection Row's checkbox column, I noticed that its generated col-id attribute was rendered as:

```ts
FAKE_PATH_ag-Grid-SelectionColumn}_0_0
```

An extra closing brace `}` was included, resulting in a malformed string.
This commit corrects the typo by removing the redundant brace:

```ts
FAKE_PATH_${col.getId()}_${i}
```

This ensures that the generated col-id values are properly formatted.
Please let me know if there's a preferred way to handle such changes — I'm happy to revise if needed.


StephenCooper:
Thank you for pointing this out. We will get it corrected.

PR #10976, opened 2025-06-23
AG-15203 - TC159 - editApi stop should only close batch editors

StephM-W:


packages/ag-grid-community/src/edit/editApi.ts
```diff
@@ -54,8 +54,22 @@ export function getEditingCells(beans: BeanCollection, params: GetEditingCellsPa
     return positions;
 }
 
-export function stopEditing({ editSvc }: BeanCollection, cancel: boolean = false): void {
-    editSvc?.stopEditing(undefined, { cancel, source: 'api' });
+export function stopEditing(beans: BeanCollection, cancel: boolean = false): void {
+    const { editSvc } = beans;
+    if (editSvc?.isBatchEditing()) {
```
sdwvit, line 59:
```suggestion
    if (editSvc?.isBatchEditing?.()) {
```

packages/ag-grid-community/src/edit/editApi.ts
```diff
@@ -54,8 +54,22 @@ export function getEditingCells(beans: BeanCollection, params: GetEditingCellsPa
...
+        if (cancel) {
+            beans.editModelSvc?.getEditPositions().forEach((cellPosition) => {
```
sdwvit, line 61: same comment as above, beans.editModelSvc?.[...] may return undefined, and the following method call would be a NPE

sdwvit (approved):
lgtm

PR #10982, opened 2025-06-23
RTI-2778 add paginationPanelHeight theme parameter

BernieSumption:


gportela85 (changes requested):
Is this not used with legacy themes?

gportela85 (approved):

PR #10988, opened 2025-06-23
Aligning selection examples with docs

AlpineJosh:


documentation/ag-grid-docs/src/content/docs/row-selection-multi-row/_examples/checkbox-selection-disable-checkboxes/provided/reactFunctionalTs/index.tsx
```diff
@@ -42,7 +42,7 @@ const GridExample = () => {
     function toggleHideCheckbox() {
         grid.current?.api.setGridOption('rowSelection', {
             mode: 'multiRow',
-            isRowSelectable: (node) => (node.data ? node.data.year <= 2004 : false),
+            isRowSelectable: (node) => (node.data ? node.data.year < 2004 : false),
```
peterjrreynolds: These should be `< 2007`, and the code snippet on the docs page updated to match

peterjrreynolds (approved):

PR #10994, opened 2025-06-23
[RTI-2784] Maintenance(grid): allow all console methods (except for console.error) in example generator

sdwvit:
- Introduced an array to match more console methods for better detection in example gen.
 - Updated getHasExampleConsoleLog to utilize the new logic.
 - Still do not allow console.error, as it is reserved for service messages like absent licence error
 - Update extreme pivot example to use console.warn

| Was | Now |
|--------|--------|
| image  | Screenshot 2025-06-23 at 16 08 49  | 

Also fixes 
https://ag-grid.atlassian.net/jira/software/projects/RTI/boards/23?selectedIssue=RTI-2824

| Was | Now |
|--------|--------|
| Screenshot 2025-06-23 at 16 24 10 | Screenshot 2025-06-23 at 16 22 47  |

### DIFF
--- a/plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/fileUtils.ts
+++ b/plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/fileUtils.ts
@@ -174,9 +174,13 @@ export function convertTsxToJsx(fileStr: string): string {
     jsxFile = jsxFile.replaceAll('// empty line', '');
     return jsxFile;
 }
+const consoleMethods = Object.getOwnPropertyNames(console);
+const consoleRegexp = new RegExp(`console\.(${consoleMethods.join('|')})`, 'g');
 
 export const getHasExampleConsoleLog = ({ contents }: { contents?: string }) => {
-    return contents?.includes('console.log');
+    const res = consoleRegexp.test(contents || '');
+    consoleRegexp.lastIndex = 0; // reset lastIndex to ensure it can be reused
+    return res;
 };
 
 export const getHasSimpleHtml = ({ contents }: { contents?: string }) => {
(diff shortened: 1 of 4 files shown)


eliasmalik (approved):

plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/getConsoleLogSnippet.ts
```diff
@@ -161,21 +161,23 @@ function patchConsoleLog() {
...
-                exampleName: '${exampleName}',
-                data: getConsoleLogData(args),
-            });
-        } catch(error) {
-            // Posting is best-effort and shouldn't block normal console logging.
-            ${logError ? 'console.error(error);' : undefined}
-        }
-        originalConsoleLog(...args);
-    };
+    // we ignore console.error to avoid printing license messages
```
taktran: It's also so that there isn't an infinite loop if the try/catch fails

plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/fileUtils.ts
```diff
@@ -174,9 +174,13 @@ export function convertTsxToJsx(fileStr: string): string {
     jsxFile = jsxFile.replaceAll('// empty line', '');
     return jsxFile;
 }
+const consoleMethods = Object.getOwnPropertyNames(console);
+const consoleRegexp = new RegExp(`console.(${consoleMethods.join('|')})`, 'g');
```
taktran, line 178: I don't think we want to support all of what `console` supports, as the parsing logic would need to change as well. An allow list of `log` and `warn` should be sufficient.

plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/getConsoleLogSnippet.ts
sdwvit: yep

taktran (approved):
Good stuff 👍🏼

PR #11000, opened 2025-06-23
RTI-2791 - tool panel next up links

peterjrreynolds:


documentation/ag-grid-docs/src/content/docs/tool-panel-filters/index.mdoc
```diff
@@ -1,4 +1,4 @@
----
+    ---
```
taktran: I think this is a mistake?

```suggestion
---
```

taktran (changes requested):

taktran (approved):
👍🏼

PR #11029, opened 2025-06-23
RTI-2847 - ensure editor gets focus in

StephM-W:


packages/ag-grid-community/src/edit/cellEditors/largeTextCellEditor.ts
```diff
@@ -69,10 +69,14 @@ export class LargeTextCellEditor extends AgAbstractCellEditor<ILargeTextEditorPa
         this.eEditor.setInputAriaLabel(translate('ariaInputEditor', 'Input Editor'));
 
         if (this.focusAfterAttached) {
-            this.eEditor.getFocusableElement().focus();
+            this.focusIn();
```
gportela85: when `this.eEditor.getFocusableElement().focus()` is focusing the editor but then a call from
`baseEditStrategy.setFocusInOnEditor()` is moving the focus back to the cell that owns the editor

the following condition is wrong
```
if (editor?.focusIn) {
            // if the editor is present, then we just focus it
            editor.focusIn();
        } else {
            // if the editor is not present, it means async cell editor (e.g. React)
            // and we are trying to set focus before the cell editor is present, so we
            // focus the cell instead
            cellCtrl.focusCell(true);
            cellCtrl.onEditorAttachedFuncs.push(() => comp?.getCellEditor()?.focusIn?.());
        }
```

According to the docs of iCellEditor, `focusIn` should only work with Full Row Edit. And the `else` above is assuming that the editor doesn't exist and trying to focus the cell, when the editor is actually there.

StephM-W: The else block in your snippet doesn't get invoked because the presence of `focusIn` on the editor causes it to take the first branch. The cell won't be focused.

StephM-W: Found a different approach that doesn't require modifying the editor, resolving.

gportela85 (approved):

PR #11053, opened 2025-06-24
RTI-2829 Custom props example

eliasmalik:
https://ag-grid.atlassian.net/browse/RTI-2829


documentation/ag-grid-docs/src/content/docs/component-cell-renderer/_examples/custom-props/customButtonComponent_typescript.ts
```diff
@@ -13,7 +13,8 @@ export class CustomButtonComponent {
         this.eGui = document.createElement('div');
         const eButton = document.createElement('button');
         eButton.className = 'btn-simple';
-        eButton.textContent = 'Launch!';
+        const company = params.data?.company;
```
StephenCooper, line 16: Can we do the same change for all of the frameworks?

StephenCooper (approved):

PR #11055, opened 2025-06-24
AG-8515 - Pivot result cols update when agg col does

AndrewGlazier:


gportela85 (changes requested):
This PR adds some breaking changes

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -384,6 +378,35 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
+        newColDef.columnGroupShow = colDef.columnGroupShow;
+        newColDef.colId = colDef.colId;
+        newColDef.valueGetter = colDef.valueGetter;
+        newColDef.aggFunc = colDef.aggFunc;
+        newColDef.valueGetter = colDef.valueGetter;
+
+        const process = this.gos.get('processPivotResultColDef');
+        process?.(colDef);
+
+        return newColDef;
```
gportela85, line 403: We should probably avoid reading `colDef` so many times:

```
public recreateColDef(colDef: ColDef): ColDef {
    if (!colDef.pivotValueColumn) {
        return colDef;
    }

    const {
        pivotValueColumn,
        headerName,
        pivotKeys,
        pivotTotalColumnIds,
        columnGroupShow,
        colId,
        valueGetter,
        aggFunc
    } = colDef;

    const newColDef = this.createColDef(
        pivotValueColumn as AgColumn,
        headerName,
        pivotKeys,
        !!pivotTotalColumnIds
    );

    Object.assign(newColDef, {
        columnGroupShow,
        colId,
        valueGetter,
        aggFunc
    });

    this.gos.get('processPivotResultColDef')?.(colDef);

    return newColDef;
}
```

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -384,6 +378,35 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
+
+        // don't overwrite these
+        newColDef.columnGroupShow = colDef.columnGroupShow;
+        newColDef.colId = colDef.colId;
+        newColDef.valueGetter = colDef.valueGetter;
+        newColDef.aggFunc = colDef.aggFunc;
+        newColDef.valueGetter = colDef.valueGetter;
+
+        const process = this.gos.get('processPivotResultColDef');
+        process?.(colDef);
```
peterjrreynolds, line 405: `newColDef`?

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
AndrewGlazier: I'm curious about this - I appreciate this is probably for bundle sizing but destructuring is worse for performance, this path could potentially be quite hot - so where do we prioritise bundle size over peformance?

To note I'm not sure how much of an impact we'd have on bundle size here either, considering the properties (newColDef/colDef) are local/private and will be minified, which is different to for example repeating non-minifiable bean names.

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -384,6 +370,34 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
+        );
+
+        // don't overwrite these
+        newColDef.columnGroupShow = colDef.columnGroupShow;
+        newColDef.colId = colDef.colId;
+        newColDef.valueGetter = colDef.valueGetter;
+        newColDef.aggFunc = colDef.aggFunc;
+        newColDef.valueGetter = colDef.valueGetter;
+
+        this.gos.get('processPivotResultColDef')?.(colDef);
```
peterjrreynolds: `newColDef`?

AndrewGlazier: Nice catch, fixed thanks

peterjrreynolds (approved):

gportela85 (approved):

PR #11059, opened 2025-06-24
RTI-2779 - Prevent homepage buttons clipping on smaller screens

markdurrant:


kylerphillips:
Seems good, may want something a little more bullet proof viewport wise post release?

kylerphillips (approved):

taktran (approved):

PR #11062, opened 2025-06-24
Adding theme to body for scroll bars

AlpineJosh:


markdurrant (changes requested):
As this is a fix for one specific example I think it might be safer to apply this fix at the example level with a specific styles.css file just for this example.

I can't remember why exactly, but I'm sure we restrict `color-scheme: dark;` to non-grid things for a reason. I think something todo with integrated charts.

Safer to only apply this fix to the single offending example rather than all examples.

markdurrant (approved):
LGTM 👍

PR #11067, opened 2025-06-24
RTI-2879 - include cellClear source in setDataValue handling flow

StephM-W:


sdwvit (approved):
code looks fine to me, we are basically reverting to the old behavior

PR #11075, opened 2025-06-24
RTI-2881 - fix full row edit read only cell

peterjrreynolds:


packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
```diff
@@ -164,6 +164,9 @@ export class CellKeyboardListenerFeature extends BeanStub {
                 editSvc?.stopEditing(cellCtrl, {
                     event,
                 });
+            } else if (rowEditing) {
```
gportela85: Wouldn't it make more sense if this condition was `else if (!cellCtrl.isCellEditable()) {`?

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
```diff
@@ -164,6 +164,9 @@ export class CellKeyboardListenerFeature extends BeanStub {
...
+                // must be on a read only cell
+                editSvc?.stopEditing(undefined, { event });
```
StephM-W: Please pass in `{rowNode}` to limit the stop to that row

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
StephM-W: Do we need this specificity? If we know the row is editing and our cell hasn't got an open editor, we know the outcome for Enter will be to stop editing?

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
peterjrreynolds: Done

packages/ag-grid-community/src/rendering/cell/cellKeyboardListenerFeature.ts
gportela85: @Stephanemw I'd rather have meaningful conditions than "assumption" conditions. But I'm ok if you want it merged

StephM-W: Stick them both in there and be certain

peterjrreynolds: Done

gportela85 (approved):

PR #11079, opened 2025-06-24
RTI-2843 - handle column ref data lookup failure

StephM-W:


gportela85 (changes requested):
I believe the value should be `undefined`

gportela85 (approved):

PR #11080, opened 2025-06-24
angular tree data drag drop example not loading due to ts compile error

SalvatorePreviti:
examples should not have import { type ... } but import type { ... and angular does not like it


seanlandsman (approved):
lgtm

PR #11082, opened 2025-06-24
RTI-2890 Port over CSS editing fix from theming API to legacy

eliasmalik:
See https://ag-grid.atlassian.net/browse/RTI-2890

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_batch-edit.scss
+++ b/community-modules/styles/src/internal/base/parts/_batch-edit.scss
@@ -6,6 +6,14 @@
         color: var(--ag-cell-batch-edit-text-color);
     }
 
+    .ag-cell.ag-cell-inline-editing {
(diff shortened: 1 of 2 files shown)


community-modules/styles/src/internal/base/parts/_batch-edit.scss
```diff
@@ -6,6 +6,14 @@
         color: var(--ag-cell-batch-edit-text-color);
     }
 
+    .ag-cell.ag-cell-inline-editing {
```
eliasmalik, line 9: Same approach as in https://github.com/ag-grid/ag-grid/blob/cfff37437aecab9f27c751f7ac856b514aeaaab9/packages/ag-grid-community/src/theming/parts/theme/material-adjustments.css#L22-L26

gportela85 (approved):

PR #11122, opened 2025-06-26
AG-15192 Run System calls from one file

StephenCooper:
The following error leads to flakey tests.

```
   Received: "Example Error: Error: (SystemJS) XHR error (404 Not Found) loading https://grid-staging.ag-grid.com/examples/find/find-full-width/typescript/traceur
            Error loading https://grid-staging.ag-grid.com/examples/find/find-full-width/typescript/traceur
```

This happens when we call `System.import()` before we have called `System.config`. At the moment these are run in separate scripts so I believe there is a race condition. 

This change inlines both into the same file.


seanlandsman (approved):
lgtm

PR #11160, opened 2025-06-30
fix https://github.com/ag-grid/ag-grid/issues/10745

### LINKED ISSUE #10745: [Bug]: Cannot read properties of undefined (reading 'getSelector')

NA

### Describe the bug

**Issue:**
After upgrading Ag-Grid from v31 to v33.3.1, we occasionally encounter the following runtime error during certain React component re-renders:

`Cannot read properties of undefined (reading 'getSelector')
`
This occurs specifically at:

`gridHeaderDropZonesSelector: beans.registry.getSelector('AG-GRID-HEADER-DROP-ZONES'),
`
Upon investigation, we found beans.registry is sometimes undefined during these re-renders.

**Observation:**
Interestingly, there's already an undefined check implemented for the status bar selector:

`statusBarSelector: beans.registry?.getSelector('AG-STATUS-BAR'),
`
but such a check is absent for the gridHeaderDropZonesSelector, leading us to confusion.

**Questions:**
Could you clarify why beans.registry might become undefined intermittently in v33.3.1?
If it was an issue on our end, why does Ag-Grid's internal implementation include an undefined check for the status bar selector but omit it for the header drop zones selector?

https://github.com/ag-grid/ag-grid/blob/33ed644f297d4aee9fcd1944aa0ea58ed4ae92c4/packages/ag-grid-community/src/gridComp/gridCtrl.ts#L70

**Expected Behavior:**
Consistent null or undefined checks for beans.registry to ensure robust behavior across re-renders and avoid runtime errors.

Image
Image

### Version

33.3.1

### Does the issue occur for a specific framework only?

React

### Is the issue only observable on a specific browser?

All browsers

chungwong:
This fix https://github.com/ag-grid/ag-grid/issues/10745
```diff
-            gridHeaderDropZonesSelector: beans.registry.getSelector('AG-GRID-HEADER-DROP-ZONES'),
+            gridHeaderDropZonesSelector: beans.registry?.getSelector('AG-GRID-HEADER-DROP-ZONES'),
```

The bug was introduced at https://github.com/ag-grid/ag-grid/commit/f31f90cf3917ec734f356b387bd3c3ab9214e7d3#diff-4931422f7b1bfbdec0e5616785313361d0b8a6613848004c70082119d335b64e where the optional chaining `?.` was removed

By  the way, can AG-GRID be built locally? I don't see a way as it is using a private registry and I cannot verify anything before submitting this PR.

### DIFF
--- a/packages/ag-grid-community/src/gridComp/gridCtrl.ts
+++ b/packages/ag-grid-community/src/gridComp/gridCtrl.ts
@@ -67,7 +67,7 @@ export class GridCtrl extends BeanStub {
         const beans = this.beans;
         return {
             paginationSelector: beans.pagination?.getPaginationSelector(),
-            gridHeaderDropZonesSelector: beans.registry.getSelector('AG-GRID-HEADER-DROP-ZONES'),
+            gridHeaderDropZonesSelector: beans.registry?.getSelector('AG-GRID-HEADER-DROP-ZONES'),
             sideBarSelector: beans.sideBar?.getSelector(),
             statusBarSelector: beans.registry?.getSelector('AG-STATUS-BAR'),
             watermarkSelector: (beans.licenseManager as IWatermark)?.getWatermarkSelector(),


StephenCooper:
Fixed in AG-15364 in 34.1

PR #11165, opened 2025-06-30
AG-15305 Update examples

StephenCooper:


documentation/ag-grid-docs/src/content/docs/cell-editing-batch/_examples/batch-editing-api/main.ts
```diff
@@ -69,14 +69,17 @@ function getEditingCells() {
 }
 
 function startBatchEdit() {
+    console.log('Starting batch edit');
```
StephenCooper, line 72: Added logging as there is no other visual feedback that this click has worked.

StephM-W (approved):

PR #11187, opened 2025-06-30
AG-14937 - Fix column move behaviours

AndrewGlazier:


packages/ag-grid-community/src/columnMove/internalColumnMoveUtils.ts
```diff
@@ -51,7 +51,7 @@ export function getBestColumnMoveIndexFromXPosition(
     let { allMovingColumns } = params;
     if (isFromHeader) {
         // If the columns we're dragging are the only visible columns of their group, move the hidden ones too
-        const newCols: AgColumn[] = [];
+        const newCols: AgColumn[] = [...allMovingColumns];
```
gportela85: On line 84 we are also cloning this array? Do we need to do this twice? 
Also, can we break up this monster? 

Some like
```
    let { allMovingColumns } = params;

    if (isFromHeader) {
        allMovingColumns = expandColumnsForMarriedGroups(allMovingColumns);
    }
    //....
    

function expandColumnsForMarriedGroups(columns: AgColumn[]): AgColumn[] {
    const expanded: AgColumn[] = columns.slice();
    for (const col of columns) {
        let group: AgColumnGroup | null = null;
        let parent = col.getParent();

        while (parent && parent.getDisplayedLeafColumns().length === 1) {
            group = parent;
            parent = parent.getParent();
        }

        if (group) {
            const isMarried = !!group.getColGroupDef()?.marryChildren;
            const groupColumns = isMarried
                ? group.getProvidedColumnGroup().getLeafColumns()
                : group.getLeafColumns();

            groupColumns.forEach((col) => {
                if (!expanded.includes(col)) { 
                    expanded.push(col);
                }
            });
        }
    }
    return expanded;
}
```

gportela85 (approved):

PR #11202, opened 2025-07-01
AG-15342 Add a har cache for example-asset json files

StephenCooper:


sdwvit (approved):

sdwvit (approved):

StephenCooper:
This does not speed up the tests when run on GH.

PR #11212, opened 2025-07-03
AG-15304 Fix example

StephenCooper:
To fix https://github.com/ag-grid/ag-grid/actions/runs/16015067061/job/45180014659


seanlandsman (approved):
lgtm

PR #11213, opened 2025-07-03
Ag 15308 return to service campaign

kylerphillips:
Screenshot 2025-07-03 at 11 15 01
Screenshot 2025-07-03 at 11 14 57


documentation/ag-grid-docs/src/components/campaigns-components/FlipCountdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const [left, setLeft] = useState(() => getTimeLeft(new Date(Date.now() + days * 86_400_000)));
+
+    useEffect(() => {
+        const target = new Date(Date.now() + days * 86_400_000);
+        const handle = setInterval(() => setLeft(getTimeLeft(target)), 1000);
+        return () => clearInterval(handle);
+    }, [days]);
+
+    return (
+        <div style={{ display: 'flex', gap: '1em', alignItems: 'center' }}>
```
markdurrant: Let's move these inline styles into the components sass file.

documentation/ag-grid-docs/src/components/campaigns-components/FlipCountdown.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.tick-flip-panel-text-wrapper {
+    #{$selector-darkmode} & {
+        color: var(--color-bg-primary);
+    }
+}
+
+.tick-credits {
+    display: none;
+}
```
markdurrant, line 57: Looks like maybe these classes aren't used?

documentation/ag-grid-docs/src/pages-styles/campaigns.module.scss
markdurrant, line 1: We should move all the RTS campaign stuff to a separate file. See if there's a way to keep this file for more generic styles that could be used across campaigns.

markdurrant (changes requested):
Let's move RTS specific styles to a new file. 

Couple other bits of code tidy-up

documentation/ag-grid-docs/src/components/campaigns-components/FlipCountdown.scss
kylerphillips: These are used for the dark mode and removing the watermark

markdurrant (approved):
LGTM 👍

PR #11230, opened 2025-07-04
AG-15253 - fix batch bulk undo 

StephM-W:


packages/ag-grid-community/src/edit/editService.ts
```diff
@@ -879,6 +891,8 @@ export class EditService extends BeanStub implements NamedBean, IEditService {
             }
 
             this.stopEditing(undefined, { source: 'api' });
+
+            this.eventSvc.dispatchEvent({ type: 'bulkEditingStopped', changes: this.toEventChangeList(edits) });
```
eliasmalik, line 895: If we only dispatch `bulkEditingStarted` when `!this.batch`, should we use the same gate on `bulkEditingStopped`? Would it be possible to get a `...stopped` event without a `...started` event?

StephM-W: Not in this case, there's a if(batch) block a few lines up, with associated return

eliasmalik (approved):

PR #11231, opened 2025-07-04
AG-15262 - update typescript example to match other frameworks

StephM-W:


documentation/ag-grid-docs/src/content/docs/cell-editing-full-row/_examples/full-row-editing/numericCellEditor_typescript.ts
```diff
@@ -47,7 +47,8 @@ export class NumericCellEditor implements ICellEditorComp {
     }
 
     getValue() {
-        return this.eInput.value;
+        const value = this.eInput.value;
+        return value === '' || value == null ? null : parseInt(value);
```
sdwvit, line 51: a test please?

sdwvit (approved):

PR #11234, opened 2025-07-04
[AG-15345] Chore(ci): migrate full build to gha

sdwvit:
- Add full-clean-build GitHub Actions workflow

### DIFF
--- a/.github/workflows/full-clean-build.yml
+++ b/.github/workflows/full-clean-build.yml
@@ -2,7 +2,7 @@ name: Full Clean Build
 
 on:
   schedule:
-    - cron: '0 0 * * *' # Daily at midnight UTC
+    - cron: '0 0 * * *' # Daily at midnight UTC against the latest branch
   workflow_dispatch:
     inputs:
       notify:
--- a/.github/workflows/full-clean-build.yml
+++ b/.github/workflows/full-clean-build.yml
@@ -14,6 +14,10 @@ on:
 env:
   COMMIT_SHA_FILE: ./commit-sha.txt
   SLACK_FILE: ./slack.json
+  BUILD_DIRECTORY: ./.builds
+  WWW_ROOT_DIR: /var/www
+  PUBLIC_ALGOLIA_APP_ID: O1K1ESGB5K
+  LTS_VERSION: b32.3.7
 
 jobs:
   full-clean-build:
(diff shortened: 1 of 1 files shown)


.github/workflows/full-clean-build.yml
```diff
@@ -0,0 +1,230 @@
...
+          else
+            echo "::set-output name=has_changes::true"
+            echo "::set-output name=current_commit::$CURRENT_COMMIT"
+            # Write current commit to file for caching
+            echo "$CURRENT_COMMIT" > {{ env.COMMIT_SHA_FILE }}
+            echo "Changes detected. Proceeding with build."
+          fi
+        id: change-check
+
+      - name: Skip Build if No Changes
```
seanlandsman: this job should run regardless of changes
at least part of its reason is to check for upstream updates that might break things for us

.github/workflows/full-clean-build.yml
```diff
@@ -0,0 +1,199 @@
...
+
+      - name: Sanity Check Base Url
+        run: |
+          node ./scripts/deployments/sanityCheckBaseUrl.js staging
+
+      - name: Build Docs For Staging
+        run: |
+          CHECK_LINKS=true CHECK_REDIRECTS=true yarn nx run ag-grid-docs:build -c staging
+
+      - name: Deploy to Staging
```
seanlandsman, line 103: we should only attempt to deploy to staging if the build and docs build steps pass

.github/workflows/full-clean-build.yml
```diff
@@ -0,0 +1,199 @@
+name: Full Clean Build
+
+on:
```
seanlandsman, line 3: we should probably be explicit about what branch this runs on (ie only on latest)

.github/workflows/full-clean-build.yml
```diff
@@ -0,0 +1,199 @@
...
+  schedule:
+    - cron: '0 0 * * *' # Daily at midnight UTC
+  workflow_dispatch:
```
seanlandsman, line 6: make sure this job runs and finishes *before* the docs tests run otherwise the latter may fail while we update staging

.github/workflows/full-clean-build.yml
```diff
@@ -0,0 +1,199 @@
...
+        default: true
+        required: true
+
+env:
+  COMMIT_SHA_FILE: ./commit-sha.txt
+  SLACK_FILE: ./slack.json
+  BUILD_DIRECTORY: ./.builds
+  WWW_ROOT_DIR: /var/www
+  PUBLIC_ALGOLIA_APP_ID: O1K1ESGB5K
+  LTS_VERSION: b32.3.7
```
seanlandsman, line 20: fine for now but I'd query npm for the latest `lts` version 

not exactly what you want but it might be a useful place to start: https://github.com/ag-grid/ag-grid/blob/latest/scripts/deployments/getLatestGridVersionMatching.js

seanlandsman (approved):

.github/workflows/full-clean-build.yml
sdwvit: docs build runs at 2:30am, and there is a timeout of 2h for this job. Should be enough?

.github/workflows/full-clean-build.yml
sdwvit: good stuff, ty

PR #11249, opened 2025-07-07
AG-15409 - Move VPRM nodes rather than forcing data

AndrewGlazier:


gportela85 (changes requested):
The overall approach looks good, but this code could use some cleanup.
We have the same condition twice and I would favor `for of` instead of forEach.

gportela85 (approved):

PR #11250, opened 2025-07-07
AG-15344 pasting into selection column

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-15344

### DIFF
--- a/packages/ag-grid-community/src/columns/columnUtils.ts
+++ b/packages/ag-grid-community/src/columns/columnUtils.ts
@@ -83,6 +83,10 @@ export function isRowNumberCol(col: ColKey): boolean {
     return id?.startsWith(ROW_NUMBERS_COLUMN_ID) ?? false;
 }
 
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/columns/columnUtils.ts
```diff
@@ -83,6 +83,10 @@ export function isRowNumberCol(col: ColKey): boolean {
     return id?.startsWith(ROW_NUMBERS_COLUMN_ID) ?? false;
 }
 
+export function isSpecialCol(col: ColKey): boolean {
+    return isColumnSelectionCol(col) || isRowNumberCol(col);
+}
+
```
gportela85, line 89: We should probably update these methods to use this util. 
`gridSerializer.getColumnsToExport`
`columnAutoSizeService.autoSizeCols`
`findService.refresh`

gportela85 (changes requested):

gportela85 (approved):

PR #11270, opened 2025-07-08
AG-15308 Add return to service pages

kylerphillips:


taktran (approved):
LGTM 👍🏼 Runs locally too

PR #11286, opened 2025-07-09
AG-15344 prevent pasting into selection column

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-15344

### DIFF
--- a/packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
+++ b/packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
@@ -368,7 +368,9 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
             const currentRowData: (string | null)[] = clipboardData[index - indexOffset];
 
             const selectionColIdx = columns.findIndex(isSpecialCol);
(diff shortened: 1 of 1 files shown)


packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
```diff
@@ -372,35 +372,37 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
-
-                if (changedPath) {
-                    changedPath.addParentNode(rowNode.parent, [column]);
-                }
-
-                const { rowIndex, rowPinned } = currentRow;
-                const cellId = _createCellId({ rowIndex, column, rowPinned });
-                cellsToFlash[cellId] = true;
-            });
+            columns
```
gportela85: Can we change this to be

```
for (let idx = 0; i < columns.length; idx++) {
    const column = columns[idx];
    
    if (
        isSpecialCol(column) ||
        !column.isCellEditable(rowNode) ||
        column.isSuppressPaste(rowNode)
    ) { continue; }
    
    // ... code
}
```

gportela85 (approved):

PR #11287, opened 2025-07-09
Fix(ci): move files around

sdwvit:
- Moved create-jira-ticket.mjs , ctrf-report-to-slack-blocks.mjs, simple-to-slack-blocks.mjs to scripts/test/ directory
 - Updated corresponding gh actions to point to these files


StephenCooper (approved):

seanlandsman (approved):
I'd probably rename the `test` directory to something like actions to be clearer as to where and how its used

sdwvit:
@seanlandsman will do in a follow up

PR #11299, opened 2025-07-10
[AG-15453] Maintenance(ci): trigger doc tests only when full build finishes

sdwvit:
- Comment out the daily schedule for the docs-tests workflow
 - Add conditional dispatch for docs-tests workflow based on build success


seanlandsman (approved):
lgtm

PR #11304, opened 2025-07-10
AG-15451 - Add content api

taktran:
* `/content/index.json`
* `/content/versions.json`
* `/content/changelog.json`
* `/content/frameworks.json`
* `/content/modules.json`
* `/content/types.json`
* `/content/docs/[framework]/index.json`
* `/content/migrations/[framework]/index.json`
* `/content/api/[framework]/index.json`
* `/content/examples/[framework]/[frameworkType]/index.json`

### DIFF
--- a/documentation/ag-grid-docs/src/content.config.ts
+++ b/documentation/ag-grid-docs/src/content.config.ts
@@ -255,16 +255,11 @@ const reactLandingPage = defineCollection({
 const contentApi = defineCollection({
     loader: glob({ base: './src/content/content-api', pattern: 'content-api.json' }),
     schema: z.object({
+        supportedVersionFrom: z.string(),
         root: z.array(
             z.object({
                 id: z.string(),
                 url: z.string(),
-            })
-        ),
-        // Previous versions supported by the content API
-        previousVersions: z.array(
-            z.object({
-                version: z.string(),
             })
         ),
     }),
--- a/documentation/ag-grid-docs/src/content.config.ts
+++ b/documentation/ag-grid-docs/src/content.config.ts
@@ -252,6 +252,24 @@ const reactLandingPage = defineCollection({
     ),
 });
 
+const contentApi = defineCollection({
+    loader: glob({ base: './src/content/content-api', pattern: 'content-api.json' }),
+    schema: z.object({
+        root: z.array(
+            z.object({
+                id: z.string(),
+                url: z.string(),
+            })
+        ),
+        // Previous versions supported by the content API
+        previousVersions: z.array(
+            z.object({
+                version: z.string(),
+            })
+        ),
+    }),
+});
+
 export const collections = {
     docs,
     apiDocumentation,
@@ -269,4 +287,5 @@ export const collections = {
     siteHeader,
     seedProjects,
     reactLandingPage,
+    contentApi,
 };
--- a/documentation/ag-grid-docs/src/pages/content/versions.json.ts
+++ b/documentation/ag-grid-docs/src/pages/content/versions.json.ts
@@ -1,0 +1,60 @@
+import { getArchiveUrl } from '@ag-website-shared/utils/getArchiveUrl';
+import { type CollectionEntry, getEntry } from 'astro:content';
+
+const site = 'grid';
+
+function parseDateString(dateString: string): string {
+    if (!dateString) {
+        return '';
+    }
+    const cleanDateString = dateString.replace(/(\d+)(st|nd|rd|th)/, '$1');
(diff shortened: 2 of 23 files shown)


documentation/ag-grid-docs/src/content.config.ts
```diff
@@ -252,6 +252,24 @@ const reactLandingPage = defineCollection({
...
+                id: z.string(),
+                url: z.string(),
+            })
+        ),
+        // Previous versions supported by the content API
+        previousVersions: z.array(
+            z.object({
+                version: z.string(),
+            })
+        ),
```
AlpineJosh: Is this the right approach here? It would require us to update it on every release. Wouldn't a version cut off be better? e.g. earliestVersion: "34.1.0"

documentation/ag-grid-docs/src/pages/content/versions.json.ts
```diff
@@ -0,0 +1,58 @@
...
+    const date = new Date(cleanDateString);
+
+    if (Number.isNaN(date as any)) {
+        throw new Error('Invalid date format');
+    }
+    const year = date.getFullYear();
+    const month = String(date.getMonth() + 1).padStart(2, '0');
+    const day = String(date.getDate()).padStart(2, '0');
+
+    return `${year}-${month}-${day}`;
```
AlpineJosh, line 19: Would: `date.toISOString().split('T')[0]` be nicer

external/ag-website-shared/src/utils/content-api/urlPaths.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function getContentApiMigrationsUrl({ framework }: { framework: Framework }) {
+    const contentApiExamplesUrl = pathJoin('migrations', framework, 'index.json');
+    return getContentApiPrefix(contentApiExamplesUrl);
+}
+
+export function getContentApiApiUrl({ framework }: { framework: Framework }) {
+    const contentApiExamplesUrl = pathJoin('api', framework, 'index.json');
+    return getContentApiPrefix(contentApiExamplesUrl);
+}
```
AlpineJosh: I don't like this name but I don't think there's much you can do. Unless we rename API to interfaces in the output? e.g. getContentApiInterfacesUrl()?

AlpineJosh (changes requested):
A few minor comments but generally looks good.

documentation/ag-grid-docs/src/pages/content/versions.json.ts
taktran: That code is more reliable, due to weird things with timezones, so keeping it

eg,

```
> new Date("July 8, 2025")
Tue Jul 08 2025 00:00:00 GMT+0100 (British Summer Time)
> new Date("July 8, 2025").toISOString()
'2025-07-07T23:00:00.000Z'
```

documentation/ag-grid-docs/src/content.config.ts
taktran: Good shout, I've updated it to use `supportedVersionFrom` instead of `perviousVersions`. We can refine this in the next iterations as well when it's actually used.

external/ag-website-shared/src/utils/content-api/urlPaths.ts
taktran: I renamed it to be `getContentApiApiDocsUrl` - a bit better?

I'd rather keep it as `api` as that's what the website calls it. It's an internal API anyway, so we can update in future if we come up with a better name.

documentation/ag-grid-docs/src/pages/content/versions.json.ts
```diff
@@ -20,35 +21,54 @@ function parseDateString(dateString: string): string {
...
+                parsedVersion.minor < supportedVersionFrom.minor
+            ) {
+                return false;
+            } else if (
+                parsedVersion.major === supportedVersionFrom.major &&
+                parsedVersion.minor === supportedVersionFrom.minor &&
+                parsedVersion.patch < supportedVersionFrom.patch
+            ) {
+                return false;
+            }
```
AlpineJosh, line 45:
```
if (parsedVersion.major !== supportedVersionFrom.major) return parsedVersion.major > supportedVersionFrom.major;
  else if (parsedVersion.minor !== supportedVersionFrom.minor) return parsedVersion.minor > supportedVersionFrom.minor;
  else return parsedVersion.patch >= supportedVersionFrom.patch;
```

Bit nicer

AlpineJosh (approved):

PR #11309, opened 2025-07-10
AG-12866 access grid api from DOM node

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-12866

### DIFF
--- a/packages/ag-grid-community/src/grid.ts
+++ b/packages/ag-grid-community/src/grid.ts
@@ -306,14 +306,15 @@ function getDefaultRowModelType(passedRowModelType?: RowModelType): RowModelType
  */
 export function getGridApi(gridElement: Element | string | null | undefined): GridApi | undefined {
     if (typeof gridElement === 'string') {
+        const gridIdSelector = `[grid-id="${gridElement}"]`;
         try {
             gridElement =
-                document.querySelector(`[grid-id="${gridElement}"]`)?.parentElement ??
+                document.querySelector(gridIdSelector)?.parentElement ??
                 document.querySelector(gridElement)?.firstElementChild ??
                 document.getElementById(gridElement)?.firstElementChild;
         } catch {
             try {
-                gridElement = document.querySelector(`[grid-id="${gridElement}"]`)?.parentElement;
+                gridElement = document.querySelector(gridIdSelector)?.parentElement;
             } catch {
                 gridElement = null;
             }
--- a/packages/ag-grid-community/src/grid.ts
+++ b/packages/ag-grid-community/src/grid.ts
@@ -108,8 +108,10 @@ export function createGrid<TData>(
         destroyCallback
     );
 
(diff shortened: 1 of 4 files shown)


packages/ag-grid-community/src/grid.ts
```diff
@@ -282,3 +295,28 @@ export class GridCoreCreator {
...
+/**
+ * Returns a `GridApi` instance that is associated with the grid rendered in `gridElement`.
+ *
+ * The `gridElement` argument can be a DOM node or a CSS selector string. It is recommended
+ * to use the selector string argument. This selector must refer to the element passed to `createGrid`.
+ *
+ * If passing a DOM node as an argument, this DOM node must be an immediate child of the element passed
+ * to `createGrid`. This is to support the case where multiple grids are instantiated in a single element.
+ */
+export function getGridApi(gridElement: Element | string | null | undefined): GridApi | undefined {
```
StephenCooper, line 311: I think we are missing the framework usecase i.e when using <AgGridReact> there is no string value passed to createGrid. 
In this case the only string value that might make sense to use is the gridId. 

We could support both, by adding another attempt to get the element via the grid-id attribute. Just need to make sure we are returning a consistent element in the component tree.

StephenCooper (changes requested):
Need to consider the frameworks case.

packages/ag-grid-community/src/grid.ts
```diff
@@ -282,3 +294,37 @@ export class GridCoreCreator {
...
+    if (typeof gridElement === 'string') {
+        try {
+            gridElement =
+                document.querySelector(`[grid-id="${gridElement}"]`)?.parentElement ??
+                document.querySelector(gridElement)?.firstElementChild ??
+                document.getElementById(gridElement)?.firstElementChild;
+        } catch {
+            try {
+                gridElement = document.querySelector(`[grid-id="${gridElement}"]`)?.parentElement;
```
StephenCooper: Why are we repeating the call here?

packages/ag-grid-community/src/grid.ts
```diff
@@ -282,3 +294,37 @@ export class GridCoreCreator {
 function getDefaultRowModelType(passedRowModelType?: RowModelType): RowModelType {
     return passedRowModelType ?? 'clientSide';
 }
+
+/**
+ * Returns a `GridApi` instance that is associated with the grid rendered in `gridElement`.
+ *
+ * The `gridElement` argument can be a DOM node or a CSS selector string. It is recommended
```
StephenCooper: This text needs updating to include gridId usecase.

packages/ag-grid-community/src/grid.ts
eliasmalik: remnant from when the `grid-id` selector was ordered last, we don't need this anymore, thanks for spotting

StephenCooper (approved):

StephenCooper (approved):

PR #11324, opened 2025-07-11
[AG-15334] Maintenance(ci): integrate jira gha into accessibility workflow

sdwvit:
- Add step to create Jira ticket on accessibility workflow failure


seanlandsman (approved):
lgtm

PR #11334, opened 2025-07-14
AG-5652 add timeout back in for angular case

eliasmalik:


testing/public-recipes/e2e/src/simple.spec.ts
```diff
@@ -19,7 +19,9 @@ test.describe('Simple e2e testing examples', () => {
         test(`can load the example and select row in ${fw}`, async ({ page }) => {
             await loadE2ETestingExample(page, fw);
 
-            await expect(page.getByTestId(agTestIdFor.rowNode('row-group-country-South Korea'))).toBeVisible();
+            await expect(page.getByTestId(agTestIdFor.rowNode('row-group-country-South Korea'))).toBeVisible({
+                timeout: 20_000,
```
sdwvit, line 23: Pretty aggressive, perhaps an event listener would be better? just a nit, lgtm otherwise

sdwvit (approved):

PR #11353, opened 2025-07-15
AG-5652 more follow-ups

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-5652

### DIFF
--- a/packages/ag-grid-community/src/testing/testIdUtils.ts
+++ b/packages/ag-grid-community/src/testing/testIdUtils.ts
@@ -7,6 +7,10 @@ function formatTestId(name: string, attributes: Record<string, string | number |
 }
 
 export const agTestIdFor = {
+    root(gridId: string): string {
+        return formatTestId('ag-grid-root', { ['grid-id']: gridId });
+    },
+
     /** Headers */
 
     headerGroupCell(colId: string | null): string {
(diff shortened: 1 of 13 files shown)


packages/ag-grid-community/src/testing/testIdUtils.ts
```diff
@@ -201,3 +201,7 @@ export const agTestIdFor = {
         return formatTestId('ag-paging-page-summary-panel-total-page', { value });
     },
 };
+
+export function getGridRootNode(id: string): Element | null {
```
StephenCooper: Can we add a test id to the root wrapper and have it share the grid-id value. Then we can drop this function and just have a helper above.

testing/module-size/moduleDefinitions.ts
```diff
@@ -48,7 +48,6 @@ export const AllGridCommunityModules: Record<`${CommunityModuleName}Module`, num
     ValidationModule: 69.88,
     ValueCacheModule: 0.65,
     CellSpanModule: 8.08,
-    TestingModule: 11.54,
```
eliasmalik, line 51: @StephenCooper Worth noting: If we don't export the `TestingModule` from main, we can't run bundle size tests on it. Since it's a non-production module, I think that's ok, and we'd prefer not exposing it?

eliasmalik: nvm, I had to re-expose the module. Can't define a module that isn't specified in the types, and if it's specified in the types it needs to be in the module def list here.

StephenCooper (approved):
Just need to simplify the module and then this is good to go.

PR #11354, opened 2025-07-15
AG-1752 tree data master detail

SalvatorePreviti:
Enables master-detail with tree data groups and leafs.

- change the masterDetail to work also with tree data
    - allow masterDetailSvc to run with treeData
    - do not use groupDefaultExpanded when using masterDetail + treeData
    - change the flatten stage to properly insert detail for groups in the right place when treeData
    - fix the default expanded logic for treeData to handle properly also master rows
- refactoring / improvement:
    - centralize the _getRowDefaultExpanded logic between treeData and group
    - remove the beforeRefreshModel event and instead talk directly to the services
- testing
    - added a test for treeData and master detail
    - regenerate the snapshots that contained treeData + master/detail


packages/ag-grid-enterprise/src/rowHierarchy/flattenStage.ts
```diff
@@ -126,6 +130,13 @@ export class FlattenStage extends BeanStub implements IRowNodeStage<RowNode[]>,
                         this.addRowNodeToRowsToDisplay(details, rowNode.sibling, result, uiLevelForChildren);
                     }
 
+                    if (treeDataMasterDetail) {
```
AndrewGlazier: Do we need this check? Surely getDetail will return false if row isn't master anyway

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -25,22 +25,16 @@ export class MasterDetailService extends BeanStub implements NamedBean, IMasterD
...
 
     public postConstruct(): void {
         if (_isClientSideRowModel(this.gos)) {
             this.enabled = this.isEnabled();
-            this.addManagedEventListeners({ beforeRefreshModel: this.beforeRefreshModel.bind(this) });
         }
     }
 
-    private beforeRefreshModel({ params }: BeforeRefreshModelEvent) {
+    public refreshModel(params: RefreshModelParams) {
```
AndrewGlazier, line 39: It looks like this would be better served by just listening for
`rowDataUpdated`
and
gridOption `masterDetail` changed

This approach (refreshModel) is very coupled to the CSRM, this doesn't look like it'll work for SSRM?

 When either of these change this service should dispatch a masterDetailChange event and CSRM and SSRM could listen for this and recalculate row indices (flatten stage and updateRowIndexesAndBounds respectively)

AndrewGlazier:
Note - we haven't captured the SSRM use case and I don't see any real reason to omit it - the work is small

See blockUtils#setDataIntoRowNode and blockUtils#UpdateDataIntoRowNode

AndrewGlazier (approved):

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
SalvatorePreviti: As MasterDetailService requires the set of changed nodes to properly invoke isRowMaster only on the added or updated rows, as per documentation and existing behaviour, we agreed that this refactoring for now can be delayed and we can just use this refreshModel

PR #11365, opened 2025-07-16
AG-15491 set topLevelRowCount correctly when grand total row is pinned

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-15491

### DIFF
--- a/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
+++ b/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
@@ -459,10 +459,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
 
         // we use the childrenAfterSort as postSortRows is occasionally used to reduce row count.
         const filteredChildren = rootNode.childrenAfterSort;
-        const grandTotalRow = _getGrandTotalRow(this.gos);
-        // const totalFooterInc = rootNode.sibling ? 1 : 0;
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -458,7 +459,10 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
 
         // we use the childrenAfterSort as postSortRows is occasionally used to reduce row count.
         const filteredChildren = rootNode.childrenAfterSort;
-        const totalFooterInc = rootNode.sibling ? 1 : 0;
+        const grandTotalRow = _getGrandTotalRow(this.gos);
```
AndrewGlazier: Rather than calling the user callback again would we be able to just check rootNode.sibling.displayed?

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -458,7 +459,10 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
+        // const totalFooterInc = rootNode.sibling ? 1 : 0;
```
AndrewGlazier: I think this is unintended as being left in

AndrewGlazier (approved):

PR #11366, opened 2025-07-16
Dynamically enable test Ids for all examples via url params

StephenCooper:
Example url with test ids enabled

https://localhost:4610/examples/column-updating-definitions/add-remove-columns/vue3?enableTestIds=true

### DIFF
--- a/documentation/ag-grid-docs/public/example-runner/grid-vue3-boilerplate/systemjs.config.dev.js
+++ b/documentation/ag-grid-docs/public/example-runner/grid-vue3-boilerplate/systemjs.config.dev.js
@@ -35,7 +35,7 @@
             typescript: 'npm:typescript@5.4.5/lib/typescript.min.js',
 
             vue: `npm:vue@${VUE_VERSION}/dist/vue.esm-browser.js`,
-            '@vue/reactivity': `npm:@vue/reactivity@${VUE_VERSION}/dist/reactivity.esm-browser.prod.js`,
+            '@vue/reactivity': `npm:@vue/reactivity@${VUE_VERSION}/dist/reactivity.esm-browser.js`,
             // vue class component
             'vue-class-component': 'npm:vue-class-component@^8.0.0-beta.3/dist/vue-class-component.cjs.js',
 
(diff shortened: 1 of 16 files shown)


documentation/ag-grid-docs/public/example-runner/grid-vue3-boilerplate/systemjs.config.dev.js
```diff
@@ -28,8 +34,8 @@
             tslib: 'npm:tslib@2.3.1/tslib.js',
             typescript: 'npm:typescript@5.4.5/lib/typescript.min.js',
 
-            vue: 'npm:vue@3.5.0/dist/vue.esm-browser.js',
-            '@vue/reactivity': 'npm:@vue/reactivity@3.0.0/dist/reactivity.esm-browser.js',
+            vue: `npm:vue@${VUE_VERSION}/dist/vue.esm-browser.js`,
+            '@vue/reactivity': `npm:@vue/reactivity@${VUE_VERSION}/dist/reactivity.esm-browser.prod.js`,
```
StephenCooper:
```suggestion
            '@vue/reactivity': `npm:@vue/reactivity@${VUE_VERSION}/dist/reactivity.esm-browser.js`,
```

plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/frameworkFilesGenerator.ts
```diff
@@ -83,6 +84,8 @@ export const frameworkFilesGenerator: Partial<Record<InternalFramework, ConfigGe
             mainJs = transformEntryFile({ entryFile: mainJs });
         }
 
+        mainJs = getEnableAGTestIdLogic(true) + '\n\n' + mainJs;
```
eliasmalik, line 87: Not sure what this codepath is for exactly, just double checking it's deliberate that we're passing `true` in here?

eliasmalik (approved):

plugins/ag-grid-generate-example-files/src/executors/generate/generator/utils/frameworkFilesGenerator.ts
StephenCooper: yes this is where we handle the javascript code, so passing true to use the umd syntax.

PR #11368, opened 2025-07-16
AG-15342 reduce duplication of cross browser testing

StephenCooper:
Reduce the number of tests we run by only testing `vanilla` and `react TS` across all three browsers. This gives us coverage on the two separate rendering engines but drops the number of total tests dramatically.


seanlandsman (approved):
lgtm

PR #11377, opened 2025-07-16
AG-1752 ssrm tree data master detail

SalvatorePreviti:
enable tree data with SSRM master detail.

- fix in blockUtils and expandListener to handle the combination of treeData and masterDetail
- rebuild the snapshots for auto group column combinatorial tests, the only changed are about tree data and master detail and ssrm
- add an ssrm test for tree data and master detail
- improve and fix the GridRows testing "framework" to handle better ssrm and auto column

Documentation

- change the order of treeData and masterDetail in the sidebar, as it makes sense to add this as last now that is supported by both grouping and treeData
- add in the documentation treeData + masterDetail and an example, that can be used by QA to validate


packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
```diff
@@ -313,25 +333,26 @@ export class BlockUtils extends BeanStub implements NamedBean {
...
-            rowTop: currentRowNode.rowTop!,
-        });
-
         if (rowNode.rowIndex === index) {
             return extractRowBounds(rowNode);
         }
 
+        let result: RowBounds | undefined;
+
+        if (rowNode.master && rowNode.expanded && rowNode.detailNode) {
```
AndrewGlazier: What is this change fixing?

packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
```diff
@@ -268,7 +287,8 @@ export class BlockUtils extends BeanStub implements NamedBean {
         nextRowTop: { value: number },
         uiLevel: number
     ): void {
-        const isUnbalancedGroup = this.gos.get('groupAllowUnbalanced') && rowNode.group && rowNode.key === '';
+        const isUnbalancedGroup =
```
AndrewGlazier: Does treeData need to be explicitly checked here? I don't think groupAllowUnbalanced is supported by treeData regardless? Seems like the check is unnecessary

SalvatorePreviti: this is unfortunately needed for the case the user uses an empty string key for tree data with getDataPath - is a very edge case scenario but is possible. Tree data has not unbalanced groups and I don't think there is other fool proof way to check if this is a unbalanced group but I will think about it

packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
SalvatorePreviti: before we were sowing or detail row or children, if i understand this code correctly the previous implementation was returning only the row height of the detail node - but will investigate further

packages/ag-grid-enterprise/src/serverSideRowModel/listeners/expandListener.ts
```diff
@@ -30,7 +30,8 @@ export class ExpandListener extends BeanStub implements NamedBean {
         if (rowNode.expanded) {
             if (rowNode.master) {
                 this.createDetailNode(rowNode);
-            } else if (_missing(rowNode.childStore)) {
+            }
+            if (_missing(rowNode.childStore)) {
```
AndrewGlazier: Probably also want to check if .hasChildren is true, so we don't create stores for leaf master rows

AndrewGlazier (approved):

PR #11383, opened 2025-07-17
Ag 15342 try caching example assets

StephenCooper:


StephenCooper:
This speeds up the tests locally, but actually slows them down on Github.

PR #11413, opened 2025-07-21
Fix(ci): update jira ticket less aggressive

sdwvit:
- Refined regex to match JIRA tickets more accurately in PR body
 - Removed redundant prefixes from the regex pattern
 - Ensured compatibility with both 'fix' and 'fixes' keywords


markdurrant (approved):
LGTM 👍

PR #11417, opened 2025-07-21
Ag 14964 hide columns until data

StephenCooper:
Fix [AG-13402](https://ag-grid.atlassian.net/browse/AG-13402)
Fix [AG-14964](https://ag-grid.atlassian.net/browse/AG-14964)


documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/column-resizing/main.ts
```diff
@@ -45,16 +45,11 @@ const gridOptions: GridOptions<IOlympicData> = {
...
     },
 };
-
 function autoSizeAll(skipHeader: boolean) {
-    const colIds: string[] = [];
-    gridApi!.getColumns()!.forEach((column) => {
-        colIds.push(column.getId());
-    });
-
-    gridApi!.autoSizeColumns({ colIds, skipHeader, defaultMaxWidth: 150, defaultMinWidth: 80 });
```
StephenCooper, line 57: I removed the colIds as it is matching the default behaviour seeing that it is providing all of them.

packages/ag-grid-community/src/columns/columnFlexService.ts
```diff
@@ -79,8 +75,12 @@ export class ColumnFlexService extends BeanStub implements NamedBean {
                 targetSize: 0,
             };
         });
+        // hide all columns and cells because we are going to flex them after they are displayed
+        if (hasFlexItems) {
+            colDelayRenderSvc?.hideColumns('colFlex');
+        }
 
-        if (!hasFlexItems) {
+        if (!totalSpace || !hasFlexItems) {
```
StephenCooper, line 83: Changed the order so that on the first pass before the gridBody has rendered we know that we will need to hide the columns and so can set the flag.

packages/ag-grid-community/src/gridBodyComp/scrollVisibleService.ts
```diff
@@ -31,6 +31,9 @@ export class ScrollVisibleService extends BeanStub implements NamedBean {
     public verticalScrollGap: boolean;
 
     public postConstruct(): void {
+        this.horizontalScrollShowing = this.gos.get('alwaysShowHorizontalScroll') === true;
+        this.verticalScrollShowing = this.gos.get('alwaysShowVerticalScroll') === true;
+
         // sets an initial calculation for the scrollbar width
```
StephenCooper, line 37: By defaulting to true if these properties have been set, it enables users to avoid the scroll flicker behaviour where it changes the width slightly.

packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
```diff
@@ -474,16 +494,12 @@ export class ColumnAutosizeService extends BeanStub implements NamedBean {
...
-        const autoSizeStrategy = this.gos.get('autoSizeStrategy');
-        if (autoSizeStrategy?.type !== 'fitCellContents') {
-            return;
-        }
-
-        const { colIds: columns, skipHeader, defaultMaxWidth, defaultMinWidth, columnLimits } = autoSizeStrategy;
+    private onFirstDataRendered(strategy: SizeColumnsToContentStrategy): void {
+        const { type, colIds: columns, skipHeader, defaultMaxWidth, defaultMinWidth, columnLimits } = strategy;
         // ensure render has finished
         setTimeout(() => {
```
SalvatorePreviti, line 507: so, this problem pre-existed and might be worse now - we should or clear the timer or check if the grid was destroyed before running stuff - there is the risk the grid was destroyed when executing both timers in this class

SalvatorePreviti (approved):

PR #11437, opened 2025-07-22
AG-10372 - add suppress mouse event callback

peterjrreynolds:
Fix [AG-10372](https://ag-grid.atlassian.net/browse/AG-10372)


gportela85 (approved):

packages/ag-grid-community/src/rendering/cell/cellMouseListenerFeature.ts
```diff
@@ -134,6 +147,20 @@ export class CellMouseListenerFeature extends BeanStub {
...
+        const fireMouseDownEvent = () => {
+            const cellMouseDownEvent = cellCtrl.createEvent(mouseEvent, 'cellMouseDown') as CellMouseDownEvent;
+            cellMouseDownEvent.isEventHandlingSuppressed = suppressMouseEvent;
+            eventSvc.dispatchEvent(cellMouseDownEvent);
+        };
+
+        if (suppressMouseEvent) {
+            fireMouseDownEvent();
+            return;
+        }
```
StephenCooper, line 164: Is this logic correct? If so can we have a comment as it is confusing having suppressMouseEvent true to then fire the mouse event.

peterjrreynolds: Yes - suppress just prevents the grid handling; it does not prevent the events being passed to applications (but we do set the suppress property in them).

StephenCooper (approved):

PR #11442, opened 2025-07-22
AG-15532 - Fix for changed col groups not causing a viewport update

AndrewGlazier:


packages/ag-grid-community/src/columns/columnViewportService.ts
```diff
@@ -267,9 +271,10 @@ export class ColumnViewportService extends BeanStub implements NamedBean {
 
         this.extractViewportColumns();
         const newHash = this.getViewportColumns().map(hashColumn).join('#');
-        const changed = this.colsWithinViewportHash !== newHash;
+        const changed = this.forceRefreshNextRender || this.colsWithinViewportHash !== newHash;
```
gportela85: I think it would be simpler to make `colsWithinViewportHash` public and set it to `''` in `ColumnGroupService.setupParentsIntoCols`. This way we wouldn't need this additional flag.

gportela85:
Broken checks

gportela85 (approved):

PR #11453, opened 2025-07-23
Fix(ci): jira automated message format for api v3

sdwvit:
Extract jiraLink to shared utility and use in JIRA scripts

- Add jiraLink to _utils.mjs for centralized link formatting
- Replace inline link creation with jiraLink in create-jira-ticket.mjs
- Remove redundant jiraLink implementation from gen-jira-files.mjs

Follow up for ag-grid/ag-grid#11446


taktran (approved):
👍🏼

PR #11459, opened 2025-07-23
[AG-15421] Fix(ssrm): when selecting a cell range spanning many rows pressing ctrl c to copy the cell range performance is slow

sdwvit:
Fix server-side row model handling in clipboard and events (only in-memory cells are copied)

Fixes [AG-15421](https://ag-grid.atlassian.net/browse/AG-15421)
Fixes [AG-15559](https://ag-grid.atlassian.net/browse/AG-15559)
Fixes [AG-15557](https://ag-grid.atlassian.net/browse/AG-15557)

| Before | After |
|--------|--------|
| Screenshot 2025-07-23 at 17 10 38 | Screenshot_20250725_134444 | 
| [Trace-Before.json](https://github.com/user-attachments/files/21391511/Trace-20250723T150308.json) | [Trace-after.json](https://github.com/user-attachments/files/21391508/Trace-20250723T170959.json) |
| Screenshot_20250725_150132 | Screenshot_20250725_150135 |
| Screenshot_20250725_150748 | Screenshot_20250725_150717 |


packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
```diff
@@ -952,10 +955,19 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
         const lastRow = rangeSvc.getRangeEndRow(range);
-
+        const isSSRM = _isServerSideRowModel(this.gos, this.beans.rowModel);
+        const isInCache = (rowPosition: RowPosition | null): rowPosition is RowPosition => {
+            if (isSSRM && rowPosition) {
+                return !!(this.beans.rowModel as ServerSideRowModel)
+                    .getRootStore()
+                    ?.getCache()
+                    .getNodes()
+                    .getBy('index', rowPosition.rowIndex);
```
AndrewGlazier: The problem with this approach is that this doesn't handle recursive stores; so I assume copying the contents of a group here may not work anymore

packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
```diff
@@ -952,10 +955,19 @@ export class ClipboardService extends BeanStub implements NamedBean, IClipboardS
...
+            }
+            return !!rowPosition;
+        };
         let node: RowPosition | null = startRow;
-
-        while (node) {
+        while (isInCache(node)) {
```
AndrewGlazier: This might also be a problem, if using maxBlocksInCache and I scroll to the bottom of the grid, do ctrl A and then copy, the first rows won't be in the cache but the latter ones will, but this assumes rows are always loaded contiguously from position 0

AndrewGlazier (changes requested):

packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
AndrewGlazier: Also note; does RowPosition contain a pinned flag? If so the pinned row may not be handled by the row model

packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
sdwvit: Right, so we need to just rely on cache and copy its contents

packages/ag-grid-enterprise/src/clipboard/clipboardService.ts
sdwvit: oh you are implying that we need to fetch data recurcively on ctrl+c, even if the group is not expanded?

AndrewGlazier: Not if the group isn't expanded, but the root cache only contains the rowIndices for the top level group, so even an expanded group wouldn't be captured by this (I think?)

AndrewGlazier (approved):

PR #11468, opened 2025-07-24
AG-15562 Enable playwright tests for examples

StephenCooper:
Enable adding playwright tests alongside documentation examples.


documentation/ag-grid-docs/tsconfig.json
```diff
@@ -33,12 +33,12 @@
       "@ag-website-shared/*": ["../../external/ag-website-shared/src/*"],
 
       // Get auto completion in _examples
-      "ag-grid-community": ["../../packages/ag-grid-community/src/main.ts"],
-      "ag-grid-enterprise": ["../../packages/ag-grid-enterprise/src/main.ts"],
+      // "ag-grid-community": ["../../packages/ag-grid-community/src/main.ts"],
```
taktran: Is this intentional?

taktran (approved):
Looks good - just a comment about TS config

documentation/ag-grid-docs/tsconfig.json
StephenCooper: yes, I will actually remove these comments. From my local tests moving the ag grid deps from devDeps to deps fixes the typing issues that these were added for in the first place. 

These path imports cause playwright to import ag-grid as commonJs which did not work.

.github/workflows/doc-tests.yml
```diff
@@ -24,11 +24,11 @@ on:
         type: boolean
         required: true
         default: true
-      skip_reactFunctional:
-        description: 'Skip React Functional tests'
-        type: boolean
-        required: true
-        default: true
+      # skip_reactFunctional:
```
eliasmalik, line 27: do we want to keep all the react tests commented out?

documentation/ag-grid-docs/project.json
```diff
@@ -136,6 +136,35 @@
...
+      "options": {
+        "commands": ["npx playwright install --with-deps chromium"],
+        "parallel": false,
+        "cwd": "documentation/ag-grid-docs"
+      },
+      "cache": true
+    },
+    "test:e2e": {
+      "executor": "nx:run-commands",
+      "dependsOn": ["test:e2e:setup", "ag-grid-community:build"],
```
eliasmalik: Just double checking: we depend on community build so we have access to `setupAgTestIds` and the other helpers?

documentation/ag-grid-docs/src/content/docs/column-moving/_examples/moving-simple/example.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { expect, test } from '@playwright/test';
+import { testAllFrameworks } from '@utils/grid/test-utils';
+
+import { wrapAgTestIdFor } from 'ag-grid-community';
+
+testAllFrameworks('', 'column-moving/moving-simple', async (page) => {
+    test.skip(true, 'This test is skipped until the issue with column moving is resolved');
```
eliasmalik: Is this the column virtualization issue you mentioned? is it not fixed now?

documentation/ag-grid-docs/src/content/docs/aggregation/_examples/aggregation-overview/example.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { expect } from '@playwright/test';
+import { testAllFrameworks } from '@utils/grid/test-utils';
+
+import { wrapAgTestIdFor } from 'ag-grid-community';
+
+const pageExampleUrl = 'aggregation/aggregation-overview';
+testAllFrameworks('', pageExampleUrl, async (page) => {
```
eliasmalik: All the example tests have no names?

StephenCooper: Names will only be required if you have multiple test cases. I could add a generic string there for now as that might be less confusing for people.

documentation/ag-grid-docs/project.json
StephenCooper: Yes thats correct. Without it it fails.

.github/workflows/doc-tests.yml
StephenCooper: Yes, can clean it up later if we are confident this is correct.

documentation/ag-grid-docs/src/content/docs/column-moving/_examples/moving-simple/example.spec.ts
StephenCooper: Updated to be react specific which is a real code bug.

PR #11469, opened 2025-07-24
AG-13120 coldefs update do not update auto group cols

SalvatorePreviti:
The check for understanding if a full regroup after columns change was too simple to handle the case of relevant colDef fields changing
This replaces the old detection logic with a new more cohomprensive check implemented in makeGroupColumns and groupColumnsChanged both for columns with group:true and with autoGroupColumns

Add also a test to verify this case


packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupedColumnts.ts
```diff
@@ -0,0 +1,57 @@
+import type { AgColumn, ColDef } from 'ag-grid-community';
```
AndrewGlazier: File name has a typo (Columnts.ts)

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupedColumnts.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (len !== columns?.length) {
+        return true;
+    }
+    for (let i = 0; i < len; i++) {
+        const a = groupedColumns[i];
+        const b = columns[i];
+        const bColDef = b.getColDef();
+        if (
+            a.col !== b ||
+            a.id !== b.getId() ||
```
AndrewGlazier: This check is unnecessary as we already check the refs

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupedColumnts.ts
```diff
@@ -0,0 +1,57 @@
...
+            a.field !== bColDef.field ||
+            a.type !== bColDef.type ||
+            a.cellDataType !== bColDef.cellDataType ||
```
AndrewGlazier: Why does cellDataType or type changing impact our data grouping?

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -74,8 +71,8 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
     // rowNode.childrenMapped: string=>RowNode = children mapped by group key (when groups) or an empty map if leaf group (this is then used by pivot)
     // for leaf groups, rowNode.childrenAfterGroup = rowNode.allLeafChildren;
 
-    private oldGroupingDetails: GroupingDetails;
-    private oldGroupDisplayColIds: string;
+    private groupCols: GroupColumn[] | null = null;
```
AndrewGlazier: This naming is less clear than the old one, oldGroupCols or lastGroupCols or prevGroupCols would be more clear

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -138,22 +135,41 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
...
+        }
 
-        const groupedCols = this.rowGroupColsSvc?.columns;
+        let showGroupColsChanged = false;
+        const showCols = this.showRowGroupCols.getShowRowGroupCols();
+        const showGroupCols = this.showGroupCols;
+        if (
+            !showGroupCols ||
+            ((groupColsChanged || afterColumnsChanged || !changedRowNodes) &&
+                groupColumnsChanged(showGroupCols, showCols))
```
AndrewGlazier: Why did we move this logic here? It's a bit harder to me to follow the code flow as the code path is now more ambiguous, before it was clear it only ran on shotgunReset (non-transaction), now this has to be inferred implicitly from the if statement

AndrewGlazier (changes requested):

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupedColumnts.ts
SalvatorePreviti: oh wow :) do I need glasses

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupedColumnts.ts
SalvatorePreviti: not required, i can remove it

AndrewGlazier (approved):

PR #11524, opened 2025-07-28
RTI-3004 - Fix SSRM transactions example

AndrewGlazier:
Fix RTI-3004

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-updating-transactions/_examples/transactions-grouping/data.ts
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-updating-transactions/_examples/transactions-grouping/data.ts
@@ -83,7 +83,11 @@ export let data = [
 
 export function deletePortfolioOnServer(portfolio) {
     const oldDataSize = data.length;
-    data = data.filter((record) => record.portfolio !== portfolio);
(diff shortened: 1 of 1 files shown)


documentation/ag-grid-docs/src/content/docs/server-side-model-updating-transactions/_examples/transactions-grouping/data.ts
```diff
@@ -83,7 +83,11 @@ export let data = [
 
 export function deletePortfolioOnServer(portfolio) {
     const oldDataSize = data.length;
-    data = data.filter((record) => record.portfolio !== portfolio);
+    const filteredData = data.filter((record) => record.portfolio !== portfolio);
+    // need to maintain original data reference
+    data.length = 0;
```
sdwvit, line 88: I am not a fan of this syntax. What if there is a reference to the element at position higher than filteredData length? If that case in not valid then ignore this comment.

```suggestion
    A.splice(0, A.length);
```

sdwvit: ok just retested. it doesn't apply and elements after filteredData.length get cleared

sdwvit (approved):

PR #11530, opened 2025-07-28
Adding button styles for examples

AlpineJosh:


taktran (approved):
LGTM, but asking @markdurrant for his opinion too

documentation/ag-grid-docs/src/components/example-runner/components/exampleRuntimeInjectedStyles.jsx
```diff
@@ -59,6 +59,16 @@ html button:not(.ag-root-wrapper button, .ag-chart button, button[class*='ag-'])
...
+html button:not(.ag-root-wrapper button, .ag-chart button, button[class*='ag-']):disabled {
+    background-color: var(--background-100);
+    opacity: 0.75
+}
+
+html button:not(.ag-root-wrapper button, .ag-chart button, button[class*='ag-']):disabled:hover {
+    background-color: var(--background-100);
+    opacity: 0.75
+}
+
```
markdurrant: Good improvement. Couple of ammends I think we can make. 


```CSS
html button:not(.ag-root-wrapper button, .ag-chart button, button[class*='ag-']):disabled {
    opacity: 0.66; /* Make effect more noticeable in lightmode */
    pointer-events: none; /* Remove need for :hover selector, remove { pointer: cursor } */
}

```

Don't think background colour does anything here?

markdurrant (changes requested):
Good idea @AlpineJosh. Think CSS can be tweaked a little

documentation/ag-grid-docs/src/components/example-runner/components/exampleRuntimeInjectedStyles.jsx
AlpineJosh: Thanks Mark. Don't think I knew that css pointer-events removed the need for hover styles. Very useful.

markdurrant (approved):
All LGTM 👍

documentation/ag-grid-docs/src/components/example-runner/components/exampleRuntimeInjectedStyles.jsx
markdurrant: NP 👍

PR #11549, opened 2025-07-29
AG-15626 - Use `app.component.ts` for adding console logger code for provided examples

taktran:
Fixes bug where angular provided examples were not logging.

### DIFF
--- a/plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
+++ b/plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
@@ -411,8 +411,9 @@ async function processProvidedFiles(
     }
 
     // Transform entry file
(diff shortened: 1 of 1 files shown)


plugins/ag-grid-generate-example-files/src/executors/generate/executor.ts
```diff
@@ -411,8 +411,9 @@ async function processProvidedFiles(
     }
 
     // Transform entry file
-    provideFrameworkFiles[entryFileName] = transformEntryFile({
-        entryFile: provideFrameworkFiles[entryFileName],
+    const transformMainFile = internalFramework === 'angular' ? mainFileName : entryFileName;
```
taktran, line 414: This only applies to provided examples

StephenCooper (approved):

PR #11556, opened 2025-07-29
AG-14577 synchronous set loading overlay text

SalvatorePreviti:
The loading text should not be set asynchronously as it would prevent the loading screen to appear if there is some sync processing going on that would block the main thread after setting loading=true.
There was a setTimeout added to force screen readers to announce the grid is loading, so, here there is another proposal of a solution to handle both cases.

This relates also to https://ag-grid.atlassian.net/browse/AG-15086


StephenCooper:
Closing as all Jiras related to this work have been marked as Done

PR #11614, opened 2025-07-31
AG-4648 Date hierarchy

eliasmalik:
See https://ag-grid.atlassian.net/browse/AG-4648


packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -93,7 +93,16 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
-        const { valueCache, colAutosize, rowGroupColsSvc, pivotColsSvc, valueColsSvc, visibleCols, eventSvc } = beans;
+        const {
+            valueCache,
+            colAutosize,
+            rowGroupColsSvc,
+            pivotColsSvc,
+            valueColsSvc,
+            visibleCols,
+            eventSvc,
+            dateHierarchyColSvc,
```
AndrewGlazier: This won't necessarily just be used for dates, so maybe a different name would be better

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -180,7 +194,7 @@ export class ColumnModel extends BeanStub implements NamedBean {
 
         const cols = this.selectCols(pivotResultCols, this.colDefCols);
 
-        this.createColumnsForService([autoColSvc, selectionColSvc, rowNumbersSvc], cols);
+        this.createColumnsForService([autoColSvc, selectionColSvc, rowNumbersSvc, dateHierarchyColSvc], cols);
```
AndrewGlazier: Why does this need done here? Isn't this just done after primary cols are set?

packages/ag-grid-enterprise/src/rowHierarchy/dateHierarchyColService.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    public isDateHierarchyColsEnabled(cols: _ColumnCollections): boolean {
+        return cols.list.some((col) => this.isDateHierarchyColsEnabledForCol(col));
+    }
+
+    public isDateHierarchyColsEnabledForCol(col: AgColumn): boolean {
+        const { dataTypeSvc } = this.beans;
+        const def = col.getColDef();
+        const isDateCol = dataTypeSvc?.getBaseDataType(col)?.includes('date');
```
AndrewGlazier: There's a weird race condition to be careful of here; cellDataTypes "auto" waits for row data to be set before they are set on the column, not sure if this is an issue or not, but keep in mind

packages/ag-grid-enterprise/src/rowHierarchy/dateHierarchyColService.ts
```diff
@@ -0,0 +1,254 @@
...
+                return `${sourceName} (${part})`;
+            }
+            return '';
+        };
+
+        switch (part) {
+            case 'year':
+                return {
+                    ...base,
+                    colId: `${sourceColDef.colId ?? sourceColDef.field}-year`,
```
AndrewGlazier: As these are auto cols, we should probably prefix with something like `ag-Grid-HierarchyColumn`

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -1461,6 +1461,12 @@ export interface GridOptions<TData = any> {
      */
     suppressGroupRowsSticky?: boolean;
 
+    /**
+     * Custom group hierarchy components can be defined here for later use in `colDef.rowGroupingHierarchy`
+     * @agModule `RowGroupingModule` / `TreeDataModule`
```
AndrewGlazier: Does TreeDataModule make sense? You can't group with tree data module enabled

documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
```diff
@@ -481,7 +481,8 @@
                 "name": "Custom Group Columns",
                 "url": "./grouping-custom-group-columns/"
             }
-        }
+        },
+        "rowGroupingHierarchy": {}
```
AndrewGlazier: Can you link this through to the doc section

packages/ag-grid-community/src/columns/baseColsService.ts
```diff
@@ -287,6 +287,13 @@ export abstract class BaseColsService extends BeanStub implements IColsService {
         // so we are preserving order of current grouping of columns that simply have rowGroup=true
         previousCols.forEach((col) => {
             if (colsWithValue.indexOf(col) >= 0) {
+                // ...with the exception that if we have columns with associated virtual columns, they need
```
AndrewGlazier: TODO SELF: Need to dive into understanding this more

packages/ag-grid-community/src/interfaces/iGroupHierarchyColService.ts
```diff
@@ -0,0 +1,9 @@
+import type { ColumnCollections } from '../columns/columnModel';
+import type { AgColumn } from '../entities/agColumn';
+import type { IColumnCollectionService } from './iColumnCollectionService';
+
+export interface IGroupHierarchyColService extends IColumnCollectionService {
+    isDateHierarchyColsEnabled(cols: ColumnCollections): boolean;
```
AndrewGlazier: Do we still want the date convention here?

packages/ag-grid-enterprise/src/rowGrouping/rowGroupColsSvc.ts
```diff
@@ -96,14 +96,28 @@ export class RowGroupColsSvc extends BaseColsService implements NamedBean, ICols
 
         this.setColRowGroupActive(column, active, source);
 
-        if (_shouldUpdateColVisibilityAfterGroup(this.gos, active)) {
+        if (
+            _shouldUpdateColVisibilityAfterGroup(this.gos, active) &&
+            !this.beans.groupHierarchyColSvc?.getColumn(column)
```
AndrewGlazier: Can you add a comment explaining this - I needed to dig for why so we probably could use a comment - 

Might also be worth checking with Zoheil that we _don't_ want to show the col when the col has been un grouped, I'd be tempted to say it still should

packages/ag-grid-enterprise/src/rowGrouping/rowGroupColsSvc.ts
```diff
@@ -96,14 +96,28 @@ export class RowGroupColsSvc extends BaseColsService implements NamedBean, ICols
...
             this.colModel.setColsVisible([column], !active, source);
         }
     }
 
     private setColRowGroupActive(column: AgColumn, rowGroup: boolean, source: ColumnEventType): void {
         if (column.rowGroupActive !== rowGroup) {
             column.rowGroupActive = rowGroup;
+
+            const { groupHierarchyColSvc } = this.beans;
+            const vCols = groupHierarchyColSvc?.getVirtualColumnsForColumn(column) ?? [];
```
AndrewGlazier: nit: can we rename this to hierarchyCols or something? I think other than the `groupHierarchyColSvc` reading this bit of code will be a bit hard to contextually ingest if you don't know what you're looking at already

packages/ag-grid-enterprise/src/rowGrouping/rowGroupColsSvc.ts
```diff
@@ -96,14 +96,28 @@ export class RowGroupColsSvc extends BaseColsService implements NamedBean, ICols
...
+            if (rowGroup && vCols.length > 0) {
+                for (let i = vCols.length - 1; i >= 0; i--) {
+                    if (!this.columns.includes(vCols[i])) {
+                        this.columns.unshift(vCols[i]);
```
AndrewGlazier: nit, as performance here probably doesn't matter: unshift can be pretty slow, I assume it's optimised to only have to shift things once if multiple args are provided, so building an array and unshifting once with a spread is _probably_ better

packages/ag-grid-enterprise/src/rowHierarchy/groupHierarchyColService.ts
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    public getVirtualColumnsForColumn(col: AgColumn): AgColumn[] {
+        if (this.isDateHierarchyColsEnabledForCol(col)) {
+            return this.sourceColumnMap.get(col) ?? [];
+        }
+        return [];
+    }
+
+    public isDateHierarchyColsEnabled(cols: _ColumnCollections): boolean {
```
AndrewGlazier: nit: lots of 'date' references

packages/ag-grid-enterprise/src/rowHierarchy/groupHierarchyColService.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    private createDateHierarchyColDefs(sourceCol: AgColumn): ColDef[] {
+        const colDefs: ColDef[] = [];
+        const sourceColDef = sourceCol.getColDef();
+
+        if (!sourceColDef.rowGroupingHierarchy) {
+            return colDefs;
+        }
+
+        if (this.isDateHierarchyColsEnabledForCol(sourceCol)) {
```
AndrewGlazier: nit: might be nice to keep this shallow by early returning the inverse

packages/ag-grid-enterprise/src/rowHierarchy/groupHierarchyColService.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    public isDateHierarchyColsEnabled(cols: _ColumnCollections): boolean {
+        return cols.list.some((col) => this.isDateHierarchyColsEnabledForCol(col));
+    }
+
+    public isDateHierarchyColsEnabledForCol(col: AgColumn): boolean {
+        const { dataTypeSvc } = this.beans;
+        const def = col.getColDef();
+        const isDateCol =
+            dataTypeSvc?.isColPendingInference(col.getColId()) || dataTypeSvc?.getBaseDataType(col)?.includes('date');
```
AndrewGlazier: I don't think we need to use cellDataTypes for now, it should be sufficient to check the new flag (rowGroupingHierarchy), as if that's present it doesn't matter what the underlying type is, users can provide their own drill down

packages/ag-grid-enterprise/src/rowHierarchy/groupHierarchyColService.ts
```diff
@@ -0,0 +1,271 @@
...
+                rowGroup: sourceColDef.rowGroup,
+                enablePivot: sourceColDef.enablePivot,
+                hide: true,
+                editable: false,
+            },
+            'dummy',
+            true
+        );
+
+        const getDate = (node: IRowNode | null): Date | null => {
```
AndrewGlazier: Might be a good idea to split these into some sort of utils files, and keep the groupHierarchyColsSvc about the actual cols rather than date processing

packages/ag-grid-enterprise/src/rowHierarchy/groupHierarchyColService.ts
```diff
@@ -0,0 +1,271 @@
...
+    _addColumnDefaultAndTypes,
+    _areColIdsEqual,
+    _columnsMatch,
+    _destroyColumnTree,
+    _getDateParts,
+    _parseDateTimeFromString,
+    _updateColsMap,
+} from 'ag-grid-community';
+
+export class GroupHierarchyColService extends BeanStub implements NamedBean, IGroupHierarchyColService {
```
AndrewGlazier: What's the impact of this new svc on bundle size? If it's significant should we consider making this a new module?

AndrewGlazier (changes requested):

AndrewGlazier (approved):

PR #11616, opened 2025-07-31
AG-5867 ensure refresh cells invokes row drag callback

SalvatorePreviti:
- add a new method `refresh` in rowDragComp that is invoked by cellComp to update the visibility of the drag handle
- simplify the rowDragComp logic and complexity by removing the two strategies classes and implement the logic in a centralised refresh method
- remove the call to workoutVisibility from postConstruct as is in any case called by refresh cells
- add unit tests to verify rowDrag is called and works as intended

### DIFF
--- a/packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
+++ b/packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
@@ -88,14 +88,19 @@ export class RowDragComp extends Component {
             return;
         }
 
+        const column = this.column;
+
         // if shown sometimes, them some rows can have drag handle while other don't,
         // so we use setVisible to keep the handles horizontally aligned (as _setVisible
         // keeps the empty space, whereas setDisplayed looses the space)
-        const shownSometimes = typeof this.column?.getColDef().rowDrag === 'function';
+        let shownSometimes = typeof column?.getColDef().rowDrag === 'function';
+        let visible = !column || this.isCustomGui() || column.isRowDrag(this.rowNode);
+        if (visible && this.rowNode.footer && this.gos.get('rowDragManaged')) {
+            visible = false; // We hide footer rows in row drag managed mode
+            shownSometimes = true;
+        }
 
-        const visible = this.isVisible();
-        const displayed = shownSometimes || visible;
-        this.setDisplayed(displayed, displayedOptions);
+        this.setDisplayed(shownSometimes || visible, displayedOptions);
         this.setVisible(visible, displayedOptions);
     }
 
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
```diff
@@ -60,6 +71,53 @@ export class RowDragComp extends Component {
...
+        let alwaysHidden = false;
+
+        if (isManaged) {
+            // Managed: only show if not prevented and not suppressed, or if there are external drop zones
+            const shouldPreventRowMove = rowDragSvc!.rowDragFeature?.shouldPreventRowMove();
+            const hasExternalDropZones = dragAndDrop?.hasExternalDropZones();
+            neverDisplayed = (shouldPreventRowMove && !hasExternalDropZones) || suppressRowDrag;
+            alwaysHidden = !!this.rowNode.footer;
+        } else {
+            // Non-managed: only show if not suppressed
```
peterjrreynolds: If source is `sortChanged`, `filterChanged` or `columnRowGroupChanged`, there is no need to run any code below this point (it didn't run previously).

packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
```diff
@@ -60,6 +71,53 @@ export class RowDragComp extends Component {
         this.addDragSource(dragStartPixels);
     }
 
+    public refresh(): void {
+        if (this.suppressVisibilityChange) {
```
gportela85: Ideally we should break up this method to look something like this:
```
public refresh(): void {
    if (this.suppressVisibilityChange) {
        return;
    }

    const displayedOptions = { skipAriaHidden: true };

    if (this.shouldNeverDisplay()) {
        this.setDisplayed(false, displayedOptions);
        return;
    }

    const { shown, isShownSometimes } = this.computeVisibilityStates();

    if (isShownSometimes) {
        this.setDisplayed(true, displayedOptions);
        this.setVisible(shown, displayedOptions);
    } else {
        this.setDisplayed(shown, displayedOptions);
        this.setVisible(shown, displayedOptions);
    }
}
```

packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
SalvatorePreviti: ok will put back the condition in postConstruct to select a different set of events to listen to as it was before

peterjrreynolds (approved):

gportela85 (approved):

PR #11622, opened 2025-07-31
AG-15476 - simplify edit overlay

StephM-W:
Fix AG-15476

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/cell-editing-batch/_examples/batch-editing-api/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/cell-editing-batch/_examples/batch-editing-api/example.spec.ts
@@ -1,0 +1,49 @@
+import { expect } from '@playwright/test';
+// Import the test helper from test-utils
+import { testAllFrameworks } from '@utils/grid/test-utils';
+
+const testUrl = 'cell-editing-batch/batch-editing-api';
+
+testAllFrameworks('Edit + Dependent Renderer', testUrl, async ({ page, agIdFor }) => {
+    const cell = agIdFor.cell('0', 'gold');
+
+    // initiate cell editing by double clicking the cell
+    await cell.dblclick();
+    const cellEditor = cell.locator('input');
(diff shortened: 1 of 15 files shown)


packages/ag-grid-community/src/interfaces/iEditModelService.ts
```diff
@@ -29,6 +29,11 @@ export type GetEditsParams = {
     withOpenEditor?: boolean;
 };
 
+export interface GetEditingRowValuesParams {
+    /** The row node to get the editing values from */
+    rowNode: IRowNode;
```
peterjrreynolds: Only one property, so don't need params

StephM-W: Simplified, thanks

documentation/ag-grid-docs/src/content/docs/cell-editing-batch/_examples/batch-editing-api/example.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    await cell.dblclick();
+    const cellEditor = cell.locator('input');
+    await expect(cellEditor).toBeVisible();
+
+    await page.keyboard.type('100'); // type in a new value
+    await page.keyboard.press('Enter'); // press Enter to save the value
+
+    await expect(cellEditor).toHaveCount(0); // verify the cell editor is closed
+    await expect(cell).toHaveText('100'); // verify the cell has the new value
+    await expect(cell).not.toHaveClass(/ag-cell-batch-edit/);
```
eliasmalik, line 20: from what I can tell this test doesn't go into batch edit mode so would this assertion always be false?

StephM-W: We've had a regression on this, where the style was misapplied, so this is more to validate it doesnt reoccur.

peterjrreynolds (approved):

PR #11642, opened 2025-08-01
Add sonarqube steps to the CI

seanlandsman:


.github/workflows/ci.yml
```diff
@@ -611,3 +608,52 @@ jobs:
             echo "Workflow failed, failing the build."
             exit 1
 
+  sonar_community:
```
taktran, line 611: Maybe worth doing it as a shared action, so we can use it on charts?

seanlandsman: Not an option - we have them separate due to historical reasons

taktran (approved):
LGTM

PR #11643, opened 2025-08-01
(angular) : Lazy load modules ag-grid-angular

medbenmakhlouf:
💡 Why This Change?

This PR updates the documentation to demonstrate how to lazy load **Ag Grid modules in Angular 17+** using the new provideAgGrid` API.

🆕 What's Included

  - Example using modules and options as lazy-loaded promises with import().
  - Initialization via provideAppInitializer (Angular 18+) or APP_INITIALIZER (<18).

✅ Benefits

  - Smarter lazy-loading: theme selection can affect which assets are included.
  - Smaller initial bundle size: no need to eagerly include all themes or configuration up front.
  - Aligns with Angular 17+ best practices.
  - Encourages module-level tree-shaking and performance optimization.
  
  
Here is the link to the full example on StackBlitz: 🔗 https://stackblitz.com/edit/stackblitz-starters-esfzg2pw?file=src%2Fmain.ts

Without lazy loading, the compiled main.js file exceeds 1 MB and includes all Ag Grid modules

Screenshot 2025-08-02 at 01 29 32

With lazy loading enabled, we can clearly observe that the Ag Grid modules are split into a separate chunk, which helps reduce the initial bundle size and improves load performance.

Screenshot 2025-08-02 at 01 28 16

### DIFF
--- a/packages/ag-grid-angular/projects/ag-grid-angular/README.md
+++ b/packages/ag-grid-angular/projects/ag-grid-angular/README.md
@@ -211,8 +211,8 @@ ModuleRegistry.registerModules([AllCommunityModule]);
 To take advantage of Angular’s lazy provider loading and reduce bundle size, use the `provideAgGrid` helper as shown below.
 
 ```ts
-import { provideAgGrid, AgGridService } from 'ag-grid-angular';
-import { ApplicationConfig, provideAppInitializer, inject } from '@angular/core';
+import { provideAgGrid } from 'ag-grid-angular';
+import { ApplicationConfig } from '@angular/core';
 
 export const appConfig: ApplicationConfig = {
   providers: [
@@ -226,18 +226,6 @@ export const appConfig: ApplicationConfig = {
       options: () => Promise.resolve({ domLayout: 'autoHeight' }),
       licenseKey: '-----' // Add your license key for enterprise
     }),
-    // angular 18+ uses `provideAppInitializer` to load the grid service
-    provideAppInitializer(async () => {
-      const agGridService = inject(AgGridService);
-      await agGridService.load();
-    }),
-    // Angular < 18 (Traditional APP_INITIALIZER)
-    {
-      provide: APP_INITIALIZER,
-      useFactory: (agGridService: AgGridService) => async () => await agGridService.load(),
-      deps: [AgGridService],
-      multi: true
-    }
   ]
 };
 ```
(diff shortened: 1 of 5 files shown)


medbenmakhlouf:
Hello @StephenCooper  @AndrewGlazier , I hope you don’t mind me tagging you here.
I’ve opened this PR and thought you might be the right people to review it.
Whenever you have time, your feedback would be greatly appreciated. Thanks in advance! 🙏

StephenCooper:
Hi @medbenmakhlouf,

Can I ask how this is different from using the `defer` approach within the application. That works without internal changes to AG Grid.

medbenmakhlouf:
Hi @StephenCooper — thanks for taking the time to look at this, and great point about @defer.

`@defer` is indeed perfect when I want to lazy-instantiate a **specific grid instance** at a component boundary.
However, the use case I’m addressing in this PR is different: **set up project-level defaults** (theme, defaultColDef, icons, license, registered modules) once in main.ts — similar to how Angular apps now centralize framework features with provideRouter, etc.

Here’s why this approach adds value:

1. **Single source of truth (app-level)** – Teams typically define AG Grid defaults and theme in one place (**main.ts**). With `@defer`, I’d still need per-component/module registration or an extra service that runs at first use, which scatters the config and increases the chance of missing it in lazy routes.

2. **Consistency with Angular 17+ patterns** – The providePluginName() style is now the norm across the ecosystem. It improves DX, keeps AG Grid aligned with modern Angular conventions, and makes it feel like a first-class Angular citizen.

3. **SSR / Hydration & early usage** – `@defer` only triggers on viewport or interaction. If grid services, types, or defaults need to be available before the first render (e.g., SSR, prerendering, non-visual usage), `@defer` can be too late.

4. **Bundle planning & explicitness** – Registering modules in main.ts makes their impact on the initial chunk explicit (as my screenshot shows). That allows for conscious trade-offs: I can still lazy-load actual grid usage per feature without duplicating registration code.

This approach is complementary to `@defer` they can work together as a great combination:
use main.ts for app-wide defaults and `@defer` for per-feature or on-demand loading when needed.

StephenCooper:
Thanks for the additional detail. I have linked to this PR for us to evaluate fully as part of AG-13371 which is an existing issue to evaluate if there are framework specific improvements we can make to module registration. 

Are you using this pattern yourself in your applications?

medbenmakhlouf:
Yes, I’m actively using this pattern in my applications.

StephenCooper:
Closing PR as this is linked to the Jira and will be taking into account as part of that feature work.

medbenmakhlouf:
@StephenCooper Thanks for the update!
Out of curiosity, could you share what solution will be considered as part of the Jira work? I’m interested in the final approach and would love to follow along.
If possible, please link the Jira ticket or reference this PR once the work starts, I’d really like to see how the final solution is implemented.

Thanks!

StephenCooper:
Jira will be AG-16458.

PR #11661, opened 2025-08-04
AG-13434 - [excel-export] - row outline should work with `groupRemoveSingleChildren`

gportela85:
fix [AG-13434](https://ag-grid.atlassian.net/browse/AG-13434)

### DIFF
--- a/packages/ag-grid-enterprise/src/excelExport/excelSerializingSession.ts
+++ b/packages/ag-grid-enterprise/src/excelExport/excelSerializingSession.ts
@@ -249,10 +249,13 @@ export class ExcelSerializingSession extends BaseGridSerializingSession<ExcelRow
         const padding = node.footer ? 1 : 0;
         const currentRow = _last(this.rows);
 
-        // Excel only supports up to 7 levels of outline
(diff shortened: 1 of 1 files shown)


packages/ag-grid-enterprise/src/excelExport/excelSerializingSession.ts
```diff
@@ -249,10 +249,13 @@ export class ExcelSerializingSession extends BaseGridSerializingSession<ExcelRow
         const padding = node.footer ? 1 : 0;
         const currentRow = _last(this.rows);
 
-        // Excel only supports up to 7 levels of outline
-        const outlineLevel = Math.min(node.level + padding, 7);
-
-        currentRow.outlineLevel = outlineLevel;
+        // if level is different than uiLevel, the parent is hidden
+        // due to `groupHideParentOfSingleChild`
+        if (node.uiLevel == null || node.level === node.uiLevel) {
```
SalvatorePreviti, line 254: are we sure this is the right way to do it? I am thinking few circumstances where uiLevel is not what we can expect, in pivoting and master-detail and total rows

SalvatorePreviti (approved):

PR #11665, opened 2025-08-04
[AG-15555] Feature(ssrm): add support for expand all collapse all

sdwvit:
- Add a prop to gridOptions to enable expandAll/collapseAll on all rows, not just loaded ones
 - Update error messages to include feature context for `getRowId` validation
 - Document changed behavior in ag-grid-docs for server-side grouping
 - bonus: Allows using optional props in warning/error messages

Fix [AG-15555](https://ag-grid.atlassian.net/browse/AG-15555)
Fix [AG-15693](https://ag-grid.atlassian.net/browse/AG-15693)

Screenshot 2025-08-12 at 14 28 34

Screenshot 2025-08-12 at 11 13 28


packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,21 +11,27 @@ import type {
...
         this.serverSideRowModel = beans.rowModel as ServerSideRowModel;
     }
 
-    private queuedRowIds: Set<string> = new Set();
-
     public postConstruct(): void {
         this.addManagedEventListeners({
             columnRowGroupChanged: () => {
-                this.queuedRowIds.clear();
+                this.expandedState.toggledNodes.clear();
```
AndrewGlazier: Changing this would be breaking but food for thought: (don't make any change, but maybe worth looping in Zoheil and asking what he thinks/if we should do something about this)
What if I went from grouping Country, to grouping Country and Sport, is it right to clear all row grouping state or should we try to maintain because we've simply added a new deeper dimension which doesn't impact which groups are available at the top level?

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -59,18 +65,15 @@ export class ServerSideExpansionService extends BaseExpansionService implements
     }
 
     public expandRows(rowIdsToExpand: string[], rowIdsToCollapse?: string[]): void {
-        const { serverSideRowModel, queuedRowIds } = this;
         const processNodes = (rowIds: string[], expanded: boolean) => {
```
AndrewGlazier: This is coming from initialState/stateService, we probably need to also store selectAll in our initial state now (and probably find a backwards-compatible way to handle old states)

Note when you fix AG-15693 we also need to somehow store _that_ state (i.e what nodes are expanded by default vs which ones have been explicitly closed, even if expandAll is false)

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -35,8 +41,8 @@ export class ServerSideExpansionService extends BaseExpansionService implements
             return;
         }
 
-        if (this.queuedRowIds.has(rowNode.id!)) {
-            this.queuedRowIds.delete(rowNode.id!);
+        const shouldExpand = this.expandedState.expandAll !== this.expandedState.toggledNodes.has(rowNode.id!);
+        if (shouldExpand) {
```
AndrewGlazier: I've opened AG-15693 to tackle an issue related to this - don't fix now as we can keep this PR small, but I've assigned you that ticket

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,21 +11,27 @@ import type {
...
+    /** Base expansion state for all nodes, whether they have been loaded or not */
+    expandAll: boolean;
+    /** RowNode IDs of those nodes whose expansion state differs from the base state */
+    toggledNodes: Set<string>;
+}
+
 export class ServerSideExpansionService extends BaseExpansionService implements NamedBean, IExpansionService {
     beanName = 'expansionSvc' as const;
 
+    private expandedState: ExpansionState = { expandAll: false, toggledNodes: new Set() };
```
AndrewGlazier: I can't see a request for this, but if possible it'd be a good idea to add set/get state and serialize this - as it'll be a good way for users to update the expanded rows - if you have time can we scope creep this?

I'm not sure why it's not a ticket already

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -82,8 +85,17 @@ export class ServerSideExpansionService extends BaseExpansionService implements
...
-        this.serverSideRowModel.expandAll(value);
-
+        this.expandedState.expandAll = value;
+        this.expandedState.toggledNodes.clear();
+        this.serverSideRowModel.expandAllTransactional((node) => {
+            if (node.stub) {
+                this.expandedState.toggledNodes.add(node.id!);
+            } else {
+                if (node.hasChildren()) {
+                    node.setExpanded(value);
```
AndrewGlazier: This pattern isn't great - I know it's not yours and this is scope creep, but if you want to implement an important fix:
Look at how setExpanded works, (calls setRowNodeExpanded on the parent class) it dispatches a `rowGroupOpened` event synchronously every time
This event is listened to by lots of things, stateService (which hacks in a debounce), editService (which doesn't debounce it at all) and the `expandListener` (SSRM only)
Every time we do this, we subsequently fire a `storeUpdated` event which then recomputes all of the row indexes in the SSRM synchronously (which is very inefficient)

If you fancy taking up the task (we can do in a different PR on the same ticket if you want) it'd be good to introduce a new pattern for bulk expanding rows and detach the internal grid logic from the synchronous, per-row `rowGroupOpened` event


For expand-all this currently works fine due to the flag we have, but that's not a great pattern and we don't offer users any way to bulk expand rows (that isn't expandAll) - so when doing the setter/getter we probably want to improve this

packages/ag-grid-enterprise/src/serverSideRowModel/serverSideRowModel.ts
```diff
@@ -464,20 +465,12 @@ export class ServerSideRowModel extends BeanStub implements NamedBean, IServerSi
         this.pauseStoreUpdateListening = paused;
     }
 
-    public expandAll(value: boolean): void {
+    public expandAllTransactional(cb: Parameters<IRowModel['forEachNode']>[0]): void {
```
AndrewGlazier: Don't love the change in this type, but have a look at the other comments and see if you can change this back as a result of any of them, if you don't feel you can then it's fine

AndrewGlazier (changes requested):

AndrewGlazier (commented):
A few notes

packages/ag-grid-enterprise/src/serverSideRowModel/serverSideRowModel.ts
sdwvit: yea I also feel that, will replace it with something nicer and less coupled

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
sdwvit: in your opinion what should happen here? I am not sure I understand this event fully to make a call

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
sdwvit: I think this would be useful to fix that issue here too, as the approach to the whole methodology is different

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
AndrewGlazier: In my opinion this is wrong:
1/ Open https://plnkr.co/edit/cdg73coW29MrN3Bh?open=main.js
2/ Right click group by Sport
Expected: Only the contents of the country groups need refreshed
Actual: Both the expanded state of the country groups is lost, and the entire grid is refreshed

There might be some niche edge cases where aggregations would be impacted and so need the entire refresh, so it'd probably need to be behind a flag, and it'd be breaking to change this by default anyway.

We don't need to make a judgement call here, just raise the issue/scenario with Zoheil and let him do so

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
 import { BaseExpansionService } from '../../rowHierarchy/baseExpansionService';
 import type { ServerSideRowModel } from '../serverSideRowModel';
 
+interface ExpansionState {
+    /** Base expansion state for all nodes, whether they have been loaded or not */
+    expandAll: boolean;
+    /** Whether the user has interacted with all nodes */
+    interactedWithAll: boolean;
```
AndrewGlazier: Is this necessary, can we just merge this into `expandAll`

I.e, expandAll: boolean | null, where if expandAll == null it's not been interacted with, however if it's true/false we know it's be explicitly toggled

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
...
+        }
+    }
 
+    public expandAll(expanded: boolean): void {
+        this.expandedState = getDefaultExpansionState({ expandAll: expanded, interactedWithAll: true });
+        this.serverSideRowModel.forEachNodeTransactional((node) => node.setExpanded(expanded));
         this.beans.eventSvc.dispatchEvent({
             type: 'expandOrCollapseAll',
-            source: value ? 'expandAll' : 'collapseAll',
+            source: expanded ? 'expandAll' : ' collapseAll',
```
AndrewGlazier: ' collapseAll' looks like a typo

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
 import { BaseExpansionService } from '../../rowHierarchy/baseExpansionService';
 import type { ServerSideRowModel } from '../serverSideRowModel';
 
+interface ExpansionState {
```
AndrewGlazier: This could probably be a stricter type, which may make it easier to follow the behaviour for anyone unfamiliar with this, e.g:
```
interface AllExpandedState {
   expandAll:  boolean;
   toggledNodes: string[]; // I'm aware you're currently using a map for quick lookup, could be a set internally and arr exported though
   // touchedNodes not required here as it's implicit because expandAll is non-null
}

interface GranularExpandedState {
  expandAll: null; // expand all boolean/null now becomes something we can target on ExpansionState for drilling into the type
  toggledNodes: string[];
  touchedNodes: string[];
}

type ExpansionState = AllExpandedState | GranularExpandedState;
```

(This is an example of the exported state, I imagine internally we'd need to use Sets where arrays are used)

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
...
+    /** Base expansion state for all nodes, whether they have been loaded or not */
+    expandAll: boolean;
+    /** Whether the user has interacted with all nodes */
+    interactedWithAll: boolean;
+    /** RowNode IDs of those nodes whose expansion state differs from the base state */
+    toggledNodes: Record<string, boolean | undefined>; // in reality this is true | undefined
```
AndrewGlazier: For internal lookup, any reason not to use a set? I'm assuming real type is Record<string, true> rather than Record<string, true | undefined> so this could be a Set<string> (same with interactedWith below)

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
...
+            set(target, p, newValue) {
+                const rowIdC = p as keyof typeof target;
+                self.interactedWith[rowIdC] = true;
+                return self.expandAll === newValue ? delete target[rowIdC] : (target[rowIdC] = true);
+            },
+        }),
+        interactedWith: new Proxy({} as ExpansionState['interactedWith'], {
+            get: (target, p) => {
+                const rowIdC = p as keyof typeof target;
+                return self.interactedWithAll ? delete target[rowIdC] || true : target[rowIdC];
```
AndrewGlazier: This should never happen right? Because you create a fresh instance of the expansion state when expandAll is toggled, and your setter won't allow setting new values if interactedWithAll is set

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
...
+     * If `expandAll` is false, it returns true if the node has been toggled.
+     */
+    isExpanded(rowId: string): boolean;
+}
+
+/**
+ * Using a proxy here allows us to automagically update the `interactedWith` state,
+ * which is used to determine the default expansion state of the nodes. And be memory conscious as well.
+ */
+const getDefaultExpansionState = (defaultOverrides: Partial<ExpansionState> = {}) => {
```
AndrewGlazier: I'm not a fan of the proxy here because I think it adds some complexity that isn't really needed, it will make it harder for people to follow/debug the state of this object as it changes, whereas in this instance I think a simple setter func could handle this logic and be a bit cleaner, thoughts?

AndrewGlazier (changes requested):

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
```diff
@@ -134,7 +134,9 @@ api.collapseAll();
 
 Calling `expandAll()` and `collapseAll()` will impact **all loaded group nodes**, including those not visible due to their containing group been closed. This means there could potentially be a huge number of groups expanded, so this method should be used very wisely to not create massive amount of server requests and loading a large amount of data.
 
-Calling `expandAll()` and `collapseAll()` will have no impact on rows yet to be loaded.
+{% note %}
+Calling `expandAll()` and `collapseAll()` will apply to all rows, including those not yet loaded.
```
AndrewGlazier: We won't actually be able to do this by default I think, we might need to introduce a new grid option flag due to this being a breaking change, even if the old behaviour is a bit rubbish

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,127 @@ import type {
...
         this.addManagedEventListeners({
             columnRowGroupChanged: () => {
-                this.queuedRowIds.clear();
+                this.expandedState = getDefaultExpansionState();
             },
         });
     }
 
-    public checkOpenByDefault(rowNode: RowNode): void {
+    public checkOpenByDefault(rowNode: RowNode): boolean {
```
AndrewGlazier: Might be worth a rename on this one; isRowExpanded may be more applicable now the signature has changed

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -2121,6 +2121,11 @@ export interface GridOptions<TData = any> {
      * @agModule `RowGroupingModule` / `TreeDataModule`
      */
     isGroupOpenByDefault?: (params: IsGroupOpenByDefaultParams<TData>) => boolean;
+    /**
+     * Controls how groups behave.
+     * @agModule `RowGroupingModule` / `TreeDataModule`
+     */
+    groupFlags?: GroupFlags;
```
AndrewGlazier: We tend to use flat grid options (mostly) as it makes for easier config for users. E.g for this, if the user did:
<AgGridReact groupFlags={{ useSsrmNewBehaviour: true }} />

Then every time their component re-rendered the grid would need to perform reactivity updates, so they'd have to extract and memoise this object to reduce unwanted side effects and processing.

I'd suggest a name more like:
ssrmExpandAllState (though maybe check with Zoheil to confirm that we really do need a flag if you haven't already) (Open to better names still as this isn't an easy thing to name)

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
```diff
@@ -134,7 +134,9 @@ api.collapseAll();
 
 Calling `expandAll()` and `collapseAll()` will impact **all loaded group nodes**, including those not visible due to their containing group been closed. This means there could potentially be a huge number of groups expanded, so this method should be used very wisely to not create massive amount of server requests and loading a large amount of data.
 
-Calling `expandAll()` and `collapseAll()` will have no impact on rows yet to be loaded.
+{% note %}
```
AndrewGlazier: We'll want to introduce a new example which demonstrates the new feature, and the difference of enabling and suppressing this flag

e.g a checkbox for toggling the value. (similar to this https://www.ag-grid.com/javascript-data-grid/grouping-data/#hiding-parents-of-individual-rows)

An expand and collapse all button above the grid

and appropriate properties (cacheBlockSize = 0, purgeClosedRowNodes = true) to make the difference apparent when doing expand/collapse all

packages/ag-grid-enterprise/src/serverSideRowModel/serverSideRowModel.ts
```diff
@@ -157,7 +157,10 @@ export class ServerSideRowModel extends BeanStub implements NamedBean, IServerSi
 
     private verifyProps(): void {
         if (_isRowSelection(this.gos) && !this.gos.exists('getRowId')) {
-            _warn(188);
+            _warn(188, { feature: 'selection' });
+        }
+        if (this.gos.get('groupFlags')?.useSsrmNewBehaviour && !this.gos.exists('getRowId')) {
```
AndrewGlazier: This check would probably be better placed in `gridOptionsValidations` as people can drop the validations module for their production build (minor bundle size improvement)

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -11,82 +11,99 @@ import type {
 import { BaseExpansionService } from '../../rowHierarchy/baseExpansionService';
 import type { ServerSideRowModel } from '../serverSideRowModel';
 
+interface IServerSideRowExpansionState {
+    expandAll: 'notInteracted' | 'expandAll' | 'collapseAll';
```
AndrewGlazier: minor nit: don't feel the need to change this if you disagree, but as we're going to be expecting users to serialize this doing a union of undefined/boolean would be less storage impact on our users

Minor, but food for thought

AndrewGlazier (changes requested):

packages/ag-grid-community/src/entities/gridOptions.ts
sdwvit: ssrmExpandAllState doesn't indicate boolean nature of this flag. it can only be true/false. undefined is coerced to false. I'll think a bit more about the name here.

sdwvit: how about `ssrmExpandAllAffectsAllRows`?

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
sdwvit: added, thnx!

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -2127,6 +2127,13 @@ export interface GridOptions<TData = any> {
      * @agModule `RowGroupingModule` / `TreeDataModule`
      */
     isGroupOpenByDefault?: (params: IsGroupOpenByDefaultParams<TData>) => boolean;
+    /**
+     * Controls how expand/collapse operations affect all rows and group interactions.
+     * If `true`, expandAll / collapseAll applies to all rows (not just loaded ones),
+     * and interacting with the group overrides the default expansion state set by `isServerSideGroupOpenByDefault`.
+     * @agModule RowGroupingModule / TreeDataModule
+     */
+    ssrmExpandAllAffectsAllRows?: boolean | undefined;
```
AndrewGlazier, line 2136: This needs to be added to documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json so that it will show up on the api documentation page, this should also link back to the relevant docs section

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/_examples/expand-all-affects-all-rows/index.html
```diff
@@ -0,0 +1,13 @@
+<div class="example-wrapper" style="height: 100%">
+    <div class="example-header">
+        <label>
+            <span>Change gridOptions['ssrmExpandAllAffectsAllRows'] value:</span>
+            <select id="input-display-type" onchange="onOptionChange()">
```
AndrewGlazier: As it's just true/false can you make this a checkbox

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/_examples/expand-all-affects-all-rows/index.html
```diff
@@ -0,0 +1,13 @@
+<div class="example-wrapper" style="height: 100%">
+    <div class="example-header">
+        <label>
+            <span>Change gridOptions['ssrmExpandAllAffectsAllRows'] value:</span>
```
AndrewGlazier: For consistency with other docs can you just make this `ssrmExpandAllAffectsAllRows` without the pseudo typescript, not everyone will be using TS and it's not a convention we've used elsewhere

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
```diff
@@ -134,7 +134,11 @@ api.collapseAll();
...
+To change the default behaviour of `expandAll()` and `collapseAll()` to expand or collapse **all group nodes**, not only loaded group nodes, set `ssrmExpandAllAffectsAllRows` flag in `GridOptions` to `true`.
+{% /note %}
+
+{% gridExampleRunner title="expand-all-affects-all-rows" name="expand-all-affects-all-rows"  exampleHeight=540 /%}
```
AndrewGlazier: I don't think that the default docs for this Expand All / Collapse All section are very good (it's the same for all the SSRM docs, but focusing on this section), do you mind trying to refactor them? For context the Row Grouping (non SSRM docs) are an example of something I've entirely rewritten more recently, the SSRM docs are a bit outdated

We tend to go for something like this: https://www.ag-grid.com/react-data-grid/grouping-single-group-column/#checkbox-selection

Note how the style is:
Simple intro to the feature
Example
Small config example

We've made this change to remove some of the more rambling-style flow of documentation, and reduce the docs to key info.

I'd suggest something like:
```
Expand and Collapse All
Group rows can be expanded or collapsed using the api methods `expandAll` and `collapseAll`.  By default, this will only impact loaded group rows but enabling the `ssrmExpandAllAffectsAllRows` grid option changes this behaviour to instead expand groups as they are loaded.

{ EXAMPLE }

The example above is configured using the following code snippet:
const gridOptions = {
   ssrmExpandAllAffectsAllRows: true,
}
```

AndrewGlazier:
There's one additional thing we should take into account - when using column menus with the _client side_ row model, we show the Expand and Collapse all buttons in the menu.

For the SSRM we don't - I think when using the new property we should show expand/collapse all buttons the same as the CSRM.

AndrewGlazier (changes requested):

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
sdwvit: good point let me do it

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
```diff
@@ -206,7 +211,7 @@ Then the columns are set up so that country uses a `valueGetter` that uses the f
 
 ## Filters
 
-By default the grid will fully purge the grid when impacted by the change in filters. The grid can be configured to only refresh when the group has been directly impacted by enabling `serverSideOnlyRefreshFilteredGroups`. Be aware, this can mean your grid may have empty group rows. This is because the grid will not refresh the groups above the groups it deems impacted by the filter.
+By default, changing filters fully purges the grid. Thought, it can be configured to only refresh when the group has been directly impacted by enabling `serverSideOnlyRefreshFilteredGroups`. Be aware, this can mean your grid may have empty group rows. This is because the grid does not refresh the groups above the groups it deems impacted by the filter.
```
AndrewGlazier: Thought -> Though

documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
```diff
@@ -1427,6 +1427,13 @@
                 "name": "Keeping Columns Visible",
                 "url": "./grouping-group-panel/#prevent-user-grouping-from-hiding-columns"
             }
+        },
+        "ssrmExpandAllAffectsAllRows": {
+            "description": "(Server-side Row Model only) Controls whether expandAll / collapseAll operations apply to all rows (not just loaded ones), and whether group interactions override the default expansion state set by `isServerSideGroupOpenByDefault`. When true, users must provide a `getRowId` function to ensure proper row identification. Default: `false`.",
+            "more": {
+                "name": "Row IDs",
+                "url": "./row-ids/"
```
AndrewGlazier: Wrong url? This should link to the expand/collapse docs section

AndrewGlazier (changes requested):

documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
sdwvit: row ids are required for this flag to work, but I can link both perhaps

AndrewGlazier (approved):

PR #11674, opened 2025-08-05
AG-15429 - ag stack

peterjrreynolds:


packages/ag-grid-community/src/agStack/interfaces/baseEvents.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface AgTooltipShowEvent extends AgTooltipEvent<'tooltipShow'> {
+    tooltipGui: HTMLElement;
+}
+
+interface AgTooltipHideEvent extends AgTooltipEvent<'tooltipHide'> {}
+
+/** Events required by AG Stack */
+export interface BaseEvents {
```
StephenCooper, line 33: It seems like a slightly random collection of events. I imagine this is going to grow over time as more features are shared?

packages/ag-grid-community/src/agStack/utils/beanUtils.ts
```diff
@@ -0,0 +1,40 @@
+import type { AgCoreBeanCollection } from '../interfaces/agCoreBeanCollection';
```
StephenCooper: The name of this file seems different to the util methods contained.

packages/ag-grid-community/src/agStack/utils/stringUtils.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
StephenCooper: Would there be any downside to just moving all the utils functions here? This would avoid having two different locations for utils to be defined. 

If they are not used, as they are exported individually the ones not used should just be shaken away.

packages/ag-grid-community/src/agStack/interfaces/baseEvents.ts
peterjrreynolds: Yes, anything the stack requires will get added here.

packages/ag-grid-community/src/agStack/utils/beanUtils.ts
peterjrreynolds: These are utils that require beans (essentially stuff that was in `gridOptionsUtils`)

peterjrreynolds: Have renamed to `document`

StephenCooper (approved):

packages/ag-grid-community/src/agStack/utils/stringUtils.ts
peterjrreynolds: Done

gportela85 (approved):

PR #11689, opened 2025-08-06
AG-3390 - Update bundle size input to be not skipped by default

taktran:
Was only running during `workflow_dispatch`, but not `push` and `pull_request`


seanlandsman (approved):
lgtm

PR #11691, opened 2025-08-06
AG-4663 drag not allowed icon

SalvatorePreviti:
This addresses both AG-4663 and AG-15252

- invokes `isRowValidDropPosition` callback both for managed and unmanaged row data
- Ensures that the DraggingEvent is passed to the getIcon and label getter for the row drag ghost component, new parameter added
- modifies the DragAndDropService to update icon and label whenever a possible change in target or filtered rows to drop changes
- fixes and improves behaviour in the rowDragFeature
- add the rowDragFeatureNudger that is responsible to handle the new group and expand group timer and scroll timer, by using the existing drag and drop service nudge() - this is required to update the icon and label correctly.
- add a new `rowsDrop` field in the dragging event and row dragging events, this is used both internally to pass data around and can be used by the user when implementing unmanaged row dragging, or, when handling managed row dragging events. This field contains information about the filtered rows to be dropped, if the drop is allowed, and behaves exactly like managed dragging
- add the logic to handle properly row dragging when master detail is enabled, it wasn't working properly
- move the public types defined in rowDragFeature to its own rowDragTypes module.
- Add the new css class `ag-dnd-ghost-not-allowed` that is added when the ghost is not allowed
- Updates the documentation and examples

Other fixes and improvements:
- improves the _areEquals utility function to check for instance equality for performance. Discussed with Stephen, he approved the change. This function is used in DragFeature to check if the dragged rows changes between events
- Improvements of DragAndDropService in terms of clarity and performance, together with the changes to properly update icon and label
- Fixes potential race conditions with the AgPromise that is used to create the dragging ghost component
- Small refactors to improve performance and to reduce bundle size in DragAndDropService


gportela85:
@SalvatorePreviti The styles were not added to legacy themes and the new theme variable was not added to the Theming API.

gportela85 (approved):

peterjrreynolds (approved):

PR #11702, opened 2025-08-07
Backport signing of releases to LTS

seanlandsman:


scripts/deployments/release/performSigningActions.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+PACKAGE_ROOT="./dist/artifacts"
+
+if [ "$#" -ne 1 ]
+then
+    echo "You must the GPG public key as the first argument"
```
taktran, line 7: The messages haven't been updated?

taktran (changes requested):

scripts/deployments/release/performSigningActions.sh
seanlandsman: I think I did this on latest - I'm probably going to leave it as is for LTS

taktran (approved):
LGTM

PR #11708, opened 2025-08-07
AG-15706 csrm row data null items

SalvatorePreviti:
Ensures that null or undefined row data item raise a warning and do not break the grid


StephM-W (approved):
LGTM

PR #11738, opened 2025-08-11
AG-15670 Extend React 19 StrictMode Support

StephenCooper:
Fixes [AG-15670](https://ag-grid.atlassian.net/browse/AG-15670)


packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
```diff
@@ -242,11 +242,10 @@ export const AgGridReactUi = <TData,>(props: InternalAgGridReactProps<TData>) =>
         !(React as any).useSyncExternalStore || _getGridOption(props, 'renderingMode') === 'legacy'
             ? 'legacy'
             : 'default';
-
     return (
         <div style={style} className={props.className} ref={setRef}>
             <RenderModeContext.Provider value={renderMode}>
-                {context && !context.isDestroyed() ? <GridComp context={context} /> : null}
+                {context && !context.isDestroyed() ? <GridComp key={context.instanceId} context={context} /> : null}
```
StephenCooper, line 248: This is a key part of the change. It forces React to reset all the state for the GridComp which is what we require to prevent stale references to destroyed contexts and ctrls.

packages/ag-grid-react/src/reactUi/cells/cellComp.tsx
```diff
@@ -236,14 +235,15 @@ const CellComp = ({
...
         const spanReady = !cellCtrl.isCellSpanning() || eWrapper.current;
         const eRef = eGui.current;
-        compBean.current = eRef ? context.createBean(new _EmptyBean()) : context.destroyBean(compBean.current);
-        if (!eRef || !spanReady || !cellCtrl) {
+        if (!eRef || !spanReady || !cellCtrl || context.isDestroyed()) {
             // We do NOT add a check for if the cellCtrl is destroyed as when there are lots of updates React
             // can get behind our internal state and call this function after the cellCtrl has been destroyed.
             // If we were to shortcut here then cell values will flash in the first column of the grid as they will
             // not have the correct cell position / styles applied as that is set via setComp.
+            compBean.current = context.destroyBean(compBean.current);
```
StephenCooper, line 239: Need to validate that this change here does not cause what this comment is talking about.

StephenCooper: The collapsing into the first column no longer happens with this change. It did not even happen without this change on latest but adding the isAlive check. So another part of the code must have changed. Will be adding this example to react-test page.

packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
gportela85: should we introduce a new config so customers can add a prefix to this?

StephenCooper: This does not end up in the DOM, its purely a React thing.

peterjrreynolds (approved):

packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
gportela85: Yes, but is it possible for <GridComp /> to have a sibling added by customers that will have the same `key`?

PR #11747, opened 2025-08-12
AG-15670 Interactive test improvements

StephenCooper:
- Enables forcing Dev Mode via `prod=false` 
- Adds React DEV test to framework list
- Adds console warn / error test failing

### DIFF
--- a/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
+++ b/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
@@ -219,9 +219,21 @@ const frameworkTest =
  * @param testBody The test body function that will be executed for each framework.
  */
 const eachFramework = (testName: string, testBody: (fixtures: TestFixtures) => Promise<void>) => {
-    extended.describe(testName, () =>
-        ALL_FRAMEWORKS.forEach((framework) => frameworkTest(framework)(undefined, testBody))
-    );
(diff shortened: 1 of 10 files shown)


documentation/ag-grid-docs/src/utils/grid/test-utils.ts
```diff
@@ -111,20 +181,32 @@ const frameworkTest =
...
 
-            await loadPage(page, agExampleUrl, agFramework);
-            await testBody({ page, agExampleUrl, agIdFor, agFramework } as TestFixtures);
+            await loadPage(page, agExampleUrl, agFramework, loadPageOptions);
+            await testBody({ page, agExampleUrl, agIdFor, agFramework, loadPageOptions } as TestFixtures);
         };
 
         if (testName) {
             extended.describe(testName, () => {
+                let errors: string[];
```
StephM-W, line 197: this codepath only gets triggered when using `test.<framework>` directly. If we want these before/after when running for eachFramework, we also need to update that function's describe block

StephM-W (approved):

eliasmalik:
Should we add a trigger to the doc-test CI job so it runs on PRs that modify the playwright config and/or example tests and/or documentation examples? Or would that thrash our minutes too much?

StephM-W (approved):
LGTM

PR #11765, opened 2025-08-13
feat: enable configuration of `appRoot`

DerYeger:
Some environments, e.g., MFEs, may not support attaching elements to `document.body`. For such environments, it may be necessary to configure an `appRoot` element that is used instead.


StephenCooper:
Does https://ag-grid.com/angular-data-grid/grid-options/#reference-miscellaneous-getDocument not solve this already?

DerYeger:
> Does https://ag-grid.com/angular-data-grid/grid-options/#reference-miscellaneous-getDocument not solve this already?

Unfortunately, this option is not sufficient. A typical module federation MFE hat a single `document` and mounts the MFEs to regular `HTMLElement` elements. For security and conflict-prevention, we'd like to also mount all popups, dialogs, modals, etc. within those root elements of the MFEs, which is currently not possible.

StephenCooper:
Could you please raise this as an issue with a reproduction in place of this PR?

If you have an enterprise license then you could report this via Zendesk. Feel free to link to this PR if you wish.

PR #11777, opened 2025-08-14
feat: implement textContentAsValue

sharunkumar:
Added a new `textContentAsValue` flag to `ColDef` that, when set to `true`, causes the value service to return the cell’s rendered DOM `textContent` when set.

This is useful when working with custom cell renderers


StephenCooper:
If you are an enterprise customer please raise feature requests via your Zendesk account.

PR #11789, opened 2025-08-16
AG-14947 - apply safari focus handling

StephM-W:
Fix AG-14947

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/provided-cell-editors/_examples/cell-data-type-editors/example.test.ts
+++ b/documentation/ag-grid-docs/src/content/docs/provided-cell-editors/_examples/cell-data-type-editors/example.test.ts
@@ -1,0 +1,39 @@
+import { expect, test } from '@utils/grid/test-utils';
+
+import type { GridOptions } from 'ag-grid-community';
+
+test.agExample(import.meta, () => {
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/src/content/docs/provided-cell-editors/_examples/cell-data-type-editors/example.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        gridOptions = {
+            columnDefs: [
+                { field: 'string', cellEditor: 'agTextCellEditor', editable: true },
+                { field: 'number', cellEditor: 'agNumberCellEditor', editable: true },
+            ],
+            rowData: [{ string: 'red', number: 123 }],
+        };
+    });
+
+    // AG-14947 - Safari doesn't focus string input fields when typing-to-start editing
```
StephenCooper, line 18: These tests are not currently running against all the browsers. Maybe we should set that up so that this would actually be covered / validated by this test.

StephM-W: Agreed - let's discuss

StephenCooper (approved):

PR #11795, opened 2025-08-18
AG-15525 - add sequencing test support and tests

StephM-W:
Fix AG-15525

### DIFF
--- a/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
+++ b/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
@@ -3,9 +3,9 @@ import type { Locator, Page, TestType } from '@playwright/test';
 import { test as base, expect } from '@playwright/test';
 import { CacheRoute } from 'playwright-network-cache';
 
-import { wrapAgTestIdFor } from 'ag-grid-community';
+import { type AgModuleName, wrapAgTestIdFor } from 'ag-grid-community';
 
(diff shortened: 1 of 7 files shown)


documentation/ag-grid-docs/src/utils/grid/test-utils.ts
```diff
@@ -103,24 +104,31 @@ async function loadPage(
...
 
+    if (agModules && agModules.length > 0) {
+        queryOptions.modules = agModules.join(',');
+    }
+
     const queryParams = new URLSearchParams(queryOptions);
     const urlFramework = agFramework === reactFunctionalTsDev ? 'reactFunctionalTs' : agFramework;
 
-    await page.goto(`/examples/${agExampleUrl}/${urlFramework}?${queryParams.toString()}`);
+    await page.goto(`./examples/${agExampleUrl}/${urlFramework}?${queryParams.toString()}`);
```
StephenCooper, line 131: Why is this `.` required and does this work against staging?

StephM-W: Required so the urls are based off the baseURL in the playwright config. This allows it to support our main site, staging, archive and branch builds

StephM-W: Otherwise they're rooted at the host

PR #11806, opened 2025-08-19
AG-15758 - ensure editors are restored during scroll/tabs when column/row virtualisation enabled

StephM-W:
Fix AG-15758


packages/ag-grid-community/src/interfaces/iEditService.ts
```diff
@@ -101,10 +102,11 @@ export interface IEditService extends NamedBean {
...
     ): void;
-    dispatchRowEvent(
-        position: Required<EditRowPosition>,
-        type: 'rowEditingStarted' | 'rowEditingStopped' | 'rowValueChanged'
-    ): void;
+    // dispatchRowEvent(
+    //     position: Required<EditRowPosition>,
+    //     type: 'rowEditingStarted' | 'rowEditingStopped' | 'rowValueChanged',
+    //     silent?: boolean
+    // ): void;
```
StephenCooper: Does this need removing?

documentation/ag-grid-docs/src/content/docs/cell-editing-full-row-debug/index.mdoc
```diff
@@ -0,0 +1,5 @@
+---
+title: "Full Row Editing Debug"
```
StephenCooper: Can you swap -debug for -test so we just have one pattern for these pages across our Grid examples

StephM-W: Renamed

packages/ag-grid-community/src/interfaces/iEditService.ts
StephM-W: Removed

gportela85 (approved):

PR #11811, opened 2025-08-19
AG-15777-fix-react-row-drag-visibility

SalvatorePreviti:
Fix rowDragComp visibility not being updated in react as refreshVisibility was just called for the main non custom js component

- rename refresh to refreshVisibility, to make clearer what it does
- the refreshVisibility method is now invoked by js component, jsx component, and for custom row drag component inside the cellCtrl

Fix [AG-15777](https://ag-grid.atlassian.net/browse/AG-15777)


StephenCooper (changes requested):
Let's see if we can make this work from the cellCtrl only.

StephenCooper (approved):

PR #11814, opened 2025-08-19
fix(docs): resolves minor discrepancy in React inline cell renderer example

avinoamsn:
- renames callback arg `props` -> `params` (matches function body)


StephenCooper:
Thank you for reporting this has been fixed.

PR #11848, opened 2025-08-21
fix(docs): clarify filter types in server-side filtering example

sdwvit:
- Updated documentation to specify that  the **Sport** column employs a Multi Column Filter combining Set and Text filters.
 
Screenshot 2025-08-21 at 17 03 50

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
@@ -127,7 +127,8 @@ For more details on setting values, see [Supplying Filter Values](./filter-set-f
 
 The example below demonstrates Server-side Filtering using the Set Filter. Note the following:
 
-* The **Country** and **Sport** columns have Set Filters defined using `filter: 'agSetColumnFilter'`.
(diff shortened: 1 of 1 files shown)


documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
```diff
@@ -127,7 +127,8 @@ For more details on setting values, see [Supplying Filter Values](./filter-set-f
 
 The example below demonstrates Server-side Filtering using the Set Filter. Note the following:
 
-* The **Country** and **Sport** columns have Set Filters defined using `filter: 'agSetColumnFilter'`.
+* The **Country** column has a Set Filter defined using `filter: 'agSetColumnFilter'`.
+* The **Sport** column has a Multi Column Filter defined using `filter: 'agMultiColumnFilter'`, it combines a Set and a Text filters.
```
peterjrreynolds:
```suggestion
* The **Sport** column has a [Multi Filter](./filter-multi/) defined using `filter: 'agMultiColumnFilter'`, it combines a Set Filter and a [Text Filter](./filter-text/).
```

Good to link out to the features when they're mentioned for the first time on a page.

peterjrreynolds (approved):

PR #11849, opened 2025-08-21
AG-15735 - extract drag and drop to ag stack

peterjrreynolds:


packages/ag-grid-community/src/agStack/utils/document.ts
```diff
@@ -28,13 +28,51 @@ export function _getDocument(beans: AgCoreBeanCollection<any, any, any, any>): D
...
+    try {
+        rootNode = _getDocument(beans).fullscreenElement as HTMLElement;
+    } catch (e) {
+        // some environments like SalesForce will throw errors
+        // simply by trying to read the fullscreenElement property
+    } finally {
+        if (!rootNode) {
+            rootNode = _getRootNode(beans);
+        }
+        const body = rootNode.querySelector('body');
```
SalvatorePreviti, line 55: minor - document.body isn't better?

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -272,12 +273,13 @@ export class RowDragFeature extends BeanStub implements DropTarget {
 
         this.validateRowsDrop(rowsDrop, canSetParent, aboveOrBelow, dropping);
 
-        rowsDrop.changed ||= rowsDropChanged(lastDraggingEvent?.rowsDrop, rowsDrop);
+        rowsDrop.changed ||= rowsDropChanged(lastDraggingEvent?.dropTarget, rowsDrop);
```
SalvatorePreviti: rowsDrop.changed is not needed anymore maybe now as we have it in the dragging event instead? Is a new property so it was never exposed to the end user

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -165,19 +165,20 @@ export class RowDragFeature extends BeanStub implements DropTarget {
...
+        lastDraggingEvent: RowDraggingEvent | null,
+        draggingEvent: RowDraggingEvent,
         moving: boolean,
         dropping: boolean
     ): RowsDrop | null {
         const { beans, gos, clientSideRowModel } = this;
         const rowsDrop = this.newRowsDrop(draggingEvent, dropping);
-        draggingEvent.rowsDrop = rowsDrop;
+        draggingEvent.dropTarget = rowsDrop;
+        draggingEvent.changed = rowsDrop?.changed;
```
SalvatorePreviti: I would just make changed not nullable and initialize it to false when the DraggingEvent is created - and remove changed from RowsDrop as we need only one. It was never exposed to the user

packages/ag-grid-community/src/dragAndDrop/rowDragTypes.ts
```diff
@@ -75,13 +128,23 @@ export interface IsRowValidDropPositionParams<TData = any, TContext = any> exten
     allowed: boolean;
 }
 
-export interface RowsDrop<TData = any, TContext = any> extends IsRowValidDropPositionParams<TData, TContext> {
+export interface RowDraggingEvent<TData = any, TContext = any>
```
SalvatorePreviti, line 132: Wait a moment. GridDraggingEvent and RowsDrop are two different instances -  how can draggingEvent be inside RowDraggingEvent. Maybe I misunderstood something.
I imagine the hierarchy should be GridDraggingEvent { rowsDrop: RowsDrop|null }; RowsDrop { draggingEvent: GridDraggingEvent } - and we should replicate the same for the end user

SalvatorePreviti (approved):
Approved with comments

packages/ag-grid-community/src/agStack/utils/document.ts
peterjrreynolds: This was moved (no code changes made). Many tweaks have been made to this over the years to handle non-standard environments (e.g. Salesforce), so will leave this as is.

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
peterjrreynolds: Removed

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
peterjrreynolds: Done

packages/ag-grid-community/src/dragAndDrop/rowDragTypes.ts
peterjrreynolds: `draggingEvent` isn't inside `RowDraggingEvent`. The hierarchy is `RowDraggingEvent` `{ dropTarget: RowsDrop|null };` `RowsDrop` `{ draggingEvent: RowDraggingEvent }`, which is essentially the same (`RowDraggingEvent` is the version of `GridDraggingEvent` that is typed to use `RowsDrop`).

`DraggingEvent` is what's exposed to end users. This doesn't contain any form of row drop. However, the `RowDrag*Event`s contain `RowsDropParams` (the external-facing version of `RowsDrop`).

I've added a few more comments to help clarify this.

PR #11855, opened 2025-08-22
AG-15784 - Upgrade astro

taktran:
## Changes

* Upgrade Astro to 5.13.3
* Upgrade `vite-plugin-svgr`
* Remove `@nx/vite` and `@vitest/ui` and just use `vitest` command directly
* Use ts for vitest configuration


testing/behavioural/project.json
```diff
@@ -11,12 +11,11 @@
       }
     },
     "test": {
-      "executor": "@nx/vite:test",
+      "command": "npx vitest -c vitest.config.ts",
```
seanlandsman, line 14: curious as to why we can't use the nx plugin?

seanlandsman (approved):
lgtm

testing/behavioural/project.json
taktran: The nx plugin is tied to a vite/vitest version, and it makes it difficult to upgrade vite/vitest independently

PR #11868, opened 2025-08-26
Correct angular path for CodeQL

seanlandsman:


.github/codeql/codeql-config.yml
```diff
@@ -1,7 +1,7 @@
 paths:
  - 'packages/ag-grid-community/src'
  - 'packages/ag-grid-enterprise/src'
- - 'packages/ag-grid-angular/src'
+ - 'packages/packages/ag-grid-angular/projects/ag-grid-angular/src'
```
taktran: I think there is a path error here

```suggestion
 - 'packages/ag-grid-angular/projects/ag-grid-angular/src'
```

Also, does `packages/ag-grid-angular` need to be added too? There is a `package.json` and `project.json` there too

seanlandsman: Path corrected. No, we're only scanning source not deps

taktran (approved):
LGTM

PR #11872, opened 2025-08-26
AG-15803 tree data path build bug

SalvatorePreviti:
The grid breaks loading filler nodes if they are loaded in the wrong order in certain scenarios.
This ensures that filler node id calculation works independently from the order of nodes


packages/ag-grid-enterprise/src/treeData/treeGroupStrategy.ts
```diff
@@ -1,5 +1,5 @@
 import type { ChangedPath, GroupingApproach, IChangedRowNodes, StageExecuteParams } from 'ag-grid-community';
-import { RowNode, _ROW_ID_PREFIX_ROW_GROUP, _removeFromArray } from 'ag-grid-community';
+import { RowNode, _removeFromArray } from 'ag-grid-community';
```
gportela85: why two lines of imports from `ag-grid-community`?

gportela85 (approved):

PR #11879, opened 2025-08-28
AG-15579 - fix date filter model format

peterjrreynolds:
Fix [AG-15579](https://ag-grid.atlassian.net/browse/AG-15579)

### DIFF
--- a/packages/ag-grid-community/src/agStack/utils/date.ts
+++ b/packages/ag-grid-community/src/agStack/utils/date.ts
@@ -1,5 +1,8 @@
 const DATE_TIME_SEPARATOR = 'T';
 
+/** Spaces are allowed for legacy reasons to support date filter models */
+const DATE_TIME_SEPARATOR_REGEXP = new RegExp(`[${DATE_TIME_SEPARATOR} ]`);
+
 /**
  * Executing this against date produces the following:
  * ["2008-08-24T21:00:08"," 21:00:08"]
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/agStack/utils/date.ts
```diff
@@ -15,21 +18,21 @@ function _padStartWidthZeros(value: number, totalStringSize: number): string {
  * An alternative separator can be provided to be used instead of hyphens.
  * @param date The date to serialise
  * @param includeTime Whether to include the time in the serialised string
- * @param separator The separator to use between date parts, e.g. 2025-01-01 or 2025/01/01
+ * @param separator The separator to use between date and time parts (default 'T')
  */
 
-export function _serialiseDate(date: Date | null, includeTime = true, separator = '-'): string | null {
+export function _serialiseDate(date: Date | null, includeTime = true, separator = DATE_TIME_SEPARATOR): string | null {
```
peterjrreynolds, line 24: The third argument wasn't actually used anywhere, so have repurposed

packages/ag-grid-community/src/filter/provided/date/dateFilterUtils.ts
```diff
@@ -18,7 +18,7 @@ export function mapValuesFromDateFilterModel(
     //       the model. When we recreate the date again here, it's without a timezone.
     const { dateFrom, dateTo, type } = filterModel || {};
     return [
-        (dateFrom && _parseDateTimeFromString(dateFrom)) || null,
-        (dateTo && _parseDateTimeFromString(dateTo)) || null,
+        (dateFrom && _parseDateTimeFromString(dateFrom, undefined, true)) || null,
+        (dateTo && _parseDateTimeFromString(dateTo, undefined, true)) || null,
```
sdwvit, line 22: can we do an options param instead of forcing dev to pass undefined?

sdwvit (approved):
one comment, rest lgtm

packages/ag-grid-community/src/filter/provided/date/dateFilterUtils.ts
peterjrreynolds: There's a performance overhead to switching to params. The third argument isn't needed outside date filters, so would rather leave this as is.

PR #11884, opened 2025-08-29
AG-3390 - Update snyk ignore with updated dep

taktran:
Something missed in https://github.com/ag-grid/ag-grid/pull/11877


seanlandsman (approved):
lgtm

PR #11928, opened 2025-09-03
Validate Rows have Guis as part of cell check

StephenCooper:
Fix AG-13130

### DIFF
--- a/packages/ag-grid-react/src/reactUi/renderStatusService.tsx
+++ b/packages/ag-grid-react/src/reactUi/renderStatusService.tsx
@@ -38,8 +38,11 @@ export class RenderStatusService extends BeanStub implements IRenderStatusServic
     public areCellsRendered(): boolean {
         const { rowRenderer } = this.beans;
         // Check that all rows ctrls have a gui
-        // Check that all cells have a GUI. 
+        // Check that all cells have a GUI.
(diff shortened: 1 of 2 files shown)


packages/ag-grid-react/src/reactUi/renderStatusService.tsx
```diff
@@ -36,6 +36,10 @@ export class RenderStatusService extends BeanStub implements IRenderStatusServic
     }
 
     public areCellsRendered(): boolean {
-        return this.beans.rowRenderer.getAllCellCtrls().every((ctrl) => !!ctrl.eGui);
+        const { rowRenderer } = this.beans;
+        // Check that all rows ctrls have a gui
+        // Check that all cells have a GUI. 
+        // The guis are only set once React has actually rendered the row / cell.
+        return rowRenderer.getAllRowCtrls().every((row) => row.isRowRendered()) && rowRenderer.getAllCellCtrls().every((ctrl) => !!ctrl.eGui);
     }
```
StephenCooper, line 45: We need to also handle the case where row ctrls have been created but not yet rendered by React. The cells check would pass based on the old cells as the new cell ctrls had not been created yet on the first run.

peterjrreynolds (approved):

PR #11933, opened 2025-09-03
AG-13990 Remove style that is now active

StephenCooper:
With this change.
Screenshot 2025-09-03 at 11 54 11

Without this change.
Screenshot 2025-09-03 at 11 54 23

Looks like the changes I made to use the theming api / or maybe some other change in the website means that the styles applied in a file called `power-of-ag-charts.css` is no longer applying to this example. That was overriding this HR focus style meaning it was not visible.

Fix [AG-13990](https://ag-grid.atlassian.net/browse/AG-13990)


markdurrant (approved):
Updated to retain vertical cell borders. 

LGTM 👍

PR #11934, opened 2025-09-03
AG-15832-check-also-editor-for-validation-capabilities

StephM-W:
Fix AG-15832

### DIFF
--- a/packages/ag-grid-community/src/edit/utils/editors.ts
+++ b/packages/ag-grid-community/src/edit/utils/editors.ts
@@ -510,7 +510,11 @@ function _hasValidationRules(beans: BeanCollection): boolean {
             );
         });
 
-    return columnsHaveRules || getFullRowEditValidationErrors;
+    const editorsHaveRules = beans.gridApi
+        .getCellEditorInstances()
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/edit/utils/editors.ts
```diff
@@ -492,7 +492,11 @@ function dispatchEditingStopped(
     }
 }
 
-function _hasValidationRules(beans: BeanCollection): boolean {
+function _hasValidationRules(beans: BeanCollection, force?: boolean): boolean {
+    if (force) {
```
peterjrreynolds: I think this is the responsibility of the caller (e.g. remove force from here, and just do the check in `_populateModelValidationErrors`)

StephM-W: It's called from a variety of places that don't have the context of where the ultimate call is coming from

StephM-W: Apologies I misunderstood the direction of your comment, makes sense!

peterjrreynolds (approved):

PR #11944, opened 2025-09-04
[AG-11746] Fix(ssrm): docs fix

sdwvit:
- Include module documentation comment for resetRowHeights

Fix [AG-11746](https://ag-grid.atlassian.net/browse/AG-11746)


packages/ag-grid-community/src/api/gridApiFunctions.ts
```diff
@@ -291,7 +291,10 @@ export const gridApiFunctionsMap: Record<keyof GridApi, ValidationModuleName> =
         getCacheBlockState: 0,
         isLastRowIndexKnown: 0,
     }),
-
+    ...mod<_RowModelSharedApi>('RowModelSharedApi', {
```
sdwvit: unsure exactly if I need this, and what is the implication if I do. perhaps @peterjrreynolds knows?

AndrewGlazier (approved):

packages/ag-grid-community/src/api/gridApiFunctions.ts
StephenCooper: Not required, as these methods are already listed

PR #11947, opened 2025-09-04
Rename InternalTestCase to TestCaseWithMeta and update references

sdwvit:
- Update function parameters to use TestCaseWithMeta instead of InternalTestCase
 - Replace __hidden with meta property in test case structure
 - Modify benchError to access error.stack via testCase.meta
 - Update testBody to use testCase.meta for iteration metrics
 - Adjust preSetup function signature to include testCase parameter
 - Update test case setup to use meta object instead of __hidden


testing/performance/e2e/matrix/index.cron.spec.ts
```diff
@@ -18,7 +18,7 @@ test(`Performance Test - `, {
         [10, 10],
         [10, 100],
         [10, 1000],
-        [100000, 1000],
+        [100000, 1000], // index = 3
```
sdwvit:
```suggestion
        [100000, 1000],
```

StephenCooper (approved):

PR #11949, opened 2025-09-04
Ag 15738 update sidebar image

kylerphillips:
Screenshot 2025-09-04 at 15 34 45


markdurrant (approved):
Ready to go 👍 

In an ideal world there wouldn't be a difference between the images apart from light/dark mode. But it's not a blocker to release.

documentation/ag-grid-docs/src/content/docs/side-bar/index.mdoc
```diff
@@ -5,6 +5,9 @@ enterprise: true
 
 This section covers how to configure the Side Bar which contains Tool Panels.
 
+{% imageCaption imagePath="resources/sidebar.webp" alt="Side Bar" centered=true /%}
```
sdwvit, line 8: how about https://ag-grid.com/javascript-data-grid/tool-panel/#overview

taktran (approved):
LGTM

PR #11958, opened 2025-09-05
Only wait on cells if they are present and hidden

StephenCooper:


testing/ag-grid-docs/e2e/utils.ts
```diff
@@ -20,7 +20,7 @@ export async function waitForGridReady(page: Page) {
         .first()
         .waitFor({ state: 'attached' });
 
-    if (await cellLocator.first().isHidden()) {
+    if ((await cellLocator.count()) > 0 && (await cellLocator.first().isHidden())) {
         console.log('Cells are hidden, waiting for visible state', new Date());
         await cellLocator.first().waitFor({ state: 'visible' });
         console.warn('Cells are now visible at', new Date());
```
StephenCooper, line 26: This logging is intentional to see if there is an actual issue with hidden cells being revealed.

seanlandsman (approved):
lgtm

PR #11963, opened 2025-09-08
RTI-3034 Run doc tests off rc

StephenCooper:


documentation/ag-grid-docs/src/content/docs/cell-selection-fill-handle/_examples/fill-handle/example.spec.ts
```diff
@@ -3,6 +3,7 @@ import { dragOverTo, expect, test } from '@utils/grid/test-utils';
 test.agExample(import.meta, () => {
     test.describe('Fill Handle', () => {
         test.vanilla('Drag Fill', async ({ page, remoteGrid, agIdFor }) => {
+            test.skip(true, 'Skipping due to flakiness, to be re-enabled in future');
```
StephenCooper, line 6: It seems this is not going to be stable hence why I have skipped it.

testing/ag-grid-docs/playwright.config.ts
```diff
@@ -39,8 +39,7 @@ export default defineConfig({
     use: {
         /* Base URL to use in actions like `await page.goto('/')`. */
         // baseURL: 'https://grid-staging.ag-grid.com',
-        baseURL: process.env.CI ? 'https://grid-staging.ag-grid.com' : 'https://localhost:4610',
-
```
StephenCooper, line 43: We may want to make this cleaner to enable but just doing the simplist change for now.

StephM-W (approved):

PR #12014, opened 2025-09-10
AG-15757 drag scroll wheel

SalvatorePreviti:
To make scrolling work while dragging, some changes were needed

- add support for pointerEvents that behave much better than hand made mouse and touch dragging
    - they support mouse capture, so the events will be all redirected to the element being dragged
    - they get properly cancelled when switching to another tab or app in a browser without leaving the drag active
    - they handle both touch and mouse events
- add handlers for the scroll events in the document, to treat page scroll as they are mouse dragging events, so the target position gets updated while scrolling
- keep old mouse and touch handling as fallback, to not break potentially existing e2e tests from us or the user
- add pointerEvents: none dynamically to the element being scrolled if the mouse was captured via a pointerEvent
    - since we are capturing the mouse, dragging over selection or other input elements will not be a problem, will not trigger anything
- remove unused exposed public properties and methods
- refactor the logic to make it more clear keeping a state object called Dragging
- remove _alreadyProcessedByDragService that was being set to the DOM event object, and use a WeakSet instead, as is a better practice
- _isFocusableFormField was still written to support internet explorer with the method msMatchesSelector - modernize it, accepts Element instead of just HTMLElement, and fix the other function so that they can accept Element too.


packages/ag-grid-community/src/agStack/core/baseDragService.ts
```diff
@@ -312,74 +389,162 @@ export class BaseDragService<
...
+    private handlers: TemporaryEventHandler[] = [];
+
+    public constructor(
+        public readonly kind: TKind,
+        public readonly dragSource: DragListenerParams,
+        public readonly startEvent: TEvent
+    ) {
+        this.kind = kind;
+        this.dragSource = dragSource;
+        this.startEvent = startEvent;
```
peterjrreynolds: Can remove these three lines

packages/ag-grid-community/src/agStack/core/baseDragService.ts
```diff
@@ -312,74 +389,162 @@ export class BaseDragService<
...
+
+// Union type for all drag variants (no any usage)
+type Dragging = BaseDragging<'mouse', MouseEvent> | BaseDragging<'touch', Touch> | PointerDragging;
+
+const INTERACTIVE_TAG_REGEX = /^(a|textarea|input|select|button)$/i;
+
+const isOverFormFieldElement = (event: Event): boolean => {
+    const tag = (event.target as Element | null)?.tagName;
+    return !!tag && INTERACTIVE_TAG_REGEX.test(tag);
+};
```
peterjrreynolds: Can this share the `_isFocusableFormField` logic already used for touch events?

packages/ag-grid-community/src/agStack/core/baseDragService.ts
SalvatorePreviti: thank you missed it!

packages/ag-grid-community/src/agStack/core/baseDragService.ts
SalvatorePreviti: done - and refactored again all the rest
Consider that all modern browser supports pointer events and pointer capture that replaces this old mechanism

packages/ag-grid-community/src/agStack/core/baseDragAndDropService.ts
```diff
@@ -380,18 +381,21 @@ export abstract class BaseDragAndDropService<
...
+    private appendDragImageComp(component: IComponent<any> & IDragAndDropImage, pointerCaptured: boolean): void {
         const eGui = component.getGui();
         const style = eGui.style;
 
-        style.setProperty('position', 'absolute');
-        style.setProperty('z-index', '9999');
+        style.position = 'absolute';
+        style.zIndex = '9999';
+        if (pointerCaptured) {
+            style.pointerEvents = 'none';
```
gportela85, line 397: The pointerEvents cannot be `none` here. The whole idea of the DragImage is that it would capture the pointer events to prevent the selection of text outside the grid while dragging content.

When will `pointerCapture` be true or false? What are the cases here?

gportela85 (approved):

PR #12020, opened 2025-09-11
AG-15918 Global grid options updates

StephenCooper:
Fix AG-15918

### DIFF
--- a/packages/ag-grid-community/src/globalGridOptions.ts
+++ b/packages/ag-grid-community/src/globalGridOptions.ts
@@ -39,6 +39,27 @@ export class GlobalGridOptions {
 
         return mergedGridOps;
     }
+
+    /**
+     * Apply global grid option for a specific option key.
+     * If the merge strategy is 'deep' and both global and provided values are objects, they will be merged deeply.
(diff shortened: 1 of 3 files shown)


SalvatorePreviti (approved):

packages/ag-grid-community/src/globalGridOptions.ts
```diff
@@ -39,6 +39,27 @@ export class GlobalGridOptions {
...
+
+    /**
+     * Apply global grid option for a specific option key.
+     * If the merge strategy is 'deep' and both global and provided values are objects, they will be merged deeply.
+     * Otherwise, the provided value is returned as is.
+     * @param optionKey - The key of the grid option to apply.
+     * @param providedValue - The value provided to the grid instance.
+     * @returns The merged value if applicable, otherwise the provided value.
+     */
+    static applyGlobalGridOption<K extends keyof GridOptions>(
```
SalvatorePreviti, line 51: just a reflection - in general i am a bit worried of the prrformance of this implementation of this class, and the risk of infinite loops if there is for some reason a cycle - however this is a bigger problem and I don't think is an issue right now

PR #12026, opened 2025-09-11
Update index.mdoc : minor: Fix typo

nikhilknoldus:


StephenCooper:
Thanks for letting us know. We will get that updated.

PR #12031, opened 2025-09-12
AG-8818 - Fix groupData not populated when cols change

AndrewGlazier:


packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -1,6 +1,44 @@
 import type { NamedBean } from '../context/bean';
+import { _getClientSideRowModel } from '../main';
 import { AbstractClientSideNodeManager } from './abstractClientSideNodeManager';
 
 export class ClientSideNodeManager<TData> extends AbstractClientSideNodeManager<TData> implements NamedBean {
     beanName = 'csrmNodeSvc' as const;
+
+    public override updateGroupRowData(): void {
```
SalvatorePreviti, line 8: minor: I am not sure this should be in the node manager, shouldn't be this in the client side row manager as the node manager is not supposed to know anything about groups, the CSRM does the grouping

packages/ag-grid-enterprise/src/treeData/clientSideChildrenTreeNodeManager.ts
```diff
@@ -243,4 +250,27 @@ export class ClientSideChildrenTreeNodeManager<TData>
             params.rowNodesOrderChanged ||= orderChanged;
         }
     }
+
+    public override updateGroupRowData(): void {
```
SalvatorePreviti, line 254: should not be this the same implementation for both kind of nodes? nodes with children or nodes without children should not make a difference here for this

packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: maybe this might need to be part of the row grouping and not CSRM as groupData makes sense only if grouping or treeData modules are loaded

PR #12045, opened 2025-09-16
Allow single pixel diff

StephenCooper:
Test running here https://github.com/ag-grid/ag-grid/actions/runs/17763457938


seanlandsman (approved):
lgtm

PR #12051, opened 2025-09-17
AG-9704 Allow scaling up autosized columns to fill empty space

eliasmalik:
Fix AG-9704


packages/ag-grid-community/src/interfaces/autoSize.ts
```diff
@@ -69,6 +69,8 @@ export interface SizeColumnsToContentStrategy {
     defaultMaxWidth?: number;
     /** Provide to limit specific column widths when sizing. */
     columnLimits?: SizeColumnsToContentColumnLimits[];
+    /** Proportionally scale up columns after sizing to fill any empty space remaining in the grid. */
```
StephenCooper: Is there a reason there is so much duplications between these interfaces. Can one extend the other? Is this the same for all the other existing interfaces too?

packages/ag-grid-community/src/validation/rules/gridOptionsValidations.ts
```diff
@@ -526,6 +526,22 @@ const GRID_OPTION_VALIDATIONS: () => Validations<GridOptions> = () => {
...
+                const validModes: NonNullable<GridOptions['autoSizeStrategy']>['type'][] = [
+                    'fitCellContents',
+                    'fitGridWidth',
+                    'fitProvidedWidth',
+                ];
+                const type = autoSizeStrategy.type;
+                if (type !== 'fitCellContents' && type !== 'fitGridWidth' && type !== 'fitProvidedWidth') {
+                    return `Invalid Auto-size strategy. \`autoSizeStrategy\` must be one of ${validModes.map((m) => '"' + m + '"').join(', ')}, currently it's ${type}`;
+                }
+                return null;
```
StephenCooper, line 545: Can you add a validation rule for `fitGridWidth` to have the required `width` property. I believe that is the only strategy with a required property. Otherwise the grid can look very broken if they just set that type and don't provide a width.

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
```diff
@@ -638,11 +636,7 @@ export class AgGridAngular<TData = any, TColDef extends ColDef<TData> = ColDef<a
      * @initial
      * @agModule `ColumnAutoSizeModule`
      */
-    @Input() public autoSizeStrategy:
-        | SizeColumnsToFitGridStrategy
-        | SizeColumnsToFitProvidedWidthStrategy
-        | SizeColumnsToContentStrategy
-        | undefined = undefined;
+    @Input() public autoSizeStrategy: AutoSizeStrategy | undefined = undefined;
```
StephenCooper, line 639: Can we validate that this won't be a breaking change, going from the individual to the union type. I wouldn't expect it to be but can we just confirm.

documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/column-resizing/main.ts
```diff
@@ -47,8 +42,11 @@ const gridOptions: GridOptions<IOlympicData> = {
         defaultMinWidth: 80,
     },
 };
-function autoSizeAll(skipHeader: boolean) {
-    gridApi!.autoSizeColumns({ skipHeader, defaultMaxWidth: 150, defaultMinWidth: 80 });
+
+function autoSizeAll() {
+    const skipHeader = document.querySelector<HTMLInputElement>('#toggle-ignore-headers')?.checked ?? false;
+    const scaleUpToFitGridWidth = document.querySelector<HTMLInputElement>('#toggle-scale-up')?.checked ?? false;
```
StephenCooper, line 48: I am guessing this is how we do all our toggle examples? This code ends up in the frameworks too?

StephenCooper (approved):

PR #12052, opened 2025-09-17
Sorting benchmark, cleanup testmanager

StephM-W:


testing/behavioural/src/rows/benchmarks/sorting.bench.ts
```diff
@@ -0,0 +1,168 @@
...
+import type { GridApi } from 'ag-grid-community';
+import { ClientSideRowModelApiModule, ClientSideRowModelModule } from 'ag-grid-community';
+import { RowGroupingModule } from 'ag-grid-enterprise';
+
+import { SimplePRNG, TestGridsManager } from '../../test-utils';
+
+const DATA_SIZES = [20000];
+const COLUMN_COUNTS = [5];
+
+suite('row grouping', () => {
```
StephenCooper: does this need renaming

StephenCooper (approved):

testing/behavioural/src/rows/benchmarks/sorting.bench.ts
StephM-W: Updated

PR #12055, opened 2025-09-17
AG-14428 - fixed checkbox shadow when wrapped

gportela85:
fix AG-14428

### DIFF
--- a/packages/ag-grid-community/src/rendering/cell/cellComp.ts
+++ b/packages/ag-grid-community/src/rendering/cell/cellComp.ts
@@ -212,7 +212,9 @@ export class CellComp extends Component {
         const usingCellValue = !editing && usingWrapper;
         const putCellValueIn = usingCellValue && this.eCellValue == null;
         if (putCellValueIn) {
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/rendering/cell/cellComp.ts
```diff
@@ -212,7 +212,9 @@ export class CellComp extends Component {
         const usingCellValue = !editing && usingWrapper;
         const putCellValueIn = usingCellValue && this.eCellValue == null;
         if (putCellValueIn) {
-            this.eCellValue = _createElement({ tag: 'span', cls: 'ag-cell-value', role: 'presentation' });
+            const isCheckboxRenderer = this.column.getColDef().cellRenderer === 'agCheckboxCellRenderer';
```
StephenCooper: I think we could move this to the cellCtrl and either have the cellCtrl expose this as a boolean. Or go further and have a new method on the cellCtrl that will return the class string. 

Should reduce the duplication in each of the components.

StephenCooper (approved):

StephenCooper (approved):

PR #12057, opened 2025-09-18
AG-5833 touch context menu

SalvatorePreviti:
- make the touch menu work on IOs and other browsers when long pressing on a header column and group columns
    - ensures that the touchListener does not handle events already handled by other touch listeners on top of the element with a global weak set
    - improves TouchListener reentrancy ensuring everything gets disposed correctly, including the timer
    - prevents the default context menu to show
- fix previous problem with the last switch to pointer events for dragging and IOs / touch / touch+mouse
    - use a global weak event so events are processed twice globally
    - fix handling of touch when pointer events happens
    - add a retry for capture mouse that in devTools seems to happen to receive a capture lost event
- the new sonarQube rules raised warning with the old code, followed the recommentations
    - use for of instead of forEach
    - use `after` instead of insert adjacent
    - replace infexOf with includes

Tested in chrome, in chrome dev tools, safari and chrome for IOs:
- Drag and drop of rows
- Drag and drop of columns
- Context menu for header, columns and rows


packages/ag-grid-community/src/agStack/core/baseDragService.ts
```diff
@@ -126,12 +143,11 @@ export class BaseDragService<
         const beans = this.beans;
         const onScroll = (event: Event) => this.onScroll(event);
         const keydownEvent = (ev: KeyboardEvent) => this.onKeyDown(ev);
-
         const rootEl = _getRootNode(beans);
         const eDocument = _getDocument(beans);
         addTempEventHandlers(
             drag.handlers,
-            [rootEl, 'contextmenu', preventEventDefault],
+            [rootEl, 'contextmenu', preventEventDefault, { passive: false }],
```
gportela85: why is there a need to pass `passive: false` here? This shouldnt be required

gportela85 (approved):

PR #12079, opened 2025-09-23
AG-15939 split row dragging docs

SalvatorePreviti:
- Split row dragging documentations in multiple sub pages
- fix all the links around
- fix a problem with the doc type parser that seems to parse strings in unions like types to process, for example Type = 'Context' | 'LeftRight' was causing the internal Context interface to be parsed and analyized. See the fix in documentation/ag-grid-docs/src/components/reference-documentation/utils/getAllPotentialTypesFromString.ts - this was causing `'AG Charts - Possible recursion error on type:` to warn slow down and break the docs generator
- reduces the 'AG Charts - Possible recursion error on type: ' log length that was generating logs too big to be read and slowing down CI


documentation/ag-grid-docs/src/content/docs/row-dragging/index.mdoc
```diff
@@ -34,358 +34,10 @@ const gridOptions = {
 
 There are two ways in which row dragging works in the grid, managed and unmanaged:
 
-* **Managed Dragging**: This is the simplest, where the grid will rearrange rows as you drag them.
-* **Unmanaged Dragging**: This is more complex and more powerful. The grid will not rearrange rows as you drag. Instead the application is responsible for responding to events fired by the grid and rows are rearranged by the application.
+* **[Managed Row Dragging](./row-dragging-managed)**: This is the simplest, where the grid will rearrange rows as you drag them.
+* **[Unmanaged Row Dragging](./row-dragging-unmanaged)**: This is more complex and more powerful. The grid will not rearrange rows as you drag. Instead the application is responsible for responding to events fired by the grid and rows are rearranged by the application.
```
StephenCooper, line 38: Do we need to add links to the existing sub pages of External DropZone and Grid to Grid?

SalvatorePreviti: Added more "see also"

documentation/ag-grid-docs/src/components/reference-documentation/utils/getAllPotentialTypesFromString.ts
```diff
@@ -0,0 +1,109 @@
...
+ * - No other parsing or filtering is applied.
+ * - exclude browser global classes and common typescript types that should not be processed.
+ *
+ * This is required to avoid the situation where for example EnumType = 'XxX' | 'Context'; will incorrectly process type Context that is internal.
+ *
+ * Example:
+ *   Input:  "Xxx<'Context', TData> | Baz[]"
+ *   Output: Set { "Xxx", "TData", "Baz" }
+ */
+export function getAllPotentialTypesFromString(type: string): string[] {
```
StephenCooper, line 15: Good find! Hopefully this will speed things up too.

StephenCooper (approved):

PR #12105, opened 2025-09-26
AG-9704 Use relative assertions in doc tests

eliasmalik:
Don't hardcode pixel assertions to avoid platform specific differences.


eliasmalik:
Doc tests pass: https://github.com/ag-grid/ag-grid/actions/runs/18037661526

StephenCooper (approved):

PR #12107, opened 2025-09-26
Fix compile error

olegat:
I do not fully understand how this build regression happened, but I've updated the `PolarChartProxy.getSeries` method to hopefully be more resistant to this kind of build regression.


alantreadway (approved):
LGTM - thanks!

PR #12108, opened 2025-09-26
AG-15943 Doc tests improvements

StephenCooper:
Run all Doc Tests examples via single playwright config.

Required added an example.spec.ts file to every existing example that we wish to test. Each of these files contains some placeholder test logic to wait for the grid and click all buttons in the example whilst watching for warnings / errors. 

Also added support to be able to only run certain tests on CI via playwrights default test filtering which now can work based off the folder path of the example. 

Manual full run: https://github.com/ag-grid/ag-grid/actions/runs/18127380027


.github/actions/test-framework-examples/action.yml
```diff
@@ -5,9 +5,14 @@ inputs:
     description: 'Framework to use'
     required: true
     default: 'invalid-framework'
-  cache-key:
-    description: 'Cache key for all-examples-cached.json'
+  browsers:
+    description: 'Broswers to run tests on: all / chromium'
```
StephM-W, line 9: 'Browsers'

documentation/ag-grid-docs/src/content/docs/example-demos.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { expect, test } from '@playwright/test';
+import { setupConsoleExpectations } from '@utils/grid/test-utils';
+
+// These could be extended to actually interact with the examples more
+// but for now just a basic load test to ensure no errors / warnings in console
+// and the grid loads without issues
+
+test.describe(`Demo Examples`, async () => {
```
StephenCooper, line 8: Moved the Demo smoke tests to here so that they will just get run.

StephM-W (approved):

PR #12130, opened 2025-09-29
[AG-6844] Feature(master-detail): allow expandall collapseall

sdwvit:
Docs update && Corrected conditions / calls for detail grid expansion service.

Fix [AG-6844](https://ag-grid.atlassian.net/browse/AG-6844)


SalvatorePreviti (approved):

StephenCooper (approved):
Do we have an example in the docs that shows this behaviour. If we do it would be good to validate it via the example test.

sdwvit:
We don't currently, but I have adding tests on my radar as part of the whole SSRM support business

PR #12131, opened 2025-09-29
AG-15946 Suppress validation module reference for UMD code

seanlandsman:


packages/ag-grid-community/src/validation/logging.ts
```diff
@@ -122,6 +123,10 @@ export function getErrorLink(errorNum: ErrorId, args: GetErrorParams<any>) {
 
 const minifiedLog = (errorNum: ErrorId, args: GetErrorParams<any>, defaultMessage?: string) => {
     const errorLink = getErrorLink(errorNum, args);
+
+    if (_isUmd()) {
+        return errorLink;
```
StephenCooper: Do we need to include the defaultMessage here too?

StephenCooper (approved):

PR #12137, opened 2025-09-30
AG-16051 csrm node manager dynamic beans

SalvatorePreviti:
Before we had abstractClientSideNodeManager, base class of
    - ClientSideNodeManager, responsible to load flat list of data
    - ClientSideChildrenTreeNodeManager, responsible to load recursive data with the children field, that contains an array of data
They were both named beans.

By merging those 3 classes together and delegate to the groupStage and treeGroupStrategy
the management of the property changes, we can merge those class together,
as the final implementation is simpler and faster than before.

This indeed remove the need to have named beans or dynamic beans for the node manager,
we just instantiate one when the ClientSideRowModel is created.
This also reduces the bundle size considerably.

Few methods were added instead to the GroupStage to handle the state.
This creates a better separation of concern and as a result we have much less code.

Together with this:
- Move extractData to the group stage, as is not used when not using tree data (bundle size)
- Simplify the rowNode class regarding typing, removing the interfaces that mark some fields as writable, the attempt here was to make the code clearer but in reality adds complexity and too many casts, using just RowNode make the code simpler and less import/exports.
- solved all sonarqube warnings regarding forEach, complexity, null coalescing operator.
- reviewed all the loops in groupStrategy, CSRM, ClientSideNodeManager
- Fixed a bug that was registering a new destructor function (memory leak) whenever the treeData property changed
- Simplified the handling of reordered state and transactions, as moves the reordered flag to the ChangedRowNodes class
- Simplifies and optimizes the changed property detection in ClientSideRowModel for the stages, using a single map instead of an array of sets for each state
- Simplifies and shorten up the fieldGetter used by TreeData
- Adds some missing properties in IGNORED_SIBLING_PROPERTIES - that could cause memory leaks
- remove the unused __daemon flag - it was written but never read in RowNode
- removes the _getGroupingApproach function, is not needed as we can query the treeData status on the GroupStage
- remove isRowDataLoaded and use a readonly property instead
- remove the interface IChangedRowNodes, we can just use ChangedRowNodes type

I don't expect any problems with those changes in term of regression or change of behaviour, all behaviour is maintained as is - and tests have a high coverage on any part of the changed code

Fix AG-16051


StephenCooper (approved):
Discussed offline.

falsyvalues:
This change introduces breaking change to `ag-grid-community` package API, as it removes `AbstractClientSideNodeManager` and `_getGroupingApproach` exported from package.
Please avoid such actions, and clearly divide public vs internal  APIs which shouldn't be imported.

PR #12159, opened 2025-10-02
[AG-11882] Fix(master-detail): allow master rows to update whether they display an expansion button

sdwvit:
Enable dynamic expansion button visibility for master rows via isRowMaster callback

Local example SSRM: https://plnkr.co/edit/zAXiCJ2NIM0VP8mM?open=main.ts&preview
Local example CSRM: https://plnkr.co/edit/tXPXIpw4Txhnovwe?open=main.ts&preview

Fixes [AG-11882](https://ag-grid.atlassian.net/browse/AG-11882)
Fixes [AG-16107](https://ag-grid.atlassian.net/browse/AG-16107)


packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
```diff
@@ -197,12 +198,14 @@ export class BlockUtils extends BeanStub implements NamedBean {
...
+        const gos = this.gos;
+        const treeData = gos.get('treeData');
 
         rowNode.setDataAndId(data, defaultId);
         const group = rowNode.group;
-
-        if ((treeData || !group) && this.gos.get('masterDetail')) {
+        const isTreeOrNotAGroup = treeData || !group;
+        const isMasterDetail = gos.get('masterDetail');
+        if (isTreeOrNotAGroup && isMasterDetail) {
```
SalvatorePreviti: there is a better way - group nodes or tree data filler nodes cannot be masters, but tree data groups with data can be master.
Tree data can have also filler nodes, nodes without data.
I think the right way is to check if (rowNode.data) then it can be master, else it can't be

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -880,18 +880,35 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
     }
 
     public refreshRow(params: RefreshRowsParams & { newData?: boolean }): void {
-        // if the row is rendered incorrectly, as the requirements for whether this is a FW row have changed, we force re-render this row.
+        const gos = this.gos;
```
SalvatorePreviti: is really this the right place to do this? Besides the fact that seems very expensive, is not there a better place? this is also calling redraw if master field didn't change - i think this logic is already in place somewhere to refresh the row if master flag changes?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -292,21 +292,37 @@ export class RowNode<TData = any>
     }
 
     private setDataCommon(data: TData, update: boolean): void {
-        const { valueCache, eventSvc } = this.beans;
+        const { valueCache, eventSvc, gos } = this.beans;
         const oldData = this.data;
-
         this.data = data;
+
+        const hasMasterFunc = gos.get('isRowMaster');
```
SalvatorePreviti: I know this function is exposed to the end user, this adds a level of complexity for sure ... but we are invoking this callback twice in this way, in MasterDetailService and here

- or remove it from there if this is the case and completely change that implementation 
- or do it instead in SSRM when data changes

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -880,18 +880,35 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
...
+        const hasRowMasterFunc = gos.get('isRowMaster');
+        const isRowMaster = gos.get('masterDetail');
+        const redraw = () => this.beans.rowRenderer.redrawRow(this.rowNode);
```
SalvatorePreviti: this creates a new object, no need to create a new function here

packages/ag-grid-community/src/rendering/row/rowCtrl.ts
```diff
@@ -880,18 +880,35 @@ export class RowCtrl extends BeanStub<RowCtrlEvent> {
...
+
+        /**
+         * If we have isRowMaster cb, we need to always refresh, as the result of the function
+         *   could have changed, eg row was master and now isn't, or vice versa.
+         * There are two parts to this, and this is the second bit.
+         * First part is in rowNode.ts where row.master is set on same conditions.
+         */
+        if (hasRowMasterFunc && isRowMaster) {
+            this.getAllCellCtrls().forEach((cellCtrl) => cellCtrl.refreshCell(params));
```
SalvatorePreviti: sonarQube new rules says: no .forEach - we are removing all .forEach as we find them - let's not add new .forEach - better to use for loops

SalvatorePreviti:
This would really benefit from behavioural unit tests both for CSRM and SSRM also to verify we don't invoke isMaster callback multiple times.
This is much more impactful than it seems

Currently tests are failing

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -291,22 +291,30 @@ export class RowNode<TData = any>
...
+    }
+
+    private setDataCommon(data: TData, update: boolean, internal = false): void {
         const { valueCache, eventSvc } = this.beans;
         const oldData = this.data;
-
         this.data = data;
+
+        if (!internal) {
+            this.beans.masterDetailSvc?.setMaster(this, false, true);
```
SalvatorePreviti: Just an idea for clarity:

- instead of having an internal flag, I would just repeat `this.beans.masterDetailSvc?.setMaster(this, false, true);` in the two public facing methods
- instead of creating a new `updateDataInternal` method that is not clear what it does, we can just make `setDataCommon` public maybe?
- we need to add documentation that internally updateData and setData should NOT be called and instead we should call setDataCommon

However - we could consider to change approach altogether and use a flag in the rowModel - see other comments

SalvatorePreviti: we have also other calls to updateData internally that can create problems and we need to check
- the clientSideChildrenTreeNodeManager uses updateData and this needs to call the internal version instead
- SSRM is for sure already setting this flag and calling the dispatch event - if we call updateData as it is now aren't we going to have two calls to isMaster callback?
- we need to check viewPortRowModel and staticPinnedRowModel too - other places where updateData is called

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -291,22 +291,30 @@ export class RowNode<TData = any>
         this.setDataCommon(data, true);
     }
 
-    private setDataCommon(data: TData, update: boolean): void {
+    public updateDataInternal(data: TData): void {
```
SalvatorePreviti: This should work now - in theory besides checking all the places we call updateData - I don't know if it would be better and cleaner to instead add a `readonly updatingData?: boolean` in the IRowModel interface and check in the masterDetailService.setMaster if this flag is on, if so skip setting the master? Delegating the responsibility away from the RowNode class and from who calls those methods

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -50,59 +76,58 @@ export class MasterDetailService extends BeanStub implements NamedBean, IMasterD
...
-    private setMasters(changedRowNodes: IChangedRowNodes | null | undefined): void {
+    public setMaster(row: RowNode, created: boolean, updated: boolean): void {
+        const oldMaster = row.master;
         const enabled = this.isEnabled();
-        this.enabled = enabled;
+        let newMaster = enabled;
 
         const gos = this.gos;
-        const isRowMaster = gos.get('isRowMaster');
+        const isRowMaster = gos.get('isRowMaster') as ((data: any) => unknown) | undefined;
```
SalvatorePreviti: minor: is this cast needed?

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
sdwvit: good spot, not needed

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -50,59 +77,64 @@ export class MasterDetailService extends BeanStub implements NamedBean, IMasterD
...
+            row.master = newMaster;
 
-            if (newMaster !== oldMaster) {
-                row.master = newMaster;
-                row.dispatchRowEvent('masterChanged');
+            if (row.detailNode && !newMaster && _isServerSideRowModel(gos)) {
+                delete this.store[row.detailNode.id!];
+                (beans.ssrmBlockUtils as BlockUtils)?.destroyRowNode(row.detailNode);
+                row.detailNode = undefined;
             }
```
sdwvit: @SalvatorePreviti

SalvatorePreviti (approved):

SalvatorePreviti (approved):

PR #12177, opened 2025-10-06
AG-16069-drag-drop-ssrm-fix

SalvatorePreviti:
The latest changes in drag feature were not considering the RowsDrop object can be created also with a row model that is not server side when using unmanaged drag.

- Removed the non nullable field "clientSideRowModel" in the class was misleading as it indeed can be null
- Added rootNode in the RowModel interface, so we assume every row model has to have a rootNode for clarity and completeness. The exception was InfiniteRowModel and ViewPortRowModel that now provides a dummy root node. Adding a dummy rootNode seems preferred than keeping it optional, also, the public interface returned by rowsDrop at the moment exposes a not null rootNode and changing it could be considered a breaking change. We just set it to null on destruction in all rowModels
- Added a unit test for drag and drop events in SSRM

Fix AG-16069


packages/ag-grid-community/src/interfaces/iRowModel.ts
```diff
@@ -9,6 +9,9 @@ export interface RowBounds {
 export type RowModelType = 'infinite' | 'viewport' | 'clientSide' | 'serverSide';
 
 export interface IRowModel {
+    /** The root row containing all the rows */
```
StephenCooper: Can we include the fact that this can be a dummy rowNode for 2 of the row models in the JsDoc?

StephenCooper (approved):

PR #12181, opened 2025-10-07
Fix example comparator

StephenCooper:


documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/_examples/infinite-set/main.ts
SalvatorePreviti: I'd suggest to not reinvent the wheel and copy the exact version we have in core - and I think is even clearer with a simple loop

```
export function areEqual<T>(a: readonly T[] | null | undefined, b: readonly T[] | null | undefined): boolean {
    if (a === b) {
        return true; // Same instance, no need to compare
    }
    if (!a || !b) {
        return a == null && b == null;
    }
    const len = a.length;
    if (len !== b.length) {
        return false;
    }
    for (let i = 0; i < len; i++) {
        if (a[i] !== b[i]) {
            return false;
        }
    }
    return true;
}
```

SalvatorePreviti (approved):

PR #12185, opened 2025-10-07
AG-14964 Delay rendering columns when autosizing and rowData available

eliasmalik:
Fix AG-14964


packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
```diff
@@ -22,6 +22,7 @@ import type {
     SizeColumnsToContentColumnLimits,
     SizeColumnsToContentStrategy,
 } from '../interfaces/autoSize';
+import { _isClientSideRowModel } from '../main';
```
StephenCooper: Need to import directly not from main.

StephenCooper (approved):

PR #12190, opened 2025-10-08
AG-16088 Avoid scrolling triggering more refreshes

StephenCooper:
Fix AG-16088


packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -320,8 +320,11 @@ export class CellCtrl extends BeanStub {
                 resolver = resolve;
             });
 
-            this.addManagedListeners(eventSvc, {
-                bodyScrollEnd: () => resolver(),
+            const [removeBodyScrollEnd] = this.addManagedListeners(eventSvc, {
```
eliasmalik, line 323: Unrelated, but this strikes me as an odd API for `addManagedListeners`. It would be more natural to have something like
```
const { foo } = this.addManagedListeners(eventSvc, { foo: () => {} });
```

eliasmalik (approved):

PR #12191, opened 2025-10-08
AG-11882-master-detail-csrm-guard

SalvatorePreviti:
add a updatingRowData flag that prevent to call setMaster before the whole rowData is loaded by CSRM
master detail service already uses setMasters, called by CSRM, to set the row master after all the row data is loaded

We don't want to change when isMaster is called for the user while the data is loaded, and we don't want to have isMaster called twice


sdwvit:
thanks Salvo, let's keep it open until we come up with a good solution which may be yours

PR #12217, opened 2025-10-13
[AG-10043] Feature(tooltip): allow setting headerTooltipValueGetter

sdwvit:
Feature(tooltip): add headerTooltipValueGetter for customizable tooltips 
- Introduced headerTooltipValueGetter to allow dynamic tooltip content.
 - Updated colDef, colDefValidations, gridOptions, and properties.json.
 - Enhanced tooltipService to prioritize headerTooltipValueGetter over headerTooltip.

local plnkr: https://plnkr.co/edit/riwX9Rc6ecKCH9d5?open=main.js
same but with column groups: https://plnkr.co/edit/NPMeY8YQqrX18I1c?open=main.js

Screenshot_20251013_163649

Fix [AG-10043](https://ag-grid.atlassian.net/browse/AG-10043)

### DIFF
--- a/.run/Header Group Tooltip Value Getter feat.run.xml
+++ b/.run/Header Group Tooltip Value Getter feat.run.xml
@@ -1,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Header Group Tooltip Value Getter feat" type="JavascriptDebugType" uri="https://plnkr.co/edit/NPMeY8YQqrX18I1c?open=main.js">
+    <method v="2" />
+  </configuration>
+</component>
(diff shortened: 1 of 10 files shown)


eliasmalik (approved):

packages/ag-grid-community/src/tooltip/tooltipService.ts
```diff
@@ -56,14 +57,10 @@ export class TooltipService extends BeanStub implements NamedBean {
...
-                if (value != null) {
-                    return value;
-                }
-
-                const res = column.getColDef().headerTooltip;
-                return res;
-            },
+            getTooltipValue: () =>
+                value ??
+                colDef?.headerTooltipValueGetter?.(_addGridCommonParams(gos, { location: 'cell', colDef, column })) ??
```
StephenCooper: Is 'cell' the correct choice here? Also we should consider if we are going to set the value / valueFormatted params.

StephenCooper (changes requested):
I think we need to check the params object being passed to the callback.

.run/Header Group Tooltip Value Getter feat.run.xml
StephenCooper, line 1: Are these run files accidentally committed?

StephenCooper (approved):

PR #12223, opened 2025-10-14
[AG-15375] Maintenance(ide): store recipes in git

sdwvit:
IDE really wants us to store them in the .run folder in the root


taktran (approved):
LGTM

PR #12225, opened 2025-10-14
[AG-16065] Fix(ssrm-groups): pivoting expansion state is preserved incorrectly

sdwvit:
Fix pivoting expansion state handling in SSRM
 - Added event listeners for columnPivotChanged and columnPivotModeChanged to reset expand state.

Relevant plunker has been added to the project

Fix [AG-16065](https://ag-grid.atlassian.net/browse/AG-16065)


.run/Pivoting expansion state SSRM.run.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
```
StephenCooper, line 1: Can you add this folder to gitIgnore to prevent it being checked in.

SalvatorePreviti (approved):

PR #12227, opened 2025-10-14
AG-12342 computed all leaf children

SalvatorePreviti:
allLeafChildren was mantained up to date by row grouping and tree data.
The cost of maintaining it is quite high considering is a field that not all users need and we used it very rarely, on non root nodes.
Keeping it up to date made rowGrouping particularly complex with its batchRemover.
Instead of trying to keep it up to date while executing transactions or immutable row data or refreshes, the simplest option is to compute it on demand when the property is accessed on the rowNode by the end user, and us removing the use of the property completely. The property caches the last result and updates will only recursively invalidate the field.

- Convert rowNode.allLeafChildren to a computed property that delegates the loading to the CSRM GroupStage
- Replace its usage internally by recursive loops or by accessing. the _leafs field directly when using CSRM
- Change rowGrouping, treeData, to invalidate the _leafs field instead of computing it
- Remove the needs to set sibling._leafs by checking if a row is a footer and has a sibling in the getter logic
- Remove the BatchRemover and largely simplify the row grouping strategy as a result of the fact it does not need to update the allLeafChildren
- Improve ClientSideNodeManager performance by never allocating a new _leafs array for the root, but instead reusing it and updating it in place, during immutable row data or transactions
- remove the method remove from ChangedRowNodes as now is handled more specifically in CSRM to avoid redundant set operations

Performance after:
image

Performance before:
image

Unit tests have a high level of coverage on CSRM, grouping, treeData.
No change in behaviour for the end user.

Fix AG-12342


testing/behavioural/src/test-utils/testGridsManager.ts
```diff
@@ -64,7 +64,6 @@ export class TestGridsManager {
     /** Destroys all created grids, and eventually created html elements */
     public destroyAllGrids(): void {
         for (const grid of this.getAllGrids()) {
-            grid.stopEditing(true);
```
StephenCooper, line 67: Was this change required to fix tests following these changes?

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/batchRemover.ts
```diff
@@ -1,84 +0,0 @@
-import type { RowNode } from 'ag-grid-community';
-
-// doing _removeFromArray() multiple times on a large list can be a bottleneck.
-// when doing large deletes (eg removing 1,000 rows) then we would be calling _removeFromArray()
-// a thousands of times, in particular RootNode.allGroupChildren could be a large list, and
-// 1,000 removes is time consuming as each one requires traversing the full list.
-// to get around this, we do all the removes in a batch. this class manages the batch.
-//
-// This problem was brought to light by a client (AG-2879), with dataset of 20,000
```
StephenCooper, line 9: Have you validated this ticket against your changes?

testing/behavioural/src/test-utils/testGridsManager.ts
SalvatorePreviti: No - I forgot to mention about this. The issue was that running the benchmarks a lot of console warning a module is not registered are printed out - I spoke with Steph before about this but never made the change, we noticed it isn't needed as all the test run without issue also removing it

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/batchRemover.ts
SalvatorePreviti: Yes - we are still doing batching indeed. The benchmark does perform deletions as well.
The previous implementation of the BatchRemover was keeping a map of objects containing two sets with the list of children to remove, and filtered the childrenAfterGroup and allLeafChildren array based on those.

The new implementation keep instead a set of parents with removed children, and filters at the end just childrenAfterGroup every child that has been moved or removed. It does this if the child.parent !== parent || removals.has(child), the set of all deleted nodes from CSRM that includes the deleted empty groups.
If so it happens, it also invalidates the allLeafChildren property.

In this way we have a simpler and faster batch removal, but is functionally the same.

I will push some comment to specify better what's happening and make mention again to the ticket

StephenCooper (approved):

PR #12236, opened 2025-10-15
AG-15930 Remove redundant casts

StephenCooper:


packages/ag-grid-community/src/misc/state/stateUtils.ts
```diff
@@ -91,7 +91,7 @@ export function convertColumnState(
 // have invalid values resulting in sparse arrays which will break state persistence/restoration.
 // e.g. [ 'colId1', undefined, 'colId3' ] => [ 'colId1', 'colId3' ]
 function _removeEmptyValues<T>(array: T[]): T[] {
```
eliasmalik, line 93: Isn't the type index wrong here? shouldn't this be `_removeEmptyValues<T>(array: (T | undefined)[]): T[];`

packages/ag-grid-community/src/modules/moduleRegistry.ts
```diff
@@ -44,7 +44,7 @@ function runVersionChecks(module: Module) {
     if (module.validate) {
         const result = module.validate();
         if (!result.isValid) {
-            const errorResult = result as ModuleValidationInvalidResult;
+            const errorResult = result;
```
eliasmalik: Could just remove this assignment and do `_errorOnce(${result.message})` directly

eliasmalik (approved):

packages/ag-grid-community/src/misc/state/stateUtils.ts
StephenCooper: That change actually results in 
```
Type '(T | undefined)[]' is not assignable to type 'T[]'.
  Type 'T | undefined' is not assignable to type 'T'.
    'T' could be instantiated with an arbitrary type which could be unrelated to 'T | undefined'.ts(2322)
```

PR #12237, opened 2025-10-15
AG-16089 fix theming performance regression in detached grids

BernieSumption:


packages/ag-grid-community/src/agStack/core/baseEnvironment.ts
```diff
@@ -69,8 +70,13 @@ export abstract class BaseEnvironment<
     public postConstruct(): void {
         const { gos, eRootDiv } = this;
         gos.setInstanceDomData(eRootDiv);
+        const themeStyleContainer = gos.get('themeStyleContainer');
+        const isShadowRoot = eRootDiv.getRootNode() instanceof ShadowRoot;
         this.eStyleContainer =
-            gos.get('themeStyleContainer') ?? (eRootDiv.getRootNode() === document ? document.head : eRootDiv);
+            gos.get('themeStyleContainer') ?? (eRootDiv.getRootNode() instanceof ShadowRoot ? eRootDiv : document.head);
```
StephenCooper: Can we re-use the `isShadowRoot` variable in this check too?

packages/ag-grid-community/src/agStack/core/baseEnvironment.ts
```diff
@@ -304,3 +310,12 @@ export interface BaseCssChangeKeys {
 }
 
 const NO_VALUE_SENTINEL = 15538;
+
+const warnOnAttachToShadowRoot = (el: HTMLElement) => {
+    const interval = setInterval(() => {
```
StephenCooper, line 319: Can we cancel this interval when it is no longer required?

StephenCooper (approved):

PR #12243, opened 2025-10-16
AG-16159 animate column widths when autosizing

eliasmalik:
Fix [AG-16159](https://ag-grid.atlassian.net/browse/AG-16159)

### DIFF
--- a/packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
+++ b/packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
@@ -3,7 +3,7 @@
     .ag-header-group-cell,
     .ag-cell {
         transition:
-            width 0.2s,
-            left 0.2s;
+            width 0.2s ease-in-out,
+            left 0.2s ease-in-out;
     }
 }
--- a/packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
+++ b/packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
@@ -1,0 +1,9 @@
+:where(.ag-animate-autosize) {
+    .ag-header-cell,
+    .ag-header-group-cell,
+    .ag-cell {
+        transition:
+            width 0.2s,
+            left 0.2s;
+    }
+}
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
```diff
@@ -0,0 +1,9 @@
+:where(.ag-animate-autosize) {
+    .ag-header-cell,
+    .ag-header-group-cell,
+    .ag-cell {
+        transition:
+            width 0.2s,
+            left 0.2s;
```
eliasmalik: @markdurrant seeking comments on animation parameters, lmk if you'd like to make any changes to them.

packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
```diff
@@ -596,3 +612,18 @@ function getAvailableWidth({ ctrlsSvc, scrollVisibleSvc }: BeanCollection): numb
...
 }
+
+const WIDTH_ANIMATION_CLASS = 'ag-animate-autosize';
+
+function setWidthAnimation({ ctrlsSvc, gos }: BeanCollection, enable: boolean): void {
+    if (gos.get('suppressColumnResizeAnimation') || gos.get('enableRtl')) {
+        return;
+    }
+
+    const classList = ctrlsSvc.getGridBodyCtrl().eGridBody.classList;
```
StephenCooper, line 623: Is there any chance of this being called via a timeout and there not being a grid body ctrl available?

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -442,6 +442,11 @@ export interface GridOptions<TData = any> {
      * @default false
      */
     suppressColumnMoveAnimation?: boolean;
+    /**
+     * If `true`, the `ag-animate-autosize` class is not added to the grid while autosizing columns. In the default themes, this results in no animation when autosizing column widths.
+     * @default false
+     */
+    suppressColumnResizeAnimation?: boolean;
```
StephenCooper, line 449: Looks like we are missing the auto generated changes for Angular and Vue components.

packages/ag-grid-community/src/columnAutosize/columnAutoSize.css
markdurrant: Have done some testing on local and the timing seems good. I would add a the 'ease-in-out' [timing-function](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transition-timing-function). But that might just be habit. 

```
transition:
    width 0.2s ease-in-out,
    left 0.2s ease-in-out;
```

StephenCooper (approved):

PR #12244, opened 2025-10-16
AG-15930 no lonely if

StephenCooper:


packages/ag-grid-enterprise/src/charts/chartComp/menu/format/axis/cartesianAxisPanel.ts
```diff
@@ -248,20 +248,18 @@ export class CartesianAxisPanel extends Component {
...
-                        { value: 'bottom', text: this.translate('bottom') },
-                    ];
-                } else if (axisType === 'yAxis') {
-                    // Vertical axis position can be changed between left and right
-                    return [
-                        { value: 'left', text: this.translate('left') },
-                        { value: 'right', text: this.translate('right') },
-                    ];
-                }
+            } else if (axisType === 'xAxis') {
```
SalvatorePreviti: Minor - since there is a return, the else is unnecessary. Don't remember if there is a eslint rule for that too

SalvatorePreviti (approved):

PR #12245, opened 2025-10-16
AG-15858 don't try to focus header if it's already focused

eliasmalik:
Fix AG-15858

### DIFF
--- a/packages/ag-grid-community/src/focusService.ts
+++ b/packages/ag-grid-community/src/focusService.ts
@@ -15,7 +15,7 @@ import type { FilterManager } from './filter/filterManager';
 import { _getDomData } from './gridOptionsUtils';
 import { DOM_DATA_KEY_HEADER_CTRL } from './headerRendering/cells/abstractCell/abstractHeaderCellCtrl';
 import type { HeaderCellCtrl } from './headerRendering/cells/column/headerCellCtrl';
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/focusService.ts
```diff
@@ -433,6 +433,10 @@ export class FocusService extends BeanStub implements NamedBean {
         const { column, headerRowIndex } = headerPosition;
         const { filterManager, ctrlsSvc, headerNavigation } = this.beans;
 
+        if (this.focusedHeader && isHeaderPositionEqual(params.headerPosition, this.focusedHeader)) {
```
peterjrreynolds, line 436: Are we sure that the header is actually focused if `this.focusedHeader` is set? E.g. if the header was focused, then got destroyed and recreated, does `focusedHeader` get cleared?

Might be safer to get the header cell and check if it contains the active element before returning `false`.

gportela85: @peterjrreynolds `this.focusedHeader` should be null when the header is not focused.

the GridHeader ctrl has the following logic onFocusOut

```
        if (!eGui.contains(relatedTarget as HTMLElement)) {
            beans.focusSvc.focusedHeader = null;
        }
```

gportela85 (approved):

peterjrreynolds (approved):

PR #12246, opened 2025-10-16
AG-16031 toggle expand aggregated values not empty

SalvatorePreviti:
- Add a unit test that reproduced the issue
- Worked out that dispatchStateUpdatedEvent should be executed before refreshCells
- Optimized / simplified the use of the events field and the debounce logic, correctly handling service disposal
- added a destructor to free memory

Fixes [AG-16031](https://ag-grid.atlassian.net/browse/AG-16031)


packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -135,43 +134,59 @@ export class ClientSideExpansionService
...
+     * if we did not do this, then the user could call setExpanded on 100+ rows, causing the grid
+     * to re-render 100+ times, which would be a performance lag.
+     *
+     * we use animationFrameService
+     * rather than debounce() so this will get done if anyone flushes the animationFrameService
+     * (eg user calls api.ensureRowVisible(), which in turn flushes ).
+     */
     protected override dispatchExpandedEvent(event: RowGroupOpenedEvent, forceSync?: boolean): void {
-        this.events.push(event);
+        (this.events ??= []).push(event);
```
sdwvit, line 151: just a nit: we don't have to create an array here, but no biggie

SalvatorePreviti: oh now we do as now is nullable, so we need to be sure is allocated when we push

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -135,43 +134,59 @@ export class ClientSideExpansionService
...
-            // values jump between group and footer, because the footer can be callback
-            // we refresh regardless as the output of the callback could be a moving target
-            const nodes = this.events.map((e) => e.node);
-            this.beans.rowRenderer.refreshCells({ rowNodes: nodes });
+        let dispatch = this.dispatchExpandedDebounced;
+        if (!dispatch) {
+            if (!this.isAlive()) {
+                return;
+            }
+            this.dispatchExpandedDebounced = dispatch = this.debounce(() => this.dispatchExpandedEvents());
```
sdwvit: let's please split this into two lines for readability

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -135,43 +134,59 @@ export class ClientSideExpansionService
...
-        if (forceSync) {
-            func();
-        } else {
-            if (this.dispatchExpandedDebounced == null) {
-                this.dispatchExpandedDebounced = this.debounce(func);
-            }
-            this.dispatchExpandedDebounced();
+        const rowNodes = new Array<IRowNode>(eventsLen);
+        for (let i = 0; i < eventsLen; ++i) {
+            rowNodes[i] = eventsToDispatch[i].node;
```
sdwvit, line 180: shallow concat() copy would do here?

sdwvit: ah nevermind, you have to dispatch too

sdwvit (approved):
I am not sure about the fix, but code looks fine apart from 1 styling change request

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
sdwvit: lets split into two lines as well please

PR #12250, opened 2025-10-16
AG-16015 - fixed async react issues while stopping edit

gportela85:
fix [AG-16015](https://ag-grid.atlassian.net/browse/AG-16015)

### DIFF
--- a/packages/ag-grid-community/src/edit/utils/editors.ts
+++ b/packages/ag-grid-community/src/edit/utils/editors.ts
@@ -421,9 +421,9 @@ export function _destroyEditor(
         return;
     }
 
-    const { comp } = cellCtrl;
+    const { comp, isStoppingEditor } = cellCtrl;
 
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/edit/utils/editors.ts
```diff
@@ -421,9 +421,9 @@ export function _destroyEditor(
         return;
     }
 
-    const { comp } = cellCtrl;
+    const { comp, isStoppingEditor } = cellCtrl;
 
-    if (comp && !comp.getCellEditor()) {
+    if (comp && (!comp.getCellEditor() || isStoppingEditor)) {
```
StephenCooper, line 426: Can we move this into the getCellEditor call inside of React so that the when the isStoppingEditor is set it returns undefined. 
This should be set synchronously in the `comp?.setEditDetails();` when its undefined and then can we not rely on a setTimeout?

PR #12333, opened 2025-10-25
AG-8751 - fixed style and tooltip issues with formulas

gportela85:
fix AG-8751


packages/ag-grid-community/src/tooltip/tooltipService.ts
```diff
@@ -272,7 +272,7 @@ export class TooltipService extends BeanStub implements NamedBean {
                 const error = formula?.getFormulaError(cellCtrl.column, cellCtrl.rowNode);
                 return error ? error.message : undefined;
             },
-            getLocation: () => 'cell',
+            getLocation: () => 'cellFormula',
```
SalvatorePreviti, line 275: are you sure this is correct? also when there are no formulas? before it was "cell" now is "cellFormula" only

SalvatorePreviti (approved):

PR #12346, opened 2025-10-28
[AG-13490] Feature(row-sorting): allow absolute sort as a built in column sort option

sdwvit:
- Updated sortingOrder to include 'aasc' and 'adesc' options.
 - Modified _defaultComparator to accept transformation options.
 - Implemented absoluteValueTransformer for sorting.
 - Adjusted compareRowNodes to handle new sorting options.

plnkr: https://plnkr.co/edit/xfKyoULgjs34KTTo?open=main.js&preview 

Fix: [AG-13490](https://ag-grid.atlassian.net/browse/AG-13490)


packages/ag-grid-community/src/columns/columnStateUtils.ts
```diff
@@ -29,7 +29,7 @@ export interface ColumnStateParams {
     /** Column's flex if flex is set */
     flex?: number | null;
     /** Sort applied to the column */
-    sort?: 'asc' | 'desc' | null;
+    sort?: 'asc' | 'desc' | 'aasc' | 'adesc' | null;
```
eliasmalik: Can we replace this with `SortDirection`?

packages/ag-grid-community/src/agStack/utils/generic.ts
```diff
@@ -32,7 +32,13 @@ export function _jsonEquals<T1, T2>(val1: T1, val2: T2): boolean {
     return val1Json === val2Json;
 }
 
-export function _defaultComparator(valueA: any, valueB: any, accentedCompare: boolean = false): number {
+export function _defaultComparator(
+    rawValueA: any,
+    rawValueB: any,
+    options: { accentedCompare?: boolean; transform?: (val: any) => any } = {}
+): number {
+    let valueA = options.transform ? options.transform(rawValueA) : rawValueA;
```
eliasmalik, line 40: Could we have `let valueA = options.transform?.(rawValueA) ?? rawValueA` or are we also anticipating `transform` returning nullish values?

packages/ag-grid-community/src/agStack/utils/aria.ts
```diff
@@ -31,13 +31,17 @@ export function _setAriaRole(element: Element, role?: string | null) {
     }
 }
 
-export function _getAriaSortState(sortDirection: 'asc' | 'desc' | 'mixed' | null): AriaSortState {
+export function _getAriaSortState(sortDirection: 'asc' | 'desc' | 'aasc' | 'adesc' | 'mixed' | null): AriaSortState {
```
eliasmalik: Can we use `SortDirection`?

community-modules/locale/src/ar-EG.ts
```diff
@@ -224,6 +224,8 @@ export const AG_GRID_LOCALE_EG = {
     chooseColumns: 'اختيار الأعمدة',
     sortAscending: 'ترتيب تصاعدي',
     sortDescending: 'ترتيب تنازلي',
+    sortAbsoluteAscending: 'ترتيب تصاعدي مطلق',
```
eliasmalik: I would maybe get James to double check these translations since it's normally his job to update them.

packages/ag-grid-community/src/interfaces/iSortModelItem.ts
```diff
@@ -2,5 +2,5 @@ export interface SortModelItem {
     /** Column Id to apply the sort to. */
     colId: string;
     /** Sort direction */
-    sort: 'asc' | 'desc';
+    sort: 'asc' | 'desc' | 'aasc' | 'adesc';
```
eliasmalik: can we use `NonNullable<SortDirection>`?

packages/ag-grid-community/src/interfaces/iSortOption.ts
```diff
@@ -1,6 +1,6 @@
 import type { Column } from './iColumn';
 
 export interface SortOption {
-    sort: 'asc' | 'desc';
+    sort: 'asc' | 'desc' | 'aasc' | 'adesc';
```
eliasmalik: Here too

packages/ag-grid-community/src/sort/sortService.ts
```diff
@@ -128,7 +128,7 @@ export class SortService extends BeanStub implements NamedBean {
                 }
 
                 // setting to 'undefined' as null means 'none' rather than cleared, otherwise issue will arise
-                // if sort order is: ['desc', null , 'asc'], as it will start at null rather than 'desc'.
+                // if sort order is: ['desc', 'aasc', 'adesc', null , 'asc'], as it will start at null rather than 'desc'. // todo not sure about this comment
```
eliasmalik: can you clarify what you're not sure about?

packages/ag-grid-community/src/agStack/utils/generic.ts
sdwvit: It can return nullish, sure.I also don't think we are going to expose it ever.

packages/ag-grid-community/src/sort/sortService.ts
sdwvit: not sure about the order of sort types

community-modules/locale/src/ar-EG.ts
sdwvit: messaged him, thank you

packages/ag-grid-community/src/interfaces/iSortModelItem.ts
sdwvit: So there is a confusion between sort types and sort direction. Sort types are 'default' and 'absolute'. While sort directions are 'ascending', 'descending'. 

We would need to clelarly differentiate between two.

PR #12347, opened 2025-10-28
AG-1792 Select cells in columns by clicking headers

eliasmalik:
Fix [AG-1792](https://ag-grid.atlassian.net/browse/AG-1792)

### DIFF
--- a/packages/ag-grid-community/src/headerRendering/cells/column/headerCellMouseListenerFeature.ts
+++ b/packages/ag-grid-community/src/headerRendering/cells/column/headerCellMouseListenerFeature.ts
@@ -16,10 +16,12 @@ export class HeaderCellMouseListenerFeature extends BeanStub {
         this.addManagedElementListeners(this.eGui, {
             click: (e) => e && this.onClick(e),
         });
-    }
 
-    public onMovingChanged(): void {
-        this.lastMovingChanged = Date.now();
+        this.addManagedListeners(this.column, {
+            movingChanged: () => {
+                this.lastMovingChanged = Date.now();
+            },
+        });
     }
 
     public onClick(event: MouseEvent): void {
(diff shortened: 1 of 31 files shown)


packages/ag-grid-community/src/headerRendering/cells/column/headerCellMouseListenerFeature.ts
```diff
@@ -0,0 +1,55 @@
+import { BeanStub } from '../../../context/beanStub';
+import type { AgColumn } from '../../../entities/agColumn';
+import { _getSuppressColumnSelection } from '../../../gridOptionsUtils';
+
+export class HeaderCellMouseListenerFeature extends BeanStub {
```
StephenCooper, line 5: Should these be part of the SelectionModule? i.e that is how they are created. Then these will not be part of the bundle of the core?

packages/ag-grid-community/src/headerRendering/cells/columnGroup/headerGroupComp.ts
```diff
@@ -242,6 +240,11 @@ export class HeaderGroupComp extends Component implements IHeaderGroupComp {
         }
 
         this.toggleCss('ag-sticky-label', !columnGroup.getColGroupDef()?.suppressStickyLabel);
+        this.toggleCss('ag-header-group-cell-selectable', !_getSuppressColumnSelection(this.gos));
+
+        this.mouseListener ??= this.createManagedBean(
```
StephenCooper: Can we do this via the selectionService? So that it can be treeshaken.

packages/ag-grid-community/src/headerRendering/cells/column/headerCellMouseListenerFeature.ts
eliasmalik: Moved `HeaderGroupCellMouseListenerFeature` into the cell selection module but this one should stay because it's concerned with sorting as well as cell selection (and sorting is in community core).

gportela85 (approved):

PR #12361, opened 2025-10-29
AG-16211 fill handle not working after cancel edit

SalvatorePreviti:
fixes:
- change AbstractSelectionHandle so it does not register anymore mouseMove events, but listen just to dragging events, this works fine and do not break when order of event registration or state changes
- properly register stopPropagation also on pointer down

improvements:

- remove onDragStart in AbstractSelectionHandle as is not needed
- remove the cast to MouseEvent, as events can be also Touch
- avoid the cast to range service in AbstractSelectionHandle by adding the autoScroll field in the interface
- makes onDragStart, onDragStop, onDragCancel, onDragging all optional in DragListenerParams

Fix AG-16211


packages/ag-grid-enterprise/src/rangeSelection/abstractSelectionHandle.ts
```diff
@@ -23,38 +23,37 @@ export abstract class AbstractSelectionHandle extends Component {
...
             dragStartPixels: 0,
             eElement: this.getGui(),
-            onDragStart: this.onDragStart.bind(this),
-            onDragging: (e: MouseEvent | Touch) => {
-                this.dragging = true;
-                (rangeSvc as RangeService).autoScrollService.check(e as MouseEvent);
+            onDragging: (e) => {
+                if (!this.dragging) {
+                    this.dragging = true;
+                    document.body.classList.add(this.getDraggingCssClass());
```
gportela85: can you update this to use `_getPageBody` instead?

packages/ag-grid-enterprise/src/rangeSelection/abstractSelectionHandle.ts
```diff
@@ -112,6 +102,10 @@ export abstract class AbstractSelectionHandle extends Component {
         // TODO: this causes a bug where if there are multiple grids in the same page, all of them will
         // be affected by a drag on any. Move it to the root element.
         document.body.classList.remove(this.getDraggingCssClass());
```
gportela85: can you update this to use `_getPageBody` instead?

gportela85 (approved):

johannes-z:
@SalvatorePreviti double click on the fill handle is still not working, after the cell editor is closed. the cell currently needs to be re-focused manually to make the fill handle work again.
Drag works as expected though.

PR #12367, opened 2025-10-31
Fix Charts integration build.

alantreadway:
- **Add warning about build script dependency to iAgChartsExports.ts and agChartsExports.ts**
- **Fix iAgChartsExports.ts + agChartsExports.ts patching for Charts->Grid integration build.**


taktran (approved):
LGTM

PR #12371, opened 2025-10-31
AG-3390 - Remove trap caching in codeql

taktran:
Each entry is about 500mb

From https://github.com/ag-grid/ag-charts/pull/5486


alantreadway (approved):
LGTM

PR #12372, opened 2025-10-31
AG-16224 Fix theme loss on react className prop change

BernieSumption:
fix AG-16224

### DIFF
--- a/packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
+++ b/packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
@@ -105,10 +105,14 @@ export const AgGridReactUi = <TData,>(props: InternalAgGridReactProps<TData>) =>
         const splitClasses = (s = '') => s.trim().split(/\s+/g).filter(Boolean);
         if (appliedClassName.current !== classNameFromReact) {
             for (const cls of splitClasses(appliedClassName.current)) {
-                classList?.remove(cls);
+                if (classList?.contains(cls)) {
+                    classList.remove(cls);
+                }
             }
             for (const cls of splitClasses(classNameFromReact)) {
-                classList?.add(cls);
+                if (!classList?.contains(cls)) {
+                    classList?.add(cls);
+                }
             }
             appliedClassName.current = classNameFromReact;
         }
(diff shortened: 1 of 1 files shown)


packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
```diff
@@ -96,8 +96,31 @@ export const AgGridReactUi = <TData,>(props: InternalAgGridReactProps<TData>) =>
     // Hook to enable Portals to be displayed via the PortalManager
     const [, setPortalRefresher] = useState(0);
 
+    const appliedClassName = useRef<string>();
```
StephenCooper, line 99: Are we able to use the `CssClassManager` for this. You can see an example of this in the `rowComp.tsx`. 

However, that may require more changes depending on where el.classList is set outside of this component.

BernieSumption: I had a go, but CssClassManager keeps internal state tracking applied classes, so every bit of code changing classes needs to use the same manager, and the Theming API doesn't have access. Sharing the instance would require either architectural changes to enable communication between Theming API and AgGridReactUi, or something hacky like storing a reference on the DOM element.

Instead I went for:

```ts
if (classList?.contains(cls)) {
    classList.remove(cls);
}
...
if (!classList?.contains(cls)) {
    classList?.add(cls);
}
```

... which should provide essentially the same benefit of not triggering a class name change in cases where a class isn't being added or removed.

CssClassManager might do it slightly more efficiently by avoiding touching DOM APIs, but that's a micro-optimization for a very hot loop which doesn't really apply here.

packages/ag-grid-react/src/reactUi/agGridReactUi.tsx
```diff
@@ -96,8 +96,35 @@ export const AgGridReactUi = <TData,>(props: InternalAgGridReactProps<TData>) =>
...
+            for (const cls of splitClasses(classNameFromReact)) {
+                if (!classList?.contains(cls)) {
+                    classList?.add(cls);
+                }
+            }
+            appliedClassName.current = classNameFromReact;
+        }
+    };
+
+    useEffect(() => {
```
StephenCooper, line 121: It might be more suitable to use `useLayoutEffect` here to ensure the class gets applied sooner.

StephenCooper (approved):

PR #12428, opened 2025-11-06
AG-11935 - allow charting non group column when row grouping is enabled

rmc-software:
Fix AG-11935

### DIFF
--- a/packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
+++ b/packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
@@ -14,7 +14,7 @@ import type {
     SortModelItem,
     SortOption,
 } from 'ag-grid-community';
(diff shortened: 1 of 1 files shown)


packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
```diff
@@ -14,7 +14,7 @@ import type {
     SortModelItem,
     SortOption,
 } from 'ag-grid-community';
-import { BeanStub, CellRangeType, _isTreeData } from 'ag-grid-community';
+import { BeanStub, CellRangeType, _isTreeData, isColumnGroupAutoCol } from 'ag-grid-community';
```
peterjrreynolds:
```suggestion
import { BeanStub, CellRangeType, _isTreeData } from 'ag-grid-community';
```

Failing lint

peterjrreynolds (approved):

PR #12454, opened 2025-11-11
AG-8754 - improved formulas docs

gportela85:
fix AG-8754


documentation/ag-grid-docs/src/content/docs/formulas/index.mdoc
```diff
@@ -21,14 +21,15 @@ const gridOptions = {
 
 {% gridExampleRunner title="Formulas" name="formulas" exampleHeight=450 /%}
 
-When using formulas, the grid enables the following features by default:
-- Row Numbers
+When using formulas, the grid enables (and cannot be disabled) the following features by default:
```
eliasmalik:
```suggestion
When using formulas, the following features are enabled by default, and cannot be disabled:
```

eliasmalik (approved):

PR #12457, opened 2025-11-11
AG-16070 formulas filtering test + fix sorting

SalvatorePreviti:
- fix sorting algorithm for formulas, we need to resolve the value when sorting
- add unit tests for formulas
- fix the test framework to support row numbers and some extra rendering options


testing/behavioural/src/formulas/formulas-filtering.test.ts
```diff
@@ -0,0 +1,746 @@
+import type { GridOptions } from 'ag-grid-community';
+import { ClientSideRowModelModule, TextEditorModule, TextFilterModule, TooltipModule } from 'ag-grid-community';
+import { CellSelectionModule, FormulaModule, SetFilterModule } from 'ag-grid-enterprise';
+import type { SetFilter } from 'ag-grid-enterprise';
+
+import type { GridRowsOptions } from '../test-utils';
+import { GridRows, TestGridsManager, waitForEvent } from '../test-utils';
+
+describe('ag-grid master detail', () => {
```
StephenCooper: Looks like a copy paste description.

StephenCooper (approved):

PR #12458, opened 2025-11-11
AG-16140 Update chart axes to dictionary

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16140


packages/ag-grid-enterprise/src/charts/chartComp/services/chartOptionsService.ts
peterjrreynolds, line 1: The following methods also need updating as they all refer to axes as an array in either chart options or directly on the chart instance:
- `getAxisProperty`
- `setAxisThemeOverrides`
- `setCartesianAxisOptions`
- `setCartesianCategoryAxisType`

peterjrreynolds (commented):
Will also need to update `upgradeChartModel` to migrate saved chart models

peterjrreynolds:
> Will also need to update `upgradeChartModel` to migrate saved chart models

Looks like we probably won't need this bit. Will add to testing steps.

packages/ag-grid-enterprise/src/charts/chartComp/services/chartOptionsService.ts
lsjroberts: I've updated `getAxisProperty`. I think the others are ok as they either take the result of `getChartAxes()` which still returns an array, or use `Object.values(axes)` directly.

lsjroberts: Updated the others as noted, thanks!

peterjrreynolds (approved):

PR #12489, opened 2025-11-13
AG-16275 - Move calc matrix out of init workflow

taktran:
Not needed for docs build


.github/workflows/ci.yml
```diff
@@ -60,15 +60,37 @@ jobs:
...
+
+      - name: Setup
+        id: setup
+        uses: ./.github/actions/setup-nx
+        with:
+          cache_mode: 'rw'
+
+  calc_matrix:
+    name: Calculate Github Matrix
+    needs: init
```
alantreadway: Is this dependency needed, can this new job run in parallel?

taktran: Tried this, but we need `nx` and nx modules from node_modules for the calculation scripts

alantreadway (approved):
LGTM

PR #12491, opened 2025-11-13
AG-16320 only progress sort when sortable

eliasmalik:
Fix [AG-16320](https://ag-grid.atlassian.net/browse/AG-16320)


StephenCooper (approved):
I like the additional tests :)

PR #12495, opened 2025-11-14
AG-8749 allow column selection via single-click when editing formulas

eliasmalik:
For reference


eliasmalik:
This is no longer needed since the interaction model was changed and single-click is now default.

PR #12497, opened 2025-11-14
[AG-16324] Fix(rich-select): broken example due to new agrichselect changes

sdwvit:
Add isInitial flag to setValueList and adjust conditionals

 - Add isInitial parameter to setValueList call for initial value setup
 - Move listComponent check below isInitial condition to ensure proper flow
 - Update list component even when values are undefined in updateValueList

fix [AG-16324](https://ag-grid.atlassian.net/browse/AG-16324)

### DIFF
--- a/packages/ag-grid-enterprise/webpackBuild.sh
+++ b/packages/ag-grid-enterprise/webpackBuild.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 array=("production" "staging")
-NX_TASK_TARGET_CONFIGURATION=${NX_TASK_TARGET_CONFIGURATION:-development}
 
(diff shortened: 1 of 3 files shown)


packages/ag-grid-enterprise/webpackBuild.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 array=("production" "staging")
+NX_TASK_TARGET_CONFIGURATION=${NX_TASK_TARGET_CONFIGURATION:-development}
```
StephenCooper: Seems unrelated?

StephenCooper (approved):

packages/ag-grid-enterprise/webpackBuild.sh
sdwvit: removed

PR #12498, opened 2025-11-14
[AG-15375] Chore(webpack): enable dev build linux

sdwvit:
Set default NX_TASK_TARGET_CONFIGURATION in webpackBuild.sh

 - Update ag-grid-community webpackBuild.sh with default value
 - Update ag-grid-enterprise webpackBuild.sh with default value

related to [AG-15375](https://ag-grid.atlassian.net/browse/AG-15375)


sdwvit:
ok I can't confirm this works when running nx dev. something overrides the value or my bash-foo is not good

PR #12500, opened 2025-11-14
AG-6477 add validation for in-range filters

eliasmalik:
Fix [AG-6477](https://ag-grid.atlassian.net/browse/AG-6477)

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
+++ b/packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
@@ -90,8 +90,9 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
 
         const fromDate = from.getDate();
         const toDate = to.getDate();
-
-        (isFrom ? from : to).setCustomValidity(getValidityMessage(fromDate, toDate, isFrom));
+        const localeKey = getValidityMessageKey(fromDate, toDate, isFrom);
+        const message = localeKey ? this.translate(localeKey, [String(isFrom ? toDate : fromDate)]) : '';
+        (isFrom ? from : to).setCustomValidity(message);
     }
 
     private createDateCompWrapper(element: HTMLElement, position: number, fromTo: 'from' | 'to'): DateCompWrapper {
@@ -264,21 +265,25 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
         return result;
     }
 
-    protected override translate(key: FilterLocaleTextKey): string {
+    protected override translate(key: FilterLocaleTextKey, variableValues?: string[]): string {
+        let normalisedKey = key;
         if (key === 'lessThan') {
-            return super.translate('before');
+            normalisedKey = 'before';
+        } else if (key === 'greaterThan') {
+            normalisedKey = 'after';
         }
-        if (key === 'greaterThan') {
-            return super.translate('after');
-        }
-        return super.translate(key);
+        return super.translate(normalisedKey, variableValues);
     }
 }
 
-function getValidityMessage(fromDate: Date | null, toDate: Date | null, isFrom: boolean): string {
+function getValidityMessageKey(
+    fromDate: Date | null,
+    toDate: Date | null,
+    isFrom: boolean
+): FilterLocaleTextKey | null {
     const isInvalid = fromDate != null && toDate != null && fromDate > toDate;
     if (!isInvalid) {
-        return '';
+        return null;
     }
-    return `Please select a date that is ${isFrom ? 'after' : 'before'} ${isFrom ? toDate : fromDate}`;
+    return isFrom ? 'tooEarly' : 'tooLate';
 }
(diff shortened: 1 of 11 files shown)


packages/ag-grid-community/src/filter/provided/providedFilter.ts
```diff
@@ -259,6 +259,11 @@ export abstract class ProvidedFilter<
             });
         }
 
+        // Don't apply an invalid model
+        if (!state.valid) {
```
eliasmalik, line 263: This is technically wider in scope than asked for by the JIRA. However, I can't find any actual implementations of `canApply` (other than the one I added here), so I don't think this actually changes any behaviour. Do we want this as general behaviour?

packages/ag-grid-community/src/filter/provided/simpleFilter.ts
```diff
@@ -714,6 +714,10 @@ export abstract class SimpleFilter<
             return false;
         }
 
+        if (this.hasInvalidInputs()) {
```
eliasmalik, line 712: Again, wider in scope than asked for by the JIRA, and I can't find any implementations of hasInvalidInputs except for in the number filter, so I don't think this actually changes any behaviour. Do we want this as general behaviour?

packages/ag-grid-community/src/filter/provided/number/numberFilter.ts
```diff
@@ -44,11 +44,27 @@ export class NumberFilter extends SimpleFilter<
...
+                isFrom: boolean
+            ) =>
+            () => {
+                const fromValue = getNormalisedValue(parser, from);
+                const toValue = getNormalisedValue(parser, to);
+                (isFrom ? from : to).setCustomValidity(getValidityMessage(fromValue, toValue, isFrom));
+            };
+
+        from.addEventListener('fieldValueChanged', getFieldChangedListener(from, to, true));
+        to.addEventListener('fieldValueChanged', getFieldChangedListener(from, to, false));
```
peterjrreynolds: Need to remove the listeners.

Can do `from.addManagedListeners(from, { fieldValueChanged: ... })`

packages/ag-grid-community/src/filter/provided/number/numberFilter.ts
```diff
@@ -141,10 +139,48 @@ export class NumberFilter extends SimpleFilter<
...
+): number | null {
+    return processNumberFilterValue(stringToFloat(numberParser, input.getValue()));
+}
+
+function getValidityMessage(fromValue: number | null, toValue: number | null, isFrom: boolean): string {
+    const isInvalid = fromValue != null && toValue != null && fromValue > toValue;
+    if (!isInvalid) {
+        return '';
+    }
+    return `Please select a value that is no ${isFrom ? 'more' : 'less'} than ${isFrom ? toValue : fromValue}`;
```
peterjrreynolds: Needs to be localised

packages/ag-grid-community/src/filter/provided/providedFilter.ts
peterjrreynolds: The set filter implements `canApply`. The primary purpose of the valid state was to disable the apply button for the set filter when in excel mode.

The code below wouldn't run in this situation though, so there won't be any impact of doing this.

packages/ag-grid-community/src/filter/provided/simpleFilter.ts
peterjrreynolds: Looks like I removed the bit that originally called it when implementing filter handlers, as it wasn't required any more.

I think this is fine.

packages/ag-grid-community/src/filter/provided/number/numberFilter.ts
```diff
@@ -44,11 +46,35 @@ export class NumberFilter extends SimpleFilter<
...
+                const validityMessage = filterLocaleKey
+                    ? translateForFilter(this, filterLocaleKey, [String(isFrom ? toValue : fromValue)])
+                    : '';
+                (isFrom ? from : to).setCustomValidity(validityMessage);
+            };
+
+        this.addManagedListeners(from, {
+            fieldValueChanged: getFieldChangedListener(from, to, true),
+        });
+        this.addManagedListeners(to, {
```
peterjrreynolds: Add the listeners to the inputs (e.g. `to.addManagedListeners(to`).

Otherwise the list of destroy functions can get very long. The filter can be long lived, but the inputs can get regenerated each time it's opened.

packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -260,3 +274,11 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
         return super.translate(key);
     }
 }
+
+function getValidityMessage(fromDate: Date | null, toDate: Date | null, isFrom: boolean): string {
```
peterjrreynolds: Can remove this now

eliasmalik: not sure I follow, why is this no longer necessary?

peterjrreynolds: When I left the comment it was still returning the text rather than the key. Can ignore now

peterjrreynolds (approved):

PR #12511, opened 2025-11-17
AG-16275 - Add yarn cache to node setup

taktran:
Caches yarn deps in `node-setup` step.


taktran:
Closing this, as a 2.6GB cache entry for yarn modules doesn't seem like it's worth the gain + may lead to future devops maintenance issues.

PR #12514, opened 2025-11-18
AG-8761 - [formulas] fixed issues with batch edit

gportela85:
fix AG-8761 
fix AG-8751

### DIFF
--- a/packages/ag-grid-enterprise/src/formula/functions/utils.ts
+++ b/packages/ag-grid-enterprise/src/formula/functions/utils.ts
@@ -22,6 +22,10 @@ export function take<T>(values: Iterable<T>, name: string, n: number): T[] {
     return out;
 }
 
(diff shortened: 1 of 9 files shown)


packages/ag-grid-enterprise/src/formula/functions/utils.ts
```diff
@@ -22,6 +22,10 @@ export function take<T>(values: Iterable<T>, name: string, n: number): T[] {
     return out;
 }
 
+export function isFormula(value: unknown): value is `=${string}` {
```
StephenCooper: Should we move this to community so that we can use it consistently?

gportela85: community should be using the `formulaService.isFormula`

StephenCooper (approved):

PR #12516, opened 2025-11-19
AG-16208   [Regression] Changelog shows DEPRECATION NOTES and BREAKING CHANGE values as [object Object]

seanlandsman:


taktran (approved):
lgtm

PR #12525, opened 2025-11-19
AG-8818 group data not populated

SalvatorePreviti:
The computation of groupData and groupValue and the relation to sibling was not always updated correctly, was inconsistent.
It was a systematic problem that this PR fixes.

This pull request introduces significant improvements and refactoring to the row grouping logic and related APIs in AG Grid. The changes focus on better management and lazy loading of group data for row nodes, improved event handling, and streamlining the internal representation of group columns. Additionally, several APIs and internal methods have been renamed or updated to clarify their purpose and usage.

**Row Grouping & Group Data Management:**
* The `RowNode` class now lazily loads `groupData` via the GroupStage, with a new backing field `_groupData` and getter/setter logic. This enables more efficient group data retrieval and ensures footers reference their main row's group data. 
* The `IRowGroupStage` interface renames `getNode` to `getNonLeaf`, adds a `loadGroupData` method for lazy group data computation, and clarifies the contract for group node retrieval.
* The `GroupStrategy` class now maintains a `nonLeafsById` map for efficient group node lookup, implements lazy group data creation with `newGroupData`, and refactors group data update logic to remove unnecessary checks and optimize performance.
* Several fields in `RowNode` are now initialized with default values to prevent undefined behavior.
* The set of ignored properties for row node sibling copying is updated to include `_groupData` and `groupValue`. 
* Simplifies the way we generate IDs for groups in Grouping, as we don't need a loop at all but just the parent id
* The `IShowRowGroupColsService` interface and its consumers now use a direct `columns` property instead of the method `getShowRowGroupCols`, simplifying access to group columns throughout the codebase. 
* A new internal event `showRowGroupColsSetChanged` is added for tracking changes in the set of group columns, so -_groupData is always invalidated if the set of displayed row group columns changes
* The `ColumnModel` now uses a `changeEventsDispatching` flag to manage event dispatching during column refresh, improving reliability and reducing redundant events. 
* Updated ignored sibling properties to include `_groupData` and `groupValue` for proper row node copying.
* Initialized additional fields in `RowNode` to default values to prevent undefined behavior.
* Refactor the ValueService getValue method, by calling another private resolveValue method to avoid accessing the groupData property until really needed so, for performance as is a computed cached property now

This result in a simpler code overall and less code. Most of the changes and added code are in tests.

**Test changes:**
* A lot of regression snapshot testing were failing because they were using the wrong groupData! Refreshed all the snapshot for the combinatorial row group column rendering
* Added more tests, and improved the testing framework to test more assertion, and make it work better with SSRM

**Changes since last review**
* Improved loadGroupData by setting _groupData in the strategies, and null if no strategy is available
* ensured that shotgunResetEverything is not called if there are no changes in the row group columns
* optimized group creation and moveNodeInWrongPath in such a way we do not allocate new temporary arrays
* optimized insertOneNode so we do not create new temporary objects

Fix AG-8818


testing/behavioural/src/group-cell-renderer/grouping/exported/csrm/__snapshots__/exported.test.ts.snap
```diff
@@ -1340,7 +1340,7 @@ exports[`ag-grid exported group values > with rowModelType=serverSide > groupDis
 "Italy","Total 2001"
 "Total Italy",""
 "(Blanks)","2000"
-"(Blanks)","Total 2000"
+"","Total 2000"
```
StephenCooper, line 1343: Is this actually a correct change?

SalvatorePreviti: Yes it is

StephenCooper (approved):

Issue #12035, opened 2025-09-12
[Bug]: Theme CSS injection rearranges CSS layers by default


aaronjordan:
### Link to reproducible scenario

https://codesandbox.io/p/sandbox/wvdxjc

### Describe the bug

Steps to reproduce 
1. Create a grid using the `theme` and `themeCssLayer="grid"` options. (tested JS and React)
2. Define a `@layer` order where `grid` is not first in the list.
    - My team ran into this when adding `grid` after `base` with [Tailwind Preflight](https://tailwindcss.com/docs/preflight#overview).

Actual behaviour
- The AG Grid theme will always be applied with the lowest specificity of _all_ expressed `@layer`s.

Expected behaviour
- The `themeCssLayer` option should not change the layer order expressed by the application's stylesheets.

More information

Chrome layers inspector at codesandbox link:
<img width="167" height="109" alt="Image" src="https://github.com/user-attachments/assets/2db22a42-7eff-40f3-84b2-47d6544d828a" />

It is possible to work around this bug by specifying a [themeStyleContainer](https://www.ag-grid.com/javascript-data-grid/grid-options/#reference-theme-themeStyleContainer) that appears later than the `@layer` order in the document. It's the declaration of the `@layer grid` by the inserted theme styles at the beginning of `head` that lowers the specificity here.

I'm sure that inserting as early as possible could make sense when `themeCssLayer` is **not** active, as user styles _should_ override the AG Grid theme. However, this reordering issue makes me wonder if appending at the **end** of the document's head by default would be more appropriate when `themeCssLayer` is active.

I'd be happy to draft an idea for a fix, please let me know if this strategy makes sense or if there's another approach we could take. Thanks!

### Version

34.1.1

### Does the issue occur for a specific framework only?

All languages

### Is the issue only observable on a specific browser?

All browsers

ognjenmarcheta:
I can confirm this is an real bug, we also have this issue where we have combination of tailwind and ag-grid theme api. Setting custom layer order just doesn't work.

For example: @layer base, ag-grid, components, utilities;
Becomes:  ag-grid, properties, base, components, utilities, theme;


<img width="418" height="44" alt="Image" src="https://github.com/user-attachments/assets/9105e9b6-f272-416e-a817-4fa27165d563" />

<img width="205" height="185" alt="Image" src="https://github.com/user-attachments/assets/5204362f-c8bb-464f-a2a4-08d831af28b9" />

[...]

PR #12531, opened 2025-11-20
Properly remove outdated styles from DOM on theme change

tehpsalmist:
By querying from `document.head`, outdated injected styles are left behind in the DOM when using a `themeStyleContainer` that is outside of `document.head` (like `document.body`, for example). The impetus to use such selectors could come from a desire to avoid layer collisions, such as were run into in #12035 . This proposed change simply widens the search to the entire document, ensuring that the stylesheets are found. Alternatively, the query could be made from `environment.eStyleContainer.querySelectorAll(...)`, which defaults to `document.head`, but if for some reason that container identifier changed, the query would fail (weird hypothetical...but I do like just using `document` regardless).

### DIFF
--- a/packages/ag-grid-community/src/agStack/theming/inject.ts
+++ b/packages/ag-grid-community/src/agStack/theming/inject.ts
@@ -84,7 +84,7 @@ export const _unregisterInstanceUsingThemingAPI = (environment: IEnvironment) =>
     injectionState.grids.delete(environment);
     if (injectionState.grids.size === 0) {
         injectionState.map = new WeakMap();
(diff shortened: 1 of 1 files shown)


tehpsalmist:
Successfully tested this fix locally, as well.

BernieSumption:
Hi @tehpsalmist thanks for the PR.

In the next release we're planning to rework the style injection and cleanup code to further improve performance. Because of this I'm going to close this PR in favour of the new work, but I have updated our internal document with this issue you've pointed out and will ensure that the new implementation handles this case correctly.

PR #12535, opened 2025-11-22
AG-16249 - [Formulas] - Excel Exporter integration

gportela85:
fix AG-16249


packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -89,16 +89,18 @@ export class ValueService extends BeanStub implements NamedBean {
      * The values from this function are not used for sorting, filtering, or aggregation purposes.
      * Handles: groupHideOpenParents, showOpenedGroup and groupSuppressBlankHeader behaviours
```
StephenCooper, line 90: Does this comment need updating if we are using it for formulas now?

StephenCooper (approved):

PR #12565, opened 2025-11-27
AG-13321-removed-rownode-flag

SalvatorePreviti:
- Add a new `destroyed` flag to a node so it is easy to see if a node was destroyed. A destroyed node cannot be reused.
- Updates row models and grouping and tree data strategy to set this flag to true
- Optimize NodeManager and group and tree strategy using this new flag instead of the set of removed nodes
- Fix row dragging to understand correctly if a leaf node was removed while dragging

** Testing **
- Add new setRowDataChecked and expectRowNodesDestroyed and applyTransactionChecked and patch executeTransactionsAsync to execute assertions around destroyed nodes


packages/ag-grid-enterprise/src/aggregation/footerUtils.ts
```diff
@@ -28,12 +28,12 @@ export function _createRowNodeFooter(rowNode: RowNode, beans: BeanCollection): v
...
+    if (!sibling) {
         return;
     }
 
-    rowNode.sibling.setRowTop(null);
-    rowNode.sibling.setRowIndex(null);
-
+    sibling._destroy(false);
     rowNode.sibling = undefined as any;
+    sibling.sibling = undefined as any;
```
StephenCooper, line 38: Extra level of cleaing?

StephenCooper (approved):

PR #12568, opened 2025-11-28
Formula Cell Editor

gportela85:


packages/ag-grid-community/src/theming/parts/formula-style/formula-styles.ts
```diff
@@ -0,0 +1,82 @@
+import type { Part } from '../../../agStack/theming/part';
+import { createPart } from '../../../agStack/theming/partImpl';
+import type { ColorValue } from '../../../agStack/theming/themeTypes';
+
+type FormulaStyleParams = {
+    /**
+     * The color of the 1st formula field token
+     */
+    'formulaToken-1Color': ColorValue;
```
BernieSumption: These names would be better as pure camel case `formulaToken1Color`. I guess they're like this to ensure a hyphen before the number in the CSS variable name.

You could achieve this by updating the kebabCase function in themeUtils.ts as follows:

```
const kebabCase = (str: string) => str.replace(/[A-Z]|\d+/g, (m) => `-${m}`).toLowerCase();
```

This function is duplicated in the Theme Builder, theme-builder/model/utils.ts. Make sure you update it there too.

packages/ag-grid-enterprise/src/widgets/agFormulaInputField.css
```diff
@@ -0,0 +1,101 @@
+.ag-formula-token {
+    padding: 0 0.25rem;
+    margin: 0 0.25rem;
+    border-radius: var(--ag-border-radius, 2px);
+    background-color: var(--ag-selected-row-background-color, rgb(64 139 251 / 12%));
+    border: 1px solid var(--ag-border-color, #c6c6c6);
```
BernieSumption: We now have an --ag-border-width variable to set the default width of all border, use this instead of 1px. This is mainly FYI because see comment below about borders which will invalidate this one :)

packages/ag-grid-enterprise/src/widgets/agFormulaInputField.css
```diff
@@ -0,0 +1,101 @@
...
+    line-height: var(--ag-line-height, 1.6);
+    cursor: default;
+    user-select: all;
+}
+
+.ag-formula-token-color-1 {
+    background-color: var(--ag-formula-token-1-background-color);
+    border-color: var(--ag-formula-token-1-color);
```
BernieSumption: I'd suggest giving each token its own border property, using BorderValue, e.g. `formulaToken1Border: BorderValue`. This allows setting it to colors other than the text color, and also removing it. The renderer for BorderValue will also automatically use borderWidth as a default instead of 1px.

See `tooltipBorder` as an example of how this is used.

BernieSumption (changes requested):

BernieSumption (approved):

packages/ag-grid-community/src/agStack/theming/themeUtils.ts
```diff
@@ -1,6 +1,6 @@
 import type { ColorValue } from './themeTypes';
 
-const kebabCase = (str: string) => str.replace(/[A-Z]/g, (m) => `-${m}`).toLowerCase();
+const kebabCase = (str: string) => str.replace(/[A-Z]|\d+/g, (m) => `-${m}`).toLowerCase();
```
StephenCooper, line 3: Are these part of your changes?

StephenCooper (approved):

packages/ag-grid-community/src/agStack/theming/themeUtils.ts
gportela85: Yes, they were recommended by @BernieSumption above.

PR #12572, opened 2025-11-29
AG-6477 In-range filter UX follow-ups

eliasmalik:
Fix [AG-6477](https://ag-grid.atlassian.net/browse/AG-6477)


packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -143,8 +140,23 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
         return dateCompWrapper;
     }
 
+    protected override getState() {
+        // State represents non-model related UI state, so we make this equivalent to the validity state of the inputs
+        // so that changes in validity state cause updates to the UI (see `ProvidedFilter.refresh`).
+        return this.hasInvalidInputs();
```
peterjrreynolds: It might be worthwhile making this an object, e.g. `{inputsValid: boolean}`.

Otherwise if we need to introduce additional state properties in the future it will make refactoring and backwards compatibility (state is saved into initial state) awkward.

peterjrreynolds (approved):

PR #12581, opened 2025-12-01
AG-3390 - Remove unused @angular/common

taktran:


taktran:
Tried removing it, but was needed for build and testing.

seanlandsman (approved):

PR #12583, opened 2025-12-01
[AG-13490] Fix(row-sorting): abs sorting order and docs

sdwvit:
addresses TC6, TC4, TC8 and others [AG-13490](https://ag-grid.atlassian.net/browse/AG-13490)
 
Screenshot 2025-12-01 at 17 34 11

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
@@ -150,10 +150,6 @@ const gridOptions = {
 
 {% apiDocumentation source="column-properties/properties.json" section="sort" names=["sort"] /%}
 
-{% note %}
-The grid uses the `sortingOrder` property to determine if Absolute sort should be enabled for a column. Setting `sort` to `{ type: 'absolute', direction: '...' }` alone is not sufficient.
-{% /note %}
-
 ## Sorting API
 
 {% note %}
--- a/documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
@@ -139,12 +139,20 @@ In this example, the column `rankingChange` uses the following sort definition:
 const gridOptions = {
     columnDefs: [
         // ... other columns 
-        { field: 'rankingChange', sort: { type: 'absolute', direction: 'asc' } },
+        {
+            field: 'rankingChange',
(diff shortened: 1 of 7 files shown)


documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
```diff
@@ -139,13 +139,21 @@ In this example, the column `rankingChange` uses the following sort definition:
 const gridOptions = {
     columnDefs: [
         // ... other columns 
-        { field: 'rankingChange', sort: { type: 'absolute', direction: 'asc' } },
+        {
+            field: 'rankingChange',
+            sort: { direction: 'asc', type: 'absolute' },
+            sortingOrder: [null, { direction: 'asc', type: 'absolute' }, { direction: 'desc', type: 'absolute' }],
```
StephenCooper, line 145: I am wondering if we should export constants for these sorts and the default sorting order. Might be a nice syntatatic sugar for this configuration.

SalvatorePreviti (approved):

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -52,6 +52,11 @@ export function isColumn(col: Column | ColumnGroup | ProvidedColumnGroup): col i
 }
 
 const DEFAULT_SORTING_ORDER: SortDirection[] = ['asc', 'desc', null];
+const DEFAULT_ABSOLUTE_SORTING_ORDER: (SortDef | SortDirection)[] = [
+    ...DEFAULT_SORTING_ORDER,
```
StephenCooper: I would query if this is what I would expect to be the default if I only specified an absolute sort.

Thinking ahead to if we support case insensitive this same approach wouldn't make as much sense. 

So wondering if it is simpler to say the default will be asc and desc and null within the same sort type.

documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
sdwvit: since default sorting order is built in now, and is applied if user doesn't specify a custom sortingOrder, do you still think they need it?

StephenCooper (approved):

packages/ag-grid-community/src/entities/agColumn.ts
sdwvit: good idea, I'll keep it till I work on case insensitive sort

PR #12586, opened 2025-12-01
AG-5382 fixes 2

SalvatorePreviti:
- Fixes additional test cases and behaviours as described in the ticket
- Fixes also the ghost component to show "1 row" when dragging a group leaf

Fix AG-5382

### DIFF
--- a/packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
+++ b/packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
@@ -1,3 +1,4 @@
+import type { LocaleTextFunc } from '../agStack/interfaces/iLocaleService';
 import type { AgColumn } from '../entities/agColumn';
 import type { RowNode } from '../entities/rowNode';
 import type { AgEventType } from '../eventTypes';
@@ -148,16 +149,6 @@ export class RowDragComp extends Component {
         };
     }
 
-    private getRowDragText(column?: AgColumn) {
-        if (column) {
-            const colDef = column.getColDef();
-            if (colDef.rowDragText) {
-                return colDef.rowDragText;
-            }
-        }
-        return this.gos.get('rowDragText');
-    }
-
     private addDragSource(dragStartPixels: number = 4): void {
         // if this is changing the drag element, delete the previous dragSource
         if (this.dragSource) {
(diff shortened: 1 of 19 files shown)


packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
```diff
@@ -206,6 +185,27 @@ export class RowDragComp extends Component {
         this.beans.dragAndDrop!.addDragSource(this.dragSource, true);
     }
 
+    private getDragItemName(draggingEvent: RowDraggingEvent | null | undefined, translate: LocaleTextFunc): string {
+        const dragItem = draggingEvent?.dragItem || this.getDragItem();
+        const dragItemCount = (draggingEvent?.dropTarget?.rows.length ?? dragItem.rowNodes?.length) || 1;
```
sdwvit, line 190:
```suggestion
        const dragItemCount = (draggingEvent?.dropTarget?.rows.length ?? dragItem.rowNodes?.length) ?? 1;
```
do we need to do a minimum 1 item count?

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
```diff
@@ -310,6 +312,7 @@ export class RowDragFeature extends BeanStub implements DropTarget {
             suppressMoveWhenRowDragging,
             sameGrid,
             withinGrid,
+            treeData: false,
```
sdwvit, line 315: intentional?

StephenCooper (approved):

packages/ag-grid-enterprise/src/rowHierarchy/groupEditService.ts
```diff
@@ -205,26 +202,48 @@ export class GroupEditService extends BeanStub implements _IGroupEditService {
...
+        if (canExpand && this.dropGroupThrottled && !target.expanded && target.isExpandable?.()) {
             target.setExpanded(true, undefined, true);
         }
 
-        if (target.expanded && target.childrenAfterSort?.length) {
+        let canPutInside = false;
+
+        if (target.expanded) {
+            canPutInside = true;
+        } else if (target.group) {
```
sdwvit: oof that's the case when lonely if would have been better for readability 

is it possible to extract all this canPutInside business in a method please?

sdwvit (approved):

packages/ag-grid-community/src/dragAndDrop/rowDragFeature.ts
SalvatorePreviti: yes - is set later only if the module is loaded by groupEditSvc

packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
SalvatorePreviti: Yeah I would not change this - as we call a user callback like this and we do not want to add a breaking change

packages/ag-grid-enterprise/src/rowHierarchy/groupEditService.ts
SalvatorePreviti: splitted

PR #12601, opened 2025-12-02
AG-16370 group-drag-n-drop-docs

SalvatorePreviti:
- Added a new section regarding group editing
- Added a new section regarding grouping drag and drop
- Moved the docs and the example regarding managed row drag and drop from row dragging to grouping

Fix AG-16370


documentation/ag-grid-docs/src/content/docs/cell-data-types/_examples/providing-custom-cell-data-types/example.spec.ts
```diff
@@ -18,6 +18,8 @@ test.agExample(import.meta, () => {
         test.eachFramework('Example', async ({ agIdFor, page, remoteGrid }) => {
             const remoteApi = remoteGrid(page, '1');
 
+            await remoteApi.setGridOption('animateRows', false); // This is needed or duplicate test ids appear due to animation timing
```
StephenCooper: Is this still required given the other fixes that are now merged?

SalvatorePreviti: No actually it should not be required but I did not try - let's remove it and see what happens in CI

documentation/ag-grid-docs/src/content/docs-nav/nav.json
```diff
@@ -1169,6 +1169,18 @@
                             "title": "Sorting",
                             "path": "grouping-sorting",
                             "isEnterprise": true
+                        },
+                        {
+                            "type": "item",
+                            "title": "Grouping Edit",
```
StephenCooper: Should the page title be `Group Edit` as opposed to Grouping? For both pages.

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
```diff
@@ -0,0 +1,25 @@
+---
+title: "Row Grouping - Editing"
+description: "Grouping editing."
+enterprise: true
+---
+
```
StephenCooper, line 5: I think this page needs a more user focussed description instead of diving straight into the property required.

The grid by default `does not` instead of `do not`

StephenCooper (approved):

PR #12602, opened 2025-12-02
Fix Doc Tests

StephenCooper:
Resolves most of the doc test errors. Formula Data Source will need making into provide example for it to work


documentation/ag-grid-docs/src/content/docs/cell-data-types/_examples/providing-custom-cell-data-types/example.spec.ts
```diff
@@ -18,6 +18,8 @@ test.agExample(import.meta, () => {
         test.eachFramework('Example', async ({ agIdFor, page, remoteGrid }) => {
             const remoteApi = remoteGrid(page, '1');
 
+            await remoteApi.setGridOption('animateRows', false); // This is needed or duplicate test ids appear due to animation timing
```
StephenCooper, line 21: Salvo is iterating on this fix in a separate PR

eliasmalik (approved):

PR #12616, opened 2025-12-03
[AG-13490] Fix(sorting): absolute sort icon not rendered in column header

sdwvit:
Fix [AG-13490](https://ag-grid.atlassian.net/browse/AG-13490) TC12, TC13, TC14


documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
```diff
@@ -127,12 +127,16 @@ The example below shows different combinations of sorting orders as follows:
 
 {% gridExampleRunner title="Sorting Order and Animation" name="sorting-order-and-animation" /%}
 
-## Example: Absolute Value Sorting
-The example below shows how to use a sort definition to allow sorting by absolute value.
+## Absolute Sorting
+Absolute Sorting allows you to sort numeric values based on their magnitude, ignoring their sign. This is useful when you want to rank values by their size rather than by positive or negative direction.
```
StephenCooper:
```suggestion
Absolute Sorting enables sorting numeric values based on their magnitude, ignoring their sign. This can be used to rank values by their size ignoring if a value is positive or negative.
```

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -854,3 +868,26 @@ export function _normalizeSortDirection(sortDirectionLike?: unknown): SortDirect
 export function _normalizeSortType(sortTypeLike?: unknown): SortType {
     return _isSortTypeValid(sortTypeLike) ? sortTypeLike : 'default';
 }
+
+/**
+ * Returns a flat set of provided menu items names
+ */
+function flattenMenuItems(columnMainMenuItems: (DefaultMenuItem | MenuItemDef)[]) {
```
StephenCooper: This code should not be part of agColumn. It should live in the column menu and be accessed via the column menu service. 

This means the code does not bloat core and it is only required when using the column menu.

documentation/ag-grid-docs/src/content/docs/row-sorting/index.mdoc
StephenCooper: When writing docs we should try to avoid the words "you" and instead write in the passive voice. Just something to be aware of.

packages/ag-grid-community/src/entities/agColumn.ts
sdwvit: That was my initial idea. the method is static and does require additional interface for the menu service to be implemented. I'll change it a bit later to be in that module.

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -451,9 +452,22 @@ export class AgColumn<TValue = any>
...
         const res: SortType[] = [];
-        const { sort, initialSort } = this.colDef;
-        const colDefSortType = (sort as SortDef)?.type;
-        const colDefInitialSortType = (initialSort as SortDef)?.type;
+        const { colMenuFactory } = this.beans;
+        const { sort, initialSort, mainMenuItems } = this.colDef;
+
+        if (Array.isArray(mainMenuItems) && mainMenuItems.length && colMenuFactory) {
+            // eslint-disable-next-line sonarjs/no-ignored-return
+            flattenMenuItems(mainMenuItems).find((k) => {
```
SalvatorePreviti: why .find if we don't use the return value? cannot we do a for const of loop?

sdwvit: we can, but it's a bit more elegant to do an early return from a loop

sdwvit: I think you are right actually, this way I don't need to fight with sonar

StephenCooper (approved):

PR #12624, opened 2025-12-04
AG-16386 First set of overlay fixes

StephenCooper:
More to follow but want to get the Doc Tests green first.


packages/ag-grid-community/src/rendering/overlays/overlayComponent.ts
```diff
@@ -45,15 +45,12 @@ export interface INoMatchingRowsOverlayParams<TData = any, TContext = any> exten
 /**
  * Parameters available to configure the provided overlays.
  */
-export type OverlayComponentUserParams = {
```
StephenCooper, line 48: The Doc Interface could not handle this so have just gone with the simpler form

sdwvit (approved):

PR #12627, opened 2025-12-04
AG-3390 - Sync website shared for trial licence link update

taktran:
From https://github.com/ag-grid/ag-charts/pull/5755


markdurrant (approved):
LGTM 👍

PR #12629, opened 2025-12-04
AG-16383 row drag handle grayed out

SalvatorePreviti:
- disable for pivoting
- add the new css class ag-drag-handle-disabled
- add a new type RowDragVisibility and computes it in RowDragService when relevant properties or events may change that state
- add a new event rowDragVisibilityChanged so the row drag handle register just to this handler instead of all the previous ones
- change the logic of the row drag comp to have a consistent css state after every update
- add test for pivoting

Tooltips not added in this PR and should be moved to a new ticket

Fix AG-16383

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_widgets.scss
+++ b/community-modules/styles/src/internal/base/parts/_widgets.scss
@@ -167,12 +167,8 @@
     }
 
     .ag-drag-handle-disabled {
-        opacity: 0.5;
+        opacity: 0.35;
         pointer-events: none;
-
-        .ag-icon {
-            opacity: 0.5;
-        }
     }
 
     .ag-cell-wrapper > *:not(.ag-cell-value):not(.ag-group-value) {
--- a/community-modules/styles/src/internal/base/parts/_widgets.scss
+++ b/community-modules/styles/src/internal/base/parts/_widgets.scss
@@ -166,6 +166,15 @@
         color: var(--ag-secondary-foreground-color);
     }
 
+    .ag-drag-handle-disabled {
+        opacity: 0.5;
+        pointer-events: none;
+
+        .ag-icon {
+            opacity: 0.5;
+        }
+    }
+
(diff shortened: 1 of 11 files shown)


community-modules/styles/src/internal/base/parts/_widgets.scss
```diff
@@ -166,6 +166,15 @@
         color: var(--ag-secondary-foreground-color);
     }
 
+    .ag-drag-handle-disabled {
```
StephenCooper, line 169: This will need to be copied to the legacy styles too.

SalvatorePreviti: Isn't it there in _widgets.scss?

packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
```diff
@@ -88,43 +81,56 @@ export class RowDragComp extends Component {
...
+            this.setVisible(visible, SKIP_ARIA_HIDDEN);
         }
 
-        if (
-            gos.get('rowDragManaged') &&
-            !!beans.rowDragSvc!.rowDragFeature?.shouldPreventRowMove() &&
-            !beans.dragAndDrop?.hasExternalDropZones()
-        ) {
-            return true; // Managed: only show if not prevented and not suppressed, or if there are external drop zones
+        this.setDisabled(!visible || (visibility === 'disabled' && !beans.dragAndDrop?.hasExternalDropZones()));
```
StephenCooper: Can we re-use `!beans.dragAndDrop?.hasExternalDropZones()` via a shared prop in this method?

StephenCooper (approved):

packages/ag-grid-community/src/dragAndDrop/rowDragComp.ts
SalvatorePreviti: mhhh as is written is gonna be executed only once per call if is really needed - what I can do is store dragAndDrop as const to make the code less verbose

PR #12633, opened 2025-12-04
AG-16327 Correctly update initial row pinning state

eliasmalik:
Fix [AG-16327](https://ag-grid.atlassian.net/browse/AG-16327)

### DIFF
--- a/packages/ag-grid-community/src/misc/state/stateService.ts
+++ b/packages/ag-grid-community/src/misc/state/stateService.ts
@@ -241,6 +241,7 @@ export class StateService extends BeanStub implements NamedBean {
             ssrmRowGroupExpansion,
             rowSelection: rowSelectionState,
             pagination: paginationState,
+            rowPinning,
         } = state;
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -310,6 +310,7 @@ export class StateService extends BeanStub implements NamedBean {
                     updateCachedState('pagination', this.getPaginationState());
                 }
             },
+            pinnedRowsChanged: () => updateCachedState('rowPinning', this.getRowPinningState()),
```
peterjrreynolds, line 317: Should the setting of the row pinned state also move from `setFirstDataRenderedState` to `setRowCountState`?

eliasmalik: Seems sensible, done

peterjrreynolds (approved):

PR #12649, opened 2025-12-05
AG-6477 don't apply invalid models if model changes after debounce

eliasmalik:
Fix [AG-6477](https://ag-grid.atlassian.net/browse/AG-6477) TC14, TC17, TC19

### DIFF
--- a/packages/ag-grid-community/src/interfaces/dateComponent.ts
+++ b/packages/ag-grid-community/src/interfaces/dateComponent.ts
@@ -42,13 +42,17 @@ export interface BaseDateParams<TData = any, TContext = any> extends AgGridCommo
     filterParams: DateFilterParams;
 
     location?: 'filter' | 'floatingFilter';
+
+    /**
+     * Method for component to tell AG Grid that an input has been focussed.
+     * Used by the grid to refresh validation messages when moving between inputs in an "inRange" filter.
+     */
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/interfaces/dateComponent.ts
```diff
@@ -47,6 +47,8 @@ export interface BaseDateParams<TData = any, TContext = any> extends AgGridCommo
 export interface IDateParams<TData = any, TContext = any> extends BaseDateParams<TData, TContext> {
     /** Method for component to tell AG Grid that the date has changed. */
     onDateChanged: () => void;
+    /** Method for component to tell AG Grid that an input has been focussed. */
+    onFocusIn?: () => void;
```
eliasmalik: I think this is a public facing interface, so are we happy to add this method to it (albeit optionally)

peterjrreynolds: Yes, but the jsdoc should explain when it's used.

Also, this should go in `BaseDateParams` (React custom components use the `Base` version instead of the `I` version).

eliasmalik: Thanks, done

peterjrreynolds (approved):

PR #12650, opened 2025-12-05
AG-16370 fix-grouping-docs

SalvatorePreviti:
Fix AG-16370


documentation/ag-grid-docs/src/content/docs/grouping-row-dragging/_examples/managed-row-group-drag-multi-level/main.ts
```diff
@@ -13,8 +13,8 @@ import {
 } from 'ag-grid-community';
 import { BatchEditModule, RowGroupingModule } from 'ag-grid-enterprise';
 
-import type { IAthlete } from './data';
 import { getAthletesData } from './data';
+import type { IAthlete } from './types';
```
StephenCooper, line 17: You can remove this line and the types file. The example generator will see the IAthlete type in the example and add both the interface file and the import for it.

SalvatorePreviti: ohhh I have an "id" field

StephenCooper (approved):

PR #12663, opened 2025-12-05
[AG-16195] Fix(rich-select): pills dont close on click non strings

sdwvit:
Fixes [RTI-3186](https://ag-grid.atlassian.net/browse/RTI-3186)
Fixes [AG-16195](https://ag-grid.atlassian.net/browse/AG-16195)

### DIFF
--- a/packages/ag-grid-enterprise/src/widgets/AgPillContainer.ts
+++ b/packages/ag-grid-enterprise/src/widgets/AgPillContainer.ts
@@ -29,7 +29,8 @@ const AgPillContainerElement: ElementParams = {
 };
 export class AgPillContainer<TValue> extends Component {
     private params: PillRendererParams<TValue>;
-    private pills: AgPill[] = [];
+    private pills: { pill: AgPill; key: string }[] = [];
+    private getKey: { (value: TValue | TValue[]): string | null };
 
     constructor() {
         super(AgPillContainerElement);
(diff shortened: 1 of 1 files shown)


packages/ag-grid-enterprise/src/widgets/AgPillContainer.ts
```diff
@@ -143,28 +145,29 @@ export class AgPillContainer<TValue> extends Component {
...
     }
 
-    private deletePill(pill: AgPill, restoreFocusToIndex?: number): void {
-        const value = pill.getText();
-        const values = (this.params.getValue() || []).filter((val) => val !== value);
-        this.params.setValue(values);
+    private deletePill(p: AgPill, restoreFocusToIndex?: number): void {
+        const { getKey, pills, params } = this;
+        const pillKey = (pills[restoreFocusToIndex ?? -1] ?? pills.find(({ pill }) => pill === p))?.key;
+        const values = (params.getValue() || []).filter((val) => getKey(val) !== pillKey);
```
sdwvit, line 164: not a huge fan of this approach, but this works with an assumption that pill text is unique per item. Otherwise it may not work correctly.

sdwvit: other way of doing this is to ask users to provide 'key generator' callback, which must be an opt in, and people tend to just ignore it and may still complain about broken UX anyway 🤷

gportela85 (approved):

sdwvit:
merged into latest, but it is meant for b35 instead. recreating a pr now

PR #12673, opened 2025-12-08
RTI-3160 - Add enterprise icon to groups

taktran:
Fix [RTI-3160](https://ag-grid.atlassian.net/browse/RTI-3160)


markdurrant (approved):
LGTM 👍

PR #12729, opened 2025-12-09
RTI-3213 Update custom editor docs

StephenCooper:
Fix RTI-3213

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/cell-editors/_component-interface-angular.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/cell-editors/_component-interface-angular.mdoc
@@ -1,13 +1,11 @@
 {% if isFramework("angular") %}
 
-The interface for the cell editor component is as follows:
+To implement a custom cell editor implement the `ICellEditorAngularComp` interface that has two mandatory methods: 
+
+- `agInit(params: ICellEditorParams)` - called when the editor is initialised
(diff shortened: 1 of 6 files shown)


sdwvit (approved):

documentation/ag-grid-docs/src/content/docs/cell-editors/_component-interface-angular.mdoc
```diff
@@ -1,13 +1,11 @@
 {% if isFramework("angular") %}
 
-The interface for the cell editor component is as follows:
+To implement a custom cell editor implement the `ICellEditorAngularComp` interface that has two mandatory methods:
```
sdwvit:
```suggestion
To create a custom cell editor, implement the `ICellEditorAngularComp` interface, which defines two mandatory methods: 
```

packages/ag-grid-community/src/interfaces/iCellEditor.ts
```diff
@@ -54,7 +54,7 @@ export interface BaseCellEditor {
 
 export interface ICellEditor<TValue = any> extends BaseCellEditor {
     /**
-     * Return the final value - called by the grid once after editing is complete
+     * Mandatory - Return the final value - called by the grid once after editing is complete
```
sdwvit:
```suggestion
     * Mandatory - Return the final value. Called by the grid once after editing is complete.
```

PR #12732, opened 2025-12-10
[AG-16202] Fix(filtering): when using celldatatype datetime the min max valid date validations do not work

sdwvit:
improve min/max date handling in DefaultDateComponent

plunker for testing (points to local) https://plnkr.co/edit/78E3AxduPsMJgu66?open=main.js

Fixes [AG-16202](https://ag-grid.atlassian.net/browse/AG-16202)

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/date/defaultDateComponent.ts
+++ b/packages/ag-grid-community/src/filter/provided/date/defaultDateComponent.ts
@@ -110,8 +110,8 @@ export class DefaultDateComponent extends Component implements IDateComp {
         } else {
             inputElement.type = 'text';
         }
-        const parsedMinValidDate = grabDate(minValidDate, minValidYear, true);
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/filter/provided/date/defaultDateComponent.ts
```diff
@@ -192,3 +163,18 @@ export class DefaultDateComponent extends Component implements IDateComp {
         return params?.filterParams?.browserDatePicker ?? true;
     }
 }
+
+function grabDate(validationDate: unknown, validationYear: unknown, isMin: boolean): null | Date {
```
eliasmalik: nit: would maybe call this something more like `getLimitDate`, `deriveLimitDate`, `getValidationDate` or similar. Would probably also just put the actual expected types in instead of `unknown` (i.e. `string | Date | undefined` and `number | undefined`)

eliasmalik (approved):
approved with nit

PR #12739, opened 2025-12-11
AG-16448 value getter fixes

SalvatorePreviti:
There are few problems that are caused by the fact that getValue when passing source: ui or source:undefined (that default to ui) returns the value being edited and not the actual cell value.

In this PR we make source when getting the value mandatory and we apply 'ui' or 'api' whenever we want to use the actual value or the value being edited.

- Expose a new "committed?:boolean" to get cell API, so the user, and our tests, can specify if we want to read the actual value or the value being edited. By default is false to not causing breaking changes. 
- removes getKeyForNode from valueService
- it is used only by grouping and pivoting, so it can go in the grouping hierarchy utils
- replaces forEach loops in pivotStage with normal faster loops
- Make source mandatory in getValue and getValueForDisplay
- ensure the consistency of source where we use value service.
- ensure that we never cache a 'ui' value in the ValueCache, we always want to use a 'api' value there.
- remove executeFilterValueGetter as it was not needed, the now exposed executeValueGetterNoCache does the same thing
- minor refactorings in the value service
    - remove _addGridCommonParams as it is slower and unnecessary to create an object and then adding fields than just doing {} object with all the fields in - and valueSvc is a hot path

Fix AG-16448

### DIFF
--- a/packages/ag-grid-community/src/valueService/valueService.ts
+++ b/packages/ag-grid-community/src/valueService/valueService.ts
@@ -7,41 +7,30 @@ import type { NamedBean } from '../context/bean';
 import { BeanStub } from '../context/beanStub';
 import type { BeanCollection } from '../context/context';
 import type { AgColumn } from '../entities/agColumn';
-import type {
-    KeyCreatorParams,
-    ValueFormatterParams,
-    ValueGetterParams,
-    ValueParserParams,
-    ValueSetterParams,
-} from '../entities/colDef';
+import type { ValueFormatterParams, ValueGetterParams, ValueParserParams, ValueSetterParams } from '../entities/colDef';
 import type { RowNode } from '../entities/rowNode';
 import type { CellValueChangedEvent } from '../events';
-import { _addGridCommonParams, _isServerSideRowModel } from '../gridOptionsUtils';
+import { _isServerSideRowModel } from '../gridOptionsUtils';
 import type { IFormulaDataService } from '../interfaces/formulas';
 import type { IEditService } from '../interfaces/iEditService';
 import type { IRowNode } from '../interfaces/iRowNode';
 import { _warn } from '../validation/logging';
-import type { ExpressionService } from './expressionService';
 import type { ValueCache } from './valueCache';
 
 export class ValueService extends BeanStub implements NamedBean {
     beanName = 'valueSvc' as const;
 
-    private expressionSvc?: ExpressionService;
     private colModel: ColumnModel;
     private valueCache?: ValueCache;
     private dataTypeSvc?: DataTypeService;
     private editSvc?: IEditService;
-    private hasEditSvc: boolean = false;
     private formulaDataSvc?: IFormulaDataService;
 
     public wireBeans(beans: BeanCollection): void {
-        this.expressionSvc = beans.expressionSvc;
         this.colModel = beans.colModel;
         this.valueCache = beans.valueCache;
         this.dataTypeSvc = beans.dataTypeSvc;
         this.editSvc = beans.editSvc;
-        this.hasEditSvc = !!beans.editSvc;
         this.formulaDataSvc = beans.formulaDataSvc;
     }
 
@@ -54,14 +43,6 @@ export class ValueService extends BeanStub implements NamedBean {
 
     private isSsrm = false;
 
-    private executeValueGetter: (
-        // eslint-disable-next-line @typescript-eslint/ban-types
-        valueGetter: string | Function,
-        data: any,
-        column: AgColumn,
-        rowNode: IRowNode
-    ) => any;
-
     public postConstruct(): void {
         if (!this.initialised) {
             this.init();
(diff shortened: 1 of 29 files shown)


packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -64,7 +64,7 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
 
         const leafNode = _csrmFirstLeaf(node);
         const rowGroupColId = rowGroupCol.getId();
-        if (!showRowGroupCols) {
+        if (!showRowGroupCols || !leafNode) {
```
StephenCooper, line 67: Did you discover a potential bug here that needed the new condition?

SalvatorePreviti: I checked down the line where this was used: this called a method that accepted a nullable RowNode and it was checked in that function but it does not make sense, better to force the node to not be null there and check in the caller, so, here

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -305,36 +293,44 @@ export class ValueService extends BeanStub implements NamedBean {
...
-        if (colDef.allowFormula && this.beans.formula?.isFormula(newValue)) {
+        if (colDef.allowFormula && beans.formula?.isFormula(newValue)) {
             return newValue;
         }
 
         const valueParser = colDef.valueParser;
 
         if (_exists(valueParser)) {
-            const params: ValueParserParams = _addGridCommonParams(this.gos, {
+            const params: ValueParserParams = {
```
gportela85, line 307: why are we removing `_addGridCommonParams` here and adding the `api` and `context` manually?

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -305,36 +293,44 @@ export class ValueService extends BeanStub implements NamedBean {
...
+                api: beans.gridApi,
+                context: beans.gridOptions.context,
```
StephenCooper, line 309: Can you quantify this improvement in performance? Is the valueParser really on a hot path?

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -663,82 +665,50 @@ export class ValueService extends BeanStub implements NamedBean {
...
         valueGetter: string | Function,
         data: any,
         column: AgColumn,
         rowNode: IRowNode
     ): any {
-        const params: ValueGetterParams = _addGridCommonParams(this.gos, {
+        const beans = this.beans;
+        const params: ValueGetterParams = {
+            api: beans.gridApi,
+            context: beans.gridOptions.context,
```
StephenCooper, line 696: I accept that this is a hot path, so can you quantify the performance here?

I think we should only make this change where it is actually required and if we can show it does make a difference.

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -663,82 +665,50 @@ export class ValueService extends BeanStub implements NamedBean {
...
             data: data,
             node: rowNode,
             column: column,
             colDef: column.getColDef(),
-            getValue: this.getValueCallback.bind(this, rowNode),
-        });
+            getValue: (field: string | AgColumn) => {
```
gportela85, line 701: The previous implementation with `_addGridCommonParams` and `getValueCallback`, defaulted to the UI source and picked up pending edits. Any `valueGetter` that depends on another column’s live (uncommitted) edit value will now see stale data until the edit commits

You can test this with
```
gridOptions = {
    columnDefs: [
         { field: 'a', editable: true },
         {
              field: 'b',
              valueGetter: p => p.getValue('a');   // used to read live edit value
          }
    ]
}
```

if you start a batch edit and change the value of `a`, this used to update the valueGetter in `b`. Now this won't happen anymore.

Is this the desired behavior?

SalvatorePreviti: this specific case yes it should read the live value and it should take the uncommitted value - I will write tests for this

packages/ag-grid-community/src/valueService/valueService.ts
SalvatorePreviti: a very cheap micro optimization done in other parts - as said in the description, remove _addGridCommonParams as it is slower and unnecessary to create an object and then adding fields than just doing {} object with all the fields in - and valueSvc is a hot path - for more info https://medium.com/@mivashchenko.blog/object-shapes-in-javascript-f312ac51675a

SalvatorePreviti: Is there a specific reason for _addGridCommonParams?

packages/ag-grid-community/src/valueService/valueService.ts
SalvatorePreviti: I can remove this - but is there a specific reason for _addGridCommonParams?

packages/ag-grid-community/src/valueService/valueService.ts
gportela85: Best to check if @StephenCooper but I believe it was an effort to reduce bundle size by removing repetitive addition of api and context.

PR #12746, opened 2025-12-12
AG-16448 add tests

SalvatorePreviti:
- Add new editing tests for value getter that when editing returns the current value being edited in a value getter
- Add new editing unit tests for edit that mimic and extend the existing documentation tests, so we can run them as part of our unit test bundle and do more regression checks
- Improves all the snapshots that were using printIds = false and specific list of columns in the various unit tests, preferring the GridRows without options


StephenCooper (approved):
Looking good.

PR #12750, opened 2025-12-15
AG-16182 Organise exports from AG Grid Community

StephenCooper:
Fix [AG-16182](https://ag-grid.atlassian.net/browse/AG-16182)

Main idea is to start the process of separating true public exports from those exports that are required by AG Grid Enterprise.


packages/ag-grid-community/src/main-internal.ts
```diff
@@ -0,0 +1,548 @@
...
+export { IDragService as _IDragService } from './agStack/interfaces/iDrag';
+export {
+    IDragAndDropImage,
+    AgDragSource as _AgDragSource,
+    AgDraggingEvent as _AgDraggingEvent,
+    AgDropTarget as _AgDropTarget,
+    IDragAndDropService as _IDragAndDropService,
+} from './agStack/interfaces/iDragAndDrop';
+export { AgEventService as _AgEventService, WithoutCommon as _WithoutCommon } from './agStack/interfaces/iEvent';
+export { IEventEmitter, IEventListener } from './agStack/interfaces/iEventEmitter';
```
peterjrreynolds: `IEventEmitter` is public (`Column` extends it), and by extension `IEventListener` too

packages/ag-grid-community/src/main-internal.ts
```diff
@@ -0,0 +1,548 @@
...
+    _defaultComparator,
+    _exists,
+    _jsonEquals,
+    _makeNull,
+    _missing,
+    _toStringOrNull,
+} from './agStack/utils/generic';
+export { _isEventFromPrintableCharacter } from './agStack/utils/keyboard';
+export { _getLocaleTextFromFunc, _getLocaleTextFromMap, _getLocaleTextFunc, _translate } from './agStack/utils/locale';
+export { AgPromise, _isPromise } from './agStack/utils/promise';
```
peterjrreynolds: `AgPromise` is public for custom components (from `IComponent`)

packages/ag-grid-community/src/main-internal.ts
```diff
@@ -0,0 +1,548 @@
...
+    IRowNodeFilterAggregateStage as _IRowNodeFilterAggregateStage,
+    IRowNodeFlattenStage as _IRowNodeFlattenStage,
+    IRowNodeGroupStage as _IRowNodeGroupStage,
+    IRowNodePivotStage as _IRowNodePivotStage,
+} from './interfaces/iRowNodeStage';
+export { ISelectionService, ISetNodesSelectedParams } from './interfaces/iSelectionService';
+export { IShowRowGroupColsService } from './interfaces/iShowRowGroupColsService';
+export { IShowRowGroupColsValueService } from './interfaces/iShowRowGroupColsValueService';
+export { ISideBar, ISideBarService } from './interfaces/iSideBar';
+export { SortModelItem } from './interfaces/iSortModelItem';
```
peterjrreynolds: `SortModelItem` is public (used in state)

eliasmalik (approved):

packages/ag-grid-community/src/main.ts
```diff
@@ -324,709 +86,484 @@ export {
...
+    ExcelWorksheet,
     ExcelWorksheetConfigParams,
+    IExcelCreator,
+    RowHeightCallbackParams,
 } from './interfaces/iExcelCreator';
 
-// dragAndDrop
-export type { DragAndDropService } from './dragAndDrop/dragAndDropService';
+// Drag and Drop
 export {
```
SalvatorePreviti: minor, not for this PR but - we should add a eslint warning as types should be exported as "export type" - and also, we will need slowly to understand if we are exporting concrete instances when they are supposed to be used just as types and in this case export type too

SalvatorePreviti (approved):

packages/ag-grid-community/src/main.ts
StephenCooper: Have applied this change.

PR #12773, opened 2025-12-17
AG-5459 - [excel-export] - added worksheet protection

gportela85:
fix [AG-5459](https://ag-grid.atlassian.net/browse/AG-5459)


StephenCooper (approved):
LGTM and nice that we can test this too.

PR #12777, opened 2025-12-18
AG-12960 Top level types

StephenCooper:
Fix AG-12960

Added the `'@typescript-eslint/prefer-function-type': 'error'` rule as this resolves Sonar issues as well as making function types easier to read in code and our docs. 

Added missing callback functions on ColDef and GridOptions.
Have not done the same for events as that seems less important as all of the methods have a void return type so the params type is all you need.


packages/ag-grid-community/src/columns/dataTypeService.ts
```diff
@@ -33,7 +33,7 @@ interface GroupSafeValueFormatter {
     groupSafeValueFormatter?: ValueFormatterFunc;
 }
 
-type DataTypeDefinitions = {
+type DataTypeDefinitionMap = {
```
StephenCooper, line 36: Renaming internal only type to avoid clashing in Doc Type generation.

SalvatorePreviti (approved):

PR #12784, opened 2025-12-19
AG-16157 - fixed scrollbar not disappearing

gportela85:
fix [AG-16157](https://ag-grid.atlassian.net/browse/AG-16157)

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/scrollVisibleService.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/scrollVisibleService.ts
@@ -42,7 +42,7 @@ export class ScrollVisibleService extends BeanStub implements NamedBean {
         this.addManagedEventListeners({
             displayedColumnsChanged: updateScrollVisible,
             displayedColumnsWidthChanged: updateScrollVisible,
-            columnEverythingChanged: updateScrollVisible,
+            newColumnsLoaded: updateScrollVisible,
         });
     }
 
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/gridBodyComp/scrollbarVisibilityHelper.test.ts
StephenCooper, line 1: 💯

packages/ag-grid-community/src/gridBodyComp/scrollVisibleService.ts
```diff
@@ -40,6 +40,7 @@ export class ScrollVisibleService extends BeanStub implements NamedBean {
         this.addManagedEventListeners({
             displayedColumnsChanged: this.updateScrollVisible.bind(this),
             displayedColumnsWidthChanged: this.updateScrollVisible.bind(this),
+            columnEverythingChanged: this.updateScrollVisible.bind(this),
```
StephenCooper: This event is deprecated according to our docs. 

> - `columnEverythingChanged` - deprecated, use `displayedColumnsChanged` which is fired at the same time, or one of the more specific [Column Events](./column-state/#column-events).

I see in one place we have this comment but in another location we are firing it without such a comment.

```
        // this event is not used by AG Grid, but left here for backwards compatibility,
        // in case applications use it
        eventSvc.dispatchEvent({
            type: 'columnEverythingChanged',
            source,
        });
```

Seems like we may need to clean this inconsistency up.

StephenCooper (approved):

PR #12819, opened 2026-01-05
[AG-14571] Feature(column): filters built in named relative date ranges

sdwvit:
Add built-in named and relative date ranges to Date Filter

- Add `presetDateFilterTypeRelativeFromToMap` to define relative date ranges like "Today", "Last 7 Days", etc., using `setStartOfDay`, `setEndOfDay`, and other utility functions
- Introduce `ISimpleFilterModelPresetType` type to support named date filter options in `iSimpleFilter`
- Implement `evaluateNonNullValue` override in `DateFilterHandler` to handle preset date range logic
- Enhance `DateFilter` to apply preset ranges when `type` matches a known preset
- Add locale text for all new date filter options in `filterLocaleText.ts`
- Refactor `sortIndicatorComp` and `columnMenuFactory` to use `_getDisplaySortForColumn` utility
- Export `_getDisplaySortForColumn` from `agColumn` and update imports accordingly
- Add `.run` files for new date filter test configurations
 
Fix [AG-14571](https://ag-grid.atlassian.net/browse/AG-14571)


packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -43,6 +66,133 @@ export function validateAndUpdateConditions<M>(conditions: M[], maxNumConditions
...
+    last30Days: (from, to) => {
+        return { from: setPreviousNDay(from, 30), to: setEndOfDay(to) };
+    },
+    last90Days: (from, to) => {
+        return { from: setPreviousNDay(from, 90), to: setEndOfDay(to) };
+    },
+    last6Months: (from, to) => {
+        from.setFullYear(from.getFullYear() - 1);
+        from.setMonth(from.getMonth() + 6);
+        return { from, to: setEndOfDay(to) };
```
sdwvit: from must be start of day

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -43,6 +66,133 @@ export function validateAndUpdateConditions<M>(conditions: M[], maxNumConditions
...
+    },
+    last12Months: (from, to) => {
+        from.setFullYear(from.getFullYear() - 1);
+        return { from, to };
```
sdwvit: both from and to should be start and end of day respectively

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -43,6 +66,133 @@ export function validateAndUpdateConditions<M>(conditions: M[], maxNumConditions
...
+    },
+    last24Months: (from, to) => {
+        from.setFullYear(from.getFullYear() - 2);
+        return { from, to };
```
sdwvit: both from and to should be start and end of day respectively

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,29 +180,87 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
+| Not blank      | `notBlank`    | Yes                 |
+| Choose one     | `empty`       | No                  |
 
 Note that the `empty` filter option is primarily used when creating [Custom Filter Options](./filter-conditions/#custom-filter-options). When 'Choose one' is displayed, the filter is not active.
 
 The default option for the Date Filter is `equals`.
 
 When providing filter options, the default filter option (or the first option if no default set) must be an option that displays an input or the `empty` filter option (as a filter option with no inputs would mean the filter is active by default).
 
-## Range Input Validation and Error States
```
eliasmalik, line 200: This section has been removed, is that deliberate?

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,29 +180,87 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
-Where this is not the case, the last edited input will display a red border and a hover tooltip directing the user to enter a valid date. While the filter is in an invalid state, the filter will not be applied.
-Screen readers that respond to the `aria-invalid` attribute or the `ValidityState` of the input will detect the input as invalid.
+{% note %}
+This feature is not included by default. To enable it, you must add the options to `filterOptions` in your date filter configuration.
+{% /note %}
+
+### Available Built-in Date Filter Options
+
+| Option Name        | Option Key     | Time Range                                        |
+|--------------------|----------------|---------------------------------------------------|
```
StephenCooper: I wonder if Time Range would be clearer in two columns? Making it easier to scan.

Also then the heading could have the indication that the end date is not inclusive. ie start is >= end is <

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -43,6 +66,133 @@ export function validateAndUpdateConditions<M>(conditions: M[], maxNumConditions
...
+ * Next Year               nextYear            [startOfNextYear, startOfYearAfterNext)
+ * Year to Date (YTD)      yearToDate          [startOfCurrentYear, startOfTomorrow)
+ * Last 7 days             last7Days           [startOfToday − 7 days, startOfTomorrow)
+ * Last 30 days            last30Days          [startOfToday − 30 days, startOfTomorrow)
+ * Last 90 days            last90Days          [startOfToday − 90 days, startOfTomorrow)
+ * Last 6 months           last6Months         [startOfToday − 6 months, startOfTomorrow)
+ * Last 12 months          last12Months        [startOfToday − 12 months, startOfTomorrow)
+ * Last 24 months          last24Months        [startOfToday − 24 months, startOfTomorrow)
+ */
+export const presetDateFilterTypeRelativeFromToMap: Record<
```
StephenCooper: Is this tree shakeable? Aim would be for it to be tied to the DateFilterModule.

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -43,6 +66,133 @@ export function validateAndUpdateConditions<M>(conditions: M[], maxNumConditions
...
+    ISimpleFilterModelPresetType,
+    (providedFrom: Date, providedTo: Date) => { from: Date; to: Date }
+> = {
+    today: (from, to) => {
+        return { from: setStartOfDay(from), to: setEndOfDay(to) };
+    },
+    yesterday: (from, to) => {
+        setPreviousDay(from);
+        setPreviousDay(to);
+        return presetDateFilterTypeRelativeFromToMap.today(from, to);
```
StephenCooper: I think this will minify better if `today` was a stand alone function defined outside of this object that can then be shared without accessing off `presetDateFilterTypeRelativeFromToMap`

packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -284,20 +285,31 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
...
+
+        const separator = params.useIsoSeparator ? 'T' : ' ';
         if (values.length > 0) {
             model.dateFrom = _serialiseDate(values[0], true, separator);
         }
         if (values.length > 1) {
             model.dateTo = _serialiseDate(values[1], true, separator);
         }
 
+        if (type && type in presetDateFilterTypeRelativeFromToMap) {
```
StephenCooper: Reuse the lookup value to avoid duplicating accessing it off the object?

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
sdwvit: This is tricky, since an alternative would be to use big switch-case. Which in turn would  also require all these functions.

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -22,6 +24,10 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
 export class DateFilterHandler extends ScalarFilterHandler<DateFilterModel, Date, IDateFilterParams> {
     public readonly filterType = 'date' as const;
     protected readonly FilterModelFormatterClass = DateFilterModelFormatter;
+    /**
+     * This is used to prevent desync when user scrolls in SSRM, and to materialise time into a grid state
+     */
+    public readonly beanCreationTime = new Date();
```
StephenCooper: If this filter handler is long lived does this not run the risk of becoming stale?

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
StephenCooper: In `testing/module-size` use the `module-stats` command to run this locally.

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
sdwvit: This handler gets destroyed at the same time as filter. We agreed to not track 'long life' if someone keeps their tab open for a while.

packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -284,20 +285,31 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
...
+            const { beanCreationTime } = params.getHandler() as DateFilterHandler;
+            const { from, to } = presetDateFilterTypeRelativeFromToMap[type as ISimpleFilterModelPresetType](
+                new Date(beanCreationTime),
+                new Date(beanCreationTime)
+            );
+            model.dateFrom = _serialiseDate(from, true, separator);
+            model.dateTo = _serialiseDate(to, true, separator);
```
peterjrreynolds: Based on our discussion yesterday, this shouldn't be setting any date values into the model

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -35,4 +41,27 @@ export class DateFilterHandler extends ScalarFilterHandler<DateFilterModel, Date
...
+        if (!this.isValid(cellValue)) {
+            return type === 'notEqual' || type === 'notBlank';
+        }
+
+        if (type && type in presetDateFilterTypeRelativeFromToMap) {
+            const { from, to } = presetDateFilterTypeRelativeFromToMap[type as ISimpleFilterModelPresetType](
+                new Date(this.beanCreationTime),
+                new Date(this.beanCreationTime)
+            );
+            return comparator(from, cellValue) > 0 && comparator(to, cellValue) < 0;
```
peterjrreynolds: Shouldn't this be `>= 0` for the from value?

sdwvit: yes it should, ty!

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
sdwvit: thanks for catching, is an accident

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,29 +180,87 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
+| Last 6 Months      | `last6Months`  | `Start Of Today minus 6 months`  | `Start Of Tomorrow`           |
+| Last 12 Months     | `last12Months` | `Start Of Today minus 12 months` | `Start Of Tomorrow`           |
+| Last 24 Months     | `last24Months` | `Start Of Today minus 24 months` | `Start Of Tomorrow`           |
+
+
+{% note %}
+`startOfToday`, `startOfTomorrow`, `startOfXXX` point to the beginning of the corresponding day, e.g. `00:00:00.000` in browser local timezone.   
+{% /note %}
+{% note %}
+`[A, B)` means including `A`, but excluding `B`. `[today, tomorrow)` corresponds to a range `00:00:00.000` - `23:59:59.999`
```
StephenCooper: This note can be removed now.

documentation/ag-grid-docs/src/content/docs/filter-date/_examples/built-in-date-ranges/main.ts
```diff
@@ -0,0 +1,79 @@
...
+    ClientSideRowModelModule,
+    DateFilterModule,
+    ModuleRegistry,
+    ValidationModule,
+    createGrid,
+} from 'ag-grid-community';
+
+ModuleRegistry.registerModules([
+    ClientSideRowModelModule,
+    DateFilterModule,
```
StephenCooper, line 13: The example currently requires the TextFilterModule as this is the default for the `athlete` column. If you open the filter for that column you get an example missing module error

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,23 +180,88 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
+
+
+{% note %}
+`Start Of Today`, `Start Of Tomorrow`, `Start Of XXX` point to the beginning of the corresponding day, e.g. `00:00:00.000` in browser local timezone.
+{% /note %}
+{% note %}
+`[A, B)` means including `A`, but excluding `B`. `[today, tomorrow)` corresponds to a range `00:00:00.000` - `23:59:59.999` 
+{% /note %}
+
+### Enabling Built-in Date Ranges
```
StephenCooper: I think this should be the first section and the complete reference list of options comes at the end of this section

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,23 +180,88 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
 
 Note that the `empty` filter option is primarily used when creating [Custom Filter Options](./filter-conditions/#custom-filter-options). When 'Choose one' is displayed, the filter is not active.
 
 The default option for the Date Filter is `equals`.
 
 When providing filter options, the default filter option (or the first option if no default set) must be an option that displays an input or the `empty` filter option (as a filter option with no inputs would mean the filter is active by default).
 
+## Built-in Named & Relative Date Ranges
+
+Date Filter includes a set of **predefined named and relative date ranges** to make common filtering tasks easier. These ranges are relative to the browser local time. These are ideal for users who want to filter by familiar time periods like "Last Week", "Year to Date", or "Next 30 Days" without extending a grid with a custom solution.
```
StephenCooper: I think the work "includes" is confusing with the clarification note below. Maybe we should say the Date Filter "supports"

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
StephenCooper: I think this is actually more likely to cause issues for CSRM. Use case a user filters their data and then leaves the app open. Then they change their filter to a different preset range the next day. This would then be incorrect.
We know that many users leave their apps open like this. 

As we are not passing dates to SSRM that use case has now gone. 

Instead can we not just call Date.now() in the evaluate step?

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,23 +180,89 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
 
 Note that the `empty` filter option is primarily used when creating [Custom Filter Options](./filter-conditions/#custom-filter-options). When 'Choose one' is displayed, the filter is not active.
 
 The default option for the Date Filter is `equals`.
 
 When providing filter options, the default filter option (or the first option if no default set) must be an option that displays an input or the `empty` filter option (as a filter option with no inputs would mean the filter is active by default).
 
+## Built-in Named & Relative Date Ranges
+
+Date Filter supports a set of **predefined named and relative date ranges** to make common filtering tasks easier. These ranges are relative to the browser local time. These are ideal for users who want to filter by familiar time periods like "Last Week", "Year to Date", or "Next 30 Days" without extending a grid with a custom solution.
```
StephenCooper: We don't use bold in general in the docs

documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
```diff
@@ -180,23 +180,89 @@ The Date Filter presents a list of [Filter Options](./filter-conditions/#filter-
...
+#### Example: Enable Built-in Date Ranges
+
+{% gridExampleRunner title="Built-in Date Ranges" name="built-in-date-ranges" /%}
+
+Example above uses following configuration:
+
+```ts
+// Enable built-in date ranges
+const dateFilterParams = {
+    filterOptions: [
```
StephenCooper: Mimic a colDef to show where this config needs to apply.

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -35,4 +37,239 @@ export class DateFilterHandler extends ScalarFilterHandler<DateFilterModel, Date
...
+        const comparator = this.comparator();
+
+        if (!this.isValid(cellValue)) {
+            return type === 'notEqual' || type === 'notBlank';
+        }
+        const presetDateRangeFn = this._presetDateFilterTypeRelativeFromToMap[
+            type as ISimpleFilterModelPresetType
+        ] as RelativeRangeFn;
+        if (presetDateRangeFn) {
+            const [from, to] = presetDateRangeFn(new Date(), new Date());
```
StephenCooper: Can we throttle this creation of dates.

StephenCooper: Could cache it for a run of filtering and have this cleared via a setTimeout immediately. The filtering of all the rows will be synchronous so the cached value will not be cleared until that has been completed

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -22,9 +23,27 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
...
     constructor() {
         super(mapValuesFromDateFilterModel, DEFAULT_DATE_FILTER_OPTIONS);
+        this.refreshFilterBaseDate();
+    }
+
+    private refreshFilterBaseDate(): void {
+        if (this.isAlive()) {
+            this.filterTypeToRangeCache.forEach((_, type) => {
+                const presetDateRangeFn = presetDateFilterTypeRelativeFromToMap[type] as RelativeRangeFn;
+                this.filterTypeToRangeCache.set(type, presetDateRangeFn(new Date(), new Date()));
```
sdwvit: call clear instead;

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -22,9 +23,27 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
...
+            });
+            const filterBaseDateTimeout = setTimeout(
+                () => this.refreshFilterBaseDate(),
+                12 * 60 * 60 * 1000 // 12 hours
```
sdwvit: do it till the end of the day
OR
do cleanup on filter stage complete event
OR 
do cleanup before stage

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -35,4 +54,241 @@ export class DateFilterHandler extends ScalarFilterHandler<DateFilterModel, Date
...
+
+        if (!this.isValid(cellValue)) {
+            return type === 'notEqual' || type === 'notBlank';
+        }
+        const typeAsPreset = type as ISimpleFilterModelPresetType;
+        const presetDateRangeFn = presetDateFilterTypeRelativeFromToMap[typeAsPreset] as RelativeRangeFn;
+        if (presetDateRangeFn) {
+            // indicates we are in preset time ranges space
+            const cache = this.filterTypeToRangeCache.get(typeAsPreset) || presetDateRangeFn(new Date(), new Date());
+            this.filterTypeToRangeCache.set(typeAsPreset, cache);
```
sdwvit: split it into cache not cached

jest.preset.js
```diff
@@ -1,3 +1,3 @@
+process.env.TZ = 'UTC';
```
sdwvit, line 1: this is the best place to put it do to a nature of jest forking workers after this code runs but before tests run.

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -22,9 +23,22 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
...
         super(mapValuesFromDateFilterModel, DEFAULT_DATE_FILTER_OPTIONS);
+        this.refreshFilterBaseDate();
+    }
+
+    private refreshFilterBaseDate(): void {
+        if (this.isAlive()) {
+            this.filterTypeToRangeCache.clear();
+            const filterBaseDateTimeout = setTimeout(
+                () => this.refreshFilterBaseDate(),
+                setStartOfNextDay(new Date()).getTime() - Date.now() - 60 * 1000
```
StephenCooper, line 39: Can we add a comment explaining this.

StephenCooper (approved):

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
sdwvit:
```suggestion
                // this evaluates to a number of ms between NOW and begginning of tomorrow MINUS 1 minute
                setStartOfNextDay(new Date()).getTime() - Date.now() - 60 * 1000
```

packages/ag-grid-community/src/filter/provided/date/dateFilterHandler.ts
```diff
@@ -22,9 +23,23 @@ function defaultDateComparator(filterDate: Date, cellValue: any): number {
...
         super(mapValuesFromDateFilterModel, DEFAULT_DATE_FILTER_OPTIONS);
+        this.refreshFilterBaseDate();
+    }
+
+    private refreshFilterBaseDate(): void {
+        if (this.isAlive()) {
+            this.filterTypeToRangeCache.clear();
+            const filterBaseDateTimeout = setTimeout(
+                () => this.refreshFilterBaseDate(),
+                // this evaluates to a number of ms between NOW and begginning of tomorrow MINUS 1 minute
```
sdwvit:
```suggestion
                // this evaluates to a number of ms between NOW and beginning of tomorrow MINUS 1 minute
```

PR #12820, opened 2026-01-05
AG-3390 Improve selection assertion for group selection case

eliasmalik:


testing/behavioural/src/test-utils/gridRows/validation/gridRowsValidator.ts
```diff
@@ -454,13 +454,24 @@ export class GridRowsValidator {
                         .join(', ')
             );
         }
+
+        // Hack: Want to re-use the `_getGroupSelectsDescendants` utility, since the underlying code
+        // is a bit annoying to reproduce here.
+        const gos = { get: (prop) => gridRows.api.getGridOption(prop) } as GridOptionsService;
```
SalvatorePreviti: nice trick :) but I think we should not rely on internal methods when we can avoid it, and it seems we can replicate the same logic to detect groupSelectsDescendants just replicating part of what _getGroupSelectsDescendants does as is simple and isolated enough

there are few general principles why we want to reduce access to internal functions and classes as much as we can:

1. we can refactor freely internal and private code without having to fix the test, as the test should rely only on public api and browser state, one day we might want to remove _getGroupSelectsDescendants for example or change how it works or change how gridOptions works and now we have a test that depend on the internal implementation we will have to fix. In theory, we should not need to change any test when we do just a pure refactoring as the test are supposed to test the behaviour independently from the implementation
2. if we change the implementation of _getGroupSelectsDescendants and introduce a bug we might not be able to see the mistake as the test code relies on this code itself to execute the tests, better to replicate its functionality for this case
3. we should never mock our own code under test, as if we introduce a bug in our mocked code we don't see it breaking, we should mock only things we do not write, like browser functionalities or network or external resources, not our own code

SalvatorePreviti: Pushed changes to handle this better and replicate the functionality

StephenCooper (approved):

PR #12821, opened 2026-01-05
AG-16440 - improved formula token handling

gportela85:
fix AG-16440


packages/ag-grid-enterprise/src/widgets/formulaRangeUtils.ts
```diff
@@ -0,0 +1,144 @@
...
+        return;
+    }
+
+    const { rangeClass } = getColorClassesForRef(ref, colorIndex);
+    range.colorClass = rangeClass;
+};
+
+// Range helpers
+export const getCellRangeParams = (beans: BeanCollection, ref: string) => {
+    const match = /^\$?([A-Za-z]+)\$?(\d+)(?::\$?([A-Za-z]+)\$?(\d+))?$/.exec(ref);
```
StephenCooper: Should we pull this regex out like the ones above to give it a descriptive name?

StephenCooper (approved):

PR #12827, opened 2026-01-07
AG-16302 fix positioning of suppressMoveWhenColumnDragging marker

BernieSumption:


packages/ag-grid-community/src/columnMove/columnDrag/moveColumnFeature.ts
```diff
@@ -393,11 +393,21 @@ export class MoveColumnFeature extends BeanStub implements DropListener {
...
+                position = ColumnHighlightPosition.Before;
+            } else {
+                position = ColumnHighlightPosition.After;
+                targetColumn = consideredColumns[targetIndex - 1];
+            }
         } else {
             position = ColumnHighlightPosition.After;
         }
 
+        if (this.lastHighlightedColumn?.column !== targetColumn || this.lastHighlightedColumn?.position !== position) {
```
SalvatorePreviti, line 406: I believe we should somehow simplify the logic of setColumnHighlighted that should be responsible of checking if column changed or highlight position changed and update both last highlight and lastHighlightedColumn field - however I would not worry to much right now and we can refactor this better when we run into it again

SalvatorePreviti (approved):

PR #12829, opened 2026-01-07
AG-16509 tree data sorting

SalvatorePreviti:
- fix a regression in version 35 that causes tree data filler nodes and groups to not be sorted properly
- minor optimizations in generic.ts to avoid passing an option object (checked with dash it is not used and safe to change)
- optimizations and cleanup of rowNodeSorter and sortStage
    - relying on ES2019 stable sort, we do not need to compare indices
    - reduced numbers of arrays allocated and copies and temporary objects creation
    - fixes all the rest of the project that was using the old way of passing rows and indices in a new object
- added more tests and benchmarks

Fix AG-16509

old

Screenshot 2026-01-07 at 13 48 50


packages/ag-grid-community/src/agStack/utils/generic.ts
```diff
@@ -3,12 +3,12 @@
  * @param {T} value
  * @returns {T | null}
  */
-export function _makeNull<T>(value?: T): T | null {
+export const _makeNull = <T>(value?: T): T | null => {
```
sdwvit, line 6: why this change?

packages/ag-grid-community/src/agStack/utils/generic.ts
```diff
@@ -21,72 +21,41 @@ export function _missing(value: any): boolean {
...
     // this is for aggregations sum and avg, where the result can be a number that is wrapped.
     // if we didn't do this, then the toString() value would be used, which would result in
     // the strings getting used instead of the numbers.
-    if (valueA?.toNumber) {
+    if (typeof valueA !== 'number' && valueA.toNumber) {
         valueA = valueA.toNumber();
     }
 
-    if (valueB?.toNumber) {
+    if (typeof valueB !== 'number' && valueB.toNumber) {
```
sdwvit: Are these two conditions overlapping? I can't imaging `number` having `toNumber` method on it.

packages/ag-grid-community/src/agStack/utils/generic.ts
```diff
@@ -21,72 +21,41 @@ export function _missing(value: any): boolean {
...
 }
 
-export function _toStringOrNull(value: any): string | null {
+export const _toStringOrNull = (value: any): string | null => {
     return value != null && typeof value.toString === 'function' ? value.toString() : null;
-}
+};
 
-export function _jsonEquals<T1, T2>(val1: T1, val2: T2): boolean {
+export const _jsonEquals = <T1, T2>(val1: T1, val2: T2): boolean => {
```
sdwvit, line 28: making all these arrow functions renders call frames anonymous in a call stack, making it difficult to debug

packages/ag-grid-community/src/clientSideRowModel/sortStage.ts
```diff
@@ -164,72 +172,69 @@ const doDeltaSort = (
...
     // add remaining from arr1
     while (i < arr1Length) {
-        res[k++] = arr1[i++].rowNode;
+        result[k++] = arr1[i++].node;
     }
 
     // add remaining from arr2
     while (j < arr2Length) {
-        res[k++] = arr2[j++].rowNode;
+        result[k++] = arr2[j++].node;
```
sdwvit: I would appreciate if these in-place increments were extracted into a separate line.

sdwvit: otherwise this code seems fishy.

packages/ag-grid-community/src/sort/rowNodeSorter.ts
```diff
@@ -10,58 +9,56 @@ import type { RowNode } from '../entities/rowNode';
...
 
-    public compareRowNodes(sortOptions: SortOption[], sortedNodeA: SortedRowNode, sortedNodeB: SortedRowNode): number {
-        const nodeA: RowNode = sortedNodeA.rowNode;
-        const nodeB: RowNode = sortedNodeB.rowNode;
+    public compareRowNodes(sortOptions: SortOption[], nodeA: RowNode, nodeB: RowNode): number {
+        const accentedCompare = this.accentedSort;
 
         // Iterate columns, return the first that doesn't match
-        for (let i = 0, len = sortOptions.length; i < len; i++) {
+        for (let i = 0, len = sortOptions.length; i < len; ++i) {
```
sdwvit, line 60: what is the difference?

packages/ag-grid-community/src/agStack/utils/generic.ts
SalvatorePreviti: just to be more consistent and always use const instead of function when possible, it also minifies better potentially if we or the user doesn't have a setting to automatically switch to lambda when `this` is not used during minification

packages/ag-grid-community/src/agStack/utils/generic.ts
SalvatorePreviti: it is equivalent - is not an anonymous function, is a named function as browser will use the name of the variable

packages/ag-grid-community/src/clientSideRowModel/sortStage.ts
SalvatorePreviti: mhhh this code was always here and was optimized quite a few PR ago but was always here - but is a standard implementation of sorted array merging https://en.wikipedia.org/wiki/Merge_algorithm

sdwvit: ah these cpp programmers and their optimisations :)

SalvatorePreviti: Moved increments in a new line

packages/ag-grid-community/src/agStack/utils/generic.ts
sdwvit: you are absolutely right, I must have outdated info 
Screenshot 2026-01-07 at 12 55 50

packages/ag-grid-community/src/agStack/utils/generic.ts
SalvatorePreviti: I changed now this code to fix sonar warning, and, wrote a better condition checking for typeof === object - the idea is that we should not try to access a property on a number or string as it is indeed pretty slow as it causes the value to be boxed to a String or Number object and access the prototype, better first to test if is an object

packages/ag-grid-community/src/sort/rowNodeSorter.ts
SalvatorePreviti: ask safari :D ++i seems to be faster than i++ but also trying to be consistent, I always use prefix increment when we do not use the return value of an increment

packages/ag-grid-community/src/sort/rowNodeSorter.ts
```diff
@@ -136,37 +137,38 @@ export class RowNodeSorter extends BeanStub implements NamedBean {
...
+        const beans = this.beans;
+
+        if (this.primaryColumnsSortGroups) {
+            if (node.rowGroupColumn === column) {
+                return this.getGroupDataValue(node, column);
+            }
 
-        if (node.group && column.getColDef().showRowGroup) {
-            return undefined;
+            if (node.group && column.getColDef().showRowGroup) {
```
StephenCooper, line 147: Is this the actual fix for the issue? Moving inside the primaryColumnsSortGroups flag.

SalvatorePreviti: the actual fix is in this function, the first if condition was not correct and we were not updating primaryColumnsSortGroups field if treeData grid option changed - treeData grid option is read by _isColumnsSortingCoupledToGroup

StephenCooper (approved):

PR #12835, opened 2026-01-08
Add Agentic tooling support + refactor existing shared tooling.

alantreadway:


external/ag-shared/prompts/commands/batch/lint-cleanup.md
```diff
@@ -0,0 +1,208 @@
...
+    git add -A
+    git commit -m "fix(lint): manually fix <rule-name> violations"
+    ```
+
+---
+
+### Common Auto-Fixable Rules Reference
+
+| Rule                                | Auto-Fix | Typical Change               |
+| ----------------------------------- | -------- | ---------------------------- |
```
StephenCooper, line 187: We have not got all of these rules enabled in our lint config.

alantreadway: This table should only be used when actual violations are found, so I think that should be fine - no rule violations, no action taken.

alantreadway:
Dropped in favor of https://github.com/ag-grid/ag-grid/pull/12842

PR #12836, opened 2026-01-08
AG-13371 Configure Modules and LicenceKey via React.Context

StephenCooper:
Fix AG-13371

Most of the changes are to provided examples in the documentation to use the new pattern to match the example generator update.


StephenCooper:
@codex

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/grid-vanilla-to-react-functional-ts.ts
```diff
@@ -92,6 +88,16 @@ function getImports(
     if (useFetchHook) {
         imports.push(`import { useFetchJson } from './useFetchJson';`);
     }
+
+    if (bindings.moduleRegistration) {
+        // Modules registration is different in React - we just export the modules array and pass it to AgGridProvider
+        const reactModuleRegistration = bindings.moduleRegistration.replace(
+            /ModuleRegistry\.registerModules\(\s*\[([\s\S]*?)\]\s*\);/,
```
sdwvit, line 95: can it be called multiple times? 
disregard if you know our examples are safe

sdwvit (approved):
reviewed everything outside of documenations changes

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/grid-vanilla-to-react-functional-ts.ts
StephenCooper: Yes this can be called multiple times. Beans are created when the Grid is created, modules being registered just puts them into the registry.

PR #12840, opened 2026-01-08
AG-16517 fix coldef refresh

SalvatorePreviti:
Currently we recreate total and pivot columns without checking if something changed.
This is because we were using Object.assign to inherit by copying the values of the previous colDef and we didn't implement preserving the column itself.

This PR fixes these problems
1 - keep always the inherited values in sync as they aren't copied but inherited via prototype (Object.create(valueColDef))
2 - columns recreated without the need to create the column
3 - context lost every applyColunnState or refresh that would cause the column to be recreated


StephenCooper (approved):

SalvatorePreviti:
There are still issues to solve with this

PR #12846, opened 2026-01-08
AG-16440 - Formulas Editor - TC Fixes

gportela85:
fix AG-16440


packages/ag-grid-enterprise/src/widgets/formulaInputRangeSyncFeature.ts
```diff
@@ -86,41 +109,102 @@ export class FormulaInputRangeSyncFeature extends BeanStub {
     }
 
     private clearTrackedRanges(): void {
-        const refs = Array.from(this.trackedRangeRefs);
+        const refs = Array.from(this.trackedRangeRefs.keys());
```
StephenCooper, line 112: Can we clean up the few cases where we are wrapping an array with Array.from? I will merge the PR without waiting for these changes to be made as they are not critical but will just be nice to clean up.

StephenCooper (approved):

PR #12849, opened 2026-01-09
Test module size reporting

StephenCooper:
Fix AG-16432

This PR adds a new PR workflow that reports the changes in Modules into the PR as a comment to make this visible during the review stage. 

The idea would be to disable the module size task that fails CI if this proves to be effective.


taktran (approved):
LGTM

PR #12863, opened 2026-01-12
AG-16432 Run scripts from latest

StephenCooper:


.github/actions/module-size-test/action.yml
```diff
@@ -21,16 +21,36 @@ inputs:
...
+        ref: ${{ inputs.actions-ref }}
+        sparse-checkout: .github/actions
+        path: .actions-temp
+
+    - name: Copy actions from actions-ref
+      if: inputs.actions-ref != '' && inputs.actions-ref != inputs.ref
+      shell: bash
+      run: |
+        rm -rf .github/actions
+        cp -r .actions-temp/.github/actions .github/actions
```
taktran, line 51: What if there are some actions/files required from `external`?

taktran (approved):
LGTM - let's see how it works in practice

PR #12868, opened 2026-01-12
AG-16229  - [formula-cell-editor] - bug fixes

gportela85:
fix AG-16229

### DIFF
--- a/packages/ag-grid-enterprise/src/widgets/formulaInputRangeSyncFeature.ts
+++ b/packages/ag-grid-enterprise/src/widgets/formulaInputRangeSyncFeature.ts
@@ -15,11 +15,14 @@ import {
 type TrackedRange = { ref: string; tokenIndex?: number | null };
 
 export class FormulaInputRangeSyncFeature extends BeanStub {
+    // Track the active editor instance per grid/cell to avoid overlapping syncs on editor restarts.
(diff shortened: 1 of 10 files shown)


StephenCooper:
Looking into why all the Enterprise modules have gained 13kb. I have validated that this is happening locally with this change and is not an issue with the new reporting.

Caused by https://github.com/ag-grid/ag-grid/pull/12868/changes#r2686156122

packages/ag-grid-enterprise/src/widgets/formulaInputRangeSyncFeature.ts
```diff
@@ -15,11 +15,14 @@ import {
 type TrackedRange = { ref: string; tokenIndex?: number | null };
 
 export class FormulaInputRangeSyncFeature extends BeanStub {
+    // Track the active editor instance per grid/cell to avoid overlapping syncs on editor restarts.
+    private static readonly activeEditorsByContext = new WeakMap<object, Map<string, FormulaInputRangeSyncFeature>>();
```
StephenCooper: This line of code prevents this class from being tree shakeable. The static property is the issue. 

Can we move this outside of the class just as a standalone const? This would resolve the tree shaking.

StephenCooper (approved):

PR #12888, opened 2026-01-15
AG-16541 avoid double cell editable callback call

SalvatorePreviti:
This refactoring reduces the number of times we invoke the user defined editabe callback in the colDef when we have already invoked it.

- add a 'editable:boolean' property to StartEditParams - if is not undefined, the isCellEditable method will not be invoked and its value is used instead
- pass editable option where applicable where we already invoked isCellEditable
- optimize editors code to remove unnecessary array allocations and speed up validation checks


StephM-W (approved):

packages/ag-grid-community/src/edit/strategy/fullRowEditStrategy.ts
```diff
@@ -268,7 +268,13 @@ export class FullRowEditStrategy extends BaseEditStrategy {
             if (suppressStartEditOnTab) {
                 nextCell.focusCell(true, event);
             } else {
-                this.editSvc.startEditing(nextCell, { startedEdit: true, event, source, ignoreEventKey: true });
+                this.editSvc.startEditing(nextCell, {
+                    startedEdit: true,
+                    event,
+                    source,
+                    ignoreEventKey: true,
+                    editable: nextEditable,
```
gportela85: @SalvatorePreviti  This introduces a regression where Full‑row edit navigation no longer starts editing when tabbing into a row where the next focused cell is non‑editable, because editable: nextEditable short‑circuits startEditing and skips the prior isCellEditable('api') check that allowed editing if any column in the row was editable.

To reproduce this, load this plunker: https://plnkr.co/edit/5ahB2ulo1Fhkuy4i?open=main.js

Start editing the first row, tab from the last cell into the next row.

gportela85 (changes requested):

packages/ag-grid-community/src/edit/strategy/fullRowEditStrategy.ts
SalvatorePreviti: pushed a fix and a test

gportela85 (approved):

PR #12894, opened 2026-01-15
Bundle size report updates

StephenCooper:
Removing bundle size from CI and instead running the comparison against last release on latest to keep track of it.

Updated the report to include %change with highlighting.


packages/ag-grid-enterprise/src/find/findModule.ts
```diff
@@ -49,5 +51,5 @@ export const FindModule: _ModuleWithApi<_FindApi<any>> = {
         findClearActive,
         findRefresh,
     },
-    dependsOn: [EnterpriseCoreModule, FindCoreModule],
+    dependsOn: [EnterpriseCoreModule, FindCoreModule, PivotModule],
```
StephenCooper: For testing only!

taktran (approved):
LGTM

PR #12898, opened 2026-01-16
AG-16541 group row editable

SalvatorePreviti:
- add groupRowEditable and groupRowValueSetter colDef properties so group nodes can be edited directly and cascade changes through their descendants
- stop creating data for group rows when setting the value of a column that has groupRowEditable or groupRowValueSetter - that is a legacy behaviour when using those two colDef properties
- document the workflow and ship a refreshed example plus a test 
- fix the row drag and drop to use rowNode instead of editSvc to support correctly the new property and readonly edit

All the new behavior is triggered only by the presence of groupRowEditable or groupRowValueSetter and if the row is a group

What's missing
- The default implementation for groupRowValueSetter for aggregations, group rows, autoGroupColDef
- ReadOnly edit is currently not handled and will be done in the next PRs
- undo/redo is not complete
- proper documentation, and documentation for treeData as well - discuss where is better to put both of those

FIX AG-16541


packages/ag-grid-enterprise/src/rowHierarchy/groupEditService.ts
```diff
@@ -479,8 +479,7 @@ export class GroupEditService extends BeanStub implements _IGroupEditService {
             if (parsedValue !== undefined) {
                 valueToSet = parsedValue;
             }
-            const result = editSvc?.setDataValue({ rowNode: row, column }, valueToSet, 'rowDrag');
-            const updated = result != null ? !!result : row.setDataValue(column, valueToSet, 'rowDrag');
+            const updated = row.setDataValue(column, valueToSet, 'rowDrag');
```
SalvatorePreviti, line 482: this fixes a potential bug with readonlyData and with the new row group editable - we need to invoke the row setDataValue and this will work as intended

SalvatorePreviti:
@codex

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -401,60 +402,103 @@ export class ValueService extends BeanStub implements NamedBean {
...
-            newValue,
-            params,
-            rowData: rowNode.data,
-            valueSetter: colDef.valueSetter,
-            field: colDef.field,
-        });
+        // group rows only persist through row data or a custom groupRowValueSetter; without either we intentionally skip writes
+        // this will be addressed by providing a default groupRowValueSetter for group columns, auto group columns, aggregation columns
+        const canUpdateValue =
+            !rowNode.group || rowNode.data || groupRowValueSetter == null || colDef.groupRowEditable == null;
```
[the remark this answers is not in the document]
SalvatorePreviti: this is a desired behaviour as we still do not have a default implementation for the groupRowValueSetter that will be introduced in the next pull requests. A group row without data should not by default, when using the new two propertie, create row data - and the user should be able to define their own logic when using this in a non aggregation and non auto group or row group column - this is still a work in progress

SalvatorePreviti:
@codex

StephM-W (approved):

documentation/ag-grid-docs/src/content/docs/grouping-edit/_examples/group-row-editable-totals/main.ts
```diff
@@ -0,0 +1,102 @@
...
+        },
+    },
+    defaultColDef: {
+        flex: 1,
+        sortable: true,
+        filter: true,
+        resizable: true,
+    },
+    rowData: getData(),
+    enableGroupEdit: true,
```
StephenCooper: Is this flag still required?

SalvatorePreviti: no that was a mistake! removing it - the new property takes precedence.
pushed new changes

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -401,60 +402,107 @@ export class ValueService extends BeanStub implements NamedBean {
...
-            params,
-            rowData: rowNode.data,
-            valueSetter: colDef.valueSetter,
-            field: colDef.field,
-        });
+        // group rows only persist through row data or a custom groupRowValueSetter; without either we intentionally skip writes
+        // this will be addressed by providing a default groupRowValueSetter for group columns, auto group columns, aggregation columns
+        const canUpdateValue =
+            !rowNode.group || rowNode.data || (groupRowValueSetter == null && colDef.groupRowEditable == null);
+
```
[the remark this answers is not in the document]
SalvatorePreviti: valueSetter requires data to not be null or undefined, for this reason it should not be invoked. and we should not create rowData when setting a value to one of this properties, this is let to the user.
The default implementation of groupRowValueSetter will take care of setting aggregation values for known aggregations and update row groups - this is still a work in progress but updated the documentation to reflect this

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -401,60 +402,107 @@ export class ValueService extends BeanStub implements NamedBean {
...
+                    data: rowNode.data,
+                    oldValue,
+                    newValue,
+                    colDef,
+                    column,
+                    eventSource,
+                })
+            );
+
+            if (result) {
```
StephenCooper: Should we handle the user not returning a value in the same was as the valueSetter?

SalvatorePreviti: I am actually not sure and I spent quite a bit of time thinking about this ... shall we consider no change if the user return nothing or shall we consider it a change? ok let me make it in such a way if the user return nothing we consider it a change for now and we can think about this in the future when we look again at undo-redo

StephenCooper: As this is a new api and we document the boolean we can keep things simple and just do what they tell us. The valueSetter behaviour is definitely more a legacy help. 

It might be sensible for us to leave it as it is and require users to implement the interface correctly.

testing/behavioural/src/cell-editing/group-edit/group-row-editable-aggregation.test.ts
```diff
@@ -0,0 +1,319 @@
...
+    });
+
+    test('group edits over filtered groups only adjust filtered descendants', async () => {
+        const rowData = createRowData();
+
+        const api = await gridsManager.createGridAndWait('group-row-editable-filtered', {
+            defaultColDef: {
+                cellEditor: 'agTextCellEditor',
+            },
+            enableGroupEdit: true,
```
StephenCooper: Is this required?

SalvatorePreviti: yeah in this case I'd say yes - we are testing editable is called when enableGroupEdit is true but the defaultcolDef is not needed and removing it

packages/ag-grid-community/src/valueService/valueService.ts
SalvatorePreviti: well is ok to return undefined - for now i convert it to true but is something we can rethink

packages/ag-grid-community/src/columns/dataTypeService.ts
```diff
@@ -542,8 +542,8 @@ export class DataTypeService extends BeanStub implements NamedBean {
                 cellEditor: 'agCheckboxCellEditor',
                 cellRenderer: 'agCheckboxCellRenderer',
                 getFindText: () => null,
-                suppressKeyboardEvent: (params: SuppressKeyboardEventParams<any, boolean>) =>
-                    !!params.colDef.editable && params.event.key === KeyCode.SPACE,
+                suppressKeyboardEvent: ({ colDef, node, event }: SuppressKeyboardEventParams<any, boolean>) =>
+                    !!(colDef.editable || (node.group && colDef.groupRowEditable)) && event.key === KeyCode.SPACE,
```
gportela85: Here `groupRowEditable` is treated as a plain boolean, so callbacks that return false for specific group rows still suppress space handling on all group rows

gportela85 (changes requested):
One small issue related to [dataTypeService.ts](https://github.com/ag-grid/ag-grid/pull/12898/changes/BASE..a03fcef453640b648495a655e805ba843aea20a2#diff-8797a990ac4f5039cfe428d875504c716509154d5d702c1cf5e368596cb2d6d7R546)

documentation/ag-grid-docs/src/content/docs/grouping-edit/_examples/group-row-editable-totals/main.ts
```diff
@@ -0,0 +1,101 @@
...
+    return Number.isFinite(numericValue) ? numericValue : params.oldValue ?? null;
+};
+
+const amountGroupRowValueSetter: GroupRowValueSetterFunc<SalesRecord> = ({ node, newValue, eventSource }) => {
+    const numericValue = Number(newValue);
+    if (!Number.isFinite(numericValue)) {
+        return;
+    }
+
+    const children = node.childrenAfterSort;
```
StephenCooper, line 51: I think this would benefit from a comment to say it is sharing the value across all children

Also worth commenting line 53 that this will handle recursive groups.

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
```diff
@@ -3,23 +3,67 @@ title: "Row Grouping - Editing Groups"
 enterprise: true
 ---
 
-The Grid supports editing of grouped data when using the [Client-Side Row Model](./row-models/#client-side).
-When users edit values in grouped columns, the grid can automatically refresh the grouping hierarchy to reflect the changes.
+The grid supports editing grouped data when using the
+[Client-Side Row Model](./row-models/#client-side). This page explains how to keep the grouping
```
StephenCooper: Are these intentional new lines?

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
```diff
@@ -3,23 +3,67 @@ title: "Row Grouping - Editing Groups"
...
+`valueSetter` never runs even if `groupRowEditable` is enabled. Provide a `groupRowValueSetter`
+whenever the edit needs to persist or update aggregates; only group rows that own real data objects
+run the normal value pipeline.
+
+When a column defines both `groupRowEditable` and `editable`, AG Grid only evaluates the property
+that matches the current node type, enabling separate rules for group rows and leaves.
+
+## Custom Distribution of Edits
+
+In this example, we show how to use `groupRowValueSetter` to distribute edits on group rows to their descendant rows.
```
StephenCooper: Prefer a passive voice, for example

In this example `groupRowValueSetter` is used to distribute....

documentation/ag-grid-docs/src/content/docs/grouping-edit/_examples/group-row-editable-totals/main.ts
SalvatorePreviti: done

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
SalvatorePreviti: is just code formatting they will not appear as new lines, md will add a new line only when there is an empty line

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
SalvatorePreviti: done

SalvatorePreviti:
> One small issue related to [dataTypeService.ts](https://github.com/ag-grid/ag-grid/pull/12898/changes/BASE..a03fcef453640b648495a655e805ba843aea20a2#diff-8797a990ac4f5039cfe428d875504c716509154d5d702c1cf5e368596cb2d6d7R546)

done

gportela85 (approved):

PR #12900, opened 2026-01-16
AG-16287 theming api selector performance improvements

BernieSumption:
Fix AG-16287


.stylelintrc.js
```diff
@@ -48,19 +50,5 @@ module.exports = {
                 ],
             },
         ],
-
-        // NOTE: In general we want to avoid targeting grid elements using
-        // [class^='ag-'] as customer applications can have elements with that
-        // prefix too. Sometimes it is unavoidable, e.g. for global style
-        // resets, in which case scope the selector so it is only applied within
-        // the grid root.
-        'selector-disallowed-list': [
```
StephenCooper, line 57: Do we not want this rule anymore?

packages/ag-grid-community/src/agStack/widgets/agAbstractLabel.css
```diff
@@ -8,15 +8,21 @@
...
 
 .ag-label-align-top {
     flex-direction: column;
     align-items: flex-start;
+}
 
-    > * {
+:where(.ag-label-align-top) {
+    .ag-label,
+    .ag-wrapper {
```
StephenCooper, line 25: Is this change based on what we have put inside of `ag-label-align-top`?

.stylelintrc.js
BernieSumption: [class^='ag-'] Is a low performance selector according to the new lint rule, so this is superfluous.

packages/ag-grid-community/src/agStack/widgets/agAbstractLabel.css
BernieSumption: Yes, whenever we could determine the set of classes that are valid children of an element, I replaced `> *` with the list of classes.

packages/ag-grid-community/src/filter/column-filters.css
```diff
@@ -63,16 +64,17 @@
...
     display: flex;
     width: 100%;
 
-    > * {
+    /* most filters */
+    > :where(.ag-input-field),
+    /* date filters */
+    > :where(.ag-date-floating-filter-wrapper),
+    /* nested multi filters */
+    > :where(.ag-floating-filter-input) {
```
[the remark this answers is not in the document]
StephenCooper: Can we validate if this is true?

packages/ag-grid-community/src/filter/column-filters.css
```diff
@@ -197,17 +199,14 @@
 .ag-simple-filter-body-wrapper {
     display: flex;
     flex-direction: column;
+    gap: var(--ag-widget-vertical-spacing);
```
StephenCooper, line 202: Is this equivalent?

packages/ag-grid-community/src/filter/column-filters.css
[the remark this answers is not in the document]
BernieSumption: I checked, custom floating filters are not direct children of this element so this concern does not apply.

packages/ag-grid-enterprise/src/sideBar/agSideBar.css
```diff
@@ -24,10 +24,6 @@
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
-
-    > * {
```
StephenCooper, line 28: Is this missing a replacement?

packages/ag-grid-enterprise/src/widgets/pillDropZonePanel.css
```diff
@@ -63,10 +63,6 @@
     padding-bottom: var(--ag-spacing);
     padding-right: var(--ag-spacing);
     padding-left: var(--ag-spacing);
-
-    > * {
```
StephenCooper, line 67: Missing a replacement?

packages/ag-grid-community/src/filter/column-filters.css
BernieSumption: There were several places in which I replaced a `> * { margin-(bottom|right): ... }` with a flex gap and in all those cases I was doing it looking at UI and checking that it looked the same. There are edge cases where the pattern can not work, but it should be fine with these changes.

packages/ag-grid-enterprise/src/sideBar/agSideBar.css
BernieSumption: This was dead code, `.ag-select-agg-func-item` is a leaf node containing text, as far as I can tell it never has element children.

packages/ag-grid-enterprise/src/widgets/pillDropZonePanel.css
BernieSumption: This wasn't required, the children (column drop pills) have an explicit height so don't need flex shrinking disabled. Presumably in a previous version this rule was required because the height of column drop pills was not explicit.

taktran (approved):
LGTM - stylelint plugin looks good 👍🏼

PR #12907, opened 2026-01-18
AG-16219 - Add highlight for RHS nav

markdurrant:


documentation/ag-grid-docs/src/components/pages-navigation/components/SideNavigation.tsx
```diff
@@ -32,11 +58,25 @@ export function SideNavigation({ headings, delayedScrollSpy }: Props) {
...
                         </li>
                     ))}
                 </ul>
+                <div className={styles.backToTop}>
+                    <a
+                        href="#top"
+                        className="nav-link"
+                        onClick={(event) => {
+                            event.preventDefault();
+                            scrollIntoViewById('top');
```
kylerphillips, line 51: @markdurrant I added the `scrollIntoViewById` which we use in the sideNav, what do you think? I think unless we enabled smooth scroll behaviour with css global in html the click to scroll by look "not smooth"

taktran (approved):
LGTM

documentation/ag-grid-docs/src/components/pages-navigation/components/SideNavigation.tsx
markdurrant: Yeah good shout. This looks much nicer than straight href.  👍

PR #12912, opened 2026-01-19
AG-16432 Use merge base instead of latest

StephenCooper:
Avoid the need for merging latest to get accurate results if a PR branch is stale compared to latest


taktran (approved):
LGTM

PR #12924, opened 2026-01-20
AG-16607 set value called twice

SalvatorePreviti:
The logic in EditService was not correct causing in some causes a double valueSetter execution.
Together with this, I added a full set of tests to verify valueSetter is called properly and the right number of times in different scenarios including batch editing, full row editing, paste, undo/redo.
I will need also those tests for regression testing for the upcoming works on group row edit.

The previous implementation caused also stopEditing to be invoked on cells not being edited and multiple editing events

- fix EditService set data value logic
- add tests for set value and edit

TC1:

Open https://plnkr.co/edit/iqkchfOukGvn8pBp?open=main.js
- click on one cell and copy a value with CTRL+C
- click on another cell and press CTRL+V
- open the console

Expected: only one console log line for value setter printed

Actual: two equal console log lines for value setter printed

FIX AG-16607

### DIFF
--- a/packages/ag-grid-community/src/edit/editService.ts
+++ b/packages/ag-grid-community/src/edit/editService.ts
@@ -898,13 +898,14 @@ export class EditService extends BeanStub implements NamedBean, IEditService {
 
     public setDataValue(position: Required<EditPosition>, newValue: any, eventSource?: string): boolean | undefined {
         try {
-            const editing = this.isEditing(position); // check if the cell is being edited
+            const batch = this.batch;
+            const editing = this.isEditing(batch ? undefined : position); // check if the cell is being edited
 
             if ((!editing || this.committing) && !SET_DATA_SOURCE_AS_API.has(eventSource)) {
                 return; // Ignore non-edit edits that are not treated as API sources.
             }
 
-            if (!editing && !this.batch && eventSource === 'paste') {
+            if (!editing && !batch && eventSource === 'paste') {
                 return; // Paste on non editable cells and not batching
             }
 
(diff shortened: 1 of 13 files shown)


SalvatorePreviti:
@codex

packages/ag-grid-community/src/edit/editService.ts
```diff
@@ -898,11 +898,17 @@ export class EditService extends BeanStub implements NamedBean, IEditService {
...
         try {
-            if ((!this.isEditing() || this.committing) && !SET_DATA_SOURCE_AS_API.has(eventSource)) {
-                return;
+            const editing = this.isEditing(position); // check if the cell is being edited
+
+            if ((!editing || this.committing) && !SET_DATA_SOURCE_AS_API.has(eventSource)) {
+                return; // Ignore non-edit edits that are not treated as API sources.
+            }
+
+            if (!editing && !this.batch && eventSource === 'paste') {
```
StephM-W: How does this work when using paste to 'edit' an editable cell?

SalvatorePreviti: it will not return and will continue execuing the rest of the code. The condition here if is not batching and is not editing and we are pasting we can just call setDataValue from the rowNode skipping edit svc

SalvatorePreviti: note - i pushed new changes but the logic for this is equivalent

SalvatorePreviti:
@codex

StephM-W (approved):

PR #12937, opened 2026-01-21
AG-16614 fix tree data promotion to group bug

SalvatorePreviti:
The logic to check if a leaf row can be promoted to group or expanded was not correct
This was introduced during the implementation of grouping drag-n-drop.
Test coverage didn't cover this case properly

Fix AG-16614


SalvatorePreviti:
@codex

StephenCooper (approved):

PR #12938, opened 2026-01-21
AG-16610 full row editing leaves leaves empty editor bug

SalvatorePreviti:
The problem seems to lie in the fact that the model does not track properly empty cells.
Iterating through all editor controls that are not yet destroyed and destroying them seems the right solution

FIX AG-16610

### DIFF
--- a/packages/ag-grid-community/src/edit/strategy/fullRowEditStrategy.ts
+++ b/packages/ag-grid-community/src/edit/strategy/fullRowEditStrategy.ts
@@ -6,7 +6,12 @@ import type { EditPosition, EditRowPosition, StartEditWithPositionParams } from 
 import type { IRowNode } from '../../interfaces/iRowNode';
 import type { CellCtrl } from '../../rendering/cell/cellCtrl';
 import { _getCellCtrl, _getRowCtrl } from '../utils/controllers';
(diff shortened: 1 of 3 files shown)


SalvatorePreviti:
@codex

packages/ag-grid-community/src/edit/strategy/fullRowEditStrategy.ts
```diff
@@ -188,12 +193,41 @@ export class FullRowEditStrategy extends BaseEditStrategy {
...
+    private destroyEditorsForRow(rowNode: IRowNode): void {
+        const rowCtrl = _getRowCtrl(this.beans, { rowNode });
+        if (!rowCtrl) {
+            return; // Row not rendered, no editors to destroy.
+        }
+
+        const destroyedColumns = new Set<AgColumn>();
+        for (const cellCtrl of rowCtrl.getAllCellCtrls()) {
+            const column = cellCtrl.column;
+            if (destroyedColumns.has(column)) {
```
StephenCooper, line 217: Why do we need to track columns? How can these be repeated?

SalvatorePreviti: I will remove the set as destroy is idempotent

gportela85 (approved):

PR #12942, opened 2026-01-21
AG-16509 fix delta sort algorithm

SalvatorePreviti:
Found a problem in maintaining order in the existing rows in the delta sort algorithm.
Moves the delta sort specific code in its own utils module and add more behavioural tests that covers all the edge cases


packages/ag-grid-community/src/clientSideRowModel/deltaSort.ts
```diff
@@ -0,0 +1,153 @@
+import type { RowNode } from '../entities/rowNode';
```
sdwvit, line 1: Moving this critical code out of original file makes it a bit difficult to review. Were there any changes made here? can you highlight exact fix?

testing/behavioural/src/sorting/sorting.test.ts
```diff
@@ -363,43 +363,6 @@ describe('Sorting', () => {
         `);
     });
 
-    test('delta sort keeps order', async () => {
```
sdwvit, line 366: was it just a refactoring?

packages/ag-grid-community/src/clientSideRowModel/deltaSort.ts
```diff
@@ -0,0 +1,153 @@
...
+    const untouchedRows = buildUntouchedArray(oldSortedRows, indexByNode, unsortedRowsLen - touchedRowsLen);
+
+    if (!touchedRowsLen) {
+        // No touched rows: return oldSortedRows if nothing removed, otherwise return untouched
+        return untouchedRows.length === oldSortedLen ? oldSortedRows : untouchedRows;
+    }
+
+    // Sort touched rows and keep a stable tie-breaker based on current index.
+    touchedRows.sort(
+        (a, b) => rowNodeSorter.compareRowNodes(sortOptions, a, b) || ~indexByNode.get(a)! - ~indexByNode.get(b)!
```
StephenCooper: `~indexByNode.get(a)!` does this convert to a boolean and so we are not actually comparing the index?

testing/behavioural/src/sorting/sorting.test.ts
SalvatorePreviti: I moved the tests in the new file delta-sorting.test.ts specific for delta-sorting as there are many more tests to run with more data and this file was way too big

packages/ag-grid-community/src/clientSideRowModel/deltaSort.ts
SalvatorePreviti: it swaps all the bits - we are using a single map node->32bit-number to store the index and mark the node as being or touched or untouched -
- x is >=0 if was not touched and the index can be used as is.
- x is < 0 (most significant bit is 1, so, is a negative 32 bit number) if was touched and ~x will give you the index

Note that array in javacript cannot in any case contain more than 2 billion elements so the algorithm works

packages/ag-grid-community/src/clientSideRowModel/deltaSort.ts
SalvatorePreviti: Yes there were changes in here as in some cases the original order was not respected. Looking at the diff does not help much as quite a few lines changed so looking from scratch does help better. This is by intention quite hard to read as the original algorithm that was present since the beginning was similarly complex. There is an extensive test coverage for all the corner cases with quite a lot of data and a benchmark

StephenCooper (approved):

PR #12951, opened 2026-01-22
Provide examples share top level spec

StephenCooper:


JamesSwinton (approved):
LGTM - didn't realise they shared them, thanks for catching

PR #12957, opened 2026-01-23
AG-3390 - Merge `b35.0.1` back into `latest`

taktran:


taktran:
Going to cherry pick individual changes instead of merging patch releases into latest.

PR #12961, opened 2026-01-23
AG-16587 onboarding styling guide

JamesSwinton:
Adds a tutorial for styling the grid with the theming API, custom CSS, and cell/row class rules to bring everything into one place for new users.


documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
```diff
@@ -0,0 +1,390 @@
...
+This tutorial assumes basic knowledge of AG Grid. If you haven't already, we recommend reading our [Introductory Tutorial](./deep-dive/) first.
+{% /note %}
+
+## Introduction to Styling
+
+AG Grid provides three main approaches to styling:
+
+- **Theming API:** Fully-typed API for customising default themes.
+- **Custom CSS:** Fine-grained control over specific grid elements.
+- **Cell/Row Styles:** Data-driven formatting for individual cells and rows.
```
BernieSumption, line 32: The 4th level here is custom components. Even if you don't have them in the tutorial it's worth mentioning as a next potential step for when creating unique designs.

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
```diff
@@ -0,0 +1,390 @@
...
+
+Each of these approaches can be used to customise our built-in themes: [Quartz](https://www.ag-grid.com/example), [Alpine](https://www.ag-grid.com/example?theme=alpine), [Balham](https://www.ag-grid.com/example?theme=balham), and [Material](https://www.ag-grid.com/example?theme=material).
+
+Our themes are simply JavaScript objects that define colours, spacing, fonts, and other design tokens. By default, the Quartz theme is applied; to use a custom theme, create a copy of the desired theme and pass it to the grid options:
```
BernieSumption: > create a copy of the desired theme

The code snippet below doesn't create a copy, just assigns the theme to a temporary variable. By contrast, `const myTheme = themeQuartz.withParams(...)` or `.withPart(...)` would create a copy.

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
```diff
@@ -0,0 +1,390 @@
...
+
+When you run your application, you should see the new theme applied to the grid: 
+
+{% gridExampleRunner title="Theme Parameters" name="theme-params" suppressDarkMode=true  /%}
+
+{% note %}
+Our [Theme Parameters](./theming-parameters/) docs provides a full list of available parameters and how to use them.
+{% /note %}
+
+### Theme Modes
```
BernieSumption, line 349: I see Theme Modes as a niche feature for people with a relatively rare use case where the grid is used in a website and different codebases own the grid code and the dark mode toggle code. The reason we added it is for our website to help with applying dark mode to examples, and with the thought that if we find it useful then some customers with complex use cases would too.

That said, arguably it's the easiest way to implement dark mode in many apps.

Anyway, the "official" way to implement dark mode in most apps is to make a dark and light version of a theme and switch between them by passing the correct current version into the theme prop. It makes sense to only show one of these techniques, I guess you should use whichever seems simpler and more beginner-friendly.

BernieSumption (approved):

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
```diff
@@ -0,0 +1,390 @@
...
+});
+
+const gridOptions = {
+    theme: myTheme,
+    // ... other options
+};
+```
+{% /if %}
+
+When you run your application, you should see the new theme applied to the grid:
```
BernieSumption: This example is visually pretty similar to the default theme, not sure if you want to make some more obvious changes to make a point.

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
JamesSwinton: Cell Renderers are covered in the previous tutorial. I'll add a note to this section.

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
JamesSwinton: Changed the wording slightly to "reference" rather than "copy" an existing theme

documentation/ag-grid-docs/src/content/docs/styling-tutorial/index.mdoc
JamesSwinton: Good point - I've decided to leave as-is. As you say, I think this approach is probably easier for many apps, and the concept of updating the theme by changing the grid option is pretty self-explanatory, whereas this approach needs a bit more context.

PR #12969, opened 2026-01-26
AG-16634 backpspace on unedited cell

SalvatorePreviti:
- separate getCellDataValue from getCellValueForDisplay called by valueSvc
- handle properly the edge case when value service value getter is invoked while stopping editing, by returning the correct value
- remove hasEditSvc as we don't need it anymore in valueSvc
- remove unused preferEditor from getCellDataValue - as it is always called with false, and simplify the logic reducing nullish operator
- fix _persistEditorValue to work correctly and guard against UNEDITED not being accidentally returned to the user
- remove the copy params from getEdit and instead pass the params, to reduce the number of repetitive calls to getEdit, and allow getEdit to request for siblings as well
- add a test
- improve internal testing documentation

FIX AG-16634


packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -193,16 +191,10 @@ export class ValueService extends BeanStub implements NamedBean {
 
         // pull these out to make code below easier to read
 
-        if (this.hasEditSvc && source === 'ui') {
```
StephM-W, line 196: Could you double check this? `hasEditSvc` was added for performance reasons, we'd benchmarked and `this.editSvc?.` was measurably slower than the boolean check, hence the guard.

StephM-W (approved):

packages/ag-grid-community/src/valueService/valueService.ts
SalvatorePreviti: No performance degradation - now invokes just a method moving all the logic to the editSvc

`latest` without edit module
Screenshot 2026-01-26 at 11 04 46

`latest` with editable column
Screenshot 2026-01-26 at 11 05 21

this branch without edit module
Screenshot 2026-01-26 at 11 05 02

this branch with editable column
Screenshot 2026-01-26 at 11 05 34

SalvatorePreviti:
@codex review

SalvatorePreviti:
@codex review

gportela85 (approved):

PR #12978, opened 2026-01-27
AG-16448 value getter get value during editing

SalvatorePreviti:
The issue was that the way we access which source of data when we call getValue in value service was not consistent everywhere.
We should always access the source from the data and not editing or batching temporary values as we compute computed values, aggregations, or we perform refresh in the grid.

- Force getValue and getValueForDisplay to always require a "from" argument that can be 'edit' 'batch' or 'data'
- Add from property also to the user API getCellValue so the user and our tests can ask for data or batch and not just always the value being typed (that now is the default to not cause breaking changes)
- Fix in all the project all places where we access getValue or getValueForDisplay to use the appropriate source for the given operation. For example, aggregation, computing values in custom value getters, clipboard, CSRM refresh stages, SSRM block utils ...
- fix the logic in editSvc that gets the value to display

Refactoring
- add more tests regarding edit and batch
- remove the interfaces for edit service and edit model, and just use the classes directly in this way IDE navigation becomes simpler skipping going to the interface first, and also, removes the need to update the interfaces when we change the classes.

FIX AG-16448


SalvatorePreviti:
@codex review

packages/ag-grid-community/src/filter/filterValueService.ts
```diff
@@ -44,7 +44,7 @@ export class FilterValueService extends BeanStub implements NamedBean {
             node,
             column,
             colDef,
-            getValue: valueSvc.getValueCallback.bind(valueSvc, node),
+            getValue: valueSvc.getValueCallback.bind(valueSvc, node, 'data'),
```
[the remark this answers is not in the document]
SalvatorePreviti: issue fixed

SalvatorePreviti:
@codex review

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -737,15 +759,17 @@ export class ValueService extends BeanStub implements NamedBean {
         const otherColumn = this.colModel.getColDefCol(field);
 
         if (otherColumn) {
-            return this.getValue(otherColumn, node);
+            return this.getValue(otherColumn, node, 'data');
         }
```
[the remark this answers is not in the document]
SalvatorePreviti: This is expected behaviour, computed values do not use the pending value but the committed value only. Changed chart data source to use the committed data value also.

testing/behavioural/src/cell-editing/cell-editing-batch.test.ts
```diff
@@ -229,11 +230,13 @@ describe('Cell Editing Batch', () => {
         api.refreshCells({ columns: ['b'], force: true });
         await asyncSetTimeout(1);
 
-        expect(cellB).toHaveTextContent('xx');
+        // valueGetter sees committed data, not pending value
+        expect(cellB).toHaveTextContent('initial');
```
StephenCooper, line 234: So this makes it a breaking fix compared with 35.0 or also 34?

packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -437,7 +437,7 @@ export class FindService extends BeanStub implements NamedBean, IFindService {
                 let valueToFind: string | null;
                 const getFindText = (groupRowRendererParams as FindGroupRowRendererParams)?.getFindText;
                 if (getFindText) {
-                    const value = valueSvc.getValueForDisplay({ node }).value;
+                    const value = valueSvc.getValueForDisplay({ node, from: 'data' }).value;
```
StephenCooper: Does this being data and the findCellRenderer being edit mean that the highlighting will potentially be over letters that do not match the search term?

i.e the service has the old data of 'zz' and the edit is 'yy'. If the user then searches for 'zz' with the 'yy' be highlighted in the cell?

packages/ag-grid-enterprise/src/menu/contextMenu.ts
```diff
@@ -190,7 +190,7 @@ export class ContextMenuService extends BeanStub implements NamedBean, IContextM
         let { anchorToElement, value, source } = params;
 
         if (rowNode && column && value == null) {
-            value = this.beans.valueSvc.getValueForDisplay({ column, node: rowNode }).value;
+            value = this.beans.valueSvc.getValueForDisplay({ column, node: rowNode, from: 'edit' }).value;
```
StephenCooper, line 193: I am not sure about this use case, one to talk about.

SalvatorePreviti: to not introduce a breaking change, we will leave 'edit'

packages/ag-grid-enterprise/src/find/findService.ts
SalvatorePreviti: changed this to 'edit' for the find service

testing/behavioural/src/cell-editing/cell-editing-batch.test.ts
SalvatorePreviti: yes - this fixes the issue - we can't compute value getter or pipeline stages based on pending values, we need always to use real data - we would not know what to refresh so we assume batch mode will not update anything until not committed

StephM-W (approved):

gportela85:
@codex review

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -737,15 +753,17 @@ export class ValueService extends BeanStub implements NamedBean {
         const otherColumn = this.colModel.getColDefCol(field);
 
         if (otherColumn) {
-            return this.getValue(otherColumn, node);
+            return this.getValue(otherColumn, node, 'data');
```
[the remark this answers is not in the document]
gportela85: @SalvatorePreviti this looks similar to the issue found in [filterValueService.ts](https://github.com/ag-grid/ag-grid/pull/12978/files/5af619e664e370483532c7b6a7c52b2b5821b081#diff-0e641abc48b8a931287596762f6a679921c56a8eae8e13dc750f7bcb7a1a341a) earlier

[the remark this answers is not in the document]
SalvatorePreviti: this is correct though - the other was a different problem, this is desired - this is the path to compute computed columns and it should fetch always from 'data'. Bacth ephemeral edits are not supposed to update computed fields, filtering, aggregation, sorting, grouping or anything else until the batch is committed. as agreed with Steph. This was part of the bug solved by this PR

gportela85 (approved):

PR #12983, opened 2026-01-27
AG-13743 / AG-14324 pagination panel visual fixes

BernieSumption:
Fix AG-13743
Fix AG-14324

### DIFF
--- a/documentation/ag-grid-docs/src/components/demos/examples/inventory/InventoryExample.module.css
+++ b/documentation/ag-grid-docs/src/components/demos/examples/inventory/InventoryExample.module.css
@@ -121,10 +121,6 @@
     padding: 16px;
 }
 
-:global(.ag-theme-quartz .ag-paging-panel) {
-    height: 40px;
-}
-
 :global(.ag-theme-quartz .ag-paging-panel > span:last-child) {
     margin-right: 0;
 }
--- a/packages/ag-grid-community/src/theming/core/css/_root.css
+++ b/packages/ag-grid-community/src/theming/core/css/_root.css
@@ -4,6 +4,7 @@ ag-grid-angular {
 }
 
 .ag-root-wrapper {
+    container-type: inline-size;
     position: relative;
     display: flex;
     flex-direction: column;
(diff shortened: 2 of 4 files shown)


packages/ag-grid-community/src/theming/core/css/_root.css
```diff
@@ -4,6 +4,7 @@ ag-grid-angular {
 }
 
 .ag-root-wrapper {
+    container-type: inline-size;
```
StephenCooper, line 7: Could this have side effects for custom components provided by users?

documentation/ag-grid-docs/src/components/demos/examples/inventory/InventoryExample.module.css
```diff
@@ -121,10 +121,6 @@
     padding: 16px;
 }
 
-:global(.ag-theme-quartz .ag-paging-panel) {
-    height: 40px;
```
StephenCooper, line 125: Did this have to be removed to avoid breaking the style or just to enable this fix to actually take effect?

BernieSumption: To enable the fix to take effect, the JIRA uses this example as a test case so I wanted to make sure the new styles work in it.

Incidentally this is an example of a rule a customer might have that would cause a visual change, as we've moved from setting height to min-height, but I think it's within the bounds documented under "Understanding CSS rule maintenance and breaking changes" - https://ag-grid.com/react-data-grid/theming-css/#understanding-css-rule-maintenance-and-breaking-changes

packages/ag-grid-community/src/theming/core/css/_root.css
BernieSumption: Yes, if a customer has set `container-type` to define a parent element of the grid as the query container, and relies on container queries using that as the default container, our change will alter the container that is queried.

It can be worked around, they'd need to specifically name the container with `container: my-container-name / inline-size;` and reference the name on the child.

That said, I think this is unlikely. It's far more useful/likely that a custom component will use container queries relative to the size of the component, not the size of some parent element of the grid. So I'd expect custom components to define their own containment context.

But as ever with CSS, always possible to come up with some CSS that will be broken by our change.

StephenCooper (approved):

PR #12995, opened 2026-01-29
Fix Task for New Modules

StephenCooper:
To fix https://github.com/ag-grid/ag-grid/actions/runs/21475502322


.github/workflows/module-size-vs-release.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: 'Release tag to compare against (e.g., b35.1.0) or leave empty to auto-detect latest release'
         type: string
         required: false
+      test_branch:
```
StephenCooper, line 14: Enables testing without releasing to latest.

testing/module-size/moduleDefinitions.ts
```diff
@@ -1,147 +1 @@
...
-
-export const baseModule = { modules: [], expectedSize: 525 };
-
-export const moduleCombinations: ModuleTest[] = [
-    ...commonFeatureSets,
-    // ...chartModules,
-    ...allGridCommunityModules, //.slice(0, 3),
-    ...allEnterpriseModules, //.slice(0, 3),
-];
+export * from '../shared/moduleDefinitions';
```
StephenCooper, line 1: Stop this file being duplicated.

testing/shared/moduleDefinitions.ts
```diff
@@ -4,7 +4,8 @@ import type {
     ModuleName,
 } from '../../packages/ag-grid-community/src/interfaces/iModule';
 
-export const AllGridCommunityModules: Record<`${CommunityModuleName}Module`, number> = {
+// Use satisfies for type safety (catches typos) while allowing extra modules not in release types
+export const AllGridCommunityModules: Record<`${CommunityModuleName}Module` | 'BigIntFilterModule', number> = {
```
StephenCooper, line 8: New modules have to be added explicitly to give so that it can run against the release version which does not have the new module yet.

seanlandsman (approved):

PR #12998, opened 2026-01-29
AG-16620 - [formulas] - docs restructure

gportela85:
fix AG-16620


documentation/ag-grid-docs/src/content/docs/formulas/index.mdoc
```diff
@@ -2,11 +2,11 @@
 title: "Formulas"
 enterprise: true
 ---
-Formula strings can be provided to cells in the grid, allowing for dynamic calculations based on other cell values.
+Formulas let users enter spreadsheet-style expressions into grid cells so values update automatically when referenced data changes. This is an Enterprise feature.
```
StephenCooper:
```suggestion
Formulas let users enter spreadsheet-style expressions into grid cells so values update automatically when referenced data changes.
```

StephenCooper (approved):

PR #13000, opened 2026-01-29
Remove AGENTS.md and CLAUDE.md generated files

BernieSumption:


alantreadway (changes requested):
These needs to be present and checked in for Cloud-based toolng to bootstrap correctly (e.g. Codex PR review)

I have changes in `ag-shared` to sync across that will address the local modification issue, hopefully I'll be able to work on that today

PR #13002, opened 2026-01-29
AG-16663 destroyed pinned rows fix

SalvatorePreviti:
- group/treeData filler nodes not unpinned on destroy — centralised in rowNode _destroy
- destroyed rowNodes could be pinned — pinRow now blocks this
- too many csrm refreshes and no guards present when removing a row during refresh — added guards for this
- added a method in csrm called reMapRows that executes a map if we are not sure another is going to happen soon (during/after a row nodes updated or a refresh)
- called reMap rows both in expand service and pinned row model - the problem of nested refresh could happen there too
- keep track of all the flags used in the final event modelUpdated (NewData KeepRenderedRows KeepUndoRedoStack Animate) in the class if a refresh (or remap) is called inside a refresh so they are correct in the end. This solves partially some problems we always had with nested refresh
- ensure refreshModel is reentrant in case of errors so we are sure state flags are correct
- add tests around the fixes and all the updated code regarding the flags
- add support for pinnedRows in GridRows snapshot and validation

FIX AG-16663


packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
```diff
@@ -373,6 +392,14 @@ export class ManualPinnedRowModel extends BeanStub implements IPinnedRowModel {
         this.eventSvc.dispatchEvent({ type: 'pinnedRowsChanged' });
         node?.dispatchRowEvent('rowPinned');
     }
+
+    /** Refreshes CSRM if not already refreshing */
+    private refreshCSRM(): void {
+        const csrm = this.csrm;
+        if (csrm && !csrm.refreshingModel) {
```
sdwvit: I don't think this file should do this check. my gut feeling says it should be inside refreshModel method

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -52,7 +52,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
      * which then triggers the listener here that calls refresh model again but at the filter stage
      * (which is about to be run by the original call).
      */
-    private refreshingModel: boolean = false;
+    public refreshingModel: boolean = false;
```
sdwvit: I'm a bit cautious about this. see comment below

gportela85 (approved):

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
gportela85: I do agree that we have a coupling issue by making refreshingModel public because it leaks internal state to other modules and now the `manualPinnedRowModel` knows about internal CSRM implementation.
That said, we have adopted this pattern of exposing properties as public to enhance Tree Shaking. So I'd get @StephenCooper to comment here.

Question: I see that CSRM.refreshModel already checks for `refreshingModel` and returns. Could we move this check higher up it doesn't need to be repeated here inside `manualPinnedRowModel`?

StephenCooper: I would agree with not exposing this flag as it seems to be used here in a way that the csrm should be in control of.

StephenCooper: I would agree with not exposing this flag as it seems to be used here in a way that the csrm should be in control of.

SalvatorePreviti: I think the best is to introduce a method refreshIfNotRefreshing - CSRM.refreshModel does things also if already refreshing - the flag is used in a portion of it and not for the full execution

SalvatorePreviti:
@codex review

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Changed the approach - exposed a new method reMapRows() that is supposed to safely execute refreshModel({map:true})

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
SalvatorePreviti: Changed the approach - added a reMapRows method that does this and ensures flags are ported and maintained correctly for the final dispatch event

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -502,6 +523,16 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
     }
 
+    public reMapRows(): void {
+        if (this.refreshingModel || this.refreshingData) {
+            // A refresh will happen. Mark that keepRenderedRows should be false.
+            this.pendingRerender = true;
+            return; // No need to trigger another refresh, we are doing one.
+        }
+        // No refresh in progress, so trigger one now with keepRenderedRows false.
+        this.refreshModel({ step: 'map', keepRenderedRows: false });
```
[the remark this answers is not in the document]
SalvatorePreviti: solved

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -502,6 +532,16 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         return true; // Nothing changed, or only updates with no new rows and no removals
     }
 
+    public reMapRows(): void {
+        if (this.refreshingModel || this.refreshingData) {
+            // A refresh is in progress - set flags so the final modelUpdated event uses the right values
+            this.noKeepRenderedRows = true;
+            this.noKeepUndoRedoStack = true;
+            return;
```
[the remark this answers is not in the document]
SalvatorePreviti: solved

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -522,6 +563,13 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             this.isSuppressModelUpdateAfterUpdateTransaction(params)
         ) {
             this.rowDataUpdatedPending ||= rowDataUpdated;
+            if (refreshingModel || this.refreshingData) {
+                // Capture flags from blocked refresh calls to apply to the final modelUpdated event
+                this.setPendingRefreshFlags(params);
+            } else if (started) {
+                // Not nested - must be changeEventsDispatching or suppressedByUpdateTransaction
```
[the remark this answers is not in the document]
SalvatorePreviti: solved

SalvatorePreviti:
@codex review

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -522,6 +568,13 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
             this.isSuppressModelUpdateAfterUpdateTransaction(params)
         ) {
             this.rowDataUpdatedPending ||= rowDataUpdated;
+            if (refreshingModel || this.refreshingData) {
+                // Nested refresh - capture flags for the outer refresh to use
+                this.setPendingRefreshFlags(params);
+            } else if (started) {
```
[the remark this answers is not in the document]
SalvatorePreviti: solved

SalvatorePreviti:
@codex review

StephenCooper (approved):

PR #13007, opened 2026-01-30
Correct default pagination code

StephenCooper:


packages/ag-grid-community/src/pagination/paginationService.ts
```diff
@@ -20,7 +22,6 @@ export class PaginationService extends BeanStub implements NamedBean {
     private pageSizeFromPageSizeSelector?: number; // When user selects page size from page size selector.
     private pageSizeFromInitialState?: number; // When the initial grid state is loaded, and a page size rehydrated
     private pageSizeFromGridOptions?: number; // When user sets gridOptions.paginationPageSize.
-    private readonly defaultPageSize: 100; // When nothing else set, default page size is 100.
```
StephenCooper, line 23: This was never actually set to 100 given the type syntax.

StephM-W (approved):

PR #13009, opened 2026-01-30
Fix margin on wrapped pagination panel

BernieSumption:


markdurrant (approved):
Checked on local. Row pagination design updates LGTM 👍

PR #13016, opened 2026-01-30
[AG-12920] Fix(bigint): qa follow up

sdwvit:
Follow up for bigint feature

- BigInt status bar support
- Status panel value typing
- Excel export BigInt handling
- Docs + nav updates
- New BigInt filter docs + example

fix [AG-12920](https://ag-grid.atlassian.net/browse/AG-12920)
Screenshot-2026-01-30_18-25-22-739
Screenshot-2026-01-30_18-25-34-416


packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -213,18 +294,39 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
-                                min = value;
+                        if (typeof value === 'number' && !isNaN(value)) {
+                            if (useBigIntAggregation) {
+                                if (Number.isInteger(value)) {
+                                    addBigIntValue(BigInt(value));
+                                    return;
+                                }
+                                hasNonIntegerNumber = true;
+                                if (useBigIntAggregation) {
+                                    disableBigIntAggregation();
```
gportela85: When addressing the comments from codex above, this line was introduced which effectively disables bigInt aggregation when a non integer is present. Is this desirable?
I'll leave some comments around that could enable a mix of non-integers and bigInts

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -153,9 +153,90 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
+                const minValue = min ?? 0;
+                const maxValue = max ?? 0;
+                bigIntSum = BigInt(sum);
+                bigIntMin = BigInt(minValue);
+                bigIntMax = BigInt(maxValue);
+                bigIntCount = numberCount;
+            }
+        };
+
+        const disableBigIntAggregation = (): void => {
```
gportela85: You could completely remove this function

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -153,9 +153,90 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
+            }
+
+            if (min === null || value < min) {
+                min = value;
+            }
+
+            numberCount++;
+        };
+
+        const enableBigIntAggregation = (): void => {
```
gportela85: This function could also be removed

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -213,18 +294,39 @@ export class AggregationComp extends Component implements IStatusPanelComp {
                             value = Number(value);
                         }
 
-                        if (typeof value === 'number' && !isNaN(value)) {
-                            sum += value;
+                        if (typeof value === 'bigint') {
```
gportela85: This condition could be changed to
```
if (typeof value === 'bigint') {
    hasBigInt = true;
    if (!hasNonIntegerNumber) {
        addBigIntValue(value);
    }
    return;
}
```

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -213,18 +294,39 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
+                                addBigIntValue(value);
+                            } else {
+                                addNumberValue(Number(value));
                             }
+                            return;
+                        }
 
-                            if (min === null || value < min) {
-                                min = value;
+                        if (typeof value === 'number' && !isNaN(value)) {
```
gportela85: This could be changed to
```
if (typeof value === 'number' && !isNaN(value)) {
    addNumberValue(value);

    if (!Number.isInteger(value)) {
        hasNonIntegerNumber = true;
        return;
    }

    if (!hasNonIntegerNumber) {
        addBigIntValue(BigInt(value));
    }
}
```

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -234,15 +336,24 @@ export class AggregationComp extends Component implements IStatusPanelComp {
         }
 
         const gotResult = count > 1;
```
gportela85: This end block could be:

```
const hasResult = count > 1;
const canShowBigInt = hasBigInt && !hasNonIntegerNumber;
const hasNumberResult = !hasBigInt && numberCount > 1;
const hasBigIntResult = canShowBigInt && bigIntCount > 1;

this.setAggregationComponentValue('count', count, hasResult);

if (hasBigIntResult) {
    const avg = bigIntSum != null && bigIntCount > 0 ? bigIntSum / BigInt(bigIntCount) : null;
    this.setAggregationComponentValue('sum', bigIntSum, true);
    this.setAggregationComponentValue('min', bigIntMin, true);
    this.setAggregationComponentValue('max', bigIntMax, true);
    this.setAggregationComponentValue('avg', avg, true);
} else {
    let avg: number | null = null;
    if (hasNumberResult) {
        avg = sum / numberCount;
    }
    this.setAggregationComponentValue('sum', sum, hasNumberResult);
    this.setAggregationComponentValue('min', min, hasNumberResult);
    this.setAggregationComponentValue('max', max, hasNumberResult);
    this.setAggregationComponentValue('avg', avg, hasNumberResult);
}
```

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -153,9 +153,90 @@ export class AggregationComp extends Component implements IStatusPanelComp {
         let numberCount = 0;
         let min: number | null = null;
         let max: number | null = null;
+        let hasNonIntegerNumber = false;
+
+        let useBigIntAggregation = false;
```
gportela85: rename `useBigIntAggregation` to `hasBigInt`.

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -110,13 +110,13 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
         aggFuncName: AggregationStatusPanelAggFunc,
-        value: number | null,
+        value: number | bigint | null,
         visible: boolean
     ) {
         const statusBarValueComponent = this.getAllowedAggregationValueComponent(aggFuncName);
         const totalRow = _getTotalRowCount(this.beans.rowModel);
         if (_exists(statusBarValueComponent) && statusBarValueComponent) {
-            statusBarValueComponent.setValue(value!, totalRow);
+            statusBarValueComponent.setValue(value, totalRow);
```
gportela85, line 120: I think it's fair to assume when visible is false the value here is null
```
statusBarValueComponent.setValue(visible ? value : null, totalRow);
```

gportela85 (changes requested):

packages/ag-grid-enterprise/src/statusBar/providedPanels/agNameValue.ts
```diff
@@ -28,7 +28,12 @@ export class AgNameValue extends Component {
         this.eLabel.textContent = this.getLocaleTextFunc()(key, defaultValue);
     }
 
-    public setValue(value: number, totalRows: number): void {
+    public setValue(value: number | bigint | null, totalRows: number): void {
+        if (typeof value === 'bigint') {
```
gportela85, line 32: This early return for bigInt means that AggregationStatusPanelParams.valueFormatter is ignored and users can’t format BigInt values. Is this by design?

documentation/ag-grid-docs/src/content/docs/cell-data-types/index.mdoc
```diff
@@ -110,6 +110,14 @@ BigInt behaviour and limitations:
 
 If you need a custom input format, provide a custom Value Parser/Formatter by [Overriding the Pre-Defined Cell Data Type Definition](#overriding-the-pre-defined-cell-data-type-definitions).
 
+#### Server-Side Row Model (SSRM) and State
+
+When using the Server-Side Row Model, BigInt values in filter models, sort models, row group keys, pivot models, and grid state are stored using a serialisable representation (string form of the integer). Values are rehydrated back into `bigint` when restoring state or constructing SSRM requests. No implicit conversion to `number` occurs during serialisation, and grid state containing BigInt values is JSON-serialisable and safe to persist and restore.
```
gportela85: This section now claims SSRM/state “rehydrates” BigInt values, but I didn’t find code SSRM/state serialization/rehydration logic. Was this done previously and is already working.

sdwvit: Yes, by rehydration I meant restoring state. It was already implemented in a separate pr.

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
StephenCooper: If possible can we add some test cases that validate that numbers and bigInts behave in the same way with these mixed inputs?

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
sdwvit: these are codependent, so visible is true when value is non null and  vice versa

packages/ag-grid-enterprise/src/statusBar/providedPanels/agNameValue.ts
sdwvit: value formatter doesn't accept bigint. What's your opinion should we support that for V1?

sdwvit:
rewrote the whole bigint aggregation business in aggregationComp.ts, please re-review @gportela85 @StephenCooper

packages/ag-grid-enterprise/src/statusBar/providedPanels/agNameValue.ts
gportela85: I think we should update IStatusPanelValueFormatterParams to accept BigInt as a values. It doesn't seem right that we have a type of value that is now supported but we don't allow the customer to format.

@StephenCooper What do you think? This would be a breaking change but I don't think we should leave it out.

gportela85 (approved):

gportela85:
@codex review

packages/ag-grid-enterprise/src/statusBar/providedPanels/agNameValue.ts
StephenCooper: What if we add a third generic property TValue with default of number | null. This can then be overridden by the user to add bigInt support if they need it. 

We then have the option to enable this as the default type in the next major. The downside of this is that we would then be left with a third generic property long term which did not add value. However, this seems like a way forward here?

packages/ag-grid-enterprise/src/statusBar/providedPanels/aggregationComp.ts
```diff
@@ -210,21 +268,30 @@ export class AggregationComp extends Component implements IStatusPanelComp {
...
-                            if (min === null || value < min) {
-                                min = value;
+                            const asNumber = Number(trimmedValue);
+                            if (!Number.isFinite(asNumber)) {
+                                return;
+                            }
+                            if (
+                                sum + asNumber >= Number.MAX_SAFE_INTEGER ||
+                                sum + asNumber <= Number.MIN_SAFE_INTEGER
+                            ) {
```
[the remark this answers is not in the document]
sdwvit: corner case, but fixed anyway

PR #13026, opened 2026-02-02
AG-3390 - Sync mdoc prettier formatting

taktran:
Sync with https://github.com/ag-grid/ag-charts/pull/6080

## TODO

- [x] Add formatting commit to [.git-blame-ignore-revs](https://github.com/ag-grid/ag-grid/blob/latest/.git-blame-ignore-revs) - https://github.com/ag-grid/ag-grid/pull/13032


StephenCooper (approved):
Should we hide this commit from Git Blame?

PR #13048, opened 2026-02-04
[AG-14571] Fix(docs): relative date ranges

sdwvit:
Docs: align date range filter guidance with SSRM docs

  - Replace SSRM subsection in Date Filter with Grid State preset range model note and link to SSRM date ranges
  - Add import snippet for PresetDateRangeFilterModel in SSRM filtering docs for discoverability

Fixes [AG-14571](https://ag-grid.atlassian.net/browse/AG-14571)

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
@@ -80,6 +80,12 @@ When using a Date Filter with built-in date ranges (for example, Today or Last 7
 
 {% interfaceDocumentation interfaceName="PresetDateRangeFilterModel" config={"description":""} /%}
 
+The type above is exported from `ag-grid-community`:
+
+```js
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/src/content/docs/server-side-model-filtering/index.mdoc
```diff
@@ -80,6 +80,12 @@ When using a Date Filter with built-in date ranges (for example, Today or Last 7
 
 {% interfaceDocumentation interfaceName="PresetDateRangeFilterModel" config={"description":""} /%}
 
+The type above is exported from `ag-grid-community`:
```
StephenCooper: I don't think we need to add this. There is no other way for a person to import a type so this seems excessive documentation.

sdwvit: they won't see the name of the interface otherwise

StephenCooper (approved):

PR #13051, opened 2026-02-04
[AG-16686] Fix(cell): data types bigint support formulas

sdwvit:
BigInt – Support formulas

Fixes [AG-16686](https://ag-grid.atlassian.net/browse/AG-16686)

Screenshot-2026-02-04_15-14-16-048
Screenshot-2026-02-04_15-14-12-244

### DIFF
--- a/packages/ag-grid-enterprise/src/formula/functions/numbers/functions.ts
+++ b/packages/ag-grid-enterprise/src/formula/functions/numbers/functions.ts
@@ -2,13 +2,25 @@ import type { FormulaFunctionParams } from 'ag-grid-community';
 
 import { FormulaError } from '../../ast/utils';
 import { criteriaToPredicate, isRangeParam, isValueParam, iterableWithoutBlanks, take, takeBetween } from '../utils';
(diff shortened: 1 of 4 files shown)


packages/ag-grid-enterprise/src/formula/functions/numbers/functions.ts
```diff
@@ -2,13 +2,25 @@ import type { FormulaFunctionParams } from 'ag-grid-community';
...
+        if (typeof n === 'bigint') {
+            accBigInt ??= coerceBigInt('PRODUCT', acc);
+            if (n === 0n) {
+                return 0n;
+            }
+            accBigInt *= n;
+            continue;
+        }
+
+        if (accBigInt != null) {
```
gportela85, line 22: shouldn't this be `if (accBigInt != null || n == null)` ?

sdwvit: no, it's never null, it throws instead of assigning null

gportela85 (approved):

PR #13058, opened 2026-02-04
AG-12324 - [group cell renderer] - fixed tooltips when truncated

gportela85:
fix AG-12324

### DIFF
--- a/packages/ag-grid-community/src/tooltip/tooltipService.ts
+++ b/packages/ag-grid-community/src/tooltip/tooltipService.ts
@@ -46,8 +46,27 @@ const getEditErrorsForPosition = (
 
 const getCellTruncationCheck = (beans: BeanCollection, ctrl: CellCtrl): (() => boolean) | undefined => {
     const isTooltipWhenTruncated = _isShowTooltipWhenTruncated(beans.gos);
-    if (!isTooltipWhenTruncated || ctrl.isCellRenderer()) {
+
+    if (!isTooltipWhenTruncated) {
         return undefined;
+    }
+
+    if (ctrl.isCellRenderer()) {
+        const colDef = ctrl.column.getColDef();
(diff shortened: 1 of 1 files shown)


packages/ag-grid-community/src/tooltip/tooltipService.ts
```diff
@@ -46,10 +46,29 @@ const getEditErrorsForPosition = (
...
+        const isGroupCellRenderer = !!colDef.showRowGroup || colDef.cellRenderer === 'agGroupCellRenderer';
+        if (!isGroupCellRenderer) {
+            return undefined;
+        }
+
+        return _isElementOverflowingCallback(() => {
+            const eCell = ctrl.eGui;
+            return (
+                (eCell.querySelector('.ag-group-value') as HTMLElement | undefined) ||
+                (eCell.querySelector('.ag-cell-value') as HTMLElement | undefined) ||
```
StephenCooper, line 66: Is this slightly duplicated with the check below? They seem to be doing a similar resolution of the eCell or .ag-cell-value. 

Not a blocker

StephenCooper (approved):

packages/ag-grid-community/src/tooltip/tooltipService.ts
gportela85: The logic below only queries the cell for `ag-cell-value` if eCell.children.length is > 0.

We could unify the logic to something like 

```
return (
        (includeGroupValue ? (eCell.querySelector('.ag-group-value') as HTMLElement) : null) ||
        (eCell.querySelector('.ag-cell-value') as HTMLElement | null) ||
        eCell
    );
 ```

But then we would query for `ag-cell-value` more often and I'd like to avoid that if possible, even if we have some code duplication

PR #13069, opened 2026-02-05
[AG-16487] Fix(ssrm): tc2

sdwvit:
SSRM tree data with `serverSideEnableClientSideSort` could lose the sorted order after `refreshServerSide({ purge: true })`. The store was rebuilt and reloaded, but the client-side sort was only re-applied when the “refresh finished” path ran. In the purge case, that path isn’t reached, so the rows were displayed in the original server order while the column header still showed an active sort.

  Fix this by reapplying client-side sort after a load completes when:
  - client-side sorting is enabled
  - a sort is active
  - and the store is fully loaded (including the “all rows loaded and no stubs” case where `isStoreFullyLoaded()` can still be false).

Fix [AG-16487](https://ag-grid.atlassian.net/browse/AG-16487)

### DIFF
--- a/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
+++ b/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
@@ -895,6 +895,18 @@ export class LazyCache extends BeanStub {
             lazyNodesAfterStoreEnd.forEach((lazyNode) => this.destroyRowAtIndex(lazyNode.index));
         }
 
+        const hasActiveSort = (this.sortSvc?.getSortOptions() ?? []).some((opt) => opt.sort != null);
+        const hasStubNodes = this.nodeMap.find((lazyNode) => !!lazyNode.node.stub) != null;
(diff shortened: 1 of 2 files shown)


packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
```diff
@@ -895,6 +895,18 @@ export class LazyCache extends BeanStub {
             lazyNodesAfterStoreEnd.forEach((lazyNode) => this.destroyRowAtIndex(lazyNode.index));
         }
 
+        const hasActiveSort = (this.sortSvc?.getSortOptions() ?? []).some((opt) => opt.sort != null);
+        const hasStubNodes = this.nodeMap.find((lazyNode) => !!lazyNode.node.stub) != null;
+        const allRowsLoaded = this.isLastRowKnown && this.nodeMap.getSize() === this.numberOfRows && !hasStubNodes;
+        const shouldClientSideSortOnLoad =
```
sdwvit: move this check higher

SalvatorePreviti (approved):

PR #13075, opened 2026-02-05
AG-16448 find and copy use batch data

SalvatorePreviti:
We want 'batch' for both find and clipboard copy, that means, the values that are in the batch, if we are in batch mode, or in the data, if not, but not in the editor cells

- Fix the exporter to accept a valueFrom to specify if we want to export 'edit' 'batch' or 'data', optional, needed by cliphoard
- Fix a bug in the find service not registering to batch edit events so not resetting the cache
- Change the find servide to use 'batch' instead of 'edit'
- Add tests
- Improve instructions around testing

FIX AG-16448

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,26 @@
+Please also reference the following rules as needed. The list below is provided in TOON format, and `@` stands for the project root directory.
+
+rules[5]:
+  - path: @.codex/memories/benchmarks.md
+    description: Running and creating performance benchmarks for AG Grid
+    applyTo[2]: testing/performance/**/*,**/benchmark*
+  - path: @.codex/memories/code-quality.md
+    description: "Code quality practices including avoiding bloat, comment guidelines, and review practices"
+    applyTo[1]: packages/*/src/**/*.ts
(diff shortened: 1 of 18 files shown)


AGENTS.md
```diff
@@ -1,3 +1,26 @@
+Please also reference the following rules as needed. The list below is provided in TOON format, and `@` stands for the project root directory.
+
+rules[5]:
```
StephenCooper, line 3: Are these updates required?

StephenCooper (approved):

AGENTS.md
SalvatorePreviti: I am merging for now and letting @alantreadway review too - we can raise another PR to adjust them

alantreadway: We should avoid committing these TOON statements, they are referencing tool-specific paths which won't be present a) locally, as its dependent on installed tools, and b) on initial checkout in the cloud

Unfortunately I didn't find a robust solution to disable generation of these for this update case (for yarn postinstall we're now stashing local edits, running `rulesync` then restoring and popping the stashed edits)

PR #13077, opened 2026-02-05
[AG-16227] Fix(react): in ssrm tabbing into the grid from the bottom loses focus

sdwvit:
Fix: SSRM tabbing from bottom loses focus when tab guard holds focus

  When Shift+Tabbing into the grid from the bottom in SSRM, the bottom tab guard can keep browser focus while rows load, so FocusService doesn’t think the grid has focus and skips restoring the focused cell. The result is focus loss after the data arrives.

  Update FocusService to treat the tab guard as a valid focus holder, allowing focus recovery to complete once the row renders.

Fixes [AG-16227](https://ag-grid.atlassian.net/browse/AG-16227)


SalvatorePreviti (approved):

sdwvit:
@gportela85 had some comments about this, don't merge

PR #13107, opened 2026-02-09
RTI-3227 Remove deprecated props

StephenCooper:
Fix [RTI-3227](https://ag-grid.atlassian.net/browse/RTI-3227)


documentation/ag-grid-docs/src/content/docs/server-side-model-selection/_examples/api-group-selects-children/fakeServer.ts
```diff
@@ -96,10 +96,9 @@ export function FakeServer(allData) {
 
     function createFilterSql(mapper, key, item) {
         if (item.operator) {
-            const condition1 = mapper(key, item.condition1);
-            const condition2 = mapper(key, item.condition2);
+            const conditions = item.conditions.map((condition) => mapper(key, condition));
```
StephenCooper, line 99: Match all the other implementations for the current approach.

peterjrreynolds (approved):

PR #13108, opened 2026-02-09
Documentation Update | Change row selection mode to 'multiRow'

Toscan0:
On current documentation:
Image

It should say `multiRow` as it says in other examples

When using `multiple`, the feature doesn't work and we get a warning

Image


StephenCooper:
Thank you for reporting we will get this fixed in our next release.

PR #13144, opened 2026-02-12
AG-15019 update Algolia ranking

BernieSumption:
Fix [AG-15019](https://ag-grid.atlassian.net/browse/AG-15019)

This PR is split into 2 commits, the first updates the Algolia indexing process, and the second removes all metaTag from documentation source, since it is made redundant by the first change.

Changes to Algolia indexing:

1. Automatic extraction of searchable words from documentation
   * We now automatically extract "code words" from documentation text and make them searchable. A code word is any word in camelCase or PascalCase, or any code sample formatted as code using backticks, e.g. `api.someMethodCall(param)`
   * Using automated analysis, ensured all manually maintained metaTags are now automatically extracted so any search that previously brought up a page through metaTags now brings up the same page through code words.
   * **IMPORTANT** There is one exception, previously we used meta tags to ensure that searching for a removed APIs like `setPinnedBottomRowData` continued to bring up the page that once documented that API. Now, the only result for a "setPinnedBottomRowData" search will be the migration page documenting its removal and replacement. I think this is more appropriate behaviour.
   * Removed metaTag directives from all documentation pages
2. Properties of some types are now searchable
   * For example, `defaultExcelExportParams` is a grid option of type `ExcelExportParams`. This interface contains a `headerFooterConfig` property. Searching "headerFooterConfig" now brings up defaultExcelExportParams.
   * Added configuration and used it to exclude some types and properties where this generates noise, e.g. we exclude `IRowNode` here because lots of API methods take a rowNode and we don't want all of these to show up if you search for any of IRowNode's 50 members.


documentation/ag-grid-docs/src/content/api-nav/nav.json
```diff
@@ -2,6 +2,7 @@
     "sections": [
         {
             "hideTitle": true,
+            "searchOrder": 60,
```
taktran: Could you update the [navigation content schema](https://github.com/ag-grid/ag-grid/blob/latest/documentation/ag-grid-docs/src/content.config.ts#L118-L128) as well, while you are at it eg,

```
searchOrder: z.number().optional(),
```

taktran (dismissed):
Looks good 👍🏼 , just a minor update request.

StephenCooper (approved):

StephenCooper (approved):

PR #13150, opened 2026-02-13
[AG-3471] Feature(sidebar): tool panel deferred apply mode for grouping pivoting aggregations

sdwvit:
This PR introduces Deferred Apply Mode for the Columns Tool Panel.
  With `toolPanelParams.deferApply = true`, tool panel interactions no longer mutate grid column state immediately. Instead, changes are staged in a dedicated deferred state model and only committed on `Apply` (or discarded on `Cancel`).

  ### Architecture Overview

  - Added a deferred state layer (`ColumnToolPanelDeferredService`) that tracks:
      - the current applied grid state
      - the pending patch produced by tool panel interactions
  - Tool panel UI components (checkboxes, drop zones, context menu, pivot mode toggle) are wired to read/write pending state when deferred mode is enabled.
  - Existing immediate-apply flows remain unchanged; deferred behavior is opt-in via `deferApply`.
  - On `Apply`, pending state is materialized into a single column-state application path.
  - On `Cancel`, pending state is dropped and the panel is rebuilt from applied state.
  - Deferred state is also synchronized with external grid events so staged edits remain coherent if column state changes outside the panel.

  ### Staged Changes Covered

  When deferred mode is enabled, pending state includes:

  - Pivot mode toggling
  - Column visibility selection
  - Row Group / Pivot / Values assignments
  - Value column ordering
  - Aggregation function changes

  Fixes [AG-3471](https://ag-grid.atlassian.net/browse/AG-3471).

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/deferred-apply-mode/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/deferred-apply-mode/main.ts
@@ -1,84 +1,97 @@
-import type { ColDef, GridApi, GridOptions, IServerSideDatasource, IServerSideGetRowsRequest } from 'ag-grid-community';
-import { ModuleRegistry, ValidationModule, createGrid } from 'ag-grid-community';
+import type { ColDef, GridApi, GridOptions } from 'ag-grid-community';
+import { ClientSideRowModelModule, ModuleRegistry, ValidationModule, createGrid } from 'ag-grid-community';
 import {
     ColumnMenuModule,
     ColumnsToolPanelModule,
     ContextMenuModule,
     PivotModule,
-    ServerSideRowModelModule,
+    RowGroupingPanelModule,
 } from 'ag-grid-enterprise';
 
 ModuleRegistry.registerModules([
+    ClientSideRowModelModule,
     ColumnsToolPanelModule,
     ColumnMenuModule,
     ContextMenuModule,
     PivotModule,
-    ServerSideRowModelModule,
+    RowGroupingPanelModule,
     ...(process.env.NODE_ENV !== 'production' ? [ValidationModule] : []),
 ]);
 
 const columnDefs: ColDef[] = [
-    { field: 'athlete', minWidth: 180, enableRowGroup: true, enablePivot: true },
-    { field: 'age', enableRowGroup: true, enablePivot: true },
-    { field: 'country', minWidth: 160, enableRowGroup: true, enablePivot: true },
-    { field: 'year', enableRowGroup: true, enablePivot: true },
-    { field: 'sport', minWidth: 180, enableRowGroup: true, enablePivot: true },
-    { field: 'gold', enableValue: true },
-    { field: 'silver', enableValue: true },
-    { field: 'bronze', enableValue: true },
-    { field: 'total', enableValue: true },
+    {
+        field: 'athlete',
+        minWidth: 200,
+        enableRowGroup: true,
+        enablePivot: true,
+    },
+    {
+        field: 'age',
+        enableValue: true,
+    },
+    {
+        field: 'country',
+        minWidth: 200,
+        enableRowGroup: true,
+        enablePivot: true,
+        rowGroupIndex: 1,
+    },
+    {
+        field: 'year',
+        enableRowGroup: true,
+        enablePivot: true,
+        pivotIndex: 1,
+    },
+    {
+        field: 'date',
+        minWidth: 180,
+        enableRowGroup: true,
+        enablePivot: true,
+    },
+    {
+        field: 'sport',
+        minWidth: 200,
+        enableRowGroup: true,
+        enablePivot: true,
+        rowGroupIndex: 2,
+    },
+    {
+        field: 'gold',
+        hide: true,
+        enableValue: true,
+    },
+    {
+        field: 'silver',
+        hide: true,
+        enableValue: true,
+        aggFunc: 'sum',
+    },
+    {
+        field: 'bronze',
+        hide: true,
+        enableValue: true,
+        aggFunc: 'sum',
+    },
+    {
+        headerName: 'Total',
+        field: 'total',
+    },
 ];
 
 let gridApi: GridApi<IOlympicData>;
-let requestSequence = 0;
-
-const logColumnStateSnapshot = (source: string) => {
-    try {
-        if (!gridApi) {
-            return;
-        }
-
-        const getColIdSafe = (col: any): string => {
-            if (!col) {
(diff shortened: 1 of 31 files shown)


packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -291,20 +291,26 @@ export class ColumnModel extends BeanStub implements NamedBean {
         const showRowNumbers = _isRowNumbers(beans);
         const valueColumns = valueColsSvc?.columns;
 
-        const res = cols.list.filter((col) => {
-            const isAutoGroupCol = isColumnGroupAutoCol(col);
-            if (showAutoGroupAndValuesOnly) {
-                const isValueCol = valueColumns?.includes(col);
+        if (showAutoGroupAndValuesOnly) {
```
sdwvit, line 297: this change fixes ordering in the pivot-mode/no-pivot-result display path so value (aggregation) columns follow valueColsSvc.columns order, instead of whichever order cols.list happened to provide.

sdwvit: maybe related to [AG-5601](https://ag-grid.atlassian.net/browse/AG-5601)

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/deferred-apply-mode/main.ts
```diff
@@ -0,0 +1,113 @@
...
+    columnDefs,
+    defaultColDef: {
+        flex: 1,
+        minWidth: 150,
+    },
+    autoGroupColumnDef: {
+        minWidth: 250,
+    },
+    pivotMode: true,
+    rowModelType: 'serverSide',
```
StephenCooper, line 79: I feel like this SSRM example belongs in the SSRM docs. I assume this is being done because the most value of this feature comes when using SSRM. 

I would prefer this example to be the simpler client side and move this one to a good place in the SSRM docs.

packages/ag-grid-community/src/interfaces/iToolPanel.ts
```diff
@@ -52,6 +52,10 @@ export interface IToolPanelColumnCompParams {
     contractColumnSelection: boolean;
     /** Suppress updating the layout of columns as they are rearranged in the grid */
     suppressSyncLayoutWithGrid: boolean;
+    /** Defer applying Columns Tool Panel changes until an explicit apply action */
+    deferApply?: boolean;
+    /** Action buttons shown in the Columns Tool Panel when `deferApply` is enabled */
+    buttons?: ('apply' | 'cancel')[];
```
StephenCooper, line 58: Is buttons required? 

`[]` - an empty array would mean no changes would ever be applied
`['apply]` - there is no way to undo the changes without manually undoing them
`['cancel']` - there is no way to apply the changes

packages/ag-grid-enterprise/src/columnToolPanel/agPrimaryColsList.ts
```diff
@@ -549,14 +552,27 @@ export class AgPrimaryColsList extends Component<AgPrimaryColsListEvent> {
     }
 
     public doSetSelectedAll(selectAllChecked: boolean): void {
-        selectAllChildren(this.beans, this.allColsTree, selectAllChecked, this.eventType);
+        const { beans, allColsTree, eventType, params } = this;
+        // In deferred mode, stage checkbox actions as pending changes instead of applying immediately.
+        selectAllChildren(
```
StephenCooper, line 557: Should we switch this to a params object to make it clearer?

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -291,23 +294,27 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
-                // keep col if a) it's auto-group or b) it's visible
-                return isAutoGroupCol || col.isVisible();
-            }
-        });
+            });
+
+            const valueColumnsInDisplayOrder = (valueColumns ?? []).filter(
+                (col) => cols.map[col.getColId()] !== undefined
+            );
+            return [...nonValueColumns, ...valueColumnsInDisplayOrder];
```
StephenCooper, line 310: I believe this code is used in Pivoting and this change may have changed the order of columns in the grid.

Do we have to make a change here to support the Apply feature?

packages/ag-grid-enterprise/src/columnToolPanel/agPrimaryColsList.ts
```diff
@@ -571,19 +587,25 @@ export class AgPrimaryColsList extends Component<AgPrimaryColsListEvent> {
...
+            // Header checkbox state (checked/indeterminate) should match staged row-group/pivot/value/visibility
+            // state in deferred mode.
             if (pivotMode) {
                 const noPivotModeOptionsAllowed =
                     !column.isAllowPivot() && !column.isAllowRowGroup() && !column.isAllowValue();
                 if (noPivotModeOptionsAllowed) {
                     return;
                 }
-                checked = column.isValueActive() || column.isPivotActive() || column.isRowGroupActive();
+                checked = this.params.isColumnCheckedInToolPanel
```
StephenCooper, line 598: It feels like these repeated checks should be centralised behind some wrapper which switches in either the defered implementation or the underlying values.

packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
```diff
@@ -112,6 +178,25 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
             childDestroyFuncs.push(() => pivotModeListener());
         }
 
+        if (mergedParams.deferApply) {
+            /**
+             * Deferred apply can rebuild this panel on refresh/cancel.
+             * Keep all destroy callbacks so old column listeners are removed.
+             */
+            const deferredSyncListeners = this.addManagedEventListeners({
+                newColumnsLoaded: this.syncDeferredFromApplied.bind(this),
```
StephenCooper, line 187: Code Style: Extract the shared function and pass to each listener as this will minify better.

packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
```diff
@@ -261,3 +692,15 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
         super.destroy();
     }
 }
+
+function areStringArraysEqual(a: string[], b: string[]): boolean {
```
StephenCooper, line 696: Do we not already have a utils method for this?

StephenCooper (changes requested):
We need to re-work this design.

PR #13156, opened 2026-02-16
AG-15888 - fixed RowResizer manipulating the DOM directly

gportela85:
fix [AG-15888](https://ag-grid.atlassian.net/browse/AG-15888)
fix [AG-14608](https://ag-grid.atlassian.net/browse/AG-14608)

### DIFF
--- a/packages/ag-grid-community/src/interfaces/rowNumbers.ts
+++ b/packages/ag-grid-community/src/interfaces/rowNumbers.ts
@@ -26,6 +26,9 @@ export interface RowNumbersOptions
             | 'valueGetter'
             | 'valueFormatter'
             | 'maxWidth'
+            | 'cellRenderer'
+            | 'cellRendererSelector'
+            | 'cellRendererParams'
         >
     > {
     /**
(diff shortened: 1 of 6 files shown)


peterjrreynolds (approved):

packages/ag-grid-enterprise/src/rowNumbers/rowNumbersService.ts
```diff
@@ -237,6 +237,7 @@ export class RowNumbersService extends BeanStub implements NamedBean, IRowNumber
     private refreshRowNumberOverrides(): void {
         const rowNumbers = _isRowNumbers(this.beans);
         this.rowNumberOverrides = {};
+        this.isSuppressCellSelectionIntegration = false;
```
StephenCooper, line 240: What is this for?

packages/ag-grid-community/src/interfaces/rowNumbers.ts
```diff
@@ -26,6 +26,9 @@ export interface RowNumbersOptions
             | 'valueGetter'
             | 'valueFormatter'
             | 'maxWidth'
+            | 'cellRenderer'
```
StephenCooper: Is this for a different ticket to add support for row number cell components?

packages/ag-grid-enterprise/src/rowNumbers/rowNumbersService.ts
gportela85: We were not resetting the property correctly before, so calling `refreshRowNumberOverrides` only worked the first time. After this property was set to true, it would no longer be reactive

StephenCooper (approved):

PR #13162, opened 2026-02-17
AG-14419  - Added keyboard navigation to Row Numbers

gportela85:
fix [AG-14419](https://ag-grid.atlassian.net/browse/AG-14419)


community-modules/locale/src/de-DE.ts
```diff
@@ -684,7 +684,7 @@ export const AG_GRID_LOCALE_DE = {
     ariaFilterColumn: 'Drücken Sie STRG EINGABETASTE, um den Filter zu öffnen',
     ariaRowDeselect: 'Drücken Sie LEERTASTE, um diese Zeile abzuwaehlen',
     ariaHeaderSelection: 'Spalte mit Header-Auswahl',
-    ariaSelectAllCells: 'Drücken Sie die Leertaste, um alle Zellen auszuwählen',
+    ariaSelectAllCells: 'Drücken Sie die Leertaste oder Enter, um alle Zellen auszuwählen',
```
StephenCooper: These do not seem to be consistent as Enter has not been translated but Space has. 

Maybe let's just update the US file and let the current process update the locales via the script?

packages/ag-grid-enterprise/src/rowNumbers/rowNumbersService.ts
```diff
@@ -282,13 +319,15 @@ export class RowNumbersService extends BeanStub implements NamedBean, IRowNumber
     }
 
     private onHeaderFocus(): void {
-        this.beans.ariaAnnounce?.announceValue('Press Space to select all cells', 'ariaSelectAllCells');
+        this.beans.ariaAnnounce?.announceValue('Press Space or Enter to select all cells', 'ariaSelectAllCells');
```
StephenCooper: Is there a reason for now also supporting Enter?

packages/ag-grid-enterprise/src/rowNumbers/rowNumbersService.ts
```diff
@@ -182,6 +182,42 @@ export class RowNumbersService extends BeanStub implements NamedBean, IRowNumber
...
+        const { rangeSvc } = this.beans;
+
+        if (!rangeSvc) {
+            return;
+        }
+
+        rangeSvc.handleCellMouseDown(this.createRowSelectionMouseEvent(keyboardEvent), cellPosition);
+    }
+
+    private createRowSelectionMouseEvent(event: KeyboardEvent): MouseEvent {
```
StephenCooper: Do we need to create a fake event to get the range service to do the right thing?

packages/ag-grid-community/src/focusService.ts
```diff
@@ -603,7 +605,9 @@ export class FocusService extends BeanStub implements NamedBean {
                 forceBrowserFocus: true,
             });
 
-            this.beans.rangeSvc?.setRangeToCell({ rowIndex, rowPinned, column });
+            if (!isRowNumberCol(column)) {
```
StephenCooper, line 608: Is there an alternative to leaking this rowNumber check across a number of services?

gportela85: Not at the moment, we have this check for column type because there are certain things we do around the grid that should behave differently for a column that is actually a header. I don't think it's worth refactoring a good portion of the services just to avoid this call.

packages/ag-grid-enterprise/src/rowNumbers/rowNumbersService.ts
gportela85: Yes, our default for selecting columns is pressing Enter, so I also added pressing Enter to be the way to select rows. So it felt wrong to leave Enter for Columns and Rows but Space to Select All.

StephenCooper (approved):

PR #13183, opened 2026-02-20
AG-13467 - [keyboard-navigation] - Added tabToNextGridContainer

gportela85:
fix [AG-13467](https://ag-grid.atlassian.net/browse/AG-13467)


packages/ag-grid-community/src/gridComp/gridCtrl.ts
```diff
@@ -94,60 +114,124 @@ export class GridCtrl extends BeanStub {
...
+                        return this.focusGridBodyDefault(backwards) || undefined;
+                    }
+
+                    const targetContainer = focusableContainers.find(
+                        (container) => container.getFocusableContainerName() === userResult
+                    );
+                    if (!targetContainer) {
+                        const translate = this.getLocaleTextFunc();
+                        _consoleWarn(
+                            translate(
```
peterjrreynolds: Does this need localising? I don't think we localise console messages anywhere else

peterjrreynolds (approved):

packages/ag-grid-community/src/gridComp/gridCtrl.ts
gportela85: Agreed, I'll remove it.

packages/ag-grid-community/src/gridComp/gridComp.ts
```diff
@@ -132,9 +140,12 @@ export class GridComp extends TabGuardComp {
...
 
     protected getFocusableContainers(): FocusableContainer[] {
-        const focusableContainers: FocusableContainer[] = [this.gridBody];
+        const focusableContainers: FocusableContainer[] = [
+            ...(this.gridHeaderDropZones?.getFocusableContainers?.() ?? []),
+            this.gridBody,
+        ];
 
-        for (const comp of [this.sideBar, this.pagination]) {
+        for (const comp of [this.sideBar, this.statusBar, this.pagination]) {
```
StephenCooper, line 148: with this returning more containers have we changed the defualt behaviour?

packages/ag-grid-community/src/utils/gridFocus.ts
```diff
@@ -1,15 +1,18 @@
...
 import type { CellPosition } from '../interfaces/iCellPosition';
+import type { FocusableContainer } from '../interfaces/iFocusableContainer';
 import type { Component } from '../widgets/component';
+import { _isStopPropagationForAgGrid } from './gridEvent';
 
 export function _addFocusableContainerListener(beans: BeanCollection, comp: Component, eGui: HTMLElement): void {
     comp.addManagedElementListeners(eGui, {
         keydown: (e: KeyboardEvent) => {
-            if (!e.defaultPrevented && e.key === KeyCode.TAB) {
+            if (!e.defaultPrevented && !_isStopPropagationForAgGrid(e) && e.key === KeyCode.TAB) {
```
StephenCooper, line 15: Was this missing previously then?

packages/ag-grid-community/src/utils/gridFocus.ts
```diff
@@ -39,11 +42,18 @@ export function _focusNextGridCoreContainer(
...
+
+    if (focusResult === true) {
         return true;
     }
 
-    if (forceOut || (!backwards && !gridCtrl.isDetailGrid())) {
+    // false from tabToNextGridContainer means browser-default tab flow.
+    if (focusResult === false) {
+        return focusResult;
+    }
```
StephenCooper, line 54: Nit: Should we collapse these into one boolean check?

packages/ag-grid-community/src/gridComp/gridComp.ts
gportela85: Yes, if the statusBar contains a focusable element, it is now part of the focus pipeline. This was an accessibility gap we had.

packages/ag-grid-community/src/utils/gridFocus.ts
gportela85: `!_isStopPropagationForAgGrid(e)` was added here to avoid handling the same Tab event twice. Some containers (like the sidebar paths) already handle Tab in their own managed handler and intentionally mark the event with `_stopPropagationForAgGrid(e)`. Without this guard, the generic `_addFocusableContainerListener` would still run and do a second `_focusNextGridCoreContainer(...)` evaluation, which caused regressions. That said, before `tabToNextGridContainer` was created, this usually didn’t hurt because duplicate handling often ended in the same force-out behavior. So it only became apparent with this new feature.

packages/ag-grid-community/src/utils/gridFocus.ts
gportela85: I'd prefer to keep this one as is, it helps debugging.

StephenCooper (approved):

SalvatorePreviti (approved):

PR #13186, opened 2026-02-23
AG-16513 Added Theming API parameters page in API docs

BernieSumption:
Fix [AG-16513](https://ag-grid.atlassian.net/browse/AG-16513)


taktran (approved):
LGTM

PR #13191, opened 2026-02-24
AG-16788 Update Rich select docs

StephenCooper:
Fix [AG-16788](https://ag-grid.atlassian.net/browse/AG-16788)

Split into three pages, 
  - overview
  - customisation
  - async values


documentation/ag-grid-docs/src/content/docs/provided-cell-editors-rich-select-customisation/_examples/rich-select-search-values/main.ts
```diff
@@ -22,36 +21,27 @@ const columnDefs: ColDef[] = [
...
             values: colors,
-            cellRenderer: ColourCellRenderer,
             searchType: 'match',
-            valueListMaxHeight: 220,
         } as IRichCellEditorParams,
     },
     {
         headerName: 'Match Any Search',
         field: 'color',
-        cellRenderer: ColourCellRenderer,
```
StephenCooper, line 48: I remove the cell renderer as I feel this is extra noise for the search values example and also it is not compatible according to the JsDoc for fuzzy matching.

gportela85 (approved):

PR #13192, opened 2026-02-24
AG-4324 Add groupHideColumnsUntilExpanded option for multi-column group display

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-4324

Adds a new `groupHideColumnsUntilExpanded` grid option that works with `groupDisplayType='multipleColumns'`. When enabled, group columns are shown dynamically based on expansion state — only the top-level group column is initially visible, and each subsequent level becomes visible when at least one group at the preceding level is expanded.

Fix [AG-4324](https://ag-grid.atlassian.net/browse/AG-4324)


packages/ag-grid-enterprise/src/rowHierarchy/autoColService.ts
```diff
@@ -273,6 +286,77 @@ export class AutoColService extends BeanStub implements NamedBean, IColumnCollec
         return res;
     }
 
+    // Recursively walks expanded group nodes to find the deepest level with an expanded node,
+    // up to maxLevel. Only descends into a node's children if that node itself is expanded,
+    // so collapsed parents never contribute to the result. Exits early once maxLevel is reached.
+    private getDeepestExpandedLevel(nodes: RowNode[] | null | undefined, maxLevel: number): number {
```
gportela85, line 294: I don't think we need all of these comment above this method. It's pretty obvious what is going on. Can we also keep the one-line curlies?

Something like

```ts
private getDeepestExpandedLevel(nodes: RowNode[] | null | undefined, maxLevel: number): number {
    let deepest = -1;

    if (!nodes) {
        return deepest;
    }

    for (const node of nodes) {
        if (!node.group || !node.expanded) {
            continue;
        }

        if (node.level > deepest) {
            deepest = node.level;
        }

        if (deepest >= maxLevel) {
            return deepest;
        }

        // only expanded nodes recurse into their child groups; collapsed branches are skipped.
        const childDeepest = this.getDeepestExpandedLevel(node.childrenAfterGroup, maxLevel);
        if (childDeepest > deepest) {
            deepest = childDeepest;
        }
        if (deepest >= maxLevel) {
            return deepest;
        }
    }

    return deepest;
}
```

packages/ag-grid-enterprise/src/rowHierarchy/autoColService.ts
```diff
@@ -273,6 +286,77 @@ export class AutoColService extends BeanStub implements NamedBean, IColumnCollec
...
+                if (childDeepest > deepest) {
+                    deepest = childDeepest;
+                    if (deepest >= maxLevel) return deepest;
+                }
+            }
+        }
+        return deepest;
+    }
+
+    private updateGroupColumnVisibility(): void {
```
gportela85, line 327: Here, I think this would be easier to read (just personal preference)

```ts
private updateGroupColumnVisibility(): void {
    const columns = this.columns?.list;

    if (!columns || columns.length === 0) {
        return;
    }

    const { gos, visibleCols, rowModel } = this.beans;
    const isFeatureEnabled = _isGroupHideColumnsUntilExpanded(gos);

    let changed = false;
    const setColVisible = (col: AgColumn, visible: boolean): void => {
        if (visible !== col.isVisible()) {
            col.setVisible(visible, 'api');
            changed = true;
        }
    };

    const setAllColumnsVisible = (): void => {
        for (const col of columns) {
            setColVisible(col, true);
        }
    };

    if (!isFeatureEnabled) {
        setAllColumnsVisible();
    } else if (columns.length > 1) {
        // Feature only applies when there are multiple columns to show/hide;
        // the first column is always visible so a single column needs no adjustment.
        const maxLevel = columns.length - 2;
        const rootChildren = rowModel?.rootNode?.childrenAfterGroup;
        const deepestExpandedLevel = this.getDeepestExpandedLevel(rootChildren, maxLevel);

        if (deepestExpandedLevel >= maxLevel) {
            setAllColumnsVisible();
        } else {
            for (let level = 0; level < columns.length - 1; level++) {
                setColVisible(columns[level + 1], deepestExpandedLevel >= level);
            }
        }
    }

    if (changed) {
        visibleCols.refresh('api');
    }
}
```

gportela85 (approved):
Just some small code simplification and needs to fix lint/format issues.

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
```diff
@@ -1421,6 +1421,14 @@ export class AgGridAngular<TData = any, TColDef extends ColDef<TData> = ColDef<a
      * @agModule `RowGroupingModule`
      */
     @Input({ transform: booleanAttribute }) public groupHideOpenParents: boolean | undefined = undefined;
+    /** When using `groupDisplayType='multipleColumns'`, hides group columns for levels
```
StephenCooper: Update jsdoc to mention groupHideOpenParents too

PR #13205, opened 2026-02-25
Fix typo

leviv:
I'm not an LLM

Noticed this typo in the [docs](https://www.ag-grid.com/javascript-data-grid/column-moving/) and wanted to fix it


StephenCooper:
Thank you for reporting we will get this fixed for our next release.

PR #13207, opened 2026-02-26
AG-15521 add sideBarPanelAnimationDuration theme parameter

BernieSumption:
Fix [AG-15521](https://ag-grid.atlassian.net/browse/AG-15521)


packages/ag-grid-enterprise/src/sideBar/toolPanelWrapper.ts
```diff
@@ -111,4 +122,52 @@ export class ToolPanelWrapper extends Component {
...
+                eContent.style.minWidth = '';
+                eGui.style.width = savedInlineWidth;
+            }
+        };
+
+        // Don't rely on the transition end event alone for cleanup because
+        // transitions might have been disabled by application or user CSS
+        const fallbackTimeout = setTimeout(cleanup, 300);
+        eGui.addEventListener('transitionstart', () => clearTimeout(fallbackTimeout), { once: true });
+        eGui.addEventListener('transitionend', cleanup, { once: true });
```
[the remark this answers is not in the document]
BernieSumption: Intentional - if we're animating, we'll get a transitionstart and cancel the setTimeout - 300ms is just supposed to be long enough to get the transitionstart event on a slow computer.

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
```diff
@@ -269,4 +269,16 @@ The example below shows two custom layouts for the Columns Tool Panel. Note the
 
 {% gridExampleRunner title="Custom Column Layout" name="custom-layout"  exampleHeight=640 /%}
 
+## Animation
```
StephenCooper: Should this be on the top level Side Bar page?

https://ag-grid.com/javascript-data-grid/side-bar/

I assume this will also apply to the filters tool panel and any custom panels provided within the side bar?

packages/ag-grid-enterprise/src/sideBar/toolPanelWrapper.ts
```diff
@@ -111,4 +122,53 @@ export class ToolPanelWrapper extends Component {
...
+
+        // disable transition, force reflow, and re-enable to synchronously start an animation
+        eGui.style.transition = 'none';
+        eGui.style.width = `${fromWidth}px`;
+        const _ = eGui.offsetWidth;
+        eGui.style.transition = '';
+
+        eGui.style.width = `${toWidth}px`;
+
+        const cleanup = () => {
```
StephenCooper, line 159: Do we need to guard against this being run after the grid / component has been destroyed?

BernieSumption: I wouldn't have thought so - we keep a reference to the element, I'd assume that if it's been removed from the DOM it's still safe to update, unless there's some agStack-specific reason it's not safe?

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
BernieSumption: Yes!

StephenCooper (approved):

PR #13214, opened 2026-02-27
AG-16821 Whitelist Bryntum

seanlandsman:
Fix AG-16821

### DIFF
--- a/packages/ag-grid-enterprise/src/license/shared/licenseManager.ts
+++ b/packages/ag-grid-enterprise/src/license/shared/licenseManager.ts
@@ -198,10 +198,7 @@ export class LicenseManager {
 
     private isWebsiteUrl(): boolean {
         const hostname = this.getHostname();
-        return (
-            hostname.match(/^((?:[\w-]+\.)?ag-grid\.com)$/) !== null ||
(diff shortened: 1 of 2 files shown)


StephenCooper (approved):

packages/ag-grid-enterprise/src/license/shared/licenseManager.ts
```diff
@@ -198,7 +198,10 @@ export class LicenseManager {
 
     private isWebsiteUrl(): boolean {
         const hostname = this.getHostname();
-        return hostname.match(/^((?:[\w-]+\.)?ag-grid\.com)$/) !== null;
+        return (
+            hostname.match(/^((?:[\w-]+\.)?ag-grid\.com)$/) !== null ||
```
StephenCooper: Shall we update to do the or in the regex?

StephenCooper (changes requested):

StephenCooper (approved):

packages/ag-grid-enterprise/src/license/shared/licenseManager.ts
seanlandsman: done - and added tests

PR #13216, opened 2026-02-27
AG-7444 - [sparklines] - added aria announcements

gportela85:
fix [AG-7444](https://ag-grid.atlassian.net/browse/AG-7444)

### DIFF
--- a/packages/ag-grid-enterprise/src/sparkline/sparklinesUtils.ts
+++ b/packages/ag-grid-enterprise/src/sparkline/sparklinesUtils.ts
@@ -121,13 +121,8 @@ const getYValue = (datum: any, yKey: string): number | null => {
         return Number.isFinite(datum) ? datum : null;
     }
 
-    if (Array.isArray(datum)) {
-        const yValue = datum[1];
-        return typeof yValue === 'number' && Number.isFinite(yValue) ? yValue : null;
-    }
-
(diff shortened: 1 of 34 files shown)


packages/ag-grid-enterprise/src/sparkline/sparklinesUtils.ts
```diff
@@ -14,3 +20,116 @@ export const wrapFn = (fn: FunctionParams, wrapperFn: WrapperFunctionParams) =>
...
+}
+
+const hasCompleteSparklineSummary = (summary: SparklineSummary): summary is CompleteSparklineSummary =>
+    summary.count > 0 && summary.min != null && summary.max != null && summary.start != null && summary.end != null;
+
+export const interpolateTemplate = (template: string, values: SparklineTemplateValues): string =>
+    template.replace(/\$\{([^}]+)\}/g, (match, token: string) => values[token] ?? match);
+
+const getYValue = (datum: any, yKey: string): number | null => {
+    if (typeof datum === 'number') {
```
StephenCooper, line 120: Could we reduce the duplication in this method, seems like once we have something that looks like a number it repeats the same code.

gportela85: done

StephenCooper (approved):

PR #13225, opened 2026-03-02
AG-16708 set data value in batch

SalvatorePreviti:
API changes
- rowNode.setDataValue(col, value, eventSource?) — eventSource now accepts 'edit' (push value into open editor in-place), 'batch' (stage pending edit, close editor), and 'data' (write directly to data, bypass batch entirely)
- rowNode.getDataValue(col, from?) — new from parameter: 'data' (default), 'value' (resolves agg wrappers to scalars), 'edit' (live editor → pending → committed), 'batch' (pending → committed), 'data-raw' (raw data, bypasses agg/formulas)
- IAggFuncResult<TAggValue> — exported as a public type

Bug fixes

- cellEditingStopped no longer fires for cells that never had an open editor
- rowEditingStopped no longer fires multiple times during full-row edit with column virtualisation

Tests
- Behavioural tests for all new API paths, event symmetry fixes, and a full-row virtualisation scroll cycle test that verifies pending values survive recycling

FIX [AG-16708](https://ag-grid.atlassian.net/browse/AG-16708)
FIX [AG-16804](https://ag-grid.atlassian.net/browse/AG-16804)


documentation/ag-grid-docs/src/content/docs/cell-editing-batch/index.mdoc
```diff
@@ -47,50 +47,141 @@ The grid API methods `startEditingCell()` and `stopEditing()` can be used to con
...
 
-The `api.getCellValue()` method retrieves the display value for a cell. It can retrieve values at different stages using the `from` parameter:
+Both accept a `from` parameter:
 
--   `'edit'` (default): Current editing value, including live typing and pending batch values
--   `'batch'`: Pending batch value, excluding live typing
--   `'data'`: Stored data value, ignoring all edit state
+| `from`    | Returns                                                            |
+| --------- | ------------------------------------------------------------------ |
+| `'edit'`  | Live editor typing → pending batch value → committed data          |
```
StephenCooper: Could we make this table clearer by having a "Fallbacks" column next to the Returns?

packages/ag-grid-community/src/interfaces/iCellEditor.ts
```diff
@@ -52,6 +52,18 @@ export interface BaseCellEditor {
     getValidationErrors?(): string[] | null;
 }
 
+/**
+ * AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time.
```
StephenCooper:
```suggestion
 * @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time.
```

StephenCooper: But that is not technically required as this interface is not exported from community.

documentation/ag-grid-docs/src/content/docs/cell-editing-batch/index.mdoc
SalvatorePreviti: improved the docs

packages/ag-grid-community/src/interfaces/iCellEditor.ts
SalvatorePreviti: fixed and leaving to make it clear is internal

StephenCooper (approved):

packages/ag-grid-community/src/valueService/valueService.ts
gportela85, line 306: Docs say 'data-raw' bypasses valueGetters, but rowNode.getDataValue calls valueSvc.getValue(..., ignoreAggData=true), and this still executes colDef.valueGetter in normal cases.

packages/ag-grid-community/src/interfaces/iRowNode.ts
```diff
@@ -315,36 +328,75 @@ export interface IRowNode<TData = any> extends BaseRowNode<TData>, GroupRowNode<
...
      *
-     * @param colKey The column where the value should be updated
-     * @param newValue The new value
-     * @param eventSource The source of the event
-     * @returns `true` if the value was changed, otherwise `false`.
+     * - **(default)** — Batch active: stages a pending batch edit; any open editor is closed.
+     *   Batch inactive: closes the editor and writes to data, or writes directly if no editor is open.
+     * - **`'edit'`** — Batch active: updates the live editor value without committing; stages pending if no editor.
+     *   Batch inactive: updates the live editor value; writes directly if no editor is open.
+     * - **`'batch'`** — Batch active: stages a pending batch edit; any open editor is closed.
```
gportela85: This says batch source closes open editors, but runtime/docs in rowNode.ts  and implementation in editService.ts#applyDirectValue keep them open.

packages/ag-grid-community/src/edit/cellEditors/agAbstractCellEditor.ts
```diff
@@ -46,6 +46,11 @@ export abstract class AgAbstractCellEditor<P extends ICellEditorParams = any, TV
         this.eEditor.onValueChange(() => params.validate());
     }
 
+    public agSetEditValue(value: TValue | null | undefined): void {
+        this.params.value = value;
+        this.eEditor.setValue(value as any, true);
```
gportela85: Not a big fan of casting as any here. I understand that some editors have different editor-value vs cell-value types (eg. Date cell value but string input in simple/date editors). So a single TValue generic for both sides does not always fit.

If it's not possible to fix this now because it would be too much work, can we look into adding a second generic for editor value type (TEditorValue) and wire it through eEditor?

packages/ag-grid-community/src/interfaces/iRowNode.ts
SalvatorePreviti: yes was the old behaviour and fixed it and forgot to update the comment, fixed

packages/ag-grid-community/src/valueService/valueService.ts
SalvatorePreviti: yes was the old behaviour and fixed it and forgot to update the comment, fixed

packages/ag-grid-community/src/edit/cellEditors/agAbstractCellEditor.ts
SalvatorePreviti: improved generics and typing in the editors and added more tests

PR #13232, opened 2026-03-03
AG-1190: Maintain row group expansion state across column changes and add resetRowGroupExpansion API

StephenCooper:
Fix [AG-1190](https://ag-grid.atlassian.net/browse/AG-1190), [AG-16321](https://ag-grid.atlassian.net/browse/AG-16321)

## Summary

- **New API: `resetRowGroupExpansion()`** — Discards all user-initiated expand/collapse overrides and re-evaluates each group against the configured defaults (`groupDefaultExpanded`, `isGroupOpenByDefault`, or `isServerSideGroupOpenByDefault`). Works across both CSRM and SSRM.
- **Preserve expansion state across CSRM group column changes** — When row group columns are added, removed, or reordered via `setGridOption('columnDefs')` or column API, groups whose IDs survive the change retain their expanded/collapsed state. Groups at or beyond the first dirty level fall back to defaults. Uses an expansion snapshot during `shotgunResetEverything` to restore state for nodes with unchanged IDs.
- **Fix `ssrmRowGroupExpansion` state restore** — `setState()` and `initialState` now correctly restore `ssrmRowGroupExpansion` (the `ExpandAllStrategy` bulk state), which was previously ignored.
- **Fix stale cached state after CSRM regroup** — `newColumnsLoaded` now triggers a debounced expansion state refresh so `getState()` reflects post-regroup node IDs rather than stale pre-regroup IDs.

> **Note:** SSRM-specific "maintain expansion state across group column changes" has been split into a separate branch (`AG-1190/ssrm-maintain-expansion-state`) for independent review.

### Core API
- `gridApi.ts` — Added `resetRowGroupExpansion()` to `_CsrmSsrmSharedGridApi`
- `rowModelSharedApi.ts` / `sharedApiModule.ts` — Wired up the new API function
- `iExpansionService.ts` — Added `isExpanded()` and `resetExpansion()` to the expansion service interface

### RowNode
- `rowNode.ts` — `_expanded` backing field with getter/setter for `expanded`, enabling lazy default resolution via `isExpanded()` on the expansion service

### CSRM
- `clientSideExpansionService.ts` — Implements `isExpanded()` and `resetExpansion()` by iterating all group/master nodes and re-evaluating defaults
- `groupStrategy.ts` — Takes an `ExpansionSnapshot` during group column changes to preserve expansion state through `shotgunResetEverything`; extracted pivot leaf-group check into `_getGroupNodeDefaultExpanded`
- `rowHierarchyUtils.ts` — New `_getGroupNodeDefaultExpanded()` function that handles pivot leaf groups, snapshot lookups, then falls back to `_getRowDefaultExpanded`

### SSRM
- `serverSideExpansionService.ts` — Implements `isExpanded()` with lazy default resolution via `_expanded`; `resetExpansion()` creates a fresh strategy and re-evaluates all nodes; `setExpansionState` dispatches `storeUpdated` when nodes change
- `blockUtils.ts` — Updated to use `_expanded` backing field

### State Service
- `stateService.ts` — Extracted `updateGroupExpansionState()` to deduplicate ssrm/csrm state caching; `shouldSetState` now checks `ssrmRowGroupExpansion`; `setRowGroupExpansionState` passes through ssrm state; `newColumnsLoaded` triggers expansion state refresh for CSRM

### Documentation
- Added `resetRowGroupExpansion` to the API docs with cross-references
- New "Reset Group Expansion" example for CSRM grouping docs

### Tests
- `grid-state.test.ts` — Tests for CSRM/SSRM state capture and restore, `ssrmRowGroupExpansion` bulk state, `initialState` with bulk expansion, and stale state after column changes
- `grouping-change-column.test.ts` — Comprehensive CSRM tests for adding/removing group columns, top-level column removal, reordering, `groupDefaultExpanded` interaction, pivot mode, and `expandAll` preservation
- `row-node-expanded.test.ts` — Tests for `_expanded` backing field and lazy default resolution


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -249,7 +245,10 @@ export class StateService extends BeanStub implements NamedBean {
         if (shouldSetState('filter', filterState)) {
             this.setFilterState(filterState);
         }
-        if (shouldSetState('rowGroupExpansion', rowGroupExpansionState)) {
+        if (
+            shouldSetState('rowGroupExpansion', rowGroupExpansionState) ||
+            shouldSetState('ssrmRowGroupExpansion', ssrmRowGroupExpansion)
```
StephenCooper, line 250: This fixes a bug with restoring SSRM expansion state from initial state.

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -31,22 +31,66 @@ export class ServerSideExpansionService
...
+                const newGroupColCount = newGroupColIds.length;
+                const maxValidLevel = this.beans.colModel.isPivotMode() ? newGroupColCount - 1 : newGroupColCount;
+                const upperBound = Math.min(maxValidLevel, firstDirtyLevel);
+                if (upperBound >= 0) {
+                    const isValidLevel = (id: string) => {
+                        const level = oldStrategy.getNodeLevel(id);
+                        return level != null && level < upperBound;
+                    };
+                    expandedRowGroupIds = expandedRowGroupIds.filter(isValidLevel);
+                    collapsedRowGroupIds = collapsedRowGroupIds?.filter(isValidLevel);
```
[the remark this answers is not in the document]
StephenCooper: Correct, added a new test to confirm this and then fixed.

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -31,22 +31,69 @@ export class ServerSideExpansionService
...
+                }
+
+                if (expandedRowGroupIds.length || collapsedRowGroupIds?.length) {
+                    newStrategy.setExpandedState({ expandedRowGroupIds, collapsedRowGroupIds });
+                }
+            }
+            this.strategy = newStrategy;
+        };
```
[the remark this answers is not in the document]
StephenCooper: Updated tests to validate this.

packages/ag-grid-enterprise/src/serverSideRowModel/services/expansion/strategies/defaultStrategy.ts
```diff
@@ -53,6 +54,7 @@ export class ExpandStrategy extends BeanStub implements IExpansionStrategy<RowGr
      */
     public setRowExpanded(row: IRowNode, expanded: boolean) {
         const id = row.id!;
+        this.nodeLevels.set(id, row.level);
```
SalvatorePreviti: this seems very expensive - do we need this map?

SalvatorePreviti:
TreeData filler nodes is not covered in this PR - not sure if this is a problem, or something for the future

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -98,6 +105,20 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
         this.beans.selectionSvc?.updateSelectableAfterGrouping(changedPath);
     }
 
+    /** When group columns change, snapshot expanded and collapsed group node IDs so they can be restored after rebuild */
+    private getExpansionSnapshot(): ExpansionSnapshot {
```
SalvatorePreviti: ok - this is quite expensive - since nonLeafsById is kept alive and we do not recreate nodes when not needed - maybe we need a property or a flag in treeNodeFlags instead of keeping 2 sets? not sure if that would help

But - the problem of going by ID is that adding or removing columns lose the state:
Group nodes gets recreated if they change ID as by contract we cannot change an ID of a node, and here we have the same issue - if an ID change we lose the state - does this solve the user problem?

the issue is that when we add or remove a grouping column the ID will not be the same as they are constructed from their path - and we keep filler nodes alive if they maintain the same ID - so is this code really solving the issue, and, do we need those sets? When we do not destroy and recreate the group as the id did not change, we could understand why we lose the expanded state instead if that is the problem

StephenCooper: This has been addressed by Salvo's commits.

packages/ag-grid-enterprise/src/serverSideRowModel/services/expansion/strategies/defaultStrategy.ts
StephenCooper: Have pulled SSRM changes out into a seperate branch.

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -100,6 +109,12 @@ export class ServerSideExpansionService
         this.updateExpandedState(node);
     }
 
+    public resetExpansion(): void {
+        this.strategy = this.createManagedBean(new ExpandStrategy());
```
SalvatorePreviti: destroy old bean

packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
```diff
@@ -413,6 +412,7 @@ const adjustAddIndexForDataPath = <TData>(allLeafs: RowNode<TData>[], addIndex:
 const initRootNode = <TData = any>(rootNode: RowNode<TData>): RowNode<TData> => {
     rootNode.group = true;
     rootNode.level = -1;
+    rootNode._expanded = true;
```
sdwvit, line 415: why is it underscored?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -244,8 +244,23 @@ export class RowNode<TData = any>
     /** Server Side Row Model Only - the children are in an infinite cache. */
     public childStore: IServerSideStore | null;
 
-    /** `true` if group is expanded, otherwise `false`. */
-    public expanded: boolean;
+    /**
+     * Backing field for `expanded` property.
+     * - `true`/`false`: explicit expansion state.
+     * - `null`: lazy — in CSRM, getter resolves the default on first access and caches it.
```
sdwvit: what does null mean in SSRM?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -244,8 +244,23 @@ export class RowNode<TData = any>
...
+     */
+    public _expanded: boolean | null | undefined = undefined;
+
+    /** `true` if group or master row is expanded. */
+    public get expanded(): boolean {
+        const expansionSvc = this.beans.expansionSvc;
+        return expansionSvc ? expansionSvc.isExpanded(this) : this.level === -1 ? true : !!this._expanded;
+    }
+
+    public set expanded(value: boolean) {
```
sdwvit, line 261: do you need a setter?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -244,8 +244,23 @@ export class RowNode<TData = any>
...
 
-    /** `true` if group is expanded, otherwise `false`. */
-    public expanded: boolean;
+    /**
+     * Backing field for `expanded` property.
+     * - `true`/`false`: explicit expansion state.
+     * - `null`: lazy — in CSRM, getter resolves the default on first access and caches it.
+     * - `undefined`: uninitialized, means false.
+     */
+    public _expanded: boolean | null | undefined = undefined;
```
sdwvit, line 253: there is setter and getter, yet the prop is public?

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -836,12 +823,18 @@ export class StateService extends BeanStub implements NamedBean {
     }
 
     private getRowGroupExpansionState(): RowGroupExpansionState | RowGroupBulkExpansionState | undefined {
```
sdwvit: probably can inline and get rid of this method

SalvatorePreviti (approved):

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -862,13 +855,11 @@ export class StateService extends BeanStub implements NamedBean {
...
     ): void {
-        const expansionSvc = this.beans.expansionSvc;
-        if (!expansionSvc) {
-            return;
-        }
-
-        const newExpansionState = rowGroupExpansionState ?? { expandedRowGroupIds: [], collapsedRowGroupIds: [] };
-        expansionSvc.setExpansionState(newExpansionState, source);
+        this.beans.expansionSvc?.setExpansionState(
+            ssrmRowGroupExpansionState ??
```
sdwvit: lets extract this into a const for readability

packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -543,7 +543,7 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
                     return;
                 }
 
-                if (!node.expanded && !node.footer) {
+                if (!node.footer && !node.expanded) {
```
sdwvit, line 546: what is the reasons of switching places here?

sdwvit: oh expanded is more expensive?

packages/ag-grid-community/src/valueService/valueService.ts
```diff
@@ -235,19 +235,22 @@ export class ValueService extends BeanStub implements NamedBean {
...
-            return false; // isPivotLeaf - not an opened group
+            return false;
         }
-
-        // isOpenedGroup is true. Now check if groupShowsAggData.
-        // groupShowsAggData = this.gos.get('groupSuppressBlankHeader') || !node.sibling
-        // We return true only if !groupShowsAggData, i.e., !groupSuppressBlankHeader && node.sibling
-        return !!node.sibling && !this.gos.get('groupSuppressBlankHeader');
+        // node.expanded (getter with side effects) evaluated last
+        return !!node.expanded;
```
sdwvit: does it need !! ?

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -96,15 +94,14 @@ export class MasterDetailService extends BeanStub implements NamedBean, IMasterD
...
-                row.expanded = _getRowDefaultExpanded(beans, row, level, false);
-            } else if (!newMaster && oldMaster) {
-                // if changing AWAY from master, then un-expand, otherwise next time it's shown it is expanded again
-                row.expanded = false;
+            if (
+                (newMaster && created) ||
+                // if changing AWAY from master, forget current state
+                (!newMaster && oldMaster)
+            ) {
+                row._expanded ??= null;
```
sdwvit, line 104: can we avoid checking ?? ?

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -137,28 +137,31 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
...
+                    return 'skip'; // no change to grouping
                 }
             } else {
-                // Group columns changed.
-                params.animate = false; // if grouping columns change, we don't animate the regrouping
-                makeGroupColumns(cols, this.groupCols);
+                params.animate = false;
+                // If the top-level group column changed, every existing node ID will differ after
+                // rebuild, so no group nodes can be reused. Check before makeGroupColumns overwrites groupCols.
+                const topLevelChanged = groupCols[0]?.col.getId() !== cols?.[0]?.getId();
```
sdwvit, line 158: should this be extracted into a const ? 

```
const topLevelGroupId = groupCols[0]?.col.getId();
```
can you pls elaborate why topLevelGroupId !== cols?.[0]?.getId() indicates that top level group column has changed?

packages/ag-grid-community/src/selection/selectionService.ts
SalvatorePreviti: expanded now is more expensive - as is a computed property that lazily compute the initial state if the backing property _expanded is null - we want to delay it as much as possible

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: the field is public - is a pattern we used in other fields as well as some parts of the code want to directly access the backing field in some cases or for performance or for implementing, like in this case, specific logic. Is still hidden to the TS user as RowNode class is internal IRowNode is public

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -431,11 +440,37 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
...
+        let nodesToDeselect: RowNode[] | undefined;
+        for (const [id, node] of groupsById) {
+            if (node.childrenAfterGroup !== null) {
+                continue;
+            }
+            if (selectionSvc && node.isSelected()) {
+                (nodesToDeselect ??= []).push(node);
+            }
+            groupsById.delete(id);
+            node._destroy(false);
```
sdwvit, line 460: why does destroy have underscore in front of it? does it also have getter / setter?

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -485,54 +520,69 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
...
-
-        groupNode.key = key;
-        groupNode.id = id;
-
-        groupNode.level = level;
-        groupNode.leafGroup = isLeafLevel;
+        let node = groupsById.get(id);
+        let reused = false;
+
+        if (node !== undefined) {
```
sdwvit: lets flip if-else please for readability, 
or simplify to `if (node)`

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
```diff
@@ -485,54 +520,69 @@ export class GroupStrategy extends BeanStub implements IRowGroupingStrategy {
...
-        groupNode.field = groupCol.field ?? null;
-        groupNode.rowGroupColumn = col;
-
-        groupNode.key = key;
-        groupNode.id = id;
-
-        groupNode.level = level;
-        groupNode.leafGroup = isLeafLevel;
+        let node = groupsById.get(id);
+        let reused = false;
```
sdwvit:
```suggestion
        let singleUse = true;
```
for readability

packages/ag-grid-enterprise/src/rowHierarchy/baseExpansionService.ts
```diff
@@ -51,7 +72,7 @@ export abstract class BaseExpansionService extends BeanStub {
 
     private updateExpandedCss(rowCtrl: RowCtrl, rowNode: RowNode): void {
         const expandable = rowNode.isExpandable();
-        const expanded = rowNode.expanded == true;
+        const expanded = !!rowNode.expanded;
```
sdwvit: do we need !! here?

packages/ag-grid-enterprise/src/rowHierarchy/baseExpansionService.ts
```diff
@@ -37,6 +37,27 @@ export abstract class BaseExpansionService extends BeanStub {
...
+            return groupDefaultExpanded === -1 || level < groupDefaultExpanded;
+        }
+        const params = _addGridCommonParams<IsGroupOpenByDefaultParams>(gos, {
+            rowNode,
+            field: rowNode.field!,
+            key: rowNode.key!,
+            level,
+            rowGroupColumn: rowNode.rowGroupColumn!,
+        });
+        return isGroupOpenByDefault(params) == true;
```
sdwvit:
```suggestion
        return !!isGroupOpenByDefault(params);
```

sdwvit: or even drop !! as name suggests it is already a bool

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -62,6 +62,37 @@ export class ClientSideExpansionService
         return this.getInternalExpansionState();
     }
 
+    public isExpanded(rowNode: RowNode<any>): boolean {
```
sdwvit, line 65: why does it live here and not on rowNode itself?

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -62,6 +62,37 @@ export class ClientSideExpansionService
...
+        let value = rowNode._expanded;
+        if (value === null) {
+            // Lazy resolution of the default expansion state via the enterprise expansion service.
+            value = this.defaultExpanded(rowNode) ?? false;
+            rowNode._expanded = value;
+        }
+        return !!value;
+    }
+
+    public resetExpansion(): void {
```
sdwvit, line 83: same question

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
```diff
@@ -62,6 +62,37 @@ export class ClientSideExpansionService
...
+        const { rowModel } = this.beans;
+
+        rowModel.forEachNode((node) => {
+            if (!node.group && !node.master) {
+                return;
+            }
+            node._expanded = null; // null triggers lazy default resolution in the expanded getter
```
sdwvit, line 90: i am wondering if there is a way to integrate this explanation somewhere so it's not repeated. Jsdoc for the _expanded prop?

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
SalvatorePreviti: 1) ssrm and csrm have two different implementations and is much cleaner using method override than putting the condition on the rowNode, so the rowNode delegates it to the service
2) bundle size - csrm and ssrm are two modules never loaded together, and, community only render flat grids so expanded is always true for the root and false for the rest

packages/ag-grid-enterprise/src/rowHierarchy/flattenUtils.ts
```diff
@@ -57,19 +57,28 @@ export function _shouldRowBeRendered(
...
+    }
+    if (isRemovedSingleChildrenGroup || isRemovedLowestSingleChildrenGroup) {
+        return false;
+    }
 
     // hide open parents means when group is open, we don't show it. we also need to make sure the
     // group is expandable in the first place (as leaf groups are not expandable if pivot mode is on).
     // the UI will never allow expanding leaf groups, however the user might via the API (or menu option 'expand all row groups')
-    const neverAllowToExpand = skipLeafNodes && rowNode.leafGroup;
+    if (!details.hideOpenParents) {
```
sdwvit, line 70: lets try to avoid negated conditions in here. see next comment

packages/ag-grid-enterprise/src/rowHierarchy/flattenUtils.ts
```diff
@@ -57,19 +57,28 @@ export function _shouldRowBeRendered(
...
+        return true;
+    }
 
-    const isHiddenOpenParent = details.hideOpenParents && rowNode.expanded && !rowNode.master && !neverAllowToExpand;
+    if (!rowNode.master && rowNode.level !== -1) {
```
sdwvit: possible to flip the check? 

```suggestion
    if (rowNode.master || rowNode.level === -1) {
       return true;
    }
    ...
```

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
SalvatorePreviti: no is just a private method that the user should not call - when we add new things we add _ so the user know is internal

packages/ag-grid-enterprise/src/rowHierarchy/flattenUtils.ts
sdwvit: and i'd extract 
```
const isRoot = rowNode.level === -1;
```

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
SalvatorePreviti: well this is correct as is - before we were invoking the user callback here - now we just set the node to null indicating the expansion service have to fetch and call the callback when the property is accessed. This is correct as is

packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
sdwvit: because of getters and setters

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: no but we need to be compatible with the original interface and we don't want to break the contract

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: is lazy too - maybe @StephenCooper review the jsdoc here

packages/ag-grid-community/src/clientSideRowModel/clientSideNodeManager.ts
SalvatorePreviti: private - and field access is faster than property setter, up to you if you want to change it but for me is fine, is an internal field

sdwvit (approved):

SalvatorePreviti:
There is a sonarqube warning to fix

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
StephenCooper: This likely comes from the time before IRowNode interface existed.

packages/ag-grid-enterprise/src/rowHierarchy/baseExpansionService.ts
StephenCooper: Not anymore following update the expanded getter.

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
StephenCooper: Answered offline

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
StephenCooper: Now done consistently via a helper.

packages/ag-grid-enterprise/src/rowHierarchy/clientSideExpansionService.ts
StephenCooper: updated

packages/ag-grid-enterprise/src/rowGrouping/groupStrategy/groupStrategy.ts
StephenCooper: The first item in the list is the top most column. If this id has changed then we need to reset all the groups with nothing to maintain.

PR #13257, opened 2026-03-06
AG-3390 - Allow alternative ai api urls and dev token to be used

taktran:


seanlandsman (approved):

taktran:
I'm going to attempt this again, so that there is less boilerplate code for example files

PR #13262, opened 2026-03-09
Fix JIRA transition applying move/mention per-ticket instead of globally

BernieSumption:
Previously, if any ticket in the PR body had a "Fix" prefix, all mentioned tickets were transitioned. Now each ticket is independently classified based on whether its own reference includes "Fix".

Fix [AG-16866](https://ag-grid.atlassian.net/browse/AG-16866)

Additionally, AG-16867 should receive a comment when tis PR is merged

### DIFF
--- a/.github/workflows/update-jira-ticket.yml
+++ b/.github/workflows/update-jira-ticket.yml
@@ -107,8 +107,13 @@ jobs:
                 }
                 const matches = pr.body?.match(regex);
                 if (matches) {
-                    const isMove = !!pr.body.match(/fix(es)? \[?((AG|RTI)-\d{3,})/gi);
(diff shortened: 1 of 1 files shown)


.github/workflows/update-jira-ticket.yml
```diff
@@ -97,7 +103,7 @@ jobs:
                 }
                 const pr = pull_number ? (await github.rest.pulls.get({owner, repo, pull_number})).data : {};
                 const is_merged = !!pr.merged_at;
-                if (!is_merged) {
+                if (!is_merged && process.env.DEBUG_SKIP_MERGE_CHECK !== 'true') {
```
sdwvit:
```suggestion
                if (!is_merged && !process.env.DEBUG_SKIP_MERGE_CHECK) {
```

sdwvit (approved):

.github/workflows/update-jira-ticket.yml
BernieSumption: done

PR #13272, opened 2026-03-10
AG-16842 fix extra border on columns tool panel

BernieSumption:
Fix [AG-16842](https://ag-grid.atlassian.net/browse/AG-16842)

### DIFF
--- a/packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
+++ b/packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
@@ -184,7 +184,9 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
             }
             const eGui = panel.getGui();
             panel.toggleResizable(
-                !eGui.classList.contains('ag-last-column-drop') && !eGui.classList.contains('ag-hidden')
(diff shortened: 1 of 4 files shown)


packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
```diff
@@ -205,6 +207,7 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
             lastVisible.classList.add('ag-last-column-drop');
         }
 
+        this.primaryColsPanel.getGui().classList.toggle('ag-last-visible-child', !lastVisible);
```
StephenCooper, line 210: Is this separate from the lastVisible path above due to the fact that we might need to toggle it off an item that was previously the bottom but now not?

StephenCooper (approved):
If no issue with my comment feel free to merge.

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
BernieSumption: Yes

PR #13276, opened 2026-03-11
AG-16586 ChangedPath, CSRM aggregation pipeline cleanup and optimisation

SalvatorePreviti:
- Split `ChangedPath` into `ChangedRowsPath` (rows-only) and `ChangedCellsPath` (row+column bitmask tracking), replacing the previous monolithic class. Column-level tracking allows the aggregation stage to skip re-aggregating unchanged columns on unchanged group rows.
    - ChangedRowsPath — lightweight Set-based tracker (rows only, no column tracking)
    - ChangedCellsPath — bitmask-based tracker for per-cell (row + column) change tracking

image

image

What we can note from the graph:
    - The pressure on GC is much lower (the red dots on the left shows random timing fluctuations caused by GC - that can slow down at random times the user application itself.
    - The graph shows a fully linear behaviour lower for small grids and huge grids, small trees and big trees, small set of changes and big

This is much faster than the previous ChangedPath that was creating new objects for every row and replicating the tree structure - this use flat arrays and a extremely fast linear O(n) sort by level.
This also correctly excludes leaf nodes and destroyed nodes, that should have not been processed by the stages.
Removing of "active" flag in ChangedPath makes the logic and the full understanding of the flow simple, and makes it faster: now if ChangedPath is undefined, is disabled. If is not undefined, is active.
We now allocate a avg O(1) objects for the change path detection instead of O(n) objects/maps - and by using maps and sets we avoid the extremely slow megamorphic objects for big trees as maps in the ChangedPath

- CSRM stages: now we process only group nodes, the only special case is when a leaf node in treeData is demoted from group to leaf, that now is handled in their stages themselves by clearing aggData and children arrays properly. This avoids to run the expensive execute methods of the stages on leaf nodes, they run only on group nodes now, this is a big performance and clarity win. This makes the flow faster and much cleaner.

- The aggregation stage is also substantially rewritten — flattened from a class-heavy OOP style into module-level functions with pre-resolved column metadata and direct aggregation loops.
    - This considerably reduce the number of RowNode getValue - that is the most expensive thing we currently have and need a full refactoring of the logic in another PR together with improving column resolution by object, that is unfortunately O(n).
    - pre-resolve value-column metadata once per execute call (colId, colDef, aggFunc, colSlot) to avoid repeated property lookups inside the per-group hot loop; unify the child-value collect loop into two branches (aggData / getValue) to keep the common path tight; extract `setAggData` / `setAggDataWithSiblings` / `fireAggDataChangedEvents` into a separate `aggDataUtils.ts` module; split pivot column resolution into a dedicated `resolvePivotColumns` helper that uses a pre-sorted ordered list from the pivot service (pre-partitioned first value columns then totals).
    - we were keeping stale aggData when tree data groups gets demoted to leaf or when switching aggregation - this was a bug and is fixed now

- `PivotResultColsService`: add `getAggregationOrderedList()` which returns pivot result columns pre-partitioned (regular first, totals after) and cached, avoiding repeated sort/partition on every aggregation refresh.

- `RowNode`: initialise children arrays and `allChildrenCount` to `null` (instead of `undefined`) so all null checks can use strict equality, and, the resulting grid is fully respecting the typescript definition with no undefined in those arrays. Initializing in advance can improve the V8 hidden class shape.

- Tests: add comprehensive unit tests for `ChangedRowsPath`, `ChangedCellsPath`, and `sortNodesByDepthFirst`; add behavioural tests for aggregation with changed columns, pivot aggregation, and filter-aggregates stage; add benchmarks for grouping aggregation, pivot aggregation, and changedPath; update snapshot expectations for `null`-initialised children fields.

- Benchmarks: added more benchmarks. Benchmarks showed an improvement of +4 ... 6% avg in various scenarios with medium or big grids
    - Measured that the highest cost we pay is in RowNode getValue - and we have some quick wins to make it much faster, this can improve massively the performance of the grid, maybe another 15% in the benches - in another PR in the future
    - Also treeData now can be massively improved - in another PR in the future


packages/ag-grid-community/src/utils/changedPathImpl/changedCellsPath.ts
```diff
@@ -0,0 +1,248 @@
...
+     * ```
+     */
+    public hasCellBySlot(rowSlot: number, colSlot: number): boolean {
+        if (rowSlot < 0) {
+            return true;
+        }
+        if (colSlot < 0) {
+            return false;
+        }
+        const word = colSlot < 32 ? this.bits : this.extraBits![(colSlot >>> 5) - 1];
```
StephenCooper: Is this logic repeated above on purpose for performance?

packages/ag-grid-community/src/utils/changedPathImpl/changedCellsPath.ts
```diff
@@ -0,0 +1,248 @@
...
+        return (word[rowSlot] & (1 << (colSlot & 31))) !== 0;
+    }
+
+    /** Registers a new row and all its unregistered ancestors. Returns the row's bitmask index.
+     * Time: O(D × ⌈C/32⌉) where D = depth, C = number of tracked columns. In practice O(D) because
+     * C < 32 is the common case (single bitmask word per row, no extraBits loop).
+     * Space: O(D × ⌈C/32⌉).
+     */
+    private ensureRow(rowNode: RowNode): number {
+        const slots = this.slots;
```
StephenCooper, line 196: Is this faster than a single deconstruction from `this`?

packages/ag-grid-community/src/utils/changedPath.ts
```diff
@@ -1,203 +1,66 @@
...
-    children: PathItem[] | null; // children of this node - will be a subset of all the nodes children
-}
-
-// when doing transactions, or change detection, and grouping is present
-// in the data, there is no need for the ClientSideRowModel to update each
-// group after an update, ony parts that were impacted by the change.
-// this class keeps track of all groups that were impacted by a transaction.
-// the the different CSRM operations (filter, sort etc) use the forEach method
-// to visit each group that was changed.
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
```
StephenCooper, line 5: These Jsdoc are redundant as the classes already have them.

packages/ag-grid-enterprise/src/aggregation/aggregationStage.ts
```diff
@@ -39,392 +62,387 @@ export class AggregationStage extends BeanStub implements NamedBean, _IRowNodeAg
...
-        }
-
-        return result;
     }
 
     public getAggregatedChildren(rowNode: RowNode | null | undefined, col: AgColumn | null | undefined): RowNode[] {
-        if (!rowNode?.group || !this.clientSide) {
-            return []; // only group nodes have aggregated children, and only supported in CSRM
+        const { gos } = this;
+        if (!rowNode?.group || !_isClientSideRowModel(gos)) {
```
StephenCooper: Is _isClientSideRowModel a bit more expensive than checking the local prop?

StephenCooper (approved):

PR #13278, opened 2026-03-11
Disable sort-root-tsconfig-paths in nx format commands

StephenCooper:
Add `--sort-root-tsconfig-paths=false` to `format-write` and `format-check` targets so the flag doesn't need to be passed on the command line. Nx defaults this to `true` which causes comments in tsconfig to be lost.

### DIFF
--- a/utilities/all/project.json
+++ b/utilities/all/project.json
@@ -166,12 +166,15 @@
         "commands": ["node external/ag-shared/scripts/watch/watch.js grid", "nx run ag-grid-docs:dev"]
       }
     },
+    "format-write": {
+      "command": "nx format:write --sort-root-tsconfig-paths=false"
+    },
     "format-check": {
       "executor": "nx:run-commands",
       "options": {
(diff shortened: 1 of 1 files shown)


utilities/all/project.json
```diff
@@ -166,12 +166,15 @@
         "commands": ["node external/ag-shared/scripts/watch/watch.js grid", "nx run ag-grid-docs:dev"]
       }
     },
+    "format-write": {
+      "command": "nx format:write --sort-root-tsconfig-paths=false"
```
taktran: Could add `--all` here too, but up to you whether you want to leave it to the user or not

taktran (approved):
LGTM - just a minor comment

utilities/all/project.json
StephenCooper:
```suggestion
      "command": "nx format:write --all --sort-root-tsconfig-paths=false"
```

PR #13291, opened 2026-03-13
AG-16586 batch aggregation

SalvatorePreviti:
### New APIs ###

- IRowNode.primaryRow: Resolves footer and pinned sibling relationships to the canonical row node. For footer rows returns the parent group; for pinned rows returns the source row; for pinned footers follows both links.
- RowGroupingEditModule: New enterprise module for group row editing functionality.
- IRowModel.hierarchical: True when grouping or tree data is active (group nodes may exist below root).

- removes IRowModel as export from main - IRowModel is internal

### Change Detection and Aggregation Batching ###

ChangeDetectionService rewritten with beginDeferred()/endDeferred() batching. When a deferred block is active, cellValueChanged events accumulate into a ChangedPath and a leaf-node set rather than triggering individual aggregation + refresh passes. On endDeferred(), a single aggregation pass runs over all accumulated changes in depth first order, then affected rows are refreshed in depth-first order. Supports nesting via a depth counter and handles re-entrant events in a follow-up pass. This is not just an optimization but required for correct functioning and avoid the risk of infinite loops if things are changed mid way

Callers wrapped in beginDeferred()/endDeferred():
- UndoRedoService.processAction: multi-cell undo/redo
- EditService: batch editing end commit
- ValueService.setDataValue: group row cascading writes
- GroupEditService: group-level value setting with child propagation
- AgFillHandle: fill-handle drag operations
- RangeService: cut operations on cell ranges

### ChangedPath Moved to Enterprise ###

ChangedCellsPath and ChangedRowsPath implementations moved from community to enterprise (rowHierarchy/changedPathImpl/). Community retains only interface definitions and _forEachChangedGroupDepthFirst utility. New ChangedPathModule provides ChangedPathFactory bean.

This frees ag-grid-community of 1.8k

### Optimisations ###

- CellCtrl.refreshCell: Optional params object instead of always-allocated destructured arg.
- RowRenderer: Added getRowCtrlByNode (O(1) lookup by RowNode, O(k) for sticky rows) and refreshRowByNode (null-safe single-row refresh) - much faster for ChangedDetectionService flow.
- FilterAggregatesStage: Fix issues and improve perf.

FIX AG-16586


packages/ag-grid-community/src/main.ts
```diff
@@ -464,7 +464,7 @@ export type { RedrawRowsParams } from './interfaces/iRedrawRowsParams';
 export type { ICheckboxCellRendererParams } from './rendering/cellRenderers/checkboxCellRenderer';
 
 // Row Model
-export type { IRowModel, RowBounds, RowModelType } from './interfaces/iRowModel';
+export type { RowBounds, RowModelType } from './interfaces/iRowModel';
```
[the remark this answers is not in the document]
SalvatorePreviti: This is fine

documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
```diff
@@ -27,6 +27,10 @@ connecting the grid with a store.
 
 ## Editing Group Row Cells
 
+{% note %}
+`groupRowEditable` and `groupRowValueSetter` require the `RowGroupingEditModule` to be registered.
```
StephenCooper, line 31: Have you also added the module level validation for these properties?

packages/ag-grid-community/src/clientSideRowModel/filterStage.ts
```diff
@@ -94,7 +94,12 @@ export class FilterStage extends BeanStub implements IRowNodeFilterStage, NamedB
             treeDataDepthFirstFilter(this.beans.rowModel.rootNode!, false);
         } else {
             const defaultFilterCallback = (rowNode: RowNode) => filterCallback(rowNode, false);
-            _forEachChangedGroupDepthFirst(this.beans.rowModel.rootNode, changedPath, defaultFilterCallback);
+            _forEachChangedGroupDepthFirst(
+                this.beans.rowModel.rootNode,
```
StephenCooper, line 98: Should we reduce the duplication of `this.beans.rowModel` as a small tidy up in this method?

packages/ag-grid-community/src/edit/editService.ts
```diff
@@ -648,20 +648,32 @@ export class EditService extends BeanStub implements NamedBean {
...
-                const position: Required<EditPosition> = { rowNode, column };
-
-                if (_sourceAndPendingDiffer(editValue) && !hasValidationErrors) {
-                    const cellCtrl = _getCellCtrl(this.beans, position);
-                    const success = this.setNodeDataValue(rowNode, column, editValue.pendingValue, cellCtrl, source);
-                    if (!success) {
-                        editsToDelete.push(position);
+        const { changeDetectionSvc } = this.beans;
+        changeDetectionSvc?.beginDeferred();
+        try {
```
StephenCooper, line 653: What can cause this to throw?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -137,6 +137,19 @@ export class RowNode<TData = any>
      */
     public pinnedSibling?: RowNode<TData>;
 
+    /** @inheritDoc */
```
StephenCooper, line 140: Does this pull the doc from IRowNode?

StephenCooper (approved):

packages/ag-grid-community/src/edit/editService.ts
StephenCooper: Or this is just a clean pattern to ensure we always end the deferred no matter if we exit early.

SalvatorePreviti: just a clean pattern really

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: yes - is optional really but makes it clear to look to the interface for the dev for the docs

PR #13324, opened 2026-03-18
AG-8814: Add datasize URL param and modernise URL param handling

StephenCooper:
## Summary
- Read `datasize` from URL query params on page load to restore the previously selected data size (matching existing `theme` param behaviour)
- Replace fragile manual string manipulation with `URL`/`URLSearchParams` API for updating both `theme` and `datasize` URL params
- Open Monthly Breakdown column group by default with sparkline column visible when collapsed

## Test plan
- [ ] Load the example grid page — verify default data size (1,000 rows) loads
- [ ] Change data size via dropdown — verify URL updates with `?datasize=` param
- [ ] Reload the page with a `?datasize=` param — verify the correct data size is selected
- [ ] Change theme via dropdown — verify URL updates with `?theme=` param
- [ ] Verify both params coexist correctly in the URL (e.g. `?theme=balham&datasize=10000x100`)
- [ ] Verify Monthly Breakdown column group is open by default with sparkline visible when collapsed

Fix AG-8814

### DIFF
--- a/documentation/ag-grid-docs/src/components/example-grid/Example.tsx
+++ b/documentation/ag-grid-docs/src/components/example-grid/Example.tsx
@@ -137,7 +137,17 @@ const staticGridOptions: GridOptions = {
     ),
 };
 
-const ExampleInner = ({ darkMode, theme, isSmall }: { darkMode: boolean; theme: string; isSmall: boolean }) => {
+const ExampleInner = ({
+    darkMode,
+    theme,
+    isSmall,
+    datasizeStr,
+}: {
+    darkMode: boolean;
+    theme: string;
+    datasizeStr: string | undefined;
+    isSmall: boolean;
+}) => {
     const gridRef = useRef(null);
     const loadInstance = useRef(0);
     const [gridThemeStr, setGridThemeStr] = useState(theme);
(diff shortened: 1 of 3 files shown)


documentation/ag-grid-docs/src/components/example-grid/Toolbar.tsx
```diff
@@ -63,16 +69,7 @@ export const Toolbar = ({
             value: newTheme,
         });
 
-        if (!IS_SSR) {
```
StephenCooper, line 66: Need to restore the !IS_SSR check into the utility method

documentation/ag-grid-docs/src/components/example-grid/Example.tsx
```diff
@@ -429,12 +445,17 @@ const ExampleInner = ({ darkMode, theme, isSmall }: { darkMode: boolean; theme:
 
 const Example = () => {
     const [darkMode] = useDarkmode();
-    const [gridThemeStr] = useState<string>(() => new URLSearchParams(window.location.search).get('theme') ?? 'quartz');
+    const [gridThemeStr] = useState<string>(() =>
+        IS_SSR ? 'quartz' : new URLSearchParams(window.location.search).get('theme') ?? 'quartz'
+    );
+    const [datasizeStr] = useState<string | undefined>(() =>
```
taktran: Nit pick:

```suggestion
    const [dataSizeStr] = useState<string | undefined>(() =>
```

documentation/ag-grid-docs/src/components/example-grid/Toolbar.tsx
```diff
@@ -48,6 +57,8 @@ export const Toolbar = ({
             type: 'dataSize',
             value,
         });
+
+        updateUrlParam('datasize', value);
```
taktran: Nit pick:


```suggestion
        updateUrlParam('dataSize', value);
```

taktran (changes requested):
Looks good, just some nitpicks

taktran (approved):
LGTM

PR #13327, opened 2026-03-18
[AG-3471] Maintenance(column): tool panel deferred apply mode

sdwvit:
Replaces the deferApply: boolean param with a buttons: ColumnToolPanelAction[] param. Deferred mode is now enabled implicitly when 'apply' is included in the buttons array, aligning with the Filters Tool Panel pattern.
                                                                                                                                                                                                   
  - deferApply: true → buttons: ['apply', 'cancel']                                                                                                                                                
  - Removes the runtime defer mode toggle (DeferModeToggleComp) — mode is now developer-controlled via grid options only
  - Adds a console warning when buttons is configured without 'apply'                                                                                                                              
  - Adds isDeferredMode() helper used throughout to derive deferred state from params                                                                                                              
  - Updates styles to match Filters Tool Panel button layout      

relates to [AG-3471](https://ag-grid.atlassian.net/browse/AG-3471)


packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
```diff
@@ -131,7 +81,7 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
         const hasPivotModule = gos.isModuleRegistered('SharedPivot');
         const hasRowGroupingModule = hasPivotModule || gos.isModuleRegistered('SharedRowGrouping');
 
-        this.isDeferModeEnabled = !!mergedParams.deferApply;
+        this.isDeferModeEnabled = !!mergedParams.buttons?.includes('apply');
```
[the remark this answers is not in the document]
sdwvit: feature not released yet

sdwvit:
ignore ai review, codex is wrong here

community-modules/styles/src/internal/base/parts/_columns-tool-panel.scss
```diff
@@ -46,31 +38,32 @@
...
+    .ag-column-panel-buttons:last-child {
+        padding-bottom: var(--ag-widget-container-vertical-padding);
     }
 
     .ag-column-panel-buttons-button {
         line-height: 1.5;
     }
 
+    .ag-column-panel .ag-standard-button.ag-column-panel-buttons-apply-button {
+        color: var(--ag-filter-panel-apply-button-color);
```
StephenCooper: We should probably not re-use the same variable but add a new one with the same default value.

sdwvit:
still need to update tests.

community-modules/styles/src/internal/base/parts/_columns-tool-panel.scss
sdwvit: gonna address it in a bit, ty!

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/deferred-apply-mode/main.ts
```diff
@@ -88,7 +88,7 @@ const gridOptions: GridOptions<IOlympicData> = {
                 iconKey: 'columns',
                 toolPanel: 'agColumnsToolPanel',
                 toolPanelParams: {
-                    deferApply: true,
+                    buttons: ['apply', 'cancel'],
```
StephenCooper: Nit: Can we swap the order to match what we have in the filters tool panel example.

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -682,6 +766,14 @@ function getDraftFunctionColumnIds(
     return colIds;
 }
 
+function arraysEqual(a: string[], b: string[]): boolean {
```
StephenCooper: We have an existing util method for this.

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -246,6 +250,86 @@ class DeferredColumnStateUpdateStrategy implements ColumnStateConcreteUpdateStra
...
+                if (!column) continue;
+                if (patch.hide !== undefined && patch.hide !== !column.isVisible()) return true;
+                if (patch.rowGroup !== undefined && !!patch.rowGroup !== column.isRowGroupActive()) return true;
+                if (patch.pivot !== undefined && !!patch.pivot !== column.isPivotActive()) return true;
+                if (patch.aggFunc !== undefined && patch.aggFunc !== column.getAggFunc()) return true;
+                if (patch.sort !== undefined && patch.sort !== column.getSortDef()?.sort) return true;
+            }
+        }
+
+        if (columnOrder) {
```
StephenCooper: Maybe we can have a helper method to reduce the duplication here?

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
sdwvit: found it thnx

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
sdwvit: not as easy to do here, but I have some tricks up my sleeve for an base class

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -246,6 +246,79 @@ class DeferredColumnStateUpdateStrategy implements ColumnStateConcreteUpdateStra
...
+                const column = beans.colModel.getColDefCol(colId);
+                if (!column) {
+                    continue;
+                }
+                if (
+                    (patch.hide !== undefined && patch.hide !== !column.isVisible()) ||
+                    (patch.rowGroup !== undefined && !!patch.rowGroup !== column.isRowGroupActive()) ||
+                    (patch.pivot !== undefined && !!patch.pivot !== column.isPivotActive()) ||
+                    (patch.aggFunc !== undefined && patch.aggFunc !== column.getAggFunc()) ||
+                    (patch.aggFunc !== undefined && patch.aggFunc !== column.getAggFunc())
```
StephenCooper, line 264: aggFunc is checked twice, was this meant to be a different property?

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -246,6 +246,79 @@ class DeferredColumnStateUpdateStrategy implements ColumnStateConcreteUpdateStra
...
+                ) {
+                    return true;
+                }
+            }
+        }
+
+        if (columnOrder && !_areEqual(columnOrder.colIds, getPrimaryColumnIds(beans))) {
```
StephenCooper, line 271: Does this work with / without `suppressSyncLayoutWithGrid`

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -246,6 +246,79 @@ class DeferredColumnStateUpdateStrategy implements ColumnStateConcreteUpdateStra
...
+                const column = beans.colModel.getColDefCol(colId);
+                if (!column) {
+                    continue;
+                }
+                if ((sortDef?.direction ?? null) !== (column.getSortDef()?.direction ?? null)) {
+                    return true;
+                }
+            }
+            if (sort.baselineCleared) {
+                for (const col of getPrimaryColumns(beans)) {
```
StephenCooper: Extract `getPrimaryColumns` to const so it is not called multiple times in this function.

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -246,6 +246,79 @@ class DeferredColumnStateUpdateStrategy implements ColumnStateConcreteUpdateStra
...
+                    if (!sort.sortDefsByColId.has(col.getColId()) && col.getSortDef() !== null) {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        if (aggFuncs) {
```
StephenCooper, line 307: Is this different to the aggFunc check in the `columnState` section?

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateStrategy.ts
```diff
@@ -115,83 +78,58 @@ export class ColumnStateUpdateStrategy extends BeanStub implements IColumnStateU
...
     }
 
     public setPivotColumns(deferMode: boolean, columns: AgColumn[], eventType: ColumnEventType): void {
-        const { strategy, fallback } = this.resolveUpdates();
-        if (strategy) {
-            strategy.setPivotColumns(deferMode, columns, eventType);
-            return;
-        }
-        fallback.setPivotColumns(columns, eventType);
+        this.dispatch('setPivotColumns', deferMode, columns, eventType);
```
StephenCooper: I like the reduction in duplication 👍

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateStrategy.ts
```diff
@@ -115,83 +78,58 @@ export class ColumnStateUpdateStrategy extends BeanStub implements IColumnStateU
...
 
-    private resolveUpdates(): {
-        strategy: IColumnStateUpdateStrategy | undefined;
-        fallback: ColumnStateConcreteUpdateStrategy;
-    } {
-        return {
-            strategy: this.getUpdateStrategy(),
-            fallback: this.getFallbackUpdates(),
-        };
+    private dispatch<M extends keyof IColumnStateUpdateStrategy>(
```
StephenCooper: Can we update the name of this method. `dispatch` should be reserved for events. It was confusing seeing a method called dispatch return a value.

I see it is being used for get and set. The set seems ok with `dispatch` but mainly the `get` use case is confusing.

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateStrategy.ts
```diff
@@ -115,83 +78,58 @@ export class ColumnStateUpdateStrategy extends BeanStub implements IColumnStateU
...
+        method: M,
+        deferMode: boolean,
+        ...args: Parameters<IColumnStateUpdateStrategy[M]> extends [any, ...infer Rest] ? Rest : []
+    ): ReturnType<IColumnStateUpdateStrategy[M]> {
+        const strategy = this.getUpdateStrategy();
+        if (strategy) {
+            return (strategy as any)[method](deferMode, ...args);
```
StephenCooper: Anyway to not have the removal of type safety whilst still benefiting from the reduction in code?

packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
StephenCooper: Should this use the helper function too?

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelColumnComp.ts
```diff
@@ -297,17 +292,17 @@ export class ToolPanelColumnComp extends Component {
     private onColumnStateChanged(): void {
         this.processingColumnStateChange = true;
         const updateStrategy = this.beans.columnStateUpdateStrategy;
-        const isPivotMode = updateStrategy.getPivotMode(!!this.params.deferApply);
+        const isPivotMode = updateStrategy.getPivotMode(isDeferredMode(this.params));
```
StephenCooper, line 295: Should all of these methods just take the `params` directly to avoid them all having to use `isDeferedModel`? 

Maybe in a follow up given the static nature of Defer Mode I wonder if the CTP should own the updateStrategy bean which will then just always be sync or deferred so then the flag won't be needed at run time.

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
sdwvit: thanks for spotting

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelColumnComp.ts
sdwvit: params is too big to pass around imo. If method uses only 1 boolean then we should pass only 1 boolean

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateStrategy.ts
sdwvit: fixed

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
sdwvit: I'll add it in a separate pr, if that's ok

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
sdwvit: yes, they need to be separate, one is for answering which columns go into values, and other responds to which functions to use

StephenCooper (approved):

PR #13331, opened 2026-03-18
AG-16600 aggregation inverse

SalvatorePreviti:
## Summary

Built-in group row value distribution: editing a group row cell automatically distributes the new value to descendant rows based on the column's aggregation function. Adds `distributeGroupValue` as a public API and supports declarative options-object configuration on `colDef.groupRowValueSetter`.

## New Features

-   **Options-object `groupRowValueSetter`**: accepts `{ distribution, integerDistribution, min, max, getValue, setValue }` in addition to a callback function. Or a boolean value to enable default, or to disable it completely
-   **Auto-default**: when `groupRowEditable` is set with no `groupRowValueSetter`, the built-in distribution is used automatically.
-   **Strategies**: `uniform` (default for `sum`), `overwrite` (default for `avg` / `count` / custom), `percentage`, `increment`. Special handling for `first`/`last`/`min`/`max` aggFuncs.
-   **Per-aggFunc records**: `distribution: { sum: 'percentage', avg: { distribution: 'increment', min: 0 } }`.
-   integer rounding with remainder spreading, bigint support.
-   **`distributeGroupValue`** exported from `ag-grid-enterprise` for use in custom callbacks.
-   **`getAggregatedChildren(colKey, recursive?)`**: new optional `recursive` parameter returns all descendant leaf rows.
-   **Auto group column guard**: `groupRowEditable`/`groupRowValueSetter` no longer inherited from `defaultColDef` into auto group columns - we have another ticket to implement in the future to handle editing auto group col defs, that is more complex
- fix on clipboardService to handle paste to a group column that has rowGroupEditable 

## Refactoring

-   **`AggregatedChildrenSvc`** extracted from `AggregationStage` — separates query logic from aggregation execution. `getNodesFromMappedSet` moved to shared `aggDataUtils.ts`.
-   **`RowGroupingEditValueSvc`** — group row value setter resolution moved from `valueService.ts` (community) into enterprise service, removing ~30 lines from community code.
-   Type definitions reorganised: `GroupRowValueSetterParams` and related types moved to `colDef-groupRowValueSetter.ts`; base param types extracted to `colDef-base.ts` (re-exported for compatibility).
-   `ChangedPathModule` dependency moved from `SharedRowGroupingModule` to `GroupHierarchyModule`.
-   `columnStateUpdateExecutionStrategy.ts`: replaced `Object.entries` destructuring with `Object.keys` + lookup for correct TS narrowing.

## Documentation

-   Rewrote and expanded the Editing Group Rows page (strategies, options, per-aggFunc records, custom callbacks, pivot editing, tree data).
-   Cross-references added to aggregation pages. New examples: `distribute-group-value-sum`, `distribute-group-value-all-agg`, `pivot-editable-builtin`, `tree-data-editable`.
-   **`formatAST.js`, `documentation-helpers.ts`**: added support for @example and basic support for @link

## Tests

-   **182 distribution tests** (10 files): all strategies, bigint, numeric edge cases, hierarchy, per-aggFunc records, custom I/O, integer rounding.
-   **30 aggregatedChildren tests** (grouping + tree data, including recursive mode).
-   Fixed pre-existing `NumberEditorModule` warnings in `cell-editing-change-flash.test.ts` and `cell-editing-get-edit-row-values.test.ts`.
-   Test utils: `editCell`/`performEdit` refactored to handle jsdom cell DOM replacement; `serialiseValue` updated for bigint and non-finite numbers.

FIX AG-16600


packages/ag-grid-enterprise/src/rowGrouping/rowGroupingEditModule.ts
```diff
@@ -1,19 +1,72 @@
...
+
+        // Resolve groupRowValueSetter: explicit config, or default when groupRowEditable is set.
+        // colDef is already deep-merged with defaultColDef (via _mergeDeep in columnFactoryUtils),
+        // so object-type options inherit and merge with defaultColDef automatically.
+        const raw = colDef.groupRowValueSetter ?? (colDef.groupRowEditable != null ? distributeGroupValue : undefined);
+        if (raw == null) {
+            return undefined; // No groupRowValueSetter → caller uses normal path.
+        }
+
+        const params: GroupRowValueSetterParams = this.gos.addCommon({
```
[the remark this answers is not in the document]
SalvatorePreviti: it was a design decision but making it clearer

packages/ag-grid-enterprise/src/rowGrouping/rowGroupingEditModule.ts
```diff
@@ -1,19 +1,79 @@
...
+ * the module supplies a default that distributes the edited value to
+ * descendant rows using {@link distributeGroupValue}.
  */
 export const RowGroupingEditModule: _ModuleWithoutApi = {
     moduleName: 'RowGroupingEdit',
     version: VERSION,
-    dependsOn: [EnterpriseCoreModule, _EditCoreModule, SharedRowGroupingModule],
+    beans: [RowGroupingEditValueSvc],
+    dependsOn: [EnterpriseCoreModule, SharedRowGroupingModule],
 };
```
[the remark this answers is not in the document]
SalvatorePreviti: This is fine - it was not needed there - the library was never released with this module before

packages/ag-grid-community/src/entities/colDef-groupRowValueSetter.ts
```diff
@@ -0,0 +1,448 @@
...
+ *
+ * @example
+ * ```ts
+ * colDef.groupRowEditable = (params) => {
+ *     // Only allow editing on leaf-level groups (groups directly containing data rows)
+ *     return params.node.childrenAfterGroup?.some(child => !child.group) ?? false;
+ * };
+ * ```
+ */
+export interface GroupRowEditableCallbackParams<TData = any, TValue = any, TContext = any>
```
StephenCooper, line 18: Is this why the documentation helper has been updated. Does this now get rendered somewhere in the docs?

packages/ag-grid-community/src/entities/colDef-groupRowValueSetter.ts
```diff
@@ -0,0 +1,448 @@
...
+     * The immediate children that contribute to this group's aggregation.
+     *
+     * - **Leaf groups** (groups directly containing data rows): the data rows themselves.
+     * - **Non-leaf groups** (groups containing sub-groups): the child group rows.
+     *   Call `setDataValue()` on these child groups to cascade the edit recursively through
+     *   the full hierarchy. The built-in `distributeGroupValue` does this automatically.
+     * - **Pivot mode**: only rows matching the edited pivot column's keys are included.
+     *
+     * To retrieve the same children programmatically from any RowNode, use
+     * {@link IRowNode.getAggregatedChildren | rowNode.getAggregatedChildren(colKey)}.
```
StephenCooper: Does this result in a link that works on our doc site?

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/distributorBigInt.ts
```diff
@@ -0,0 +1,340 @@
...
+    GroupRowValueSetterFunc,
+    GroupRowValueSetterParams,
+    IRowNode,
+} from 'ag-grid-community';
+
+import type { DistributionStrategy } from './valueConversion';
+import { isNumericLike, resolveStrategy, toBigInt } from './valueConversion';
+
+/** Distributes a BigInt value to children using integer-safe arithmetic. */
+export class DistributorBigInt {
```
StephenCooper, line 15: Maybe we should try to create a BigIntModule that can be used to provide all the BigInt features. This is quite a big amount of code in the new RowGroupEditingModule if the application is not using BigInts. 

May have to align with a major version though.

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/distributorNumber.ts
```diff
@@ -0,0 +1,385 @@
...
+    GroupRowValueSetterFunc,
+    GroupRowValueSetterParams,
+    IRowNode,
+} from 'ag-grid-community';
+
+import type { DistributionStrategy } from './valueConversion';
+import { isIntegerColDef, isNumericLike, resolveStrategy, toNumber } from './valueConversion';
+
+/** Distributes a numeric value to children using the chosen strategy. */
+export class DistributorNumber {
```
StephenCooper, line 15: Is this and the bigInt version duplicating code that could be shared or has this purposefully been kept separate for cleaner code?

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/valueConversion.ts
```diff
@@ -0,0 +1,133 @@
...
+            return toNumber((raw as { value: unknown }).value);
+        }
+    }
+    return 0;
+};
+
+const numberToBigInt = (value: number): bigint => (Number.isFinite(value) ? BigInt(Math.round(value)) : 0n);
+
+/** Coerces an unknown value to a BigInt. Returns 0n for non-convertible inputs. */
+export const toBigInt = (raw: unknown): bigint => {
```
StephenCooper, line 66: Do we not already have code for this?

packages/ag-grid-community/src/entities/colDef-groupRowValueSetter.ts
SalvatorePreviti: Yes I updated the doc helper to render properly @link as code (is useful in vscode, less in docs, but we need to support it) and @example that i think can be helpful

packages/ag-grid-community/src/entities/colDef-groupRowValueSetter.ts
SalvatorePreviti: at the moment link just format it as code - it does not link anywhere right now - not sure if we can implement it really, however, is useful from vscode or other ides and before it was not rendering correctly in our docs.

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/distributorBigInt.ts
SalvatorePreviti: It is but we don't have a BigInt module :( not sure how we can do this without a breaking change for the bigint code we already have - there is more code than this around for bigint. I think we need to support it out of the box really without a module though - but the separation here makes the code cleaner

I don't think the BigIntModule is in our architecture that feasible though as is horizontal and touches many features and we would end up exporting just a huge service with mixed responsibilities for different parts with micro methods that would not save much

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/distributorNumber.ts
SalvatorePreviti: kept separate for cleaner code - code cannot be reused as for example adding +1 requires to add +1n or the code does not work! so better to keep it completely separated, and to reduce complexity of the other class that as to deal with floating point that is different

packages/ag-grid-community/src/entities/colDef-groupRowValueSetter.ts
SalvatorePreviti: It gets rendered like this: 
image

SalvatorePreviti: note - i am pushing some updates that simplify and improves the documentation and simplify the code and the API - a more generic "precision" instead of integer distribution, and removing (for now) min and max, the user can implement it with providing a custom setValue

documentation/ag-grid-docs/src/components/reference-documentation/utils/documentation-helpers.ts
```diff
@@ -236,6 +241,27 @@ export function formatJsDocString(docString: string) {
...
+            if (!descText && !code) {
+                return '';
+            }
+            let result = `\n<br>\n<strong>Example${descText ? `: ${descText}` : ''}</strong>`;
+            if (code) {
+                result += `\n<pre class="code"><code>${code}</code></pre>`;
+            }
+            return result;
+        })
         .replace(optionReg, '<li style="margin-left:1rem"> $1 </li>')
```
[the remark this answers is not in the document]
SalvatorePreviti: fixed

packages/ag-grid-enterprise/src/aggregation/aggregatedChildrenSvc.ts
```diff
@@ -0,0 +1,82 @@
...
+    result: RowNode[]
+): void => {
+    for (let i = 0, len = children.length; i < len; ++i) {
+        const child = children[i];
+        if (child.group) {
+            collectLeafDescendants(getImmediateAggChildren(child, col, gos), col, gos, result);
+        } else {
+            result.push(child);
+        }
+    }
```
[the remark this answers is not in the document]
SalvatorePreviti: not a problem! we use recursion everywhere in the grid for this

packages/ag-grid-enterprise/src/rowGrouping/distributeGroupValue/valueConversion.ts
SalvatorePreviti: Is buried in something else - there is similar code but with some differences and for different purposes - in value service and in formulas and in other places too - I did not want to refactor other code for this - we can do in another PR as this is already way too big - seems a refactor that we could do in another session - but - I would not consider high priority as this does not add much bundle pressure right now, the saving is modest, not enough

StephenCooper (approved):

PR #13340, opened 2026-03-19
[AG-3471] Support suppressSyncLayoutWithGrid in deferred apply mode

sdwvit:
- Enable column reordering in CTP when `suppressSyncLayoutWithGrid` is active with deferred mode
- Add `hasDeferredColumnOrder` to strategy interface to distinguish draft column order from live fallback
- Reset staged changes on external grid events (sort, visibility, row group, value, pivot, new columns loaded) with an exception for column moves when `suppressSyncLayoutWithGrid` is true

relates to [AG-3471](https://ag-grid.atlassian.net/browse/AG-3471)

### DIFF
--- a/packages/ag-grid-enterprise/src/columnToolPanel/agPrimaryColsList.ts
+++ b/packages/ag-grid-enterprise/src/columnToolPanel/agPrimaryColsList.ts
@@ -121,7 +121,7 @@ export class AgPrimaryColsList extends Component<AgPrimaryColsListEvent> {
 
         this.expandGroupsByDefault = !contractColumnSelection;
 
-        const isPreventMove = suppressColumnMove || suppressSyncLayoutWithGrid;
+        const isPreventMove = suppressColumnMove || (suppressSyncLayoutWithGrid && !isDeferredMode(params));
 
         const virtualList = this.createManagedBean(
             new VirtualList<ToolPanelColumnGroupComp | ToolPanelColumnComp, ColumnModelItem>({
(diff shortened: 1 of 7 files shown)


packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
```diff
@@ -130,6 +131,39 @@ export class ColumnToolPanel extends Component implements IColumnToolPanel, IToo
...
+                sortChanged: resetListener,
+                columnVisible: resetListener,
+                columnRowGroupChanged: resetListener,
+                columnValueChanged: resetListener,
+                columnPivotChanged: resetListener,
+                columnPivotModeChanged: resetListener,
+                newColumnsLoaded: resetListener,
+                ...(mergedParams.suppressSyncLayoutWithGrid ? {} : { columnMoved: resetListener }),
+            });
+            childDestroyFuncs.push(
```
sdwvit, line 137: not sure if needed, it came from codex review

StephenCooper: not needed. Managedevent listeners clean up when the bean calling it is destoryed.

StephenCooper (approved):

packages/ag-grid-enterprise/src/columnToolPanel/agPrimaryColsList.ts
```diff
@@ -121,7 +121,7 @@ export class AgPrimaryColsList extends Component<AgPrimaryColsListEvent> {
 
         this.expandGroupsByDefault = !contractColumnSelection;
 
-        const isPreventMove = suppressColumnMove || suppressSyncLayoutWithGrid;
+        const isPreventMove = suppressColumnMove || (suppressSyncLayoutWithGrid && !isDeferredMode(params));
```
StephenCooper, line 124: Oh I see the existing behaviour does stop moving for both properties. We may need to update this doc section 

https://ag-grid.com/javascript-data-grid/tool-panel-columns/#suppress-column-reordering

StephenCooper (changes requested):
Actually we need the destroy cleanup before merging.

packages/ag-grid-enterprise/src/columnToolPanel/columnToolPanel.ts
sdwvit: that is tracking beal lifecycle, but i do recreate child components manually in this file on refresh(). I think I need to keep it.

StephenCooper (approved):

PR #13350, opened 2026-03-20
AG-3471 TC 5

StephenCooper:
Relates to [AG-3471](https://ag-grid.atlassian.net/browse/AG-3471)


documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/deferred-apply-mode/main.ts
```diff
@@ -75,7 +74,6 @@ const gridOptions: GridOptions<IOlympicData> = {
     autoGroupColumnDef: {
         minWidth: 250,
     },
-    pivotMode: true,
```
StephenCooper, line 78: Don't start the example in Pivot Mode

sdwvit: why?

sdwvit:
there are more changes than needed here

sdwvit (approved):

PR #13372, opened 2026-03-23
Fix(columns): tool panel drag interaction shows pin icon instead of reorder indicator

sdwvit:
isMoveBlocked was reading pivot mode from the live grid state (colModel.isPivotMode()), so toggling pivot off in deferred mode without applying still blocked column moves and showed pin icon instead of reorder indicator.

Fixes [RTI-3259](https://ag-grid.atlassian.net/browse/RTI-3259)


sdwvit:
ignore codex, it's wrong

.run/Column Tool Panel Icon bug.run.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
```
sdwvit, line 1: intended

StephenCooper (approved):

PR #13381, opened 2026-03-23
[RTI-3259] Fix(columns): tool panel drag interaction shows pin icon instead of reorder indicator

sdwvit:
Two fixes for drag interaction in deferred columns tool panel:

1. isMoveBlocked was reading pivot mode from the live grid state, so toggling pivot off in deferred mode without applying still blocked column moves and showed pin icon instead of reorder indicator.

2. After fixing the icon, column reordering still didn't work because onColumnsChanged used buildTreeFromProvidedColumnDefs when pivot mode was active (live), ignoring the deferred column order. Now falls through to buildTreeFromWhatGridIsDisplaying when a deferred column order exists.

Fixes [RTI-3259](https://ag-grid.atlassian.net/browse/RTI-3259)


sdwvit:
codex is wrong, but i deleted the file anyway

StephenCooper (approved):

PR #13382, opened 2026-03-23
RTI-3270 Add note about Angular 21 zoneless default to NgZone docs

StephenCooper:
Fix [RTI-3270](https://ag-grid.atlassian.net/browse/RTI-3270)


sdwvit (approved):
Was asked about this another day, good change!

PR #13396, opened 2026-03-24
RTI-3282-more-docs-fixes-for-group-edit

SalvatorePreviti:
- improves the documentation
- improves the example by providing a case of a computed column with a valueGetter where the group cell is editable
- fixes a bug where the edit was blocked if a valueGetter + rowGroupValueSetter was defined but NO valueSetter is defined in isSetValueSupported

FIX RTI-3282

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
@@ -9,9 +9,9 @@ and refreshing the grouping hierarchy after edits.
 
 ## Editing Group Row Cells
 
-In the example below, double-click any group row's `Sales` cell to edit it. The new value is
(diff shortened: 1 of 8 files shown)


documentation/ag-grid-docs/src/content/docs/grouping-edit/index.mdoc
```diff
@@ -222,26 +228,81 @@ const gridOptions = {
...
+    },
+}
+```
+
 Aggregation functions not listed in the record fall through to the `default` fallback, then to
-the built-in defaults (see the strategy table above). Custom aggregation functions require an
-explicit entry or `default` to be editable.
+the built-in defaults (see the strategy table above).
+
+### The default Fallback
```
StephenCooper: Can we remove 'The' from headings.

StephenCooper (approved):

Issue #13176, opened 2026-02-19
[Bug]:  Filter input intermittently reverts typed characters (35.1.0 regression)


josgraha:
### Link to reproducible scenario

https://plnkr.co/edit/G6jbsxVRXLiG7cTS?preview

### Describe the bug

Steps to reproduce 
1.   When typing into an AG Grid column filter input, characters sometimes revert to an earlier value. Example: typing “GL” results in the input reverting to “G” or empty; the user must retype to make it
  stick. This does not happen in 35.0.1, and appears to be a regression in 35.1.0.

Expected behavior
  Filter input should retain all typed characters.

Actual behaviour

  Filter input intermittently reverts to earlier values while typing.

 Version

  - Repro: ag-grid-community / ag-grid-enterprise 35.1.0
  - Not repro: 35.0.1

More information

  Reproduction
  Plunker with minimal example (see below).

  index.html
```
  <!DOCTYPE html>
  <html lang="en">
    <head>
      <meta charset="utf-8" />
      <title>AG Grid Filter Input Repro</title>
      <style>
        html, body, #myGrid { height: 100%; margin: 0; }
      </style>

      <!-- AG Grid 35.1.0 -->
      <script src="https://unpkg.com/ag-grid-community@35.1.0/dist/ag-grid-community.min.noStyle.js"></script>
      <link rel="stylesheet" href="https://unpkg.com/ag-grid-community@35.1.0/styles/ag-grid.css" />
      <link rel="stylesheet" href="https://unpkg.com/ag-grid-community@35.1.0/styles/ag-theme-alpine.css" />
    </head>
    <body>
      <div id="myGrid" class="ag-theme-alpine"></div>
      <script src="script.js"></script>
    </body>
  </html>
```
  script.js
```
  const columnDefs = [
    { field: "colA", filter: "agTextColumnFilter" },
    { field: "colB", filter: "agTextColumnFilter" },
  ];

  const rowData = [
    { colA: "ALPHA", colB: "One" },
    { colA: "GLAM", colB: "Two" },
    { colA: "GLIDE", colB: "Three" },
    { colA: "GLOW", colB: "Four" },
    { colA: "GLOOM", colB: "Five" },
  ];

  const gridOptions = {
    columnDefs,
    rowData,
    defaultColDef: {
      filter: true,
      floatingFilter: true,
      flex: 1,
      sortable: true,
    },
  };

  const gridDiv = document.querySelector("#myGrid");
[...]

PR #13415, opened 2026-03-25
fix: floating filter input intermittently reverts typed characters (#13176)

josgraha:
## Summary

Addresses #13176 — floating filter text input can intermittently revert typed characters.

## Analysis

The reporter identified this as a 35.1.0 regression vs 35.0.1. After reviewing the diff between `release-35.0.1` and `release-35.1.0`, the floating filter text input code path (`textInputFloatingFilter.ts`, `simpleFloatingFilter.ts`, `floatingFilterTextInputService.ts`) is functionally unchanged between the two versions. The missing `silent=true` parameter exists in both.

The changes between versions in the filter area are: BigInt filter support, date preset ranges, `isTypeEditable` refactor (no behavioral change for text filters), `onModelAsStringChange` (enterprise set filter only), and CSS/theming performance improvements. None of these directly alter the text floating filter input path.

This suggests either the regression is caused by an indirect change (e.g. CSS relayout timing making the race more observable), or the issue is intermittent enough to have been present but unnoticed in 35.0.1.

Regardless, the fix is correct — `onModelUpdated` should pass `silent=true` to `setValue` to prevent re-triggering the debounced input listener during the model sync-back, consistent with the existing pattern used elsewhere in the same file.

## Root Cause

In `textInputFloatingFilter.ts`, the `onModelUpdated` method calls `inputSvc.setValue()` **without** the `silent=true` parameter. This allows the debounced input change listener to re-fire during the model sync-back loop:

1. User types characters → debounced `syncUpWithParentFilter` fires
2. Parent filter model updates → `onModelUpdated` called on floating filter
3. `setValue()` without `silent` triggers the input change listener again
4. Within the debounce window, this can reset the input to the stale model value

## Fix

Pass `silent=true` to `setValue` in `onModelUpdated`, consistent with the existing pattern used in `recreateFloatingFilterInputService` (line 99) and `syncUpWithParentFilter` (line 120) in the same file.

## Changes

- `packages/ag-grid-community/src/filter/floating/provided/textInputFloatingFilter.ts` — add `silent=true` to `setValue` call in `onModelUpdated`
- `testing/behavioural/src/filters/floating-filters.test.ts` — add behavioural tests for floating filter input retention during debounce cycle

## Validation

- `yarn nx build:types ag-grid-community` — pass
- `yarn nx lint ag-grid-community` — pass
- `yarn nx test ag-grid-community` — 1141 tests pass
- `./behave.sh "floating-filters"` — 8 tests pass (6 existing + 2 new)
- `yarn nx format` — clean

## CLA

I agree to the retention of intellectual property rights as described in the AG Grid CONTRIBUTING.md. All software rights for this contribution are granted to AG Grid Ltd.

### DIFF
--- a/packages/ag-grid-community/src/filter/floating/provided/textInputFloatingFilter.ts
+++ b/packages/ag-grid-community/src/filter/floating/provided/textInputFloatingFilter.ts
@@ -46,7 +46,7 @@ export abstract class TextInputFloatingFilter<
     protected onModelUpdated(model: M): void {
         this.setLastTypeFromModel(model);
         this.setEditable(this.canWeEditAfterModelFromParentFilter(model));
-        this.inputSvc.setValue(this.filterModelFormatter.getModelAsString(model));
+        this.inputSvc.setValue(this.filterModelFormatter.getModelAsString(model), true);
     }
 
     protected override setParams(params: TParams): void {
--- a/testing/behavioural/src/filters/floating-filters.test.ts
+++ b/testing/behavioural/src/filters/floating-filters.test.ts
@@ -1,4 +1,5 @@
-import { getByTestId } from '@testing-library/dom';
+import { getByTestId, waitFor } from '@testing-library/dom';
+import { userEvent } from '@testing-library/user-event';
 
 import {
     ClientSideRowModelModule,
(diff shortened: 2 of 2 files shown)


josgraha:
Closing this PR after deeper analysis. The `silent` parameter on `setValue` only controls whether the `fieldValueChanged` local event fires — it does not prevent the internal value update, and `fieldValueChanged` has no listeners in the floating filter code path. This means the change is effectively a no-op for text floating filters.

The reported regression between 35.0.1 and 35.1.0 likely has a different root cause outside the floating filter code path, which is functionally unchanged between these versions. Further investigation needed on the consumer side.

PR #13419, opened 2026-03-25
Improve git-conventions skill PR body and base branch detection

StephenCooper:
Two fixes to the git-conventions skill:

**PR body convention** — JIRA-linked PR bodies now require a `Fix #AG-XXXX` line (for downstream automations) and a brief change summary, rather than just a bare ticket link. Added a concrete example to make the expected format unambiguous.

**detect-base-branch.sh** — Fixed a bug where the script would incorrectly select a release branch (e.g. `b35.2.0`) when working on `latest`. The main loop was comparing consecutive release branch distances but not checking whether the candidate was actually closer than `origin/latest`, causing the wrong base to be chosen.


alantreadway (approved):
LGTM

PR #13420, opened 2026-03-25
[AG-16487] Fix(ssrm): regression calling api refreshserverside to load rows doesn t re sort and also doesn t clear the column header sort indicator

sdwvit:
## Summary

Fixes [AG-16487](https://ag-grid.atlassian.net/browse/AG-16487) — an SSRM regression from v33.0.0 where rows were not re-sorted client-side after `api.refreshServerSide({ purge: true })` with `serverSideEnableClientSideSort` enabled. Column sort indicators stayed on the header but rows rendered in server-returned order.

## Root cause

Two separate defects combined to produce the regression:

1. **Missing sort entry point after purge** — `clientSideSortRows()` was only invoked via `fireRefreshFinishedEvent`, which is only reached when `wasRefreshing` was true. After `refreshServerSide({ purge: true })` the cache is destroyed and recreated, so `nodesToRefresh` is empty on the subsequent load and the sort never runs.
2. **`isStoreFullyLoaded()` falsely reported nodes out-of-place** — The previous implementation iterated `nodeMap` in insertion order and compared each entry's `.index` against a running counter. After non-purge refreshes that move or restore nodes, insertion order no longer matches index order, so the predicate returned false even when every index `0..N-1` was present. This blocked downstream sort calls in `applyServerSideTransaction` and `refreshAfterSort` too.

## Changes

**`packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts`**

- `onLoadSuccess`: call `clientSideSortRows()` before `fireStoreUpdatedEvent` when `serverSideEnableClientSideSort` is on and the store is fully loaded, so the grid sees sorted data in a single update. Comment enumerates the other two sort entry points (`lazyStore.applyServerSideTransaction`, `lazyStore.refreshAfterSort`) and notes why a sort in `fireRefreshFinishedEvent` would double-sort.
- `isStoreFullyLoaded`: walk by index (`nodeMap.getBy('index', i)` for `i` in `[0, numberOfRows)`) instead of iterating in insertion order. Return type narrowed to explicit `boolean`; all early-return paths now return `false`.

**`testing/behavioural/src/sorting/ssrm-client-side-sorting.test.ts`** (new, 1066 lines, 24 tests)

Behavioural coverage for the fix, including:

- Purge and non-purge refresh (flat, tree, row-grouping) — mirrors QA TC1 (Olympic row grouping + aggregation) and TC2 (tree data "Refresh Everything" / "Purge Everything")
- Child-route purge and non-purge refresh
- Async datasource
- Paginated loading (`cacheBlockSize: 2`) at root and child stores, including child-route purge that reloads all blocks
- Grand-total row preservation across re-sort
- Multi-column sort
- Sort change between refreshes
- Transactions placing out-of-order rows into their sorted position
- Server returning rows in different order / with updated values
- Configurations with and without `getRowId`


packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
```diff
@@ -902,6 +902,17 @@ export class LazyCache extends BeanStub {
         if (wasRefreshing && finishedRefreshing) {
             this.fireRefreshFinishedEvent();
         }
+
+        // Apply client-side sort after purge reload (wasRefreshing is false for new caches)
```
gportela85: personally I'd go with
```
if (
  !this.gos.get('serverSideEnableClientSideSort') || 
  wasRefreshing ||
  !this.isStoreFullyLoaded() || 
  !this.clinetSideSortRows()
) {
  return;
}

this.fireStoreUpdatedEvent();
```

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
```diff
@@ -902,6 +902,17 @@ export class LazyCache extends BeanStub {
...
+        if (!wasRefreshing && this.isStoreFullyLoaded()) {
+            const isClientSideSortingEnabled = this.gos.get('serverSideEnableClientSideSort');
+            if (isClientSideSortingEnabled) {
+                const didSort = this.clientSideSortRows();
+                if (didSort) {
+                    this.fireStoreUpdatedEvent();
```
StephenCooper: We are already calling `fireStoreUpdatedEvent()` on line 898 above. So now we are calling this twice when this code path is hit. 

Does the fix work if moved above that line and so we don't need to check if the clientSideSortRows actually did anything?

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
sdwvit: updated

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
sdwvit: updated

SalvatorePreviti (approved):

PR #13429, opened 2026-03-26
AG-16674 Respect transaction rowCount for add/update, not just removals

sdwvit:
- Extend `ServerSideTransaction.rowCount` support to add and update transactions (previously only applied during removals)
- When `rowCount` is provided on an add-only or update-only transaction, the store's row count is now overridden via the existing `setRowCount` method
- Added behavioural tests for both add and update transactions honouring supplied`rowCount` 

Fix [AG-16674](https://ag-grid.atlassian.net/browse/AG-16674)

### DIFF
--- a/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
+++ b/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
@@ -1073,7 +1073,7 @@ export class LazyCache extends BeanStub {
     /**
      * Transaction Support here
      */
-    public updateRowNodes(updates: any[]): RowNode[] {
+    public updateRowNodes(updates: any[], newRowCount?: number): RowNode[] {
         const updatedNodes: RowNode[] = [];
         updates.forEach((data) => {
             const id = this.getRowId(data);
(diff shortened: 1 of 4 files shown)


packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
```diff
@@ -1083,10 +1083,17 @@ export class LazyCache extends BeanStub {
                 updatedNodes.push(lazyNode.node);
             }
         });
+
+        const isNewRowCountValid = newRowCount != null && newRowCount >= 0;
```
StephenCooper, line 1087: I would extract this into a method so it can be shared.

testing/behavioural/src/row-data/server-side-row-model-transactions.test.ts
```diff
@@ -77,6 +77,65 @@ describe('Server Side Row Model Transactions', () => {
...
+
+        await waitForEvent('firstDataRendered', api);
+        expect(api.getDisplayedRowCount()).toBe(100);
+
+        api.applyServerSideTransaction({
+            update: [{ id: 0, value: 'Updated Row 0' }],
+            rowCount: 200,
+        });
+
+        expect(api.getDisplayedRowCount()).toBe(200);
```
StephenCooper, line 135: Does this result in extra server side calls / loading rows? 

I know it is on the user to get this right but would be interested to know what happens if they provide a value too high and also too low.

PR #13434, opened 2026-03-27
AG-9654 Support reactive pivotComparator

StephenCooper:
Fix [AG-9654](https://ag-grid.atlassian.net/browse/AG-9654)

## Overview

- **Reactive `pivotComparator`**: `PivotStage` now includes `pivotComparator` in its change detection, so updating a column's `pivotComparator` via `applyColumnState` / `setColumnDefs` triggers a pivot rebuild without requiring a full grid refresh.
- **Named types for pivot options**: Introduced `PivotRowTotals` and `PivotColumnGroupTotals` string union types in `GridOptions` (exported from community `main.ts`), replacing ad-hoc `GridOptions['...']` indexed lookups.
- **Simplified grid options change detection**: The five separate `*LastTime` fields for pivot grid options are replaced with a single `gridOptionsHashLastTime` string, consistent with the existing hash-based change detection pattern used for aggregation columns, group columns, and pivot comparators.

### DIFF
--- a/packages/ag-grid-enterprise/src/pivot/pivotStage.ts
+++ b/packages/ag-grid-enterprise/src/pivot/pivotStage.ts
@@ -120,10 +120,10 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
         this.groupColumnsHashLastTime = groupColumnsHash;
 
         const pivotColumns = pivotColsSvc?.columns ?? [];
-        const anyComparators =
+        const shouldTrackPivotOrder =
             gos.get('enableStrictPivotColumnOrder') && pivotColumns.some((col) => col.getColDef().pivotComparator);
-        const pivotOrder = anyComparators ? computePivotOrder(this.uniqueValues, pivotColumns, 0) : [];
-        const pivotComparatorsChanged = !_areEqual(pivotOrder, this.pivotOrderLastTime);
+        const pivotOrder = shouldTrackPivotOrder ? computePivotOrder(this.uniqueValues, pivotColumns, 0) : [];
+        const pivotOrderChanged = !_areEqual(pivotOrder, this.pivotOrderLastTime);
         this.pivotOrderLastTime = pivotOrder;
 
         const anyGridOptionsChanged = this.refreshProps.some((p) => changedProps?.has(p));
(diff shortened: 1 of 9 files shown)


packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -126,32 +114,36 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
...
-
-        this.pivotRowTotalsLastTime = pivotRowTotals;
-        this.pivotColumnGroupTotalsLastTime = pivotColumnGroupTotals;
-        this.suppressExpandablePivotGroupsLastTime = suppressExpandablePivotGroups;
-        this.removePivotHeaderRowWhenSingleValueColumnLastTime = removePivotHeaderRowWhenSingleValueColumn;
+        const pivotComparatorsHash = (pivotColsSvc?.columns ?? [])
+            .map((column) => `${column.getId()}-${column.getColDef().pivotComparator?.toString() ?? ''}`)
+            .join('#');
+        const pivotComparatorsChanged = pivotComparatorsHash !== this.pivotComparatorsHashLastTime;
+        this.pivotComparatorsHashLastTime = pivotComparatorsHash;
```
StephenCooper: This is the main part of the fix. To now consider pivotCompartor as part of the need to refresh set

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -281,3 +260,27 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
...
+function computePivotOrder(values: Map<string, any>, pivotColumns: AgColumn[], depth: number): string[] {
+    const comparator = pivotColumns[depth]?.getColDef().pivotComparator;
+    const keys = [...values.keys()];
+    if (comparator) {
+        keys.sort(comparator);
+    }
+    if (depth === pivotColumns.length - 1) {
+        return keys;
+    }
+    return keys.flatMap((key) => {
```
SalvatorePreviti: this is a bit hard to read and is definitely slow - isn't there a simpler option to do this with a simpler loop?

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
SalvatorePreviti: the only risk is if there is some user that is always passing lazily a () => callback that would force a refresh every time with react without using useCallback?

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
StephenCooper: update this method.

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
StephenCooper: No longer using function hashes and instead evaluating their result to know if we need to rebuild.

PR #13442, opened 2026-03-30
Detect redirects that shadow existing pages

BernieSumption:
Add a build-time check in redirectsChecker that errors when a redirect's source path corresponds to an existing page in the build output.

Also removes the theming-api redirect that was shadowing the Theme Parameters API docs page


taktran (approved):
LGTM

StephenCooper (approved):

PR #13444, opened 2026-03-31
AG-17003 Optimise validation checks to avoid repeated work

StephenCooper:
## Summary

- Cache property name checks (deprecations, fuzzy string matching, row model support) per `objectName` so they run once per unique property name, not on every `processOptions` call. This avoids repeated Levenshtein distance calculations for gridOptions updates and grids with many identically-shaped colDefs.
- Pre-compute a `Set` of all valid property names (`allValidNames`) in each validator definition for O(1) lookups instead of rebuilding arrays and doing linear scans each time.
- Inline and remove `checkProperties` / `_fuzzyCheckStrings` helpers — the single cached loop now handles deprecation warnings, property name validation, and row model checks in one pass.

Fix [AG-17003](https://ag-grid.atlassian.net/browse/AG-17003)

### DIFF
--- a/packages/ag-grid-community/src/validation/validationTypes.ts
+++ b/packages/ag-grid-community/src/validation/validationTypes.ts
@@ -3,10 +3,28 @@ import type { GridOptions } from '../entities/gridOptions';
 import type { ValidationModuleName } from '../interfaces/iModule';
 import type { RowModelType } from '../interfaces/iRowModel';
 
+// Vue adds these properties to all objects, so we ignore them when checking for invalid properties
+const VUE_FRAMEWORK_PROPS = ['__ob__', '__v_skip', '__metadata__'];
+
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/validation/validationTypes.ts
```diff
@@ -3,10 +3,28 @@ import type { GridOptions } from '../entities/gridOptions';
 import type { ValidationModuleName } from '../interfaces/iModule';
 import type { RowModelType } from '../interfaces/iRowModel';
 
+// Vue adds these properties to all objects, so we ignore them when checking for invalid properties
+const VUE_FRAMEWORK_PROPS = ['__ob__', '__v_skip', '__metadata__'];
```
sdwvit, line 7: a sprinkle of prototype pollution 🪄

sdwvit (approved):

PR #13447, opened 2026-03-31
AG-8439 ssrm grand total

SalvatorePreviti:
- add support of grand total for flat and hierarchical grids in SSRM
- new needsGrandTotal in the gridRows request that the grid sends if grand total is enabled and there is not yet a cached grand total row data
- new "rowGroupFooter_ROOT_NODE_ID" and exported ROW_ID_GRAND_TOTAL from main that can be used to update the row with transactions (or remove it).
- new documentation, tests and 2 examples

Refactoring and optimization:

- make _createRowNodeFooter return the sibling so we don't need to access again node.sibling
- replace O(n^2) array with find index and splice for remove transaction with a O(n) Set

FIX AG-8439


packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -27,6 +27,14 @@ export const ROW_ID_PREFIX_TOP_PINNED = 't-';
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
 export const ROW_ID_PREFIX_BOTTOM_PINNED = 'b-';
 
+/**
+ * The row ID for the grand total footer row, "rowGroupFooter_ROOT_NODE_ID".
```
sdwvit: says footer here, but it can be placed anywhere? does it follow the correct naming convention ?

packages/ag-grid-community/src/entities/rowNode.ts
```diff
@@ -27,6 +27,14 @@ export const ROW_ID_PREFIX_TOP_PINNED = 't-';
...
+ * Used in both CSRM and SSRM to identify the grand total row.
+ * In SSRM, the server should include a row with this ID (as returned by `getRowId`)
+ * when `needsGrandTotal` is `true` in the `getRows` params.
+ */
+export const ROW_ID_GRAND_TOTAL = 'rowGroupFooter_ROOT_NODE_ID';
```
sdwvit, line 36: did you want to add `AG` here?

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: ouch - well yes internally we always called them footers - maybe i can try to make the documentation better though as we don't need to call them always footers

packages/ag-grid-community/src/entities/rowNode.ts
SalvatorePreviti: no - this must be this name as this is the hardcoded group footer also for CSRM - better to use the exact same ID - I think this is safe enough

packages/ag-grid-community/src/interfaces/iServerSideDatasource.ts
```diff
@@ -46,6 +46,17 @@ export interface IServerSideGetRowsParams<TData = any, TContext = any> extends A
...
+    /**
+     * A hint indicating the grid does not yet have grand total data cached.
+     * When `true`, the server should include grand total data in the response via `grandTotalRowData`
+     * or as a row in `rowData` whose `getRowId` returns `ROW_ID_GRAND_TOTAL`.
+     *
+     * This is only a hint — the server may always provide updated grand total data regardless of
+     * this flag, and the grid will accept it. Providing grand total data when this is `false`
+     * will update the existing grand total row.
+     */
+    needsGrandTotal: boolean;
```
sdwvit, line 58: should it be optional?

packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
```diff
@@ -135,10 +135,8 @@ export class BlockUtils extends BeanStub implements NamedBean {
         const getGroupIncludeFooter = _getGroupTotalRowCallback(this.beans.gos);
         const doesRowShowFooter = getGroupIncludeFooter({ node: rowNode });
         if (doesRowShowFooter) {
-            _createRowNodeFooter(rowNode, this.beans);
-            if (rowNode.sibling) {
-                rowNode.sibling.uiLevel = rowNode.uiLevel + 1;
-            }
+            const footerNode = _createRowNodeFooter(rowNode, this.beans);
```
sdwvit, line 138: should the method accept props overrides like uiLevel?

packages/ag-grid-community/src/interfaces/iServerSideDatasource.ts
SalvatorePreviti: No this is the input that arrives to the user - so we always set it true or false - you think this might be a problem?

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
```diff
@@ -91,14 +91,47 @@ Below shows `isServerSideGroupOpenByDefault()` and `getRoute` in action. Note th
...
+To display a grand total row, set the `grandTotalRow` property to `'top'`, `'bottom'`, `'pinnedTop'`, or `'pinnedBottom'`. This works on flat grids without any row grouping.
+
+### Providing Grand Total Data
+
+When `grandTotalRow` is set, the `needsGrandTotal` hint on `getRows` params will be `true` for root-level requests that don't yet have cached grand total data. The server may always provide updated grand total data regardless of this hint.
+
+The grand total row data can be provided in two ways:
+
+-   **`grandTotalRowData` field** (recommended): Set the `grandTotalRowData` field on the `success` callback params. `getRowId` must return `ROW_ID_GRAND_TOTAL` (value: `'rowGroupFooter_ROOT_NODE_ID'`, exported from `ag-grid-community`) for this data. Set it to `null` to explicitly remove an existing grand total row.
+-   **In `rowData` array**: Include a row whose `getRowId` returns `ROW_ID_GRAND_TOTAL` in the `rowData` array. The grid will extract it automatically. This approach requires `getRowId` to be configured.
```
StephenCooper: I am not sure we need this second approach? Would be good to understand why this would be better for the developer.

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
```diff
@@ -18,6 +18,8 @@ export class ManualPinnedRowModel extends BeanStub implements IPinnedRowModel {
     private bottom: PinnedRows;
     /** Cached CSRM reference, null if not using client-side row model */
     private csrm: IClientSideRowModel | null = null;
+    /** True if using server-side row model */
+    private ssrm: boolean = false;
```
sdwvit, line 22: there is a method to check it, do we really need this cache?

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
```diff
@@ -29,8 +31,9 @@ export class ManualPinnedRowModel extends BeanStub implements IPinnedRowModel {
         this.top = new PinnedRows(beans, 'top');
         this.bottom = new PinnedRows(beans, 'bottom');
 
-        // Cache CSRM reference if using client-side row model
+        // Cache row model references
         this.csrm = _getClientSideRowModel(beans) ?? null;
+        this.ssrm = beans.rowModel.getType() === 'serverSide';
```
sdwvit, line 36:
```suggestion
        this.ssrm = _getServerSideRowModel(beans) ?? null;
```
but I am not sure if we need csrm and ssrm cache here

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
```diff
@@ -314,12 +317,12 @@ export class ManualPinnedRowModel extends BeanStub implements IPinnedRowModel {
...
     private pinGrandTotalRow() {
-        const { csrm, beans, _grandTotalPinned: float } = this;
-        if (!csrm) {
-            return;
-        }
+        const { beans, _grandTotalPinned: float } = this;
+
+        // Grand total node is the root node's sibling in both CSRM and SSRM.
+        // Other row models (Infinite, Viewport) have a rootNode but never set sibling.
+        const sibling = this.csrm || this.ssrm ? beans.rowModel.rootNode?.sibling : undefined;
```
sdwvit, line 324:
```suggestion
        const sibling = beans.rowModel.rootNode?.sibling;
```
should be enough?

packages/ag-grid-enterprise/src/rowHierarchy/flattenStage.ts
```diff
@@ -129,9 +129,10 @@ export class FlattenStage extends BeanStub implements _IRowNodeFlattenStage, Nam
 
                     // if the parent was excluded, then ui level is that of the parent
                     const uiLevelForChildren = excludedParent ? uiLevel : uiLevel + 1;
+                    let footerNode: RowNode | undefined;
                     if (doesRowShowFooter === 'top') {
```
sdwvit, line 133: can we put 'top' / 'bottom' branches next to each other? for readability

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyBlockLoadingService.ts
```diff
@@ -162,11 +168,16 @@ export class LazyBlockLoadingService extends BeanStub implements NamedBean {
             removeNodesFromLoadingMap();
         };
 
+        const isRootStore = parentNode.level === -1;
```
sdwvit, line 171: is this enough condition?

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyCache.ts
```diff
@@ -1036,6 +1074,29 @@ export class LazyCache extends BeanStub {
         });
     }
 
+    /**
+     * Creates or updates the grand total row node from server response data.
+     * Uses _createRowNodeFooter with the same sibling pattern as CSRM.
+     */
+    public createOrUpdateGrandTotalNode(data: any): void {
```
sdwvit, line 1081: lets avoid any pls

sdwvit (approved):

packages/ag-grid-community/src/interfaces/iServerSideDatasource.ts
sdwvit: yes if user-facing interface, as it is a breaking change then

packages/ag-grid-community/src/entities/rowNode.ts
sdwvit: sure, if default is bottom placement

SalvatorePreviti: improved the comment

packages/ag-grid-enterprise/src/serverSideRowModel/blocks/blockUtils.ts
SalvatorePreviti: currently I'd say no as it would make the method slower for the other cases where is not needed

documentation/ag-grid-docs/src/content/docs/server-side-model-grouping/index.mdoc
SalvatorePreviti: I will change a bit the documentation giving less emphasis of the use of rows, is more incidental and to be fully compatible with transactions approach - the row should be processed for full consistency also when returned in the rows as we use that ID for transaction but we should advice better the user to use the returned field

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
SalvatorePreviti: the method is a bit slow - I would say better as we already do it for csrm

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
SalvatorePreviti: we have it in many places as those helper functions are not the fastest, in the future we can do better and I have a plan to do that but is low priority - future refactoring allowing to do something like "this.beans.rowModel.ssrm" "this.beans.rowModel.csrm" removing the need of checking options and those utility functions

packages/ag-grid-community/src/pinnedRowModel/manualPinnedRowModel.ts
SalvatorePreviti: no! also the other models have a rootNode - it is a dummy rootNode but it exists - we need to check if we are csrm or ssrm

packages/ag-grid-enterprise/src/rowHierarchy/flattenStage.ts
SalvatorePreviti: no the order of instructions here is important - the top must be inserted before the middle and the bottom after it

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyBlockLoadingService.ts
SalvatorePreviti: yes it should be - the only node that has level -1 is the root node and we already rely on this in other part of the code, so it is not a new pattern and there is not commonly a different way for doing this

PR #13454, opened 2026-04-01
Detect redirects that shadow existing pages (#13442)

markdurrant:
Add a build-time check in redirectsChecker that errors when a redirect's source path corresponds to an existing page in the build output. This prevents redirects from silently making real pages unreachable.

Also removes the theming-api redirect that was shadowing the Theme Parameters Reference page.


taktran (approved):
LGTM

PR #13474, opened 2026-04-06
fix: use RTL-aware scroll helpers in getHScrollPosition and header se…

liad242:
## Problem

When `enableRtl` is set to `true`, horizontal scrolling can cause the header and body viewports to desynchronize — the header scrolls but the body cells don't follow (or vice versa).

This happens because two places in the scroll sync chain read/write `scrollLeft` directly without using the RTL-aware helpers (`getScrollLeft` / `setScrollLeft`), which handle cross-browser differences in how `scrollLeft` behaves under RTL:
- Chrome/Safari use negative `scrollLeft` values in RTL
- Firefox uses positive (inverted) values
- The helpers normalize these to a consistent positive value

## Root Cause

**1. `RowContainerCtrl.getHScrollPosition()`** reads raw `this.eViewport.scrollLeft` instead of `_getScrollLeft(this.eViewport, this.enableRtl)`.

This returns browser-specific values to callers (e.g. `getHorizontalPixelRange()` API, drag-and-drop positioning) that expect normalized values.

**2. `HeaderRowContainerComp.setViewportScrollLeft()`** writes directly to `this.getGui().scrollLeft = left` instead of using `_setScrollLeft(this.getGui(), left, enableRtl)`.

The header receives a normalized scroll offset from the body's scroll handler, but writes it as a raw `scrollLeft` value. In Chrome RTL, the browser expects a negative value, but receives a positive one — causing the header to scroll to the wrong position.

## Fix

- `rowContainerCtrl.ts`: `getHScrollPosition()` now uses `_getScrollLeft(this.eViewport, this.enableRtl)`, consistent with `getCenterViewportScrollLeft()` and `getViewportScrollLeft()` in the same class.
- `headerRowContainerComp.ts`: `setViewportScrollLeft` now uses `_setScrollLeft(this.getGui(), left, enableRtl)`, consistent with `setCenterViewportScrollLeft()` in `RowContainerCtrl`.

Both fixes use the same RTL-aware helpers that are already correctly used everywhere else in the scroll sync chain.

## How to reproduce

1. Set `enableRtl: true` on the grid
2. Add enough columns so horizontal scrolling is required
3. Scroll horizontally — observe header and body cells become misaligned

Particularly reproducible in Chrome, where `scrollLeft` is negative in RTL mode.

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerCtrl.ts
@@ -489,8 +489,8 @@ export class RowContainerCtrl extends BeanStub implements ScrollPartner {
 
     public getHScrollPosition(): { left: number; right: number } {
         const res = {
-            left: this.eViewport.scrollLeft,
-            right: this.eViewport.scrollLeft + this.eViewport.offsetWidth,
+            left: _getScrollLeft(this.eViewport, this.enableRtl),
+            right: _getScrollLeft(this.eViewport, this.enableRtl) + this.eViewport.offsetWidth,
         };
         return res;
     }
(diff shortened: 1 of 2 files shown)


StephenCooper:
We are not able to reproduce this based on this information alone so I will be closing this PR.

Instead could you please open an issue with a reproduction that shows this issue. i.e fork one of our demos and update them from the [RTL](https://ag-grid.com/javascript-data-grid/rtl/) docs page. 

Maybe also be useful to attach a screen recording of what you are seeing.

PR #13492, opened 2026-04-09
[AG-12557] Feature(toolbar): add configurable quick access

sdwvit:
Adds a new enterprise `toolbar` grid option that renders a configurable quick-access bar above the grid. The toolbar supports built-in action items, custom user components, left/right alignment, display modes, keyboard navigation, RTL, and runtime reconfiguration.

a test build is in the jira comments

Fix [AG-17122](https://ag-grid.atlassian.net/browse/AG-17122)
Fix [AG-12557](https://ag-grid.atlassian.net/browse/AG-12557)


community-modules/locale/src/ar-EG.ts
```diff
@@ -334,6 +334,18 @@ export const AG_GRID_LOCALE_EG = {
     sortAbsoluteDescending: 'ترتيب تنازلي مطلق',
     sortUnSort: 'إلغاء الترتيب',
 
+    // Toolbar
+    toolbarAutoSizeAll: 'Auto Size All',
```
StephenCooper: I think we should re-use the existing locales instead of creating new toolbar specific duplicates. 

The grid already has locales for most of these and I see no reason why we should not be consistent and use those.

community-modules/styles/src/internal/base/parts/_toolbar.scss
```diff
@@ -0,0 +1,161 @@
+@use 'ag';
+
+@mixin output {
+    .ag-toolbar {
```
StephenCooper, line 4: Validated that legacy theming is working correctly?

packages/ag-grid-community/src/interfaces/iToolbar.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type ToolbarDisplay = 'icon' | 'iconAndLabel';
+
+export type Toolbar = {
+    display?: ToolbarDisplay;
+    items: (ToolbarItemDef | string)[];
+};
+
+export interface ToolbarItemDef {
+    component?: any;
```
StephenCooper: Based on StatusPanelDef and SideBarDef the component and params names should match. So instead of component it should be toolbarItem.

packages/ag-grid-enterprise/src/toolbar/providedItems/abstractToolbarItemComp.ts
```diff
@@ -0,0 +1,62 @@
...
+        const label = this.getLocaleTextFunc()(this.getLocaleKey(), this.getDefaultLabel());
+        this.eLabel.textContent = label;
+        this.getGui().setAttribute('aria-label', label);
+        this.getGui().setAttribute('title', label);
+
+        this.addManagedElementListeners(this.getGui(), { click: () => this.onAction() });
+    }
+
+    public init(params: IToolbarItemParams): void {
+        this.updateDisplay(params.display);
```
StephenCooper: Would suggest inlining the display and disabled methods into refresh and have init call that. Should save on bundle size.

packages/ag-grid-enterprise/src/toolbar/providedItems/columnsPanelToolbarItem.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    protected getDefaultLabel(): string {
+        return 'Columns';
+    }
+
+    public override init(params: IToolbarItemParams): void {
+        super.init(params);
+        if (!hasSideBarPanel(this.gos, this.panelId)) {
+            _warnOnce(
```
StephenCooper: Validation text should live in errorText and just use a code in the app code.

packages/ag-grid-enterprise/src/toolbar/providedItems/columnsPanelToolbarItem.ts
```diff
@@ -0,0 +1,39 @@
...
+                `toolbar item 'columnsPanel' requires a sidebar with the 'columns' tool panel configured. The item will not be rendered.`
+            );
+            this.setDisplayed(false);
+        }
+    }
+
+    protected onAction(): void {
+        if (this.beans.gridApi.getOpenedToolPanel() === this.panelId) {
```
StephenCooper: Reduce duplication of this.beans.gridApi.

packages/ag-grid-enterprise/src/toolbar/providedItems/exportToolbarItem.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.showExportMenu(menuItems);
+    }
+
+    private getExportMenuItems(): MenuItemDef[] {
+        const { gos, beans } = this;
+        const localeTextFunc = this.getLocaleTextFunc();
+        const items: MenuItemDef[] = [];
+
+        if (!gos.get('suppressCsvExport') && beans.csvCreator) {
+            items.push({
```
StephenCooper: We should find a way to share this with the menuItemMapper

packages/ag-grid-enterprise/src/toolbar/providedItems/searchIcon.ts
```diff
@@ -0,0 +1,8 @@
+export function createSearchIcon(): HTMLSpanElement {
+    const eIcon = document.createElement('span');
+    eIcon.className = 'ag-toolbar-input-icon';
+    eIcon.setAttribute('aria-hidden', 'true');
+    eIcon.innerHTML =
```
StephenCooper: This should be a lint warning as we have actively tried to avoid using innerHTML.

StephenCooper: Does this not need to follow the normal icon pattern?

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,428 @@
...
+    _focusNextGridCoreContainer,
+    _getActiveDomElement,
+    _removeFromParent,
+    _skipFocusableContainerListenerForAgGrid,
+} from 'ag-grid-community';
+
+import agToolbarCSS from './agToolbar.css';
+import type { ToolbarService } from './toolbarService';
+
+const BUILT_IN_ITEMS: Record<string, string> = {
```
StephenCooper: Can we not use our type for these built in items?

packages/ag-grid-enterprise/src/toolbar/providedItems/quickFilterToolbarItem.ts
```diff
@@ -0,0 +1,47 @@
...
+export class QuickFilterToolbarItem extends Component implements IToolbarItemComp {
+    private eInput!: HTMLInputElement;
+
+    constructor() {
+        super({ tag: 'div', cls: 'ag-toolbar-item ag-toolbar-input' });
+    }
+
+    public init(params: IToolbarItemParams): void {
+        const localeTextFunc = this.getLocaleTextFunc();
+        const label = localeTextFunc('toolbarQuickFilter', 'Quick Filter');
```
StephenCooper: Quick Filter does not seem like the right end user thing to see.

packages/ag-grid-enterprise/src/toolbar/providedItems/quickFilterToolbarItem.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        this.getGui().appendChild(createSearchIcon());
+
+        this.eInput = document.createElement('input');
```
gportela85: `_createElement`

packages/ag-grid-enterprise/src/toolbar/providedItems/filtersPanelToolbarItem.ts
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            _warnOnce(
+                `toolbar item 'filtersPanel' requires a sidebar with a filters tool panel configured. The item will not be rendered.`
+            );
+            this.setDisplayed(false);
+        }
+    }
+
+    protected onAction(): void {
+        if (this.beans.gridApi.getOpenedToolPanel() === this.panelId) {
```
gportela85: const { gridApi } = this.beans;

packages/ag-grid-enterprise/src/toolbar/providedItems/findToolbarItem.ts
```diff
@@ -0,0 +1,47 @@
...
+        super({ tag: 'div', cls: 'ag-toolbar-item ag-toolbar-input' });
+    }
+
+    public init(params: IToolbarItemParams): void {
+        const localeTextFunc = this.getLocaleTextFunc();
+        const label = localeTextFunc('toolbarFind', 'Find');
+
+        this.getGui().appendChild(createSearchIcon());
+
+        this.eInput = document.createElement('input');
```
gportela85: _createElement()

packages/ag-grid-enterprise/src/toolbar/providedItems/resetColumnsToolbarItem.ts
```diff
@@ -0,0 +1,22 @@
...
+    protected getIconName(): IconName {
+        return 'minimize';
+    }
+
+    protected getLocaleKey(): string {
+        return 'toolbarResetColumns';
+    }
+
+    protected getDefaultLabel(): string {
+        return 'Reset Columns';
```
gportela85: No translate?

packages/ag-grid-enterprise/src/toolbar/providedItems/exportToolbarItem.ts
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        return items;
+    }
+
+    private showExportMenu(menuItems: MenuItemDef[]): void {
+        const eGui = this.getGui();
+        const popupSvc = this.beans.popupSvc!;
+
+        const eMenu = document.createElement('div');
```
gportela85: _createElement

packages/ag-grid-enterprise/src/toolbar/providedItems/filtersPanelToolbarItem.ts
```diff
@@ -0,0 +1,44 @@
...
+    protected getIconName(): IconName {
+        return 'filtersToolPanel';
+    }
+
+    protected getLocaleKey(): string {
+        return 'toolbarFiltersPanel';
+    }
+
+    protected getDefaultLabel(): string {
+        return 'Filters';
```
gportela85: Translate

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,428 @@
...
+        );
+
+        _addFocusableContainerListener(this.beans, this, eGui);
+
+        this.addManagedElementListeners(eGui, {
+            focusin: (e: FocusEvent) => {
+                const target = e.target as HTMLElement;
+                if (target.matches('.ag-toolbar-button, .ag-toolbar-input-field')) {
+                    eGui.querySelectorAll<HTMLElement>('.ag-toolbar-button, .ag-toolbar-input-field').forEach((el) =>
+                        el.setAttribute('tabindex', '-1')
```
gportela85: what is this code doing? Why is this handling necessary? It should also respect the gridOptions.tabIndex when not using -1.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,428 @@
...
+                separator.setAttribute('role', 'separator');
+                eContainer.appendChild(separator);
+                continue;
+            }
+
+            const key = itemConfig.key || itemConfig.component;
+            const existingItem = existingItemsToReuse.get(key);
+            let promise: AgPromise<IToolbarItemComp>;
+            if (existingItem) {
+                promise = AgPromise.resolve(existingItem);
```
gportela85: In general we are trying to avoid new uses of AgPromise. Can this be replaced with a standard Promise or is it necessary because of our internals relying on `resolveNow`?

packages/ag-grid-enterprise/src/toolbar/providedItems/resetColumnsToolbarItem.ts
sdwvit: its a fallback, translate key above should cover it

packages/ag-grid-enterprise/src/toolbar/providedItems/filtersPanelToolbarItem.ts
sdwvit: same response

packages/ag-grid-enterprise/src/toolbar/providedItems/searchIcon.ts
sdwvit: yes, fixed, ty!

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,395 @@
...
+        return toolbar.items
+            .map((item) => normaliseItem(item, nextKey))
+            .filter((item) => {
+                const key = item.key ?? item.toolbarItem;
+                if (item.toolbarItem === 'separator') {
+                    return true;
+                }
+                if (seen.has(key)) {
+                    return false;
+                }
```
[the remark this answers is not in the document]
sdwvit: that's cool, up for a discussion

documentation/ag-grid-docs/src/content/docs/csv-export/index.mdoc
```diff
@@ -149,3 +149,7 @@ Note the following:
 ### CsvCellData
 
 {% interfaceDocumentation interfaceName="CsvCellData" overrideSrc="csv-export/csv.json" /%}
+
+{% note %}
+CSV Export can also be triggered from the [Quick Access Toolbar](./toolbar/) using the `csvExport` built-in toolbar item.
```
StephenCooper: Instead of this note at the bottom of the page I think we should either add to the existing heading of the page, because that is now missing this use case. 

Alternatively we add a short section at the start of the page, along the lines of How to Export which could then include links to context menu and quick access toolbar.

documentation/ag-grid-docs/src/content/docs/excel-export/index.mdoc
```diff
@@ -70,3 +70,7 @@ The sections below cover the various ways the exported Excel spreadsheets can be
 -   **[Master Detail](./excel-export-master-detail/)**: Export master detail grids to single or multiple sheets.
 -   **[Page Setup](./excel-export-page-setup/)**: Customize the page setup of the exported spreadsheets.
 -   **[Data Protection](./excel-export-data-protection/)**: Protect worksheets and unlock specific cells.
+
+{% note %}
+Excel Export can also be triggered from the [Quick Access Toolbar](./toolbar/) using the `excelExport` built-in toolbar item.
```
StephenCooper: Same as CSV above. I see this page does have links in the heading so could take that approach for this page and the csv page.

documentation/ag-grid-docs/src/content/docs/filter-quick/index.mdoc
```diff
@@ -156,3 +156,7 @@ The Quick Filter is only supported with the [Client-Side Row Model](./row-models
 ### Grid API Methods
 
 {% apiDocumentation source="grid-api/api.json" section="filter" names=["isQuickFilterPresent", "getQuickFilter"] /%}
+
+{% note %}
+A Quick Filter input can be embedded directly in the grid using the [Quick Access Toolbar](./toolbar/) with the `quickFilter` built-in toolbar item.
```
StephenCooper: I would remove this note as we will be updating this page as part of the linked work item.

documentation/ag-grid-docs/src/content/docs/find/index.mdoc
```diff
@@ -162,3 +162,7 @@ When using [Row Pagination](./row-pagination/), searching will be performed acro
 See the [Customising Find](#customising-find) section for an example of using Find with Row Grouping, as well as how to customise behaviour for Pagination.
 
 If data is mutated outside of the grid (e.g. not via grid options or API methods), Find will not re-run automatically. This would apply to situations where `api.refreshCells()` or `api.redrawRows()` are being used. To get Find to update, `api.findRefresh()` should be called after either of these API methods.
+
+{% note %}
+A Find input can be embedded directly in the grid using the [Quick Access Toolbar](./toolbar/) with the `find` built-in toolbar item.
```
StephenCooper: Same delete this note as the page will be updated in linked item.

packages/ag-grid-enterprise/src/agStack/agMenuItemRenderer.ts
```diff
@@ -92,15 +92,20 @@ export class AgMenuItemRenderer<
...
+            iconWrapper.appendChild(iconSvc.createIconNoSpan('check')!);
         } else if (icon) {
             if (_isNodeOrElement(icon)) {
                 iconWrapper.appendChild(icon);
             } else if (typeof icon === 'string') {
-                // eslint-disable-next-line no-restricted-properties -- no other way to parse custom HTML strings from the user
-                iconWrapper.innerHTML = icon;
+                let iconEl;
+                if (icon.trimStart().charAt(0) !== '<' && (iconEl = iconSvc.createIconNoSpan(icon))) {
+                    iconWrapper.appendChild(iconEl);
```
StephenCooper: Potential breaking change, if not needed as a result of MenuItem changes then we could avoid this.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,395 @@
...
+
+    private handleKeyDown(e: KeyboardEvent): void {
+        const activeEl = _getActiveDomElement(this.beans) as HTMLElement;
+        const eGui = this.getGui();
+        if (!eGui.contains(activeEl)) {
+            return;
+        }
+
+        // Don't intercept navigation keys in text inputs — allow normal caret navigation
+        if (activeEl instanceof HTMLInputElement && (activeEl.type === 'text' || activeEl.type === 'search')) {
```
StephenCooper: Check if required.

packages/ag-grid-community/src/validation/rules/userCompValidations.ts
```diff
@@ -57,4 +57,16 @@ export const USER_COMP_MODULES: Record<UserComponentName, ValidationModuleName>
     agFilteredRowCountComponent: 'StatusBar',
     agTotalAndFilteredRowCountComponent: 'StatusBar',
     agFindCellRenderer: 'Find',
+    agAutoSizeAllToolbarItem: 'Toolbar',
+    agColumnChooserToolbarItem: 'Toolbar',
+    agColumnsPanelToolbarItem: 'Toolbar',
+    agCsvExportToolbarItem: 'Toolbar',
+    agExcelExportToolbarItem: 'Toolbar',
+    agFiltersPanelToolbarItem: 'Toolbar',
+    agFindToolbarItem: 'Toolbar',
```
StephenCooper, line 61: Can this support an Array approach?

packages/ag-grid-enterprise/src/toolbar/agToolbar.css
```diff
@@ -0,0 +1,192 @@
+.ag-toolbar {
+    display: flex;
+    align-items: center;
+    gap: calc(var(--ag-spacing));
```
BernieSumption: Should remove redundant `calc()` like this and the ones that `calc(X * 1)`

packages/ag-grid-enterprise/src/toolbar/agToolbar.css
```diff
@@ -0,0 +1,192 @@
...
+
+:where(.ag-ltr) .ag-toolbar-input-icon {
+    /* rtl:ignore */
+    left: calc(var(--ag-spacing) * 1);
+}
+
+:where(.ag-rtl) .ag-toolbar-input-icon {
+    /* rtl:ignore */
+    right: calc(var(--ag-spacing) * 1);
+}
```
BernieSumption: This feature shouldn't need custom RTL styles, you should remove all of them, design for LTR mode, and in theory it should just work. If it doesn't then that's odd so let me know may need a tweak to the setup.

In general, rtl:ignore is for when you want a feature to look the same in RTL mode, or when you're doing some crazy advanced magic of the kind we try to avoid doing any more.

packages/ag-grid-enterprise/src/toolbar/agToolbar.css
```diff
@@ -0,0 +1,192 @@
...
+    font-size: var(--ag-font-size);
+    line-height: 1.5;
+    outline: none;
+    width: 100%;
+}
+
+:where(.ag-ltr) .ag-toolbar-input-field {
+    /* rtl:ignore */
+    padding: calc(var(--ag-spacing) * 0.5) calc(var(--ag-spacing) * 1) calc(var(--ag-spacing) * 0.5)
+        calc(var(--ag-spacing) * 1 + 14px + var(--ag-spacing) * 0.5);
```
BernieSumption: Since this is adding padding for an icon, you probably want `var(--ag-icon-size)` instead of 14px

Also that would add 2px more padding which I think you could use, the current padding looks a tiny bit tight

Image

community-modules/styles/src/internal/base/parts/_toolbar.scss
```diff
@@ -0,0 +1,204 @@
...
+        pointer-events: none;
+    }
+
+    .ag-ltr .ag-toolbar-input-icon {
+        left: calc(var(--ag-spacing) * 1);
+    }
+
+    .ag-rtl .ag-toolbar-input-icon {
+        right: calc(var(--ag-spacing) * 1);
+    }
```
BernieSumption: Search `ag.unthemed-rtl` for how we do this in Sass.

community-modules/styles/src/internal/base/parts/_toolbar.scss
```diff
@@ -0,0 +1,204 @@
+@use 'ag';
```
BernieSumption, line 1: Generally check that this file is up to date with the .css file, I asked Claude and got:

❯ is @community-modules/styles/src/internal/base/parts/_toolbar.scss a direct map onto @packages/ag-grid-enterprise/src/toolbar/agToolbar.css are there any substantive differences?
Read community-modules/styles/src/internal/base/parts/_toolbar.scss (205 lines)
Read packages/ag-grid-enterprise/src/toolbar/agToolbar.css (193 lines)

They're mostly equivalent but there are a few substantive differences:

1. .ag-toolbar gap value
- Legacy: gap: calc(var(--ag-spacing) * 2)
- Theming API: gap: calc(var(--ag-spacing))

2. .ag-toolbar-button padding
- Legacy: padding: calc(var(--ag-spacing) * 1) calc(var(--ag-spacing) * 1.5) (vertical/horizontal)
- Theming API: padding: calc(var(--ag-spacing)) (uniform)

3. Missing rules in Theming API CSS — these are in the legacy file but absent from agToolbar.css:
- .ag-toolbar-left (display, align, gap)
- .ag-toolbar-left:has(.ag-toolbar-panel) (flex: 1)
- .ag-toolbar-spacer (flex-grow: 1)
- .ag-toolbar-right (display, align, gap)

The remaining differences are just the expected pattern differences between the two style systems

community-modules/locale/src/en-US.ts
```diff
@@ -326,6 +326,12 @@ export const AG_GRID_LOCALE_EN = {
     sortAbsoluteDescending: 'Sort Absolute Descending',
     sortUnSort: 'Clear Sort',
 
+    // Toolbar
+    toolbarFind: 'Find',
```
StephenCooper, line 331: Add a linked Jira for the 35.3 translations if this does not exist yet.

testing/behavioural/src/toolbar/toolbar.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        });
+
+        await waitForEvent('firstDataRendered', api);
+
+        const gridDiv = TestGridsManager.getHTMLElement(api)!;
+        const toolbar = gridDiv.querySelector<HTMLElement>('.ag-toolbar');
+        expect(toolbar).toBeNull();
+    });
+
+    test('renders left and right containers', async () => {
```
StephenCooper: This test name seems outdated.

testing/behavioural/src/toolbar/toolbar-react.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        await rendered.findByText('Alice');
+
+        const container = rendered.container;
+        const toolbar = container.querySelector<HTMLElement>('.ag-toolbar');
+        expect(toolbar).toBeNull();
+    });
+
+    test('toolbar is positioned above grid body', async () => {
+        const rendered = render(
+            <AgGridReact rowData={[{ name: 'Alice' }]} columnDefs={[{ field: 'name' }]} toolbar={{ items: [] }} />
```
StephenCooper, line 46: This behaviour seems undesirable. I can expect that if the array is empty that we should not show the toolbar.

testing/behavioural/src/toolbar/toolbar-built-in-items.test.ts
```diff
@@ -0,0 +1,589 @@
+import { ClientSideRowModelModule, QuickFilterModule } from 'ag-grid-community';
```
StephenCooper: Nit: I think this could benefit from being split out per provided item. (row and pivoting combined)

packages/ag-grid-react/src/reactUi/gridComp.tsx
```diff
@@ -124,16 +124,32 @@ const GridComp = ({ context }: GridCompProps) => {
             paginationSelector,
             sideBarSelector,
             statusBarSelector,
+            toolbarSelector,
             gridHeaderDropZonesSelector,
         } = gridCtrl.getOptionalSelectors();
         const additionalEls: HTMLElement[] = [];
 
+        if (toolbarSelector) {
+            const toolbarComp = context.createBean(new toolbarSelector.component());
```
StephenCooper: Are we able to use the `addComponentToDom` helper defined below?

packages/ag-grid-enterprise/src/toolbar/toolbarModule.ts
```diff
@@ -0,0 +1,32 @@
...
+        agFindToolbarItem: FindToolbarItem,
+        agPivotPanelToolbarItem: PivotPanelToolbarItem,
+        agQuickFilterToolbarItem: QuickFilterToolbarItem,
+        agRowGroupPanelToolbarItem: RowGroupPanelToolbarItem,
+    } satisfies Record<ToolbarItemComponentName, any>,
+    icons: {
+        filter: 'filter',
+    },
+    selectors: [AgToolbarSelector],
+    dependsOn: [EnterpriseCoreModule, _KeyboardNavigationModule],
```
StephenCooper: Is the KeyboardNav module required?

packages/ag-grid-community/src/interfaces/iToolbar.ts
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Provide a custom component for the toolbar item.
+     * If omitted, the grid renders a default button using `label`, `icon` and `action`.
+     */
+    toolbarItem?: any;
+    alignment?: 'left' | 'right';
+    display?: ToolbarDisplay;
+    key?: string;
+    /** Parameters to be passed to the custom component specified in `toolbarItem`. */
+    toolbarItemParams?: any;
```
StephenCooper: Can we move this under `toolbarItem` so that in the Docs these are closely related. Will be clearer for the custom component approach. 

I am also going to keep thinking if it makes sense to make this a union type to enable us to provide stricter typing in the provided case as opposed to the custom component.

packages/ag-grid-enterprise/src/menu/exportMenuItems.ts
```diff
@@ -0,0 +1,26 @@
...
+        return null;
+    }
+    return {
+        name: localeTextFunc('csvExport', 'CSV Export'),
+        icon: _createIconNoSpan('csvExport', beans, null),
+        action: () => gridApi.exportDataAsCsv(),
+    };
+}
+
+export function getExcelExportMenuItem(beans: BeanCollection, localeTextFunc: LocaleTextFunc): MenuItemDef | null {
```
StephenCooper: If we are not providing these out of the box it might be better to revert this change as it is not required anymore.

packages/ag-grid-enterprise/src/toolbar/providedItems/buttonToolbarItem.ts
```diff
@@ -0,0 +1,81 @@
...
+    public refresh(params: ButtonToolbarItemParams): boolean {
+        this.applyParams(params);
+        return true;
+    }
+
+    private applyParams(params: ButtonToolbarItemParams): void {
+        this.params = params;
+        const eGui = this.getGui();
+
+        this.eIcon.replaceChildren();
```
StephenCooper: Can we use the shared `_clearElement` util for this. 

We may want to separately update that to use this simpler approach of calling replaceChildren but at least that will apply consistently to all usecases.

packages/ag-grid-enterprise/src/toolbar/providedItems/buttonToolbarItem.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.eLabel.textContent = params.label ?? '';
+        _setDisplayed(this.eLabel, params.display === 'iconAndLabel' && !!params.label);
+
+        if (params.label) {
+            eGui.setAttribute('aria-label', params.label);
+            eGui.setAttribute('title', params.label);
+        } else {
+            eGui.removeAttribute('aria-label');
+            eGui.removeAttribute('title');
+        }
```
StephenCooper: Could be simplified to 

```
 _setAriaLabel(eGui, params.label);
 _addOrRemoveAttribute(eGui, 'title', params.label);
```

community-modules/locale/src/en-US.ts
sdwvit: https://ag-grid.atlassian.net/browse/AG-17117

packages/ag-grid-enterprise/src/toolbar/providedItems/findToolbarItem.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const INPUT_DEBOUNCE_MS = 300;
+
+function createMatchCount(): HTMLSpanElement {
+    const eMatchCount = _createElement<HTMLSpanElement>({
+        tag: 'span',
+        cls: 'ag-toolbar-find-match-count',
+        attrs: { 'aria-live': 'polite' },
+    });
+    _setDisplayed(eMatchCount, false);
```
StephenCooper: Seems like this could be part of the `createElement` call?

packages/ag-grid-enterprise/src/toolbar/providedItems/rowGroupPanelToolbarItem.ts
```diff
@@ -0,0 +1,35 @@
...
+            _warn(302, {
+                itemName: 'rowGroupPanel',
+                moduleName: 'RowGroupingPanel',
+                ...this.gos.getModuleErrorParams(),
+            });
+            this.setDisplayed(false);
+            return;
+        }
+
+        const panel = this.createManagedBean(new RowGroupDropZonePanel(true));
```
StephenCooper: I suspect that this breaks tree shaking. We don't want class dependencies like this. The approach we have taken in other places to to define builder methods on a service that lives in the module that can be called. 

In this case we will need to add a new bean to the `RowGroupingPanelModule` whose only role is to build the row group drop zone panel and pivot drop zone panel. This will then need adding to the context for this module to then use. 

This should result in the self size of the toolbar module size dropping.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,366 @@
...
+        };
+    }
+    if (normalised.key == null) {
+        const key = typeof normalised.toolbarItem === 'string' ? normalised.toolbarItem : nextKey();
+        normalised = { ...normalised, key };
+    }
+    return normalised;
+}
+
+function getToolbarItemCompDetails(
```
StephenCooper: Just call directly, I see this only being used in one place.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,366 @@
...
+};
+
+class AgToolbar extends Component implements FocusableContainer {
+    private userCompFactory: UserComponentFactory;
+    private readonly toolbarItems: Map<string, IToolbarItemComp> = new Map();
+    private customKeyCounter: number = 0;
+    // Incremented on each rebuild so stale async resolves from a previous generation can be discarded
+    private generation: number = 0;
+
+    public wireBeans(beans: BeanCollection) {
```
StephenCooper: Just access off beans in the one location this is used to avoid the need for wireBeans and private property.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    public getFocusableContainerName(): 'toolbar' {
+        return 'toolbar';
+    }
+
+    private onTabKeyDown(_e: KeyboardEvent): void {
+        // Allow native tab order between toolbar items
+    }
+
+    private handleKeyDown(e: KeyboardEvent): void {
```
StephenCooper, line 107: I commented this method out locally and did not see the change in behaviour. Do we need this logic?

packages/ag-grid-enterprise/src/toolbar/providedItems/findToolbarItem.ts
sdwvit: If I inline this into _createElement, I'd need 'ag-hidden' plus 'aria-hidden': 'true' in attrs to match. Doesn't seem like a win to me.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
sdwvit: responded in slack

packages/ag-grid-enterprise/src/toolbar/providedItems/rowGroupPanelToolbarItem.ts
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        const panel = this.createManagedBean(builder.createRowGroupDropZone(true));
+        this.getGui().appendChild(panel.getGui());
+
+        // Keep the inner panel always visible — the wrapper controls visibility
+        panel.setDisplayed(true);
+        this.addManagedListeners(panel, {
+            displayChanged: () => panel.setDisplayed(true),
+        });
```
StephenCooper: Can we check if this is required.

packages/ag-grid-enterprise/src/toolbar/agToolbar.ts
```diff
@@ -0,0 +1,358 @@
...
+        if (!this.isAlive() || placeholder.parentNode !== this.getGui()) {
+            _removeFromParent(placeholder);
+            this.destroyBean(component);
+            return;
+        }
+
+        this.toolbarItems.set(key, component);
+        const gui = component.getGui();
+        placeholder.replaceWith(gui);
+        if (component instanceof Component) {
```
StephenCooper: Evaluate if we need this code.

StephenCooper (approved):

PR #13498, opened 2026-04-10
AG-3773 - [cell-notes] - added showDelay / hideDelay and display improvements

gportela85:
fix [AG-3773](https://ag-grid.atlassian.net/browse/AG-3773)

### DIFF
--- a/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
+++ b/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
@@ -298,15 +298,29 @@ export class CellCtrl extends BeanStub {
     }
 
     private checkFormulaError() {
+        if (!this.beans.formula) {
+            return;
+        }
         this.eGui.classList.toggle('formula-error', this.hasFormulaError());
     }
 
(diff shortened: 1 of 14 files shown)


packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -298,8 +298,15 @@ export class CellCtrl extends BeanStub {
     }
 
     private checkFormulaError() {
-        const isFormulaError = !!this.beans.formula?.getFormulaError(this.column, this.rowNode);
-        this.eGui.classList.toggle('formula-error', isFormulaError);
+        this.eGui.classList.toggle('formula-error', this.hasFormulaError());
```
StephenCooper, line 304: Can we update this code to only call toggle if this.beans.formula is actually present. I have seen this pop up in a profiling that I have been doing.

gportela85: fixed

StephenCooper (approved):

PR #13503, opened 2026-04-11
Cancel stale cell selection refresh animation frames

### LINKED ISSUE #13502: [Bug]: Cell selection refresh rAF callbacks can pile up in hidden tabs

https://plnkr.co/edit/YlSDQfPFPJPZv5Pe?open=main.js&preview

### Describe the bug

Steps to reproduce

1. Open the live Plunker sample: https://plnkr.co/edit/YlSDQfPFPJPZv5Pe?preview
2. Click Start Spam.
3. Leave the tab in the background for 10-20 seconds.
4. Return to the tab.
5. Observe the latched metrics:

last hidden scheduled
last hidden peak pending
pending on resume
fired after resume

This behavior is also observed in the live example here: https://www.ag-grid.com/example-finance/

Open the dev tools in that example, and paste the following:
```ts
(() => {
  const nativeRaf = window.requestAnimationFrame.bind(window);
  const nativeCancel = window.cancelAnimationFrame.bind(window);
  const pending = new Set();
  let scheduled = 0;
  let cancelled = 0;
  let fired = 0;

  window.requestAnimationFrame = (cb) => {
    let id;
    id = nativeRaf((ts) => {
      pending.delete(id);
      fired++;
      cb(ts);
    });
    pending.add(id);
    scheduled++;
    return id;
  };

  window.cancelAnimationFrame = (id) => {
    pending.delete(id);
    cancelled++;
    return nativeCancel(id);
  };

  window.rafStats = () => ({ scheduled, cancelled, fired, pending: pending.size });
})();

setInterval(() => {
    console.log(rafStats());
}, 1000);
```

You will see that the number of pending requestAnimationFrame continues to increase if you switch tabs.

Actual behaviour

While the tab is hidden, repeated grid refreshes continue scheduling deferred cell-selection refresh work. Those callbacks are not cancelled or deduped, so work builds up while the tab is backgrounded. When the tab becomes visible again, the backlog flushes in a burst, shown in the sample by non-zero pending on resume / fired after resume values and cancelled remaining at 0.

Expected behaviour

Only the latest pending cell-selection refresh should remain queued per cell. If another refresh is scheduled before the previous frame runs, the stale callback should be cancelled or coalesced. Returning to a previously hidden tab should not cause a burst of redundant cell-selection refresh work.

More information

Repro sample is based on the AG Grid range-handle example and uses cell selection with a preselected range.
The sample triggers gridApi.refreshCells({ force: true, suppressFlash: true }) every 100ms.
This appears to affect the cell-selection refresh path specifically; it is most visible when cell selection is enabled and a range/handle is active.

Observed locally on latest, and reproduced in the Plunker using published AG Grid 35.2.1.

### Version

35.2.1

### Does the issue occur for a specific framework only?

All languages

### Is the issue only observable on a specific browser?

All browsers

rickono:
Fixes #13502

Cancel pending cell selection refresh animation frames before scheduling a new one and when destroying a cell.

This prevents hidden tabs from accumulating stale cell selection refresh callbacks that all flush when the tab becomes visible again.

Adds unit coverage for rescheduling, explicit clearing, and destroy cleanup.

### DIFF
--- a/packages/ag-grid-community/src/agStack/utils/dom.ts
+++ b/packages/ag-grid-community/src/agStack/utils/dom.ts
@@ -443,15 +443,32 @@ export function _observeResize(
 }
 
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
-export function _requestAnimationFrame(beans: UtilBeanCollection, callback: any) {
+export function _requestAnimationFrame(beans: UtilBeanCollection, callback: any): number {
     const win = _getWindow(beans);
 
     if (win.requestAnimationFrame) {
-        win.requestAnimationFrame(callback);
+        return win.requestAnimationFrame(callback);
     } else if ((win as any).webkitRequestAnimationFrame) {
-        (win as any).webkitRequestAnimationFrame(callback);
+        return (win as any).webkitRequestAnimationFrame(callback);
     } else {
-        win.setTimeout(callback, 0);
+        return win.setTimeout(callback, 0);
+    }
+}
+
+/** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
+export function _cancelAnimationFrame(beans: UtilBeanCollection, requestId: number | null | undefined): void {
+    if (requestId == null) {
+        return;
+    }
+
+    const win = _getWindow(beans);
+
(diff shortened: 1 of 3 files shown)


StephenCooper:
Will be resolved as part of AG-17093

PR #13530, opened 2026-04-15
AG-9670 Enable Re-ordering of Pagination Controls

StephenCooper:
Fix [AG-9670](https://ag-grid.atlassian.net/browse/AG-9670)


packages/ag-grid-community/src/gridOptionsInitial.ts
```diff
@@ -39,7 +39,7 @@ export const INITIAL_GRID_OPTION_KEYS = {
     loadingOverlayComponent: true,
     suppressLoadingOverlay: true,
     noRowsOverlayComponent: true,
-    paginationPageSizeSelector: true,
```
StephenCooper, line 42: The code is responsive so no need for this to be marked initial.

PR #13551, opened 2026-04-17
AG-17093 Prevent raf accumulation in background tab

StephenCooper:
Fix [AG-17093](https://ag-grid.atlassian.net/browse/AG-17093)

Prevents unbounded requestAnimationFrame cellSelection refreshes from accumulating when tab in the background. Able to only keep the latest raf as the callback always takes the latest data.

Also made an optimisation so that the RowSpanService does not register event listeners which result in further raf calls, even when this feature is not used.

### DIFF
--- a/packages/ag-grid-community/src/interfaces/iCellRangeFeature.ts
+++ b/packages/ag-grid-community/src/interfaces/iCellRangeFeature.ts
@@ -4,7 +4,7 @@ import type { ICellComp } from '../rendering/cell/cellCtrl';
 export interface ICellRangeFeature {
     setComp(cellComp: ICellComp): void;
     unsetComp(): void;
-    refreshRangeStyleAndHandle(): void;
+    scheduleRefreshRangeStyleAndHandle(): void;
     updateRangeBordersIfRangeCount(): void;
     onCellSelectionChanged(): void;
     destroy(): void;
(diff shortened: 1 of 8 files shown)


packages/ag-grid-community/src/navigation/cellNavigationService.ts
```diff
@@ -4,25 +4,17 @@ import { _missing } from '../agStack/utils/generic';
...
 import type { CellPosition } from '../interfaces/iCellPosition';
 import type { IRowNode } from '../interfaces/iRowNode';
-import type { RowSpanService } from '../rendering/spanning/rowSpanService';
 import { _warn } from '../validation/logging';
 
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
 export class CellNavigationService extends BeanStub implements NamedBean {
     beanName = 'cellNavigation' as const;
 
-    private rowSpanSvc: RowSpanService | undefined;
```
StephenCooper, line 20: Just a small code cleanup in this file.

packages/ag-grid-community/src/interfaces/iCellRangeFeature.ts
```diff
@@ -4,7 +4,7 @@ import type { ICellComp } from '../rendering/cell/cellCtrl';
 export interface ICellRangeFeature {
     setComp(cellComp: ICellComp): void;
     unsetComp(): void;
-    refreshRangeStyleAndHandle(): void;
+    scheduleRefreshRangeStyleAndHandle(): void;
```
sdwvit, line 7: is it internal? safe to rename?

sdwvit (approved):

packages/ag-grid-community/src/interfaces/iCellRangeFeature.ts
StephenCooper: Yes this is internal. See the jsDoc on the interface.

PR #13559, opened 2026-04-17
AG-3773 - [notes] - improved positioning

gportela85:
fix [AG-3773](https://ag-grid.atlassian.net/browse/AG-3773)

### DIFF
--- a/packages/ag-grid-community/src/agStack/popup/popupPositionUtils.ts
+++ b/packages/ag-grid-community/src/agStack/popup/popupPositionUtils.ts
@@ -195,7 +195,7 @@ function normaliseFindBestPlacementOptions(
 
 function mirrorAlignment(alignment: Alignment): Alignment {
     const [targetAnchor, referenceAnchor] = alignment.split('-') as [Anchor, Anchor];
-    return `${mirrorAnchor(targetAnchor)}-${mirrorAnchor(referenceAnchor)}` as Alignment;
+    return `${mirrorAnchor(targetAnchor)}-${mirrorAnchor(referenceAnchor)}`;
 }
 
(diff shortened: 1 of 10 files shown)


packages/ag-grid-community/src/agStack/popup/popupPositionUtils.ts
```diff
@@ -154,6 +175,48 @@ function getGapDirection(targetAnchor: Anchor, refAnchor: Anchor): { dx: number;
...
+        mirrorPlacementsInRtl: gapOrOptions.mirrorPlacementsInRtl ?? true,
+    };
+}
+
+function mirrorAlignment(alignment: Alignment): Alignment {
+    const [targetAnchor, referenceAnchor] = alignment.split('-') as [Anchor, Anchor];
+    return `${mirrorAnchor(targetAnchor)}-${mirrorAnchor(referenceAnchor)}`;
+}
+
+function mirrorAnchor(anchor: Anchor): Anchor {
```
StephenCooper, line 228: Slightly less code if written as an object lookup?

packages/ag-grid-enterprise/src/notes/agNotesPopup.ts
```diff
@@ -257,3 +257,41 @@ export class AgNotesPopup extends BeanStub {
         this.params.onClosed(noteChanged, editedNote, closeEvent);
     }
 }
+
+export function findNotesPopupPosition(params: {
```
StephenCooper: Can we tidy this function up a little? Thinking the input types could be real types and maybe we could reduce some of the chained accessors in the method?

StephenCooper (approved):

PR #13572, opened 2026-04-21
feat: add api.applyAutoSizeStrategy() for runtime reapplication (#13525)

### LINKED ISSUE #13525: [Support]: Enhancement request: API to reapply configured autoSizeStrategy after initial load

_No response_

### Current behavior
`autoSizeStrategy` appears to run only during initial grid load / first data render.

After that point, there is no single API that reapplies the configured strategy from `gridOptions.autoSizeStrategy.`
You can use `sizeColumnsToFit` and `autoSizeAllColumns` but both diverge from what my `autoSizeStrategy` was... 

`
export const autoSizeStrategy: SizeColumnsToContentStrategy = {
  type: "fitCellContents",
  scaleUpToFitGridWidth: true
};
`

Proposal
Add a Grid API method to reapply the configured strategy at runtime, e.g.:

`api.applyAutoSizeStrategy()`
Uses `gridOptions.autoSizeStrategy `when present.
Optional overload:

`api.applyAutoSizeStrategy(strategyOverride)`

Behavior:
1. If a strategy exists, apply it according to type.
2. If no strategy exists, either:
- no-op, or
- fall back to current default autosize behavior (maintainer preference).

I am currently checking the size of content in a resize observer with a `debounce` and getting the size of the cell contents but would be nice to do that with only `gridRef.current.api.applyAutoSizeStrategy()`

This is what I am currently doing to get around this
```
  useEffect(() => {
    let timeoutId: NodeJS.Timeout | null = null;
    let initialized = false;

    const sizeColumns = () => {
      if (tableRef.current && containerRef.current) {
        const api = tableRef.current.api;
        const allColumns = api?.getColumns() ?? [];
        const containerWidth = containerRef.current.offsetWidth;

        // Auto-size first to get true content widths, then decide
        api?.autoSizeColumns(allColumns, false);

        const naturalContentWidth = allColumns.reduce(
          (sum, col) => sum + col.getActualWidth(),
          0
        );

        if (containerWidth > naturalContentWidth) {
          api?.sizeColumnsToFit();
        }
        // else: leave columns at natural auto-sized widths for horizontal scroll
      }
    };

    const resizeObserver = new ResizeObserver(() => {
      // Let the autoSizeStrategy run on initial mount
      if (!initialized) {
        initialized = true;
        return;
      }
      if (timeoutId) clearTimeout(timeoutId);
      // Debounce any change
      timeoutId = setTimeout(sizeColumns, 350);
    });

    if (containerRef.current) {
      resizeObserver.observe(containerRef.current);
    }

    return () => {
      if (timeoutId) clearTimeout(timeoutId);
      resizeObserver.disconnect();
    };
  }, []);
```

### Version

34.3.0

### What language are you using?

TypeScript

### What browser are you using?

Chrome

nrsundberg:
Adds a public grid API method that reapplies the configured autoSizeStrategy (or an ad-hoc override) at any time after init — enabling column sizing in response to container resizes, filter changes, or other dynamic layout events without reloading the grid.

- New method: `api.applyAutoSizeStrategy(strategyOverride?)`
- Supports all three strategy variants (`fitGridWidth`, `fitProvidedWidth`, `fitCellContents`)
- No-op when neither override nor `gridOptions.autoSizeStrategy` is set
- Columns resized by this call fire `columnResized` with `source: 'autoSizeStrategy'`
- Init-time dispatch and runtime dispatch share a single helper
- `autoSizeCols` gains a narrow opt-in `eventSource` for the final batched dispatch so the strategy's source surfaces without disturbing existing autosize call sites
- Adds behavioural test coverage and a "Reapplying the Auto-Size Strategy at Runtime" subsection under Column Sizing with an interactive example

## Review follow-ups

- Renamed the init-only `applyAutosizeStrategy` to `applyInitialAutoSizeStrategy` to remove the confusing casing collision with the public API
- Trimmed multi-line JSDocs on internal autosize service methods
- Tightened the `fitCellContents` + `colIds` behavioural test to assert at least one targeted column appears in the resize events
- Added an interactive docs example with four buttons (re-apply configured, override with each built-in variant) and cross-framework Playwright coverage

Closes #13525

### DIFF
--- a/documentation/ag-grid-docs/src/content/api-documentation/grid-api/api.json
+++ b/documentation/ag-grid-docs/src/content/api-documentation/grid-api/api.json
@@ -75,6 +75,12 @@
                 "name": "Auto-Size Columns to Fit Cell Contents",
                 "url": "./column-sizing/#auto-size-columns-to-fit-cell-contents"
             }
+        },
+        "applyAutoSizeStrategy": {
+            "more": {
+                "name": "Reapplying the Auto-Size Strategy at Runtime",
+                "url": "./column-sizing/#reapplying-the-auto-size-strategy-at-runtime"
+            }
         }
     },
     "columnMoving": {
--- a/documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/apply-auto-size-strategy/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/apply-auto-size-strategy/example.spec.ts
@@ -1,0 +1,74 @@
+import { type AgGridFixtures, expect, test, waitForGridContent } from '@utils/grid/test-utils';
+import type { Locator, Page } from 'playwright/test';
+
+async function getWidth(locator: Locator): Promise<number | undefined> {
+    return (await locator.boundingBox())?.width;
+}
+
+const COL_IDS = ['athlete', 'age', 'country', 'year', 'date', 'sport'] as const;
+
+type ColIds = (typeof COL_IDS)[number];
+
+function getHeaders(agIdFor: AgGridFixtures['agIdFor']): Record<ColIds, Locator> {
+    return Object.fromEntries(COL_IDS.map((colId) => [colId, agIdFor.headerCell(colId)])) as Record<ColIds, Locator>;
+}
+
+async function totalHeaderWidth(headers: Record<ColIds, Locator>): Promise<number> {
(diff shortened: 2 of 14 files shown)


StephenCooper:
We have an existing ticket AG-4093 [Column Sizing] Allow autoSizeStrategy to re-run on Grid Events. I have linked this to that work as additional user feedback.

I will be closing this PR.

PR #13589, opened 2026-04-22
AG-16878 better formula cache

SalvatorePreviti:
# AG-16878 — Better formula cache

Preserve parsed ASTs across row/column changes, invalidate values in O(1), slim per-cell error state, make invalidation row-change-aware.

## Core

**`cachedResult: Map<RowNode, Map<AgColumn, CellFormula>>`** (was `WeakMap`). Iteration + `size` short-circuits. Explicit cleanup at every mutation site.

**Versioned cache.** `valueCacheVersion++` invalidates every cached value in O(1); ASTs survive. `CellFormula._valueVersion` stamps each write; `isValueReady()` is an integer compare. `refreshFormulas` remains the nuclear option (`cellValueChanged`, column-instance changes).

**Error model.** `CellFormula.error: FormulaError` → scalars (`errorType` / `errorId` / `errorMessage` / `errorVariableValues`), rebuilt on demand. The old shape retained MBs of closure state via `Error` stack traces on grids with many erroring cells; four scalars per cell now. Eval loop throws `CellFormula` itself — no per-propagation `new FormulaError`.

## CSRM — `onRowsChanged`

`IFormulaService.onRowsChanged(changedRowNodes, newData)` at the end of CSRM `refreshModel`.

-   `newData` → `refreshFormulas(true)`.
-   `removals` / `updates` → `dropRow` each (covers `sibling`, `sibling.pinnedSibling`, `pinnedSibling`).
-   `reordered || removals.length || adds.size || updates.size` → `bumpValueCacheAndRefresh`. **`adds.size` included:** `ROW("N",true)` that was `#REF!` re-evaluates when row N is appended (regression test added).
-   No `changedRowNodes` (sort / filter / paginate / column change) → bump; positions may have shifted.

Replaces the old `modelUpdated` / `rowDataUpdated` listeners. `formulaDataSvc.getFormula` was N×M (per visible cell per invalidation); now N (once per `(row, col)`).

**Pinned rows.** `onPinnedRowsChanged` (for `pinnedRowDataChanged` + `pinnedRowsChanged`) drops entries where `row.rowPinned` + their group-footer siblings. Deliberately **not** `dropRow`: a pinned row's `pinnedSibling` points back to the still-valid unpinned main.

## Column listeners

-   `newColumnsLoaded`: rebuild A1 map + `refreshFormulas(true)` (column instances may be destroyed).
-   `columnMoved`: rebuild A1 map + `bumpValueCacheAndRefresh` (ASTs preserved).

`rebuildColRefMap`: in-place `.clear()`, uppercase alphabet directly (was lowercase + `toUpperCase` per col), parallel `colToRefMap: Map<AgColumn, string>` makes `getColRef(col)` O(1) — was a full entries scan called per header render / formula token highlight.

## Micro-optimizations

-   **Pre-bound callbacks** (`ensureCellFormulaBound`, `resolveAddrRef`): once per service, not per frame.
-   **Reused frame `address`** as `evalAst`'s current-cell arg — no per-frame `{ row, column }` allocation.
-   **Cached lookups**: `baseDataType` on `CellFormula`, `formulaDataSvc` in `postConstruct`, `formulaColumnsPresent` refreshed only in `setFormulasActive`, shared `REFRESH_CELLS_PARAMS` const.
-   **Hot-field placement**: declared first with initializers. `activeCtx` previously triggered a V8 hidden-class transition on first write.
-   **Indexed `for` loops** on column scan paths. New `isValidFunctionName` drops the `[...name].every(...)` spread allocation.
-   **`_isExpressionString`**: `charCodeAt(0) === 61` instead of `startsWith('=')`. Cell-render hot path.

## Structural

`CellFormula` extracted to `cellFormula.ts` (cycle broken via `import type`). `FormulaService.destroy()` clears owned state. Serializer: template literals → `+` concat, shared `LETTERS_ONLY` regex, `col.colId` over `col.getId()`, extracted `emitA1Ref` / `emitRefColPart` / `emitRefRowPart`.

## Tests & tooling

Regression test for the `adds` bump. Edge-case serializer test (operand types, precedence, parens, unary/postfix, ranges). Column-reorder benchmark. `bench-compare.mjs`: per-file grouping, certain-vs-noisy split, stable file-identity anchors.

## Measured impact

Numbers predate the micro-opt pass — treat as a floor.

↑ 3.54x cold: invalidate then evaluate sum/product/branch, 3000 rows (16.42 → 58.08 ops/s)
↑ 2.39x re-eval 200 chained + 1500 fan-out dependents (144.1 → 344.4 ops/s)
↑ 1.09x cold: invalidate then SUM over 100000 cells (48.76 → 52.99 ops/s)
↑ 1.07x swap columns then re-evaluate 6000 formulas (1154.3 → 1231.5 ops/s)
↑ 1.09x warm: re-read sum/product/branch, 3000 rows (804.0 → 878.9 ops/s)

Old cold path re-parsed every formula on every `modelUpdated`; new cold path keeps ASTs. Per-erroring-cell memory: KBs → four scalars.


packages/ag-grid-enterprise/src/formula/formulaService.ts
```diff
@@ -158,10 +208,25 @@ export class FormulaService extends BeanStub implements IFormulaService, NamedBe
...
+        };
+        const onColumnMoved = () => {
+            if (!this.active || this.cachedResult.size === 0) {
+                return;
+            }
+            this.rebuildColRefMap();
+            // Column instances are stable across a reorder - only their positions changed. Parsed
+            // ASTs keep referring to the right colIds; only absolute `COLUMN("A",true)` refs pick
+            // up different columns via colRefMap. Bumping the value version re-evaluates surviving
+            // cells while preserving their parsed ASTs.
```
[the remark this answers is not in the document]
SalvatorePreviti: fixed

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
```diff
@@ -32,155 +30,149 @@ export function colIndexFromId(colModel: ColumnModel, cols: AgColumn[], colId: s
...
-
-    // If child is unary-minus-encoded, let the unary print logic decide.
-    if (isUnaryMinusNode(child)) {
+    // Unary-minus children format as '-x' or '-(x)' which is self-contained; no outer parens.
+    if (unaryMinusInner(child)) {
         return false;
     }
 
     const childDef = getDefBySymbol(child.operation, 'infix');
-    if (!childDef || childDef.fixity !== 'infix') {
```
StephenCooper, line 183: Is this change in logic a bug fix?

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
```diff
@@ -32,155 +30,149 @@ export function colIndexFromId(colModel: ColumnModel, cols: AgColumn[], colId: s
...
     }
     return startRef;
 }
 
-function serializeCellREF(beans: BeanCollection, cell: Cell): string {
-    const colPart = (r: CellRef) => `COLUMN(${quoteString(columnValueForREF(beans, r))}${r.absolute ? ',true' : ''})`;
-    const rowPart = (r: CellRef) => `ROW(${quoteString(rowValueForREF(beans, r))}${r.absolute ? ',true' : ''})`;
+function emitRefColPart(beans: BeanCollection, ref: CellRef): string {
+    const value = quoteString(columnValueForREF(beans, ref));
+    return ref.absolute ? 'COLUMN(' + value + ',true)' : 'COLUMN(' + value + ')';
```
StephenCooper: Should we reduce duplication by moving the conditional part?

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
```diff
@@ -32,155 +30,149 @@ export function colIndexFromId(colModel: ColumnModel, cols: AgColumn[], colId: s
...
 }
 
+const LETTERS_ONLY = /^[A-Za-z]+$/;
+
 function quoteString(s: string): string {
     if (s.includes('"')) {
         throw new FormulaError(18);
     }
-    return `"${s}"`;
+    return '"' + s + '"';
```
StephenCooper, line 54: Is string interpolation slower?

SalvatorePreviti: Yes it usually is

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
SalvatorePreviti: Ok

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
SalvatorePreviti: It is dead code in the meaning that that path should be never reached, but I will have another look later to see if I missed something in general

StephenCooper (approved):

packages/ag-grid-enterprise/src/formula/ast/serializer.ts
SalvatorePreviti: getDefBySymbol does the same thing indeed

packages/ag-grid-enterprise/src/formula/formulaService.ts
```diff
@@ -146,91 +174,225 @@ export class FormulaService extends BeanStub implements IFormulaService, NamedBe
...
+            let dropped = false;
+            for (const row of cache.keys()) {
+                if (row.rowPinned) {
+                    cache.delete(row);
+                    dropped = true;
+                    const sibling = row.sibling;
+                    if (sibling) {
+                        cache.delete(sibling);
+                    }
+                }
```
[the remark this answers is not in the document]
SalvatorePreviti: we support only client side row model with formulas

StephenCooper (approved):

PR #13601, opened 2026-04-23
[AG-17112] Feature(toolbar): support menu items for grouped actions

sdwvit:
- Adds new `agMenuToolbarItem` built-in toolbar item that renders a button opening a `MenuList` popup of `MenuItemDef`s (configurable `label`, `tooltip`, `icon`, `menuItems` via `toolbarItemParams`), registered in `ToolbarModule`, `ToolbarItemComponentName`, and `USER_COMP_MODULES`; documented in `toolbar/index.mdoc`; with a new `toolbarMenu` locale entry used as the default label; and demonstrated in the built-in items example via a CSV/Excel export menu (with `CsvExportModule`/`ExcelExportModule` registered and `example.spec.ts` counts updated)
- Updates custom toolbar button examples (TS, React, Angular, Vue) to support an optional `title` separate from `label`, so tooltip text no longer has to come from the visible label
- Widens `.ag-toolbar-find` from `240px` to `280px` in both the legacy Sass part and the Theming API CSS

Fix [AG-17112](https://ag-grid.atlassian.net/browse/AG-17112)


sdwvit:
working on module size now

documentation/ag-grid-docs/src/content/docs/toolbar/index.mdoc
```diff
@@ -45,6 +45,7 @@ The following built-in items are available:
 -   `agPivotPanelToolbarItem`: Embeds the [Pivot Panel](./pivoting/) in the toolbar. Requires `RowGroupingPanelModule`. Configured independently of the `pivotPanelShow` grid option — if both are set, both will render.
 -   `agRowGroupPanelToolbarItem`: Embeds the [Row Group Panel](./grouping-group-panel/) in the toolbar. Requires `RowGroupingPanelModule`. Configured independently of the `rowGroupPanelShow` grid option — if both are set, both will render.
 -   `agQuickFilterToolbarItem`: A text input that filters grid rows using the [Quick Filter](./filter-quick/). Requires `QuickFilterModule`.
+-   `agMenuToolbarItem`: A button that opens a dropdown menu of [Menu Items](./component-menu-item/). Requires `ContextMenuModule`. Supply `menuItems` via `toolbarItemParams`, along with an optional `label`, `tooltip`, and `icon`. `menuItems` accepts the same entries as the [Context Menu](./context-menu/#built-in-menu-items) — `MenuItemDef` objects or built-in string names such as `'copy'`, `'export'`, or `'separator'`.
```
StephenCooper: Does it also work just with `ColumnMenu`? Is so we should update the docs to include both.

packages/ag-grid-community/src/interfaces/iToolbar.ts
```diff
@@ -58,14 +59,26 @@ export interface ToolbarButtonItemDef<TData = any, TContext = any> extends Toolb
...
 /**
  * Reference to a built-in toolbar item component (or `'separator'`).
  */
-export interface ToolbarBuiltInItemDef extends ToolbarItemDefBase {
+export interface ToolbarBuiltInItemDef<TData = any, TContext = any> extends ToolbarItemDefBase {
     /** A built-in toolbar item component name, or `'separator'`. */
     toolbarItem: ToolbarItemComponentName | 'separator';
-    /** Built-in items do not accept params. */
-    toolbarItemParams?: never;
+    /** Params for the built-in item. Currently only `agMenuToolbarItem` consumes params. */
```
StephenCooper: If we split this into 2 parts would that enable the `toolbarItem: 'agMenuToolbarItem'` to be the only one with params type and the rest get never like before?

packages/ag-grid-enterprise/src/toolbar/providedItems/menuToolbarItem.ts
```diff
@@ -0,0 +1,107 @@
...
+            return;
+        }
+
+        const eGui = this.getGui();
+        _setAriaExpanded(eGui, true);
+
+        contextMenuSvc.showAnchoredMenu({
+            anchorElement: eGui,
+            menuItems,
+            ariaLabel: this.params.tooltip ?? this.params.label ?? this.getLocaleTextFunc()('toolbarMenu', 'Menu'),
```
StephenCooper: Can we share this logic to avoid repeating it.

packages/ag-grid-enterprise/src/menu/contextMenu.ts
```diff
@@ -83,6 +93,57 @@ export class ContextMenuService extends BeanStub implements NamedBean, IContextM
         this.menu.hideActiveMenu();
     }
 
+    public showAnchoredMenu(params: ShowAnchoredMenuParams): void {
```
StephenCooper: I can see why the menu button is dependant on the ContextMenu now with this current implementation. 

Open questions, should we follow the more isolated approach like with the RowGroupPanelBuilder? i.e just a builder for the MenuList logic. 

Can we re-use the `ColumnMenuFactory`?

sdwvit:
Accidentally force pushed & reset to the wrong commit. Local history restored, gonna do quick review if everything is back.

sdwvit:
all ready for review again

StephenCooper (approved):

PR #13620, opened 2026-04-24
AG-16677 fix(pivot): keep re-added pivotRowTotals groups adjacent

sdwvit:
Fix bug with pivot totals column positioning during grid modifications

  - Added isPivotRowTotalColumn() to IPivotColDefService.
  - Implemented it in PivotColDefService using the PivotRowTotal_ colId prefix.
  - Updated restoreColOrder() so new pivot row total columns do not use generic sibling matching.
  - Instead, they anchor after the previous preserved column in the fresh generated list, so they stay after group/service columns and before/after pivot value columns as generated.

  Also fixed a sorting bug: 

  - PivotStage now detects pure value column reorder: same value columns, different order.
  - That signal is passed through pivotResultCols.setPivotResultCols(...).
  - ColumnModel.refreshCols(...) skips restoreColOrder() for that refresh, so the freshly generated pivot order is shown instead of restoring the stale order.
  - Added behavioural coverage for setValueColumns(['bronze', 'gold', 'silver']).

Fix [AG-16677](https://ag-grid.atlassian.net/browse/AG-16677)

Local repro https://plnkr.co/edit/CIAKpYngoZwWnVha?open=main.js

### DIFF
--- a/packages/ag-grid-community/src/columns/columnModel.ts
+++ b/packages/ag-grid-community/src/columns/columnModel.ts
@@ -461,12 +461,15 @@ export class ColumnModel extends BeanStub implements NamedBean {
         const previousSiblingPosMap: Map<AgColumn, AgColumn | AgColumn[]> = new Map();
 
         const { pivotColDefSvc } = this.beans;
-        const freshListIndex = new Map<AgColumn, number>();
-        for (let i = 0; i < cols.list.length; i++) {
-            freshListIndex.set(cols.list[i], i);
-        }
+        let freshListIndex: Map<AgColumn, number> | null = null;
 
         const getPreviousColInFreshList = (col: AgColumn): AgColumn | null => {
+            if (freshListIndex == null) {
+                freshListIndex = new Map<AgColumn, number>();
+                for (let i = 0; i < cols.list.length; i++) {
+                    freshListIndex.set(cols.list[i], i);
+                }
+            }
             const idx = freshListIndex.get(col)!;
             for (let i = idx - 1; i >= 0; i--) {
                 const candidate = cols.list[i];
--- a/packages/ag-grid-community/src/columns/columnModel.ts
+++ b/packages/ag-grid-community/src/columns/columnModel.ts
@@ -174,7 +174,7 @@ export class ColumnModel extends BeanStub implements NamedBean {
     // setPivotMode, applyColumnState,
     // functionColsService.setPrimaryColList, functionColsService.updatePrimaryColList,
     // pivotResultCols.setPivotResultCols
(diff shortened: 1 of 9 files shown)


packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -460,23 +460,54 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
             } else {
                 // if we have a single col, then we need to add the new col to the array
-                previousSiblingPosMap.set(prevSiblingIdx, [prev, col]);
+                previousSiblingPosMap.set(anchor, [prev, col]);
+            }
+        };
+
+        // for each new col, find the col it needs inserted after and store for when array is constructed
+        for (const col of additionalCols) {
+            if (pivotColDefSvc?.isPivotRowTotalColumn(col.colDef)) {
```
SalvatorePreviti, line 497: There is a better way to understand if a column is a pivot total column that will also implicitly check if pivot is enabled: colDef.pivotTotalColumnIds != null - so you don't need this method

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -460,23 +460,54 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
-            if (prevSiblingIdx == null) {
-                noSiblingsAvailable.push(col);
-                continue;
+        const { pivotColDefSvc } = this.beans;
+        const freshListIndex = new Map<AgColumn, number>();
+        for (let i = 0; i < cols.list.length; i++) {
+            freshListIndex.set(cols.list[i], i);
+        }
+
+        const getPreviousColInFreshList = (col: AgColumn): AgColumn | null => {
```
SalvatorePreviti, line 466: Might this new code make col model update slower for when pivot is not enabled? shall we check for it?

packages/ag-grid-community/src/columns/columnModel.ts
sdwvit: yeah, that was my initial implementation, but it is not broad enough it seems pivotTotalColumnIds can be nullable, while column should still be accounted for. Also I like the fact it is encapsulated into a service. I get it is more code, but it's a bit cleaner

SalvatorePreviti: mhhh ok for now then

SalvatorePreviti (approved):

packages/ag-grid-community/src/columns/columnModel.ts
sdwvit: thank you for flagging , let me fix it.

PR #13622, opened 2026-04-27
Format with updated prettier

seanlandsman:


StephenCooper (approved):
Can we ensure this commit is excluded from git blame.

PR #13632, opened 2026-04-28
chore(bench): add pivot transaction benchmark

sdwvit:
Adds a pivot transaction-update benchmark.

### DIFF
--- a/packages/ag-grid-enterprise/src/pivot/pivotStage.ts
+++ b/packages/ag-grid-enterprise/src/pivot/pivotStage.ts
@@ -143,11 +143,12 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
             // list (and the includes-scan against last time) on every transaction
             // would burn allocations on the no-op hot path.
             const aggregationColumnIds = aggregationColumns.map((column) => column.getId());
+            const lastIds = this.aggregationColumnIdsLastTime;
             const aggregationColumnsReordered =
-                this.aggregationColumnIdsLastTime != null &&
-                this.aggregationColumnIdsLastTime.length === aggregationColumnIds.length &&
-                !_areEqual(this.aggregationColumnIdsLastTime, aggregationColumnIds) &&
-                this.aggregationColumnIdsLastTime.every((id) => aggregationColumnIds.includes(id));
+                lastIds != null &&
+                lastIds.length === aggregationColumnIds.length &&
+                !_areEqual(lastIds, aggregationColumnIds) &&
+                lastIds.every((id) => aggregationColumnIds.includes(id));
             this.aggregationColumnIdsLastTime = aggregationColumnIds;
 
             const pivotColumnGroupDefs = this.pivotColDefSvc.createPivotColumnDefs(this.uniqueValues);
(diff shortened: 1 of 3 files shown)


packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -146,6 +139,17 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
             pivotOrderChanged ||
             anyGridOptionsChanged
         ) {
+            // Reorder detection only matters on the rebuild path. Computing the id
+            // list (and the includes-scan against last time) on every transaction
+            // would burn allocations on the no-op hot path.
+            const aggregationColumnIds = aggregationColumns.map((column) => column.getId());
```
StephenCooper, line 145: Is this at risk of leading to stale values in the last value if it is not kept up to date with the current state of the grid for each change?

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -146,6 +139,17 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
...
+            const aggregationColumnsReordered =
+                this.aggregationColumnIdsLastTime != null &&
+                this.aggregationColumnIdsLastTime.length === aggregationColumnIds.length &&
+                !_areEqual(this.aggregationColumnIdsLastTime, aggregationColumnIds) &&
+                this.aggregationColumnIdsLastTime.every((id) => aggregationColumnIds.includes(id));
```
SalvatorePreviti: I am unsure about this ... why do we need the every if we are executing an _areEqual? also, this is of course O(n*n) and would be better to create a set of ids so it becomes O(n), and this.aggregationColumnIdsLastTime is accessed way too many times and this doesn't get optimized, putting it in a local makes the code faster smaller and more readable

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
sdwvit: No risk of stale values.`aggregationColumnsHash` is recomputed every call, and any `add/remove/reorder/rename` of agg columns flips `aggregationColumnsChanged` to true, next it enters the rebuild branch, and `aggregationColumnIdsLastTime` gets refreshed. So the cached value can only diverge from the live columns in cases where the rebuild branch wouldn't have used the cached `aggregationColumnIdsLastTime` anyway.

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
sdwvit: extracted `this.aggregationColumnIdsLastTime` into a const, ty

sdwvit: Re _areEqual: The two checks aren't redundant — they're complementary. _areEqual is order-sensitive, so !_areEqual means "differs in order or content". every(id => current.includes(id)) (paired with the lengt check above) means "same set". The pair !_areEqual && every(includes) is specifically "same set, different order" = reorder, which is what aggregationColumnsReordered needs to flag. Dropping every would also flag content swaps like [A,B,C] → [A,B,D] as reorders.

sdwvit: Re: using Set. Fair if it was a hot path, but this branch only runs on the rebuild path, and n is typically <10. The Set allocation likely costs more than it saves at realistic sizes, so I'd rather keep the array scan.

PR #13648, opened 2026-04-29
AG-17220 fix inner header component destroyed on every column def update

StephenCooper:
## Problem

`HeaderComp.refresh()` compared the current template against the result of `workOutTemplate()`, but that function returns a **new** `ElementParams` object on every call. The reference was never equal, so `refresh()` always returned `false`, triggering a full destroy + remount of the header cell (and its `innerHeaderComponent`) on every `setColumnDefs` call.

A second, related issue: when `innerHeaderComponent.refresh()` returns `false` (signalling it cannot handle the update), the return value was silently ignored — the inner component was left stale. Worse, the outer `HeaderComp` would have been remounted unnecessarily if it returned false because of the template comparison bug above.

## Fix

- Replace the single template-reference comparison with three explicit boolean checks: custom template string, formula-active state, and sort-indicator presence. These are exactly the conditions `workOutTemplate()` inspects, so the equality check is now direct and allocation-free.
- When `innerHeaderComponent.refresh()` returns `false`, only the inner component is replaced (`replaceInnerHeaderComponent`) — the outer `HeaderComp` stays alive.
- Reorder `replaceInnerHeaderComponent` to detach the old GUI from the DOM before calling `destroyBean`, so the element is not briefly orphaned in `eText`.

## Tests

Two new behavioural test files:

- `columns-header-inner.test.ts` — vanilla-JS grid, verifies both scenarios (refresh not remount; inner component remount when refresh returns false, outer header unaffected).
- `columns-header-inner.test.tsx` — React grid, same two scenarios using `useEffect` mount/unmount counters.

Fix [AG-17220](https://ag-grid.atlassian.net/browse/AG-17220)

### DIFF
--- a/packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
+++ b/packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
@@ -1,5 +1,5 @@
 import { RefPlaceholder } from '../../../agStack/interfaces/agComponent';
-import { _removeFromParent, _setDisplayed } from '../../../agStack/utils/dom';
+import { _clearElement, _removeFromParent, _setDisplayed } from '../../../agStack/utils/dom';
 import { _toString } from '../../../agStack/utils/string';
 import { _getInnerHeaderCompDetails } from '../../../components/framework/userCompUtils';
 import type { AgColumn } from '../../../entities/agColumn';
@@ -86,6 +86,7 @@ export class HeaderComp extends Component implements IHeaderComp {
     private innerHeaderComponent: IInnerHeaderComponent | undefined;
     private currentInnerHeaderComponent: IHeaderParams['innerHeaderComponent'];
     private isLoadingInnerComponent: boolean = false;
+    private innerComponentGeneration: number = 0;
 
     private mouseListener?: HeaderCellMouseListenerFeature;
 
@@ -186,6 +187,7 @@ export class HeaderComp extends Component implements IHeaderComp {
         }
 
         this.isLoadingInnerComponent = true;
+        const generation = ++this.innerComponentGeneration;
 
         userCompDetails.newAgStackInstance().then((comp) => {
             this.isLoadingInnerComponent = false;
@@ -194,8 +196,9 @@ export class HeaderComp extends Component implements IHeaderComp {
                 return;
             }
 
-            if (this.isAlive()) {
+            if (this.isAlive() && generation === this.innerComponentGeneration) {
                 this.innerHeaderComponent = comp;
+                _clearElement(this.eText);
                 this.eText?.appendChild(comp.getGui());
             } else {
                 this.destroyBean(comp);
--- a/packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
+++ b/packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
@@ -2,7 +2,6 @@ import { RefPlaceholder } from '../../../agStack/interfaces/agComponent';
 import { _removeFromParent, _setDisplayed } from '../../../agStack/utils/dom';
 import { _toString } from '../../../agStack/utils/string';
 import { _getInnerHeaderCompDetails } from '../../../components/framework/userCompUtils';
-import type { UserComponentFactory } from '../../../components/framework/userComponentFactory';
 import type { AgColumn } from '../../../entities/agColumn';
 import { _isLegacyMenuEnabled } from '../../../gridOptionsUtils';
 import type { IHeaderComp, IHeaderParams, IInnerHeaderComponent } from '../../../interfaces/iHeader';
@@ -79,7 +78,9 @@ export class HeaderComp extends Component implements IHeaderComp {
     public params: IHeaderParams;
 
     private currentDisplayName: string;
-    private currentTemplate: ElementParams | string | null | undefined;
+    private currentTemplateCustom: string | undefined;
+    private currentFormulaActive: boolean;
+    private currentHasSortIndicator: boolean;
     private currentShowMenu: boolean;
     private currentSuppressMenuHide: boolean;
     private currentSort: boolean | undefined;
@@ -91,19 +92,29 @@ export class HeaderComp extends Component implements IHeaderComp {
     private mouseListener?: HeaderCellMouseListenerFeature;
 
     public refresh(params: IHeaderParams): boolean {
-        const oldParams = this.params;
+        const {
+            beans: { formula, sortSvc },
+            currentTemplateCustom,
+            currentFormulaActive,
+            currentShowMenu,
+            currentHasSortIndicator,
+            currentSort,
+            currentSuppressMenuHide,
+            params: oldParams,
+        } = this;
         this.params = params;
 
-        // if template changed, then recreate the whole comp, the code required to manage
-        // a changing template is to difficult for what it's worth.
+        // if template changed, then recreate the whole comp
         if (
-            this.workOutTemplate(params, !!this.beans?.sortSvc) != this.currentTemplate ||
-            this.workOutShowMenu() != this.currentShowMenu ||
-            params.enableSorting != this.currentSort ||
-            (params.column as AgColumn).formulaRef != this.currentRef ||
-            (this.currentSuppressMenuHide != null && this.shouldSuppressMenuHide() != this.currentSuppressMenuHide) ||
+            !!formula?.active !== currentFormulaActive ||
+            !!sortSvc !== currentHasSortIndicator ||
+            params.enableSorting != currentSort ||
             oldParams.enableFilterButton != params.enableFilterButton ||
-            oldParams.enableFilterIcon != params.enableFilterIcon
+            oldParams.enableFilterIcon != params.enableFilterIcon ||
+            (params.column as AgColumn).formulaRef != this.currentRef ||
+            (params.template?.trim?.() ?? params.template) !== currentTemplateCustom ||
+            this.workOutShowMenu() != currentShowMenu ||
+            (currentSuppressMenuHide != null && this.shouldSuppressMenuHide() != currentSuppressMenuHide)
         ) {
             return false;
         }
@@ -112,7 +123,9 @@ export class HeaderComp extends Component implements IHeaderComp {
             // Mimic the merging of params that happens during init of _getInnerHeaderCompDetails(userCompFactory, params, params);
             const mergedParams = { ...params };
             _mergeDeep(mergedParams, params.innerHeaderComponentParams);
-            this.innerHeaderComponent.refresh?.(mergedParams);
+            if (this.innerHeaderComponent.refresh?.(mergedParams) === false) {
+                this.replaceInnerHeaderComponent(params);
+            }
         } else {
             this.setDisplayName(params);
         }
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -112,7 +123,9 @@ export class HeaderComp extends Component implements IHeaderComp {
             // Mimic the merging of params that happens during init of _getInnerHeaderCompDetails(userCompFactory, params, params);
             const mergedParams = { ...params };
```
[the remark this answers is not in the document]
StephenCooper: Addressed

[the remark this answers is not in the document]
StephenCooper: addressed.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -79,31 +78,44 @@ export class HeaderComp extends Component implements IHeaderComp {
...
     public refresh(params: IHeaderParams): boolean {
-        const oldParams = this.params;
+        const {
+            beans: { formula, sortSvc },
+            currentTemplateCustom,
+            currentFormulaActive,
+            currentShowMenu,
+            currentHasSortIndicator,
+            currentSort,
+            currentSuppressMenuHide,
```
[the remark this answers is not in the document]
StephenCooper: Addressed even though it was a non issue as the selector is hardcoded in that method.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -112,7 +124,12 @@ export class HeaderComp extends Component implements IHeaderComp {
             // Mimic the merging of params that happens during init of _getInnerHeaderCompDetails(userCompFactory, params, params);
             const mergedParams = { ...params };
             _mergeDeep(mergedParams, params.innerHeaderComponentParams);
-            this.innerHeaderComponent.refresh?.(mergedParams);
+            if (
+                params.innerHeaderComponent !== this.currentInnerHeaderComponent ||
```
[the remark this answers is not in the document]
StephenCooper: Fixed

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -112,7 +124,12 @@ export class HeaderComp extends Component implements IHeaderComp {
...
+                !this.innerHeaderComponent.refresh?.(mergedParams)
```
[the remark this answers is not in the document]
StephenCooper: While technically correct I think this is being over defensive and not worth the added complexity and code bloat.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -79,31 +78,44 @@ export class HeaderComp extends Component implements IHeaderComp {
...
+            !!formula?.active !== currentFormulaActive ||
+            !!sortSvc?.getSortIndicatorSelector() !== currentHasSortIndicator ||
+            params.enableSorting != currentSort ||
             oldParams.enableFilterButton != params.enableFilterButton ||
-            oldParams.enableFilterIcon != params.enableFilterIcon
+            oldParams.enableFilterIcon != params.enableFilterIcon ||
+            (params.column as AgColumn).formulaRef != this.currentRef ||
+            (params.template?.trim?.() ?? params.template) !== currentTemplateCustom ||
+            this.workOutShowMenu() != currentShowMenu ||
+            (currentSuppressMenuHide != null && this.shouldSuppressMenuHide() != currentSuppressMenuHide)
```
sdwvit: can you please simplify this condition, also a lot of implicit casting with `!=` may benefit from narrowing

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
sdwvit: does it have a side effect? if so pls extract it into a variable

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -133,10 +150,13 @@ export class HeaderComp extends Component implements IHeaderComp {
     public init(params: IHeaderParams): void {
         this.params = params;
 
-        const { sortSvc, touchSvc, rowNumbersSvc, userCompFactory } = this.beans;
+        const { sortSvc, touchSvc, rowNumbersSvc, formula } = this.beans;
         const sortComp = sortSvc?.getSortIndicatorSelector();
-        this.currentTemplate = this.workOutTemplate(params, !!sortComp);
-        this.setTemplate(this.currentTemplate, sortComp ? [sortComp] : undefined);
+        const template = this.workOutTemplate(params, !!sortComp);
+        this.currentTemplateCustom = params.template?.trim?.() ?? params.template;
```
sdwvit:
```suggestion
        this.currentTemplateCustom = params.template?.trim?.();
```
should be enough?

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -133,10 +150,13 @@ export class HeaderComp extends Component implements IHeaderComp {
...
+        this.currentFormulaActive = !!formula?.active;
+        this.currentHasSortIndicator = !!sortComp;
+        this.setTemplate(template, sortComp ? [sortComp] : undefined);
```
sdwvit, line 160: why there needs to be an array?

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -79,31 +78,44 @@ export class HeaderComp extends Component implements IHeaderComp {
...
+    private currentInnerHeaderComponent: IHeaderParams['innerHeaderComponent'];
     private isLoadingInnerComponent: boolean = false;
 
     private mouseListener?: HeaderCellMouseListenerFeature;
 
     public refresh(params: IHeaderParams): boolean {
-        const oldParams = this.params;
+        const {
+            beans: { formula, sortSvc },
+            currentTemplateCustom,
```
sdwvit: more an arch question, but why do we need to keep a 'current' state in this comp?

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -174,15 +196,25 @@ export class HeaderComp extends Component implements IHeaderComp {
 
             if (this.isAlive()) {
                 this.innerHeaderComponent = comp;
-                if (this.eText) {
-                    this.eText.appendChild(comp.getGui());
-                }
+                this.eText?.appendChild(comp.getGui());
```
sdwvit, line 202: what happens if eText is null

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -174,15 +196,25 @@ export class HeaderComp extends Component implements IHeaderComp {
...
             } else {
                 this.destroyBean(comp);
             }
         });
     }
 
+    private replaceInnerHeaderComponent(params: IHeaderParams): void {
+        const oldGui = this.innerHeaderComponent!.getGui();
```
sdwvit: either the type needs refinement or ! assertion is not needed here

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
SalvatorePreviti: why do we need trim though?

SalvatorePreviti (approved):

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: To extract I would need to have a nested if statement which I don't think would work out as clean. We don't want to call refresh on the old component if it is different to the new inner component.

However, this has highlighted a missing test case. A column that does not have an innerHeaderComponent but then is updated to have one.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: I will simplify to just store params.template for tracking changes. The trim is only needed when we go to render it.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: Its what setTemplate expects.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: The same behaviour as currently. I have just inlined the optional chaining.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: Will be fixed as part of the missing test case.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: Previous implementation comment is that it was more effort to perform the selective updates then just rebuild when any part changes.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
sdwvit: if eText is null appendChild would resolve into an undefined and throw in runtime, I was referring to this :)

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: Have simplified and pulled into its own method.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -79,13 +78,13 @@ export class HeaderComp extends Component implements IHeaderComp {
     public params: IHeaderParams;
 
     private currentDisplayName: string;
-    private currentTemplate: ElementParams | string | null | undefined;
+    private currentFormulaActive: boolean | undefined;
     private currentShowMenu: boolean;
     private currentSuppressMenuHide: boolean;
-    private currentSort: boolean | undefined;
```
StephenCooper, line 85: This cannot change as it is recording the inclusion of the sortService. Modules cannot dynamically change

SalvatorePreviti (approved):

testing/behavioural/src/columns/headers/columns-header-inner.test.ts
```diff
@@ -0,0 +1,549 @@
...
+        (TrackedComp.prototype as any).refresh = function (_params: IHeaderParams): boolean {
+            TrackedComp.refreshCount++;
+            return true;
+        };
+    } else if (refreshBehaviour === 'reject') {
+        (TrackedComp.prototype as any).refresh = function (_params: IHeaderParams): boolean {
+            TrackedComp.refreshCount++;
+            return false;
+        };
+    }
```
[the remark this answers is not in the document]
SalvatorePreviti: mhh yes this is a bit weird, there is not a better way? fine with me though but maybe there is an easier / cleaner way

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
```diff
@@ -174,15 +196,27 @@ export class HeaderComp extends Component implements IHeaderComp {
...
             }
         });
     }
 
+    private replaceInnerHeaderComponent(params: IHeaderParams): void {
+        const curr = this.innerHeaderComponent;
+        if (curr) {
+            _removeFromParent(curr.getGui());
+            this.destroyBean(curr);
+        }
```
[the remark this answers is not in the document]
SalvatorePreviti: to verify

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
StephenCooper: This is not a real issue. If eText is null the full statement is skipped. appendChild is always present if eText is defined. That would be when the second one is required.

packages/ag-grid-community/src/headerRendering/cells/column/headerComp.ts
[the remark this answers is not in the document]
StephenCooper: Was correct, extended tests to cover this edge case.

testing/behavioural/src/columns/headers/columns-header-inner.test.ts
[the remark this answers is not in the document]
StephenCooper: Cleaned up

StephenCooper:
I am going to close this PR for now as the scope has grown too much for the severity of the original issue.

PR #13663, opened 2026-05-01
[AG-17073] Feature(toolbar): update related documentation to include examples and references

sdwvit:
- `filter-quick/index.mdoc`: added **Quick Access Toolbar** section above **Server-Side Data**, noting it as the recommended Enterprise approach.
- `find/index.mdoc`: added **Quick Access Toolbar** section after **Customising Find**.
- `grouping-group-panel/index.mdoc`: added **Row Group Panel in the Toolbar** section after **Side Bar**; removed trailing note with incorrect `rowGroupPanel` identifier.
- `pivoting/index.mdoc`: added **Pivot Panel in the Toolbar** section above **Pivot Mode vs Pivot Active**; removed incorrect **See Also** entry.

Fix [AG-17073](https://ag-grid.atlassian.net/browse/AG-17073)


documentation/ag-grid-docs/src/content/docs/pivoting/index.mdoc
```diff
@@ -87,6 +87,18 @@ const gridOptions = {
 };
 ```
 
+## Pivot Panel in the Toolbar
+
```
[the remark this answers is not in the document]
sdwvit: which is fine, pivoting and grouping are enterprise features

SalvatorePreviti (approved):

community-modules/styles/src/internal/base/parts/_toolbar.scss
```diff
@@ -115,6 +115,11 @@
         padding: 0;
     }
 
+    /* stylelint-disable-next-line selector-max-specificity */
```
sdwvit: I had to do this, as group panel already has :where selector, and I can't easily override it using :where

documentation/ag-grid-docs/src/content/docs/filter-quick/index.mdoc
```diff
@@ -2,14 +2,30 @@
 title: "Quick Filter"
 ---
 
-Quick Filter is a piece of text given to the grid (typically the user will type it in somewhere in your application) that is used to filter rows by comparing against the data in all columns. This can be used in addition to column-specific filtering.
+Quick Filter is a piece of text given to the grid that is used to filter rows by comparing against the data in all columns. It can be used in addition to column-specific filtering.
 
-## Setting the Quick Filter
+The filter text is split into words, and each word is compared against every column value (case insensitive). All words must match a row for it to be included. For example, searching "Tony Ireland" will only show rows containing both "Tony" and "Ireland".
```
StephenCooper: I feel like this is the wrong place for this additional detail. We should combine it with the Checking the Quick Filter and Overriding the Quick Filter Value. These could be sub sections under a Quick Filter Value heading. This would make sense for the above snippet to be there too.

documentation/ag-grid-docs/src/content/docs/filter-quick/index.mdoc
```diff
@@ -20,16 +36,16 @@ const gridOptions = {
...
 ```
 
-You can also set the initial Quick Filter text via the `quickFilterText` grid option.
+The initial Quick Filter text can also be set via the `quickFilterText` grid option.
 {% /if %}
 
-The provided Quick Filter text will be split into words, and each word will be compared against the row. A word matches the row if the string value of any of the columns contains the word (the check is case insensitive). All words must match the row for it to be included. For example, if the text is "Tony Ireland", the Quick Filter will only include rows with both "Tony" AND "Ireland" in them.
+{% gridExampleRunner title="Quick Filter" name="quick-filter" /%}
 
 ## Checking the Quick Filter
```
StephenCooper: This section needs to be updated to cover the toolbar case too.

documentation/ag-grid-docs/src/content/docs/find/_examples/customising-find/provided/reactFunctionalTs/index.tsx
```diff
@@ -128,66 +105,57 @@ const GridExample = () => {
...
+                        pinnedTopRowData={pinnedTopRowData}
+                        pinnedBottomRowData={pinnedBottomRowData}
+                        columnDefs={columnDefs}
+                        defaultColDef={defaultColDef}
+                        pagination={true}
+                        paginationPageSize={5}
+                        paginationPageSizeSelector={paginationPageSizeSelector}
+                        toolbar={toolbar}
+                        findOptions={findOptions}
+                        modules={modules}
```
StephenCooper: Please use the AgGridProvider still for module registration.

StephenCooper (approved):

PR #13695, opened 2026-05-07
[WIP] React max depth repro

StephenCooper:


packages/ag-grid-react/src/reactUi/cellRenderer/groupCellRenderer.tsx
```diff
@@ -15,11 +14,17 @@ import type {
...
     UserCompDetails,
 } from 'ag-grid-community';
-import { _toString } from 'ag-grid-community';
+import { CssClassManager, _toString } from 'ag-grid-community';
 
 import { BeansContext } from '../beansContext';
 import { showJsComp } from '../jsComp';
-import { CssClasses } from '../utils';
+
+type CssManagers = {
```
StephenCooper, line 22: Using CssClassManagers to update classes to avoid the need to trigger React re-renders when we are just changing class names. Similar to the cellComp.

StephenCooper:
Bug is resolved much more simply at a higher level https://github.com/ag-grid/ag-grid/pull/13708

PR #13683, opened 2026-05-06
[AG-17161] Feature(toolbar): design followups 3

sdwvit:
https://ag-grid.atlassian.net/browse/AG-17161

## Summary

- **Theming API — Material light/dark**: underlined toolbar inputs and find wrapper (no top/left/right borders), primary-colour focus border, 0.2s button hover transition, 6% foreground bg on hover
- **Theming API — Balham**: fixed icon-field padding formula (`iconSize + spacing × 2`) so it works with Balham's compact `spacing: 4`; shrunk Balham search icon viewBox (`-4 -4 32 32`) to match 32×32 visual weight of other Balham icons
- **Theming API — Alpine**: accent colour applied to toolbar button text and icon on hover (both via `ag-toolbar-button` and `ag-icon` overrides)
- **Legacy base**: added CSS variable shims in `_base-variables.scss` mapping new Theming API var names (`--ag-spacing`, `--ag-input-border`, `--ag-focus-shadow`, etc.) to legacy equivalents so `_toolbar.scss` works in all legacy themes; fixed icon-field padding formula; raised specificity of icon-padding and find-input border/shadow overrides with `input[class^='ag-'][type='text'].ag-toolbar-input-field` selectors to beat theme `text-input` padding rules (0-5-1 vs 0-4-1)
- **Legacy — Material**: underlined find/input wrappers, primary-colour focus border, 0.5s button hover transition, explicit find-wrapper height to align bottom borders; light hover bg ~6% foreground, dark hover bg `#000000` matching header-cell hover
- **Legacy — Alpine**: icon button hover colour → `--ag-alpine-active-color`
- **Legacy — Quartz**: icon button hover bg → 10% foreground overlay
- **Button hover propagation** (both Theming API and legacy): added `.ag-toolbar-button-wrapper:hover .ag-toolbar-button, .ag-toolbar-button-wrapper:hover .ag-toolbar-button .ag-icon` rules so hover colour overrides the explicit `color` on `.ag-toolbar-button` and `color` on `.ag-icon`
- **Focus shadow**: added `:focus:not(:focus-visible) { box-shadow: none }` to suppress shadow on click in both CSS files

Fix [AG-17161](https://ag-grid.atlassian.net/browse/AG-17161)

image
Screenshot 2026-05-06 at 16 08 56
screencapture-localhost-4610-javascript-data-grid-toolbar-2026-05-07-10_18_43
Screenshot 2026-05-07 at 10 30 40

PR #13702, opened 2026-05-08
[AG-17294] Fix(toolbar): legacy Alpine font weight and button hover

sdwvit:
- Fix legacy Alpine toolbar font weight: use direct `font-weight: 700` on `.ag-toolbar` in `_index.scss` (same pattern as `.ag-header-row`) instead of declaring `--ag-header-font-weight` variable
- Restore `--ag-icon-button-hover-color` on toolbar button hover (was hardcoded to `--ag-foreground-color`, losing the blue active-colour hover in Alpine)
- Restore `padding-left: var(--ag-cell-horizontal-padding)` to `.ag-column-drop-horizontal` in legacy themes — removed globally in #13683, breaking the standalone group panel drop zone; inside toolbar `.ag-toolbar-panel .ag-column-drop-horizontal` keeps `padding: 0` (spacing comes from `.ag-toolbar-item` margin instead)
- Remove own horizontal padding from `.ag-toolbar-panel` in legacy themes — double-spacing with the drop zone's own padding and the item margin

Fix [AG-17294](https://ag-grid.atlassian.net/browse/AG-17294)

### DIFF
--- a/community-modules/styles/src/internal/themes/alpine/_alpine-variables.scss
+++ b/community-modules/styles/src/internal/themes/alpine/_alpine-variables.scss
@@ -97,6 +97,7 @@
         -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue',
         sans-serif;
     --ag-font-size: 13px;
+    --ag-header-font-weight: 700;
     --ag-icon-font-family: agGridAlpine;
 
     // MISC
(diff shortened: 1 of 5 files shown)


community-modules/styles/src/internal/themes/alpine/_alpine-variables.scss
```diff
@@ -97,6 +97,7 @@
         -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue',
         sans-serif;
     --ag-font-size: 13px;
+    --ag-header-font-weight: 700;
```
StephenCooper: Where is this coming from? I don't see that variable being set to a concrete value for any other theme?

sdwvit: removed, it's not needed.

StephenCooper (approved):

PR #13717, opened 2026-05-11
RTI-3327 fix-tree-list-aria-warning-on-click

SalvatorePreviti:
Clicking the set-filter tree-group checkbox focused an input inside an aria-hidden wrapper; fixed by suppressing mousedown focus on that input, which is already tabindex="-1" so click/toggle and keyboard nav are unaffected.

Also, reuse (and export from main-internal) _preventEventDefault 

FIX RTI-3327

### DIFF
--- a/packages/ag-grid-enterprise/src/setFilter/setFilterListItem.ts
+++ b/packages/ag-grid-enterprise/src/setFilter/setFilterListItem.ts
@@ -23,6 +23,7 @@ import {
     _getCellRendererDetails,
     _getShouldDisplayTooltip,
     _isShowTooltipWhenTruncated,
+    _preventEventDefault,
     _setAriaChecked,
     _setAriaDescribedBy,
     _setAriaExpanded,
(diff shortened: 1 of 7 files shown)


packages/ag-grid-enterprise/src/setFilter/setFilterListItem.ts
```diff
@@ -167,12 +168,17 @@ export class SetFilterListItem<V> extends Component<SetFilterListItemEvent> {
...
             .setValue(this.isSelected, true)
             .setDisabled(!!this.params.readOnly)
-            .getInputElement()
-            .setAttribute('tabindex', '-1');
+            .getInputElement();
+        eInput.setAttribute('tabindex', '-1');
+        this.addManagedElementListeners(eInput, {
+            // Focus belongs on the outer treeitem wrapper; for aria-hidden tree groups, focusing
+            // the inner input triggers a Chromium aria-hidden warning. The click still toggles.
+            mousedown: _preventEventDefault,
```
gportela85, line 180: I'm not too sure about preventing default on the input element here. This can create some side effects that we are not anticipating and possible regressions. It might be better to push this to post release.

PR #13726, opened 2026-05-11
[RTI-3312] Fix(toolbar): row group panel and pivot panel placeholder text appears bolder than existing rgp and pp panels

sdwvit:
Fix toolbar panel placeholder text appearing bolder than row group and pivot panels

Scope font-weight to non-panel toolbar items only, so column drop panels inherit normal weight instead of the header font-weight.

Fix [RTI-3312](https://ag-grid.atlassian.net/browse/RTI-3312)

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_toolbar.scss
+++ b/community-modules/styles/src/internal/base/parts/_toolbar.scss
@@ -11,7 +11,6 @@
         color: var(--ag-toolbar-text-color);
         font-family: var(--ag-header-font-family);
         font-size: var(--ag-header-font-size);
-        font-weight: var(--ag-header-font-weight);
         white-space: nowrap;
     }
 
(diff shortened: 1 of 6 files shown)


BernieSumption (approved):

community-modules/styles/src/internal/base/parts/_toolbar.scss
```diff
@@ -24,6 +23,11 @@
         margin: 0 calc(var(--ag-grid-size, 8px) * 2);
     }
 
+    // stylelint-disable-next-line selector-max-specificity
+    .ag-toolbar-item:not(.ag-toolbar-panel) {
```
BernieSumption: Use `.ag-toolbar-item:where(:not(.ag-toolbar-panel))`

BernieSumption (approved):
Approved once the lint disable comments are removed

PR #13768, opened 2026-05-14
AG-17362 Fix checkbox test IDs disappearing in column chooser on scroll

StephenCooper:
## Summary

- `setupAllTestIds()` was only triggered by grid-level events (`sideBarUpdated`, `columnMenuVisibleChanged`, etc.), so when the column chooser virtual list scrolled and created new DOM elements via virtualization, they never received test IDs.
- Added a capture-phase `scroll` listener on the root node that detects scrolling inside column panels (`.ag-column-panel`) or chooser popups (`.ag-panel`) and re-applies test IDs via a 100ms debounce.
- Added a behavioural test that verifies checkbox test IDs persist after scrolling down and back up in the column tool panel virtual list.

## Test plan

- [x] New behavioural test `column-chooser-test-ids.test.ts` passes with the fix and fails without it
- [ ] Manual verification: open column chooser with many columns, scroll down/up, confirm `data-testid` attributes remain on checkbox elements

### DIFF
--- a/packages/ag-grid-community/src/testing/testIdService.ts
+++ b/packages/ag-grid-community/src/testing/testIdService.ts
@@ -72,7 +72,7 @@ export class TestIdService extends BeanStub implements NamedBean, ITestIdService
         };
         root.addEventListener('scroll', onScroll, { capture: true });
         this.addDestroyFunc(() =>
-            root.removeEventListener('scroll', onScroll, { capture: true } as EventListenerOptions)
+            root.removeEventListener('scroll', onScroll, { capture: true })
         );
     }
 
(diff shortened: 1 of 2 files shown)


Stephanemw (approved):

SalvatorePreviti (approved):

packages/ag-grid-community/src/testing/testIdService.ts
```diff
@@ -55,6 +55,25 @@ export class TestIdService extends BeanStub implements NamedBean, ITestIdService
...
+                target instanceof HTMLElement &&
+                target.classList.contains('ag-virtual-list-viewport') &&
+                (target.closest('.ag-column-panel') || target.closest('.ag-panel'))
+            ) {
+                onVirtualListScroll();
+            }
+        };
+        root.addEventListener('scroll', onScroll, { capture: true });
+        this.addDestroyFunc(() =>
+            root.removeEventListener('scroll', onScroll, { capture: true } as EventListenerOptions)
```
StephenCooper:
```suggestion
            root.removeEventListener('scroll', onScroll, { capture: true })
```

sdwvit (approved):

PR #13769, opened 2026-05-14
AG-17363 fix SSRM infinite loading when expanding empty group with groupTotalRow

StephenCooper:
Fix [AG-17363](https://ag-grid.atlassian.net/browse/AG-17363)

- Fixes infinite server requests when expanding a node marked `group: true` that returns empty children in SSRM Tree Data mode with `groupTotalRow: 'bottom'`
- Root cause: `isDisplayIndexInStore` used `cache.getRowCount()` (data rows only), so a child store with 0 data rows but a footer row would disown its display index — the parent cache then created a spurious stub for that index, triggering a load→empty→footer→stub loop
- Fix: use `getRowCount()` which includes the footer in its count, so the store correctly claims ownership of the footer's display index

## Test plan

- [x] New behavioural test: empty group + `groupTotalRow: 'bottom'` — verifies no infinite requests
- [x] New behavioural test: empty group without `groupTotalRow` — verifies existing behaviour preserved
- [x] New behavioural test: empty group + `groupTotalRow: 'bottom'` + `groupHideOpenParents: true` — covers the DFS lookup path (`lazyCache.ts:181`)
- [x] 749/749 SSRM tests pass
- [x] 272/272 tree-data tests pass
- [x] 2329/2329 grouping tests pass

### DIFF
--- a/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyStore.ts
+++ b/packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyStore.ts
@@ -274,10 +274,11 @@ export class LazyStore extends BeanStub implements IServerSideStore {
      * @returns the virtual size of this store
      */
     getRowCount(): number {
+        const rowCount = this.cache.getRowCount();
         if (this.parentRowNode.sibling) {
-            return this.cache.getRowCount() + 1;
+            return rowCount + 1;
         }
-        return this.cache.getRowCount();
+        return rowCount;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyStore.ts
```diff
@@ -294,7 +294,7 @@ export class LazyStore extends BeanStub implements IServerSideStore {
      * @returns whether or not the row exists within this store
      */
     isDisplayIndexInStore(displayIndex: number): boolean {
-        if (this.cache.getRowCount() === 0) {
+        if (this.getRowCount() === 0) {
```
sdwvit, line 298: this still calls cache.getRowCount, why is this a fix?

StephenCooper: Because it also handles the sibling case and adds 1. This is the key difference in this case. When the groupTotalRow is provided we need this to return 1, so that the grid does not keep trying to load the data again even though it has loaded but just got an empty [].

```
    getRowCount(): number {
        if (this.parentRowNode.sibling) {
            return this.cache.getRowCount() + 1;
        }
        return this.cache.getRowCount();
    }
```

sdwvit (approved):

PR #13786, opened 2026-05-18
AG-17382 Improve example grid Monthly Breakdown column group

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-17382

Improvements to the Monthly Breakdown column group in the example grid:

- Add an always-visible **Average** column showing the mean of all monthly values, positioned at the end of the group
- Extract shared month column config into a `monthColBase` object to reduce duplication between month columns and the average column
- Add `minWidth` to month columns, sparkline column, and average column
- Use `processPivotResultColDef` to enforce a minimum column width of 200px on pivot-generated columns
- Add a toolbar separator between the row group panel and pivot panel items

Fix [AG-17382](https://ag-grid.atlassian.net/browse/AG-17382)

### DIFF
--- a/documentation/ag-grid-docs/src/components/example-grid/Example.tsx
+++ b/documentation/ag-grid-docs/src/components/example-grid/Example.tsx
@@ -221,6 +221,7 @@ const ExampleInner = ({
                     ? []
                     : [
                           { toolbarItem: 'agRowGroupPanelToolbarItem', alignment: 'left' as const },
+                          { toolbarItem: 'separator', alignment: 'left' as const },
                           { toolbarItem: 'agPivotPanelToolbarItem', alignment: 'left' as const },
                       ]),
                 'agQuickFilterToolbarItem',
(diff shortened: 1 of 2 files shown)


StephenCooper:
## Changes in this PR

**Monthly Breakdown column group (`colDefs.ts`)**
- Added an always-visible **Average** column at the end of the group showing the mean of all 12 monthly values (`aggFunc: 'avg'`, `valueGetter` computing the average client-side)
- Added `minWidth` to the Winning Trends sparkline column, all month columns, and the Average column
- Extracted shared month column config into a `monthColBase` object — month columns and the Average column now both spread this base, removing the duplicated `cellClassRules`, `cellDataType`, `filter`, and `filterParams`

**Pivot mode fix (`Example.tsx`)**
- `minWidth` on a `ColGroupDef` is not a valid property (it lives only on `ColDef`) so the group-level constraint was silently ignored
- Added `processPivotResultColDef` to `staticGridOptions` to enforce `minWidth: 200` and an initial `width` of at least 200px on all pivot-generated columns, ensuring the aggregation header label has enough room

**Toolbar (`Example.tsx`)**
- Added a `separator` toolbar item between `agRowGroupPanelToolbarItem` and `agPivotPanelToolbarItem`

gportela85 (approved):

PR #13797, opened 2026-05-18
[PATCH] AG-17384 duplicate grand total row bug

SalvatorePreviti:
Fixes the manual pinned row model not properly handling creation and destroying of pinned rows in SSRM

- for safety and correctness, added more properties to ignore to the sibling, destroyed should not be copied, same for pinned rows when creating a sibling
- fixed pinRow in pinned row model that was executing on _destroy also for the clone node that was causing the issue
- fixes row node destruction handling correctly unpinning
- add tests to increase coverage and this specific case


StephenCooper:
If we patch this it will need to be merged into the b35.3.1 branch not the current release branch of b35.3.0. So I have converted this to a Draft until that is generated.

seanlandsman (approved):

PR #13803, opened 2026-05-19
[AG-17270] Fix(toolbar): pin built in actions not working

sdwvit:
Fall back to iterating over the current range selection for pinTop, pinBottom, and unpinRow actions when triggered from the toolbar (where no row node is provided).

Fix [AG-17270](https://ag-grid.atlassian.net/browse/AG-17270)

### DIFF
--- a/packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
+++ b/packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
@@ -5,6 +5,7 @@ import type {
     GetNoteParams,
     IAggFuncService,
     IColsService,
+    IMenuActionParams,
     INoteAccess,
     INotesService,
     LocaleTextFunc,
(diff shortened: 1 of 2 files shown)


packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -190,26 +203,41 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         ? {
                               name: localeTextFunc('pinTop', 'Pin to Top'),
                               icon: _createIconNoSpan('rowPinTop', beans, column),
-                              action: ({ node, column }) =>
-                                  node && pinnedRowModel.pinRow(node as RowNode, 'top', column as AgColumn | null),
+                              action: ({ node, column }) => {
```
gportela85: looks like this conditional is basically the same for all three calls

```ts
if (node) {
    pinnedRowModel.pinRow(node as RowNode, 'top', column as AgColumn | null);
} else {
    forEachRowInSelection((node) => pinnedRowModel.pinRow(node, 'top', null));
}
```

would be better to have a method
```ts
const processPinAction(side: 'top' | 'bottom' | null, column: AgColumn | null) => {
    if (node) {
        pinnedRowModel.pinRow(node as RowNode, side, column);
    } else {
        forEachRowInSelection((node) => pinnedRowModel.pinRow(node, side, null));
    }
}
```

gportela85 (approved):

PR #13809, opened 2026-05-20
[AG-16296] Maintenance(column): filters return date only values in filter model for celldatatype date

sdwvit:
Use `params.colDef.cellDataType` with `getDateIncludesTimeFlag` to determine whether to include the time component when serialising `dateFrom`/`dateTo` in the date filter model. Previously the time was always included, causing date-only values to carry a time suffix for `cellDataType: 'date'` columns.

Fix [AG-16296](https://ag-grid.atlassian.net/browse/AG-16296)

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
+++ b/packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
@@ -289,7 +289,8 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
         const { params, filterType, beans } = this;
         const dataTypeSvc = beans.dataTypeSvc;
         const values = this.getValues(position);
-        const includeDateTime = dataTypeSvc?.getDateIncludesTimeFlag(params.colDef.cellDataType) ?? true;
+        const includeDateTime =
+            params.includeTime ?? dataTypeSvc?.getDateIncludesTimeFlag(params.colDef.cellDataType) ?? true;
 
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
```diff
@@ -286,15 +286,17 @@ export class DateFilter extends SimpleFilter<DateFilterModel, Date, DateCompWrap
     protected createCondition(position: number): DateFilterModel {
         const type = this.getConditionType(position);
         const model: Partial<DateFilterModel> = {};
-        const { params, filterType } = this;
+        const { params, filterType, beans } = this;
+        const dataTypeSvc = beans.dataTypeSvc;
         const values = this.getValues(position);
+        const includeDateTime = dataTypeSvc?.getDateIncludesTimeFlag(params.colDef.cellDataType) ?? true;
```
peterjrreynolds: Shouldn't this also check the `includeTime` filter param?

peterjrreynolds (commented):
Also need to update the JSDoc on `DateFilterModel.dateFrom`

packages/ag-grid-community/src/filter/provided/date/dateFilter.ts
sdwvit: good point, added

peterjrreynolds (approved):

PR #13822, opened 2026-05-21
AG-17366-column-model-rewrite

SalvatorePreviti:
# AG-17366 Column Model Improvements

Faster column lookups and refresh path, less code, cleaner internal architecture. **No public API changes.**

## Behaviour changes & bug fixes vs `latest`

- **Service col instance preservation.** Auto / selection / row-numbers cols keep the same `AgColumn` instance across `autoGroupColumnDef` updates, pivot toggle and rowSelection toggle; wrappers reused via a `(col, depth)` cache. Latest regenerated them.
- **`maintainColumnOrder` + service col regen.** Regenerated service cols keep the user's non-head position instead of snapping back; separate `lastOrder` / `lastPivotOrder` survive pivot toggle round-trips.
- **`getCol` / `getColDefCol` accept `field` keys.** A `field` string or a fresh `{ field: 'name' }` ColDef now resolves to the matching column; latest only matched `colId === field` or by reference on the merged colDef.
- **No bean leaks on teardown.** Every column (primary, hierarchy virtual, service, pivot result, tree data) is destroyed exactly once across grid teardown, pivot toggle and `columnDefs` replacement.
- **Pivot result tree destruction on clear/restore.** Cleared-then-replaced pivot trees release their detached beans.
- **Hierarchy virtuals appear exactly once.** `getColumnState()` / `api.getAllGridColumns()` no longer duplicate hierarchy virtuals.
- **`pinned: true` is treated as left-pinned.** Preserved across the new partitioning helper.

## Performance

Complexity / allocation, `latest` → branch:

- **`getCol` / `getColDefCol` / `getColById`** — **O(N) → O(1).** Latest did sequential scans across `colDefCols`, `cols` and per-service fallbacks; branch uses a unified `colsById` hash and fallback lazy loaded Map.
- **`getCol` slow-path fallback** — **O(N) scan → O(1) Map lookup.** Lazy `colsByDef` Map (built on first slow-path call) indexes `colDef` / `userProvidedColDef` refs and `field` strings when `field !== colId`. Invalidated whenever `colsById` changes.
- **`getAllCols()`** — **O(N) every call → O(N) once, O(1) thereafter.** Cached `Object.values(colsById)`; latest rebuilt from multiple source arrays per call.
- **`extractCols`** — **O(N²) → O(N).** Set-based membership replaces `indexOf` / `includes` walks (hit on every rowGroup / pivot / value extraction).
- **`expandColumnInto` (hierarchy)** — **O(N²) → O(N).** Caller-owned dedup Set is shared across successive calls in the loop.
- **`compareVirtualColumns`** — **O(N²·log N) → O(N·log N).** Map-based position lookup replaces two `indexOf` per comparator call inside `Array.sort`.
- **`columnDefFactory.getColumnDefs`** — **O(N²) → O(N·log N).** Map-based position lookups for sort and rowGroup/pivot priority replace per-col `indexOf`.
- **`placeLockedColumns`** — **always O(N) + 3 partition allocs → O(N) + 0 allocs on the common path.** Returns the input unchanged when no column is lock-positioned.
- **`autoColService.refreshCols`** — **O(N) + per-call bean allocation → O(N) + 0 allocs on the no-op path.** `autoColIdsMatch` fuses the expected-colIds build with the equality compare.
- **`selectionColService.refreshCols` / `rowNumbersService.refreshCols`** — **bean alloc + replace → O(1) short-circuit on `want === have`.**
- **`ColWrapperCache`** — shared utility class (community, own file) instantiated once per service: `ColumnGroupService.serviceWrapperCache` (auto / sel / rn, evicted per `refreshCols`) and `GroupHierarchyColService.wrapperCache` (hierarchy virtuals, evicted per `applyToColDefTree`). **Per-refresh `balanceTreeForAutoCols` alloc → cached `(col, depth)` reuse;** double-destroy from tree-destroy is safe via `isAlive()`.
- **`setColumnGroupState`** — **O(N × K) → O(N + K).** One DFS builds a `groupId → group` map; the K state items then do O(1) lookups.
- **`getProvidedColGroup`** — full-tree DFS → early-exit recursive walk. Returns on first match.
- **`balanceColumnTree`** — `hasGroups` is a per-tree property, hoisted to a single scan instead of recomputed inside the per-column branch.
- **`createColumnGroups`** — caches the run's shared parent (`runParent`) across the inner loop instead of re-resolving `originalParent` of the previous node on every iteration; `originalParentOf` helper folds the `isColumnGroup ? group.providedColumnGroup : col` hop.
- **Direct-field reads** — **getter method calls → direct property access** in hot paths (`col.colId`, `col.visible`, `col.rowGroupActive`, etc.).

## Architecture

- **`colsById` is the single source of truth** for lookups across user, hierarchy, service and pivot result cols. Pivot result colIds are explicitly prefixed.
- **`lastOrder` / `lastPivotOrder` store colIds, not AgColumn refs.** Service-col replacements (same colId, new instance) resolve transparently.
- **`restoreColOrder` + `placeLockedColumns`** behind a single `restoreOrLockColumns` helper. `restoreColOrder` rewritten: colIds-based snapshot, Map-based positions, forward-emit; sibling-finder iterative.
- **`_createColumnTree`** returns `{columnTree, treeDepth, columns}` in one walk; replaces a separate `_getColumnsFromTree` DFS.
- **Hierarchy service** unified `applyToColDefTree(list, tree, treeDepth)` entry returns input refs when nothing changed. Two-phase: plan colIds (no bean alloc) → compare → rebuild on mismatch.
- **Producer pattern for service cols** (rowNumbers / selection / autoGroup) replaces the `IColumnCollectionService` shared base. Each exposes `refreshCols(source)`; ColumnModel reads `.column` / `.columns` directly in dependency order (auto → selection → rowNumbers).
- **Pivot result lifecycle** owned by `PivotResultColsService`: `pivotCols` / `pivotTree` / `pivotTreeDepth` / `savedPivotTree` + `applyPivotResultColDefs` / `clearPivotResultCols` / `destroyTrees`. Adds a clear/restore window so reapplying the same pivot defs reuses the saved tree.
- **`ColWrapperCache`** lives in `columnGroups/colWrapperCache.ts`; owns both `buildWrapper` (the AgProvidedColumnGroup chain construction) and `destroyAutoWrapperChain` (the idempotent teardown). Both consuming services initialise their instance in `postConstruct` so `this.beans` is wired before first use.

## Internal API (no public API impact)

- `IColumnCollectionService` removed; replaced by per-service interfaces (`IAutoColService`, `IRowNumbersService`, `IGroupHierarchyColService`) with only the methods actually called.
- `IAutoColService.createColumns(cols, updateOrders, source) → refreshCols(source)`. Ordering moves to ColumnModel's `lastOrder` snapshot.
- `IRowNumbersService.createColumns(...) → refreshCols()`; drops `updateColumns` / `getColumn`.
- `IGroupHierarchyColService`: `createColumns + splice` collapsed into `applyToColDefTree`; `expandColumnInto` takes a caller-owned dedup Set; `insertVirtualColumnsForCol` returns `AgColumn[] | null`; drops `updateColumns` / `getColumns` / `getColumn`.
- `RowAutoHeightService.setAutoHeightActive` removed — folded into `ColumnModel.computeColSpanAndAutoHeight`.
- `_destroyColIfAlive(col)` added in `columnUtils.ts` (null-safe, `isAlive`-checked).
- `isGroupHierarchyCol(col)` discriminator added.

## Out of scope (follow-up)

Renaming the asymmetric `colDefList` / `colDefTree` ↔ `colsList` / `colsTree` field pair (e.g. `primary*` / `display*`) would touch many call sites across tool panels and SSRM listeners for a clarity win — kept separate. Documented inline in `columnModel.ts`.

Further performance improvements and caching


StephenCooper:
@SalvatorePreviti could you attach the benchmark results please?

packages/ag-grid-community/src/columnMove/columnMoveUtils.ts
```diff
@@ -17,14 +31,36 @@ export function placeLockedColumns(cols: AgColumn[], gos: GridOptionsService): A
...
         }
-    });
+    }
 
     const isRtl = gos.get('enableRtl');
+    const leftLen = left.length;
+    const normalLen = normal.length;
+    const rightLen = right.length;
+    const result = new Array<AgColumn>(leftLen + normalLen + rightLen);
+    let pos = 0;
```
StephenCooper, line 41: Given we think that lockCols are not used as much is this performance gain worth the code bloat for this faster approach?

PR #13835, opened 2026-05-22
AG-13924 Preserve integrated chart series order when restoring with rebuilt cellRange

Stephanemw:
## Summary

- Adds `seriesColOrder?: string[]` to the `ChartModel` interface. `getChartModels()` now populates it with the current user-defined value column order.
- During `restoreChart()`, if `seriesColOrder` is present, it is used to sort `valueColState` during initialisation — independently of `cellRange.columns`, which callers typically rebuild from current grid columns in grid order.
- Updates TC5 to exercise the real customer pattern: snapshot the model, clear the chart, hide a column, rebuild `cellRange.columns` in grid order, then restore and verify the saved relative series order is preserved.

## Motivation

Customers save chart models with `getChartModels()` and later restore them after the available column set has changed. The established pattern is to rebuild `cellRange.columns` from the current visible grid columns before calling `restoreChart()`. Because this rebuild uses grid column order, the saved user-defined series order was lost — causing legend colours and series positions to shift on restore.

`seriesColOrder` is a new optional field the caller does not know to overwrite, so it survives the `cellRange.columns` rebuild and lets the grid recover the original ordering.

## Test plan

- TC1–TC5 all pass (`./docs-e2e.sh "saving-and-restoring"`)
- TC5 now directly tests the customer rebuild pattern (previously tested passive restore only)
- `yarn nx build:types ag-grid-enterprise` passes clean

Fix [AG-13924](https://ag-grid.atlassian.net/browse/AG-13924)

### DIFF
--- a/packages/ag-grid-community/src/interfaces/IChartService.ts
+++ b/packages/ag-grid-community/src/interfaces/IChartService.ts
@@ -77,6 +77,8 @@ export interface ChartModel {
     seriesChartTypes?: SeriesChartType[];
     seriesGroupType?: SeriesGroupType;
     useGroupColumnAsCategory?: boolean;
+    /** Ordered list of value column IDs reflecting the user-defined series order at the time the model was saved. */
+    seriesColOrder?: string[];
 }
 
 export interface IChartService {
--- a/packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
+++ b/packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
@@ -46,6 +46,7 @@ export interface ChartModelParams {
     seriesChartTypes?: SeriesChartType[];
     seriesGroupType?: SeriesGroupType;
     useGroupColumnAsCategory?: boolean;
+    seriesColOrder?: string[];
 }
 
 export const DEFAULT_CHART_CATEGORY = 'AG-GRID-DEFAULT-CATEGORY';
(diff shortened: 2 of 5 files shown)


packages/ag-grid-enterprise/src/charts/chartComp/model/chartDataModel.ts
```diff
@@ -450,6 +451,17 @@ export class ChartDataModel extends BeanStub {
                 cs.order = savedValueOrder.has(cs.colId) ? savedValueOrder.get(cs.colId)! : nextOrder++;
```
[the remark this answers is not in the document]
Stephanemw: False positive — `seriesChartTypes` does not reorder `valueColState`; it reads FROM it.

The execution order in `updateCellRanges()` (called from `postConstruct()`) is:
1. `resetColumnState()` — `seriesColOrder` sort applies here; `valueColState` is correctly ordered before anything else runs
2. `comboChartModel.updateSeriesChartTypes()` — maps `valueColState` to `seriesChartTypes` via `savedCustomSeriesChartTypes.find((s) => s.colId === valueCol.colId)` (lookup by colId, not index)

The combo model inherits the correct order from `valueColState` — it doesn't impose its own. The `updateModel()` path (which calls `comboChartModel.update(seriesChartTypes)`) is the non-init path and is not invoked during `postConstruct()`.

packages/ag-grid-community/src/interfaces/IChartService.ts
```diff
@@ -77,6 +77,8 @@ export interface ChartModel {
     seriesChartTypes?: SeriesChartType[];
     seriesGroupType?: SeriesGroupType;
     useGroupColumnAsCategory?: boolean;
+    /** Ordered list of value column IDs reflecting the user-defined series order at the time the model was saved. */
+    seriesColOrder?: string[];
```
peterjrreynolds, line 81: Can't we rely on the order of the columns in the range? When you reorder the series, the range is updated.

E.g.
1. Open https://plnkr.co/edit/7VW814bd0BCrGwI9?open=main.js
2. Click Save chart
Actual: Console shows `['country', 'sugar', 'fat', 'weight']`
3. Drag Weight to be the first series
4. Click Save chart
Actual: Console shows `['country', 'weight', 'sugar', 'fat']`

Stephanemw: As discussed - this feels unavoidable to support keeping the old category colours/indexes when users add/remove columns into the list at any point.

PR #13839, opened 2026-05-22
AG-17158 AG Grid - www Domain Consolidation: Findings and Actions

seanlandsman:


documentation/ag-grid-docs/src/utils/htaccess/htaccessRules.ts
```diff
@@ -1,6 +1,13 @@
 import { urlWithBaseUrl } from '../urlWithBaseUrl';
 import { SITE_301_REDIRECTS } from './redirects';
 
+
```
taktran: Extra spaces don't seem necessary

taktran (approved):

PR #13840, opened 2026-05-22
AG-3390 - Harden grid staging deploy script

taktran:
## Summary

Ports the hardening applied to the charts staging deploy script in [ag-charts PR #6893](https://github.com/ag-grid/ag-charts/pull/6893) over to grid's equivalent `updateGridStagingRemote.sh`.

The changes:

- Add `set -euo pipefail` so a failed `unzip`/`mv` halts the deploy rather than continuing with stale or missing files.
- Validate the uploaded archive (existence + `unzip -tq`) **before** touching the live site, so a corrupt or missing upload can no longer wipe the existing staging deploy.
- Extract into a temp `${STAGING_DIR}.new.$$` directory and swap into place with `mv`, keeping the previous deploy aside as `${STAGING_DIR}.old` until cleanup. The `branch-builds` subtree is preserved into the new dir before the swap, preserving the existing behaviour.

Companion `createAndDeployDocsToTCGH.sh` is intentionally **out of scope** for this PR — it can be hardened separately if desired.

## Test plan

- [ ] Manual review of the diff against [ag-charts PR #6893](https://github.com/ag-grid/ag-charts/pull/6893/files) for parity.
- [ ] Confirm the deploy succeeds end-to-end on a TC pipeline run against the staging host (verify `branch-builds` survives across deploys, archive validation rejects a corrupt upload, and previous `html` content is fully replaced).

### DIFF
--- a/scripts/deployments/updateGridStagingRemote.sh
+++ b/scripts/deployments/updateGridStagingRemote.sh
@@ -1,14 +1,35 @@
 #!/usr/bin/env bash
 
-echo "Backing up branch builds"
-cp -R @WWW_ROOT_DIR@/html/branch-builds @WWW_ROOT_DIR@/
+set -euo pipefail
 
-echo "Cleaning current grid staging"
-rm -rf @WWW_ROOT_DIR@/html/*
-mv @FILENAME@ @WWW_ROOT_DIR@/html/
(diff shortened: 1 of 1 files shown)


taktran:
There is an issue where the staging `.htaccess` file, which is manually generated would be removed on deployment - this needs to be resolved before we merge this

PR #13854, opened 2026-05-27
AG-15505 Fix example

StephenCooper:
Fix for https://github.com/ag-grid/ag-grid/actions/runs/26489575045/job/78004288900 following changes in https://github.com/ag-grid/ag-charts/pull/6862


lsjroberts (approved):
LGTM

PR #13874, opened 2026-05-28
AG-17227 Use chart-level context for sparklines, drop callback wrapping

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17227

Replace per-callback context wrapping with chart-level `context`, so the sparkline cell renderer no longer puts a function into `sparklineOptions` and the chart's structural-options cache can hit on the second cell onward.

- `sparklineCellRenderer.ts` — set `context: { data, cellData }` on the chart options at init and on every `update()`. Drop `wrapTooltipRenderer` / `wrapItemStyler` (user callbacks now receive `params.context` via the chart-side `callWithContext` path). Drop the default `tooltipRenderer` / `tooltipRendererWithXValue` install — the chart-side sparkline preset supplies an equivalent default.
- `sparklinesUtils.ts` — remove the now-unused `wrapFn` helper.

Depends on ag-grid/ag-charts#6931, which adds context to `FAST_PATH_OPTIONS`, fixes the cache-aliasing for `context`/`container`, and adds the xValue-aware default to the preset. Land that first.

Fix [AG-17227](https://ag-grid.atlassian.net/browse/AG-17227)

### DIFF
--- a/packages/ag-grid-enterprise/src/sparkline/sparklineCellRenderer.ts
+++ b/packages/ag-grid-enterprise/src/sparkline/sparklineCellRenderer.ts
@@ -121,13 +121,8 @@ export class SparklineCellRenderer extends Component implements ICellRenderer {
 
             this.sparklineOptions.type ??= 'line';
 
-            // No `tooltip.renderer` is installed when the user didn't supply one — the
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #13889, opened 2026-06-01
AG-17148 fix stale 301s

skewis6:


documentation/ag-grid-docs/src/utils/htaccess/redirects.ts
```diff
@@ -2593,11 +2585,6 @@ export const SITE_301_REDIRECTS: Redirect[] = [
     { from: '/react-data-grid/component-cell-editor/', to: '/react-data-grid/cell-editors/' },
     { from: '/vue-data-grid/component-cell-editor/', to: '/vue-data-grid/cell-editors/' },
 
-    { from: '/javascript-data-grid/component-header/', to: '/javascript-data-grid/column-headers/' },
```
taktran, line 2596: How come you are removing this?

I still see it referenced in https://github.com/ag-grid/ag-grid/blob/latest/packages/ag-grid-community/src/entities/colDef.ts#L587

taktran (approved):
LGTM

PR #13909, opened 2026-06-02
AG-17408 - [calculated columns] - API and Theming improvement

gportela85:
fix [AG-17408](https://ag-grid.atlassian.net/browse/AG-17408)


packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -682,13 +771,81 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+    }
+
+    private isValidDataType(dataType: string): boolean {
+        return this.beans.dataTypeSvc?.isDataTypeRegistered(dataType) ?? dataType in BASE_DATA_TYPE_LOCALE_KEYS;
+    }
+
+    private getDataTypeDisplayName(dataType: string): string {
+        const localeKey = BASE_DATA_TYPE_LOCALE_KEYS[dataType];
+        if (localeKey != null) {
+            return this.getLocaleTextFunc()(localeKey, this.capitaliseDataType(dataType));
```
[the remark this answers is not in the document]
gportela85: oops

packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -1025,14 +1025,32 @@ export class CellCtrl extends BeanStub {
...
-        this.comp.toggleCss(CSS_CALCULATED_COLUMN, this.isCalculatedColumn());
+    private setServiceCssClasses(): void {
+        const classes = this.beans.columnCssClassSvc.getCellClasses(this.column, this.rowNode);
+        const oldClasses = this.serviceCssClasses;
+        this.serviceCssClasses = new Set(classes);
+
+        for (const cssClass of classes) {
+            if (oldClasses.has(cssClass)) {
+                oldClasses.delete(cssClass);
+            } else {
```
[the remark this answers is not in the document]
gportela85: CellCtrl is not reused for a different row/column. Cells are keyed by instanceId, and CellCtrl is constructed with a fixed column and rowNode. Virtual column changes either keep the same CellCtrl for the same column instance or destroy it and create a new one. 
The stale-class risk would only apply if a future provider returns dynamic classes and fails to trigger refreshServiceCssClasses(). For Calculated Columns, highlight changes explicitly refresh previous and next highlighted columns, so this specific claim is not a real bug.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -682,13 +771,81 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
         );
         state.close = () => dialog.close();
+        this.setHighlightedColumn(columnToHighlight);
         const destroyDialogMouseListeners = this.addManagedElementListeners(dialog.getGui(), {
             mousedown: () => form.hideSuggestions(),
         });
         dialog.addDestroyFunc(() => destroyDialogMouseListeners.forEach((destroyFunc) => destroyFunc()));
+        dialog.addDestroyFunc(() => this.setHighlightedColumn(null));
         dialog.addEventListener('destroyed', () => this.destroyBean(form));
     }
```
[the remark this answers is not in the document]
gportela85: dataTypes: [] falls back to defaults because getValidDataTypes() returns defaults when no valid entries remain.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -682,13 +771,81 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+            text: this.getDataTypeDisplayName(dataType),
+        }));
+    }
+
+    private getValidDataTypes(dataTypes: readonly string[]): CalculatedColumnType[] {
+        const validDataTypes: CalculatedColumnType[] = [];
+        for (let i = 0, len = dataTypes.length; i < len; i++) {
+            const dataType = dataTypes[i];
+            if (validDataTypes.indexOf(dataType) >= 0) {
+                continue;
```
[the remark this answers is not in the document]
gportela85: Will use _camelCaseToHumanText instead

packages/ag-grid-community/src/styling/columnCssClassService.ts
```diff
@@ -0,0 +1,71 @@
...
+export class ColumnCssClassService extends BeanStub implements NamedBean {
+    public readonly beanName = 'columnCssClassSvc' as const;
+
+    private readonly providers: ColumnCssClassProvider[] = [];
+
+    public addProvider(provider: ColumnCssClassProvider): void {
+        this.providers.push(provider);
+    }
+
+    public getCellClasses(column: AgColumn, rowNode: RowNode): string[] {
```
StephenCooper: If this method was passed the old classes could you then return the ones to remove directly from this method to save constructing that in the cell and header?

packages/ag-grid-community/src/styling/columnCssClassService.ts
```diff
@@ -0,0 +1,71 @@
...
+
+            for (const cssClass of providerClasses) {
+                classes.push(cssClass);
+            }
+        }
+
+        return classes;
+    }
+
+    public getHeaderClasses(column: AgColumn | null, columnGroup: AgColumnGroup | null): string[] {
```
StephenCooper: Can we reduce the duplication of these methods?

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnForm.ts
```diff
@@ -201,15 +208,21 @@ export class CalculatedColumnForm extends Component {
...
-        });
-        this.addManagedElementListeners(this.eOperators, {
-            click: () => this.showSuggestions('operator', '', null, this.eOperators),
-        });
+        if (this.helperLists.has('columns')) {
+            this.addManagedElementListeners(this.eColumns, {
+                click: () => this.showSuggestions('column', '', null, this.eColumns),
+            });
+        }
+        if (this.helperLists.has('functions')) {
```
StephenCooper, line 215: I think we could reduce the duplication here too.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -100,6 +137,48 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+    }
+
+    private refreshCalculatedColumnHighlight(column: AgColumn | null): void {
+        if (column == null) {
+            return;
+        }
+
+        const cellCtrls = this.beans.rowRenderer.getCellCtrls(null, [column]);
+        for (let i = 0, len = cellCtrls.length; i < len; i++) {
+            cellCtrls[i].refreshServiceCssClasses();
```
StephenCooper: It seems like this could become less efficient as it will be re-creating classes from all services and not just those related to the calculatedColumn now.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -682,13 +771,80 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
 
+    private getDefaultDraft(): Omit<CalculatedColumnDraft, 'colId' | 'headerName'> {
+        const firstDataType = this.getDataTypeOptions()[0]?.value ?? DEFAULT_DRAFT.cellDataType;
+        return {
+            ...DEFAULT_DRAFT,
+            cellDataType: firstDataType,
+        };
+    }
+
+    private getDataTypeOptions(currentDataType?: string): CalculatedColumnDataTypeOption[] {
```
StephenCooper, line 766: I feel this validation is potentially overkill. I would suggest moving to the validation module only, or just not doing it. 

What happens if one of the options is invalid? Does the DataService itself warn when the calculated column is created? That might be good enough.

gportela85: simplified it

packages/ag-grid-community/src/styling/columnCssClassService.ts
gportela85: removed the css class service

packages/ag-grid-community/src/styling/columnCssClassService.ts
gportela85: done

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
gportela85: removed

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnForm.ts
gportela85: We could with something like:

```ts
const options = ['Columns', 'Functions', 'Operators'] as const;

options.forEach((option) => {
    const button = this[`e${option}`];
    const lowerOption = option.toLocaleLowerCase() as CalculatedColumnHelperList;
    const suggestion = `${lowerOption}s` as ColumnSuggestion['type']
    if (this.helperLists.has(lowerOption)) {
        this.addManagedElementListeners(button, {
            click: () =>
                this.showSuggestions(suggestion, '', null, button),
        });
    }
});
```

I don't think this is better

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -684,13 +746,66 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+    private getDataTypes(dataTypes: readonly string[]): CalculatedColumnType[] {
+        const uniqueDataTypes: CalculatedColumnType[] = [];
+        for (const dataType of dataTypes) {
+            if (uniqueDataTypes.indexOf(dataType) < 0) {
+                uniqueDataTypes.push(dataType);
+            }
+        }
+
+        return uniqueDataTypes;
+    }
```
StephenCooper, line 789: Can we just leave this to the user to make sure of?

StephenCooper (approved):

PR #13912, opened 2026-06-03
AG-17134 - allowlist additional CSP report-only console warnings

StephenCooper:
## Summary

- Adds two missing CSP report-only warning patterns to the E2E `excludeErrors` allowlist so doc tests stop failing on benign browser console messages
- Covers the Chromium `report-to` variant (existing entry only matched `report-uri`) and the `frame-ancestors` ignored-in-report-only warning

Closes AG-17134

## Test plan

- [ ] Run `./docs-e2e.sh` against an example that was previously failing (e.g. `accessibility/accessibility`) and confirm it passes


taktran (approved):
LGTM

PR #14014, opened 2026-06-10
Remove editing helper from core bundle

StephenCooper:
As `_populateModelValidationErrors` was directly called from CellCtrl it meant all of the editing validation was always present in core even when editing is not being used.

Instead have moved this behind the `editSvc` so that it can be dropped if editing is not enabled.


Stephanemw (approved):
LGTM

PR #14019, opened 2026-06-10
AG-17531 move grid options to client to reduce page load size

skewis6:
Fix AG-17531


documentation/ag-grid-docs/markdoc.config.ts
```diff
@@ -118,6 +118,7 @@ export default defineMarkdocConfig({
                 section: { type: String },
                 names: { type: Array },
                 config: { type: Object },
+                clientOnly: { type: Boolean },
```
taktran: Can you add a comment here for when to use this attribute for future reference

taktran (approved):

documentation/ag-grid-docs/src/stores/queryClientStore.ts
```diff
@@ -1,6 +1,8 @@
-import { QueryClient } from '@tanstack/react-query';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { atom } from 'nanostores';
 
+export { QueryClientProvider };
```
taktran: Why do you need this? Can't you just import from `@tanstack/react-query` directly?

taktran (changes requested):

taktran (approved):

skewis6:
Closing to come up with e more robust solution in future.

PR #14029, opened 2026-06-10
AG-17134 - Remove unsafe-eval from the main-site CSP

taktran:
- Replace the theme builder's eval-based string-literal unescaping with a manual unescaper — the only main-page eval use.
- Split the CSP into 'site' (no unsafe-eval) and 'examples' scopes; the generated .htaccess re-sets the relaxed policy for /examples/ and /archive/ paths via an Apache <If> override (SystemJS, Angular JIT and the Vue runtime compiler still need eval inside example documents).
- Staging enforces the split; production keeps enforcing the previous policy and reports on the split via Report-Only, gated by PRODUCTION_CSP_PHASE (flip to 'enforce' once the window is clean).
- Mirror the path-scoped split on the dev server (agDevCsp vite plugin) instead of the static dev CSP header.
- Cover with cspRules/htaccessRules unit tests and a CSP e2e spec asserting headers per path and per-framework example loads.


StephenCooper (approved):

skewis6 (approved):
LGTM

PR #14042, opened 2026-06-11
[AG-6377] Fix(pagination): page input fixes

sdwvit:
- Navigate only on Enter/blur, not on every keystroke
- Escape restores current page without navigating
- ArrowUp/Down increment/decrement page
- Add `role=spinbutton`, `aria-valuenow/min/max`, and `aria-label` kept in sync
- Right-align the page number value inside the input

Fix [AG-6377](https://ag-grid.atlassian.net/browse/AG-6377)

### DIFF
--- a/packages/ag-grid-community/src/pagination/pageSummaryComp.ts
+++ b/packages/ag-grid-community/src/pagination/pageSummaryComp.ts
@@ -1,13 +1,101 @@
+import type { AgElementParams, LocaleTextFunc } from 'ag-stack';
 import { KeyCode, RefPlaceholder, _setAriaDisabled, _setAriaLabel, _setAriaRole } from 'ag-stack';
 
 import { AgInputNumberFieldSelector } from '../agWidgets/agInputNumberField';
 import type { BeanCollection } from '../context/context';
 import type { IRowModel } from '../interfaces/iRowModel';
 import { _createIconNoSpan } from '../utils/icon';
+import type { AgComponentSelectorType } from '../widgets/component';
 import { Component } from '../widgets/component';
 import type { GridInputNumberField } from '../widgets/gridWidgetTypes';
 import type { PaginationService } from './paginationService';
 import { _formatPaginationNumber } from './paginationUtils';
+
+function buildPageSummaryTemplate(
+    idPrefix: string,
+    noInput: boolean,
+    localeTextFunc: LocaleTextFunc
+): AgElementParams<AgComponentSelectorType> {
+    const pageNumberChild = {
+        cls: 'ag-paging-number',
+        attrs: { id: `${idPrefix}-start-page-number` },
+        tag: noInput ? 'span' : 'ag-input-number-field',
+        ref: noInput ? 'lbCurrentStatic' : 'lbCurrentInput',
+    } as const;
+
+    return {
+        tag: 'span',
+        cls: 'ag-paging-page-summary-panel',
+        role: 'presentation',
+        children: [
+            {
+                tag: 'div',
+                ref: 'btFirst',
+                cls: 'ag-button ag-paging-button',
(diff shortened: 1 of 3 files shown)


packages/ag-grid-community/src/pagination/pageSummaryComp.ts
```diff
@@ -141,13 +141,34 @@ export class PageSummaryComp extends Component {
...
-            lbCurrentInput.onValueChange(this.onInputPage.bind(this));
-            this.addManagedListeners(lbCurrentInput.getInputElement(), {
-                blur: () => {
-                    if (!lbCurrentInput.getInputElement().value.trim()) {
+            const eInput = lbCurrentInput.getInputElement();
+            _setAriaRole(eInput, 'spinbutton');
+            this.addManagedListeners(eInput, {
+                keydown: (e: KeyboardEvent) => {
+                    if (e.key === KeyCode.ENTER) {
+                        e.preventDefault();
```
StephenCooper: If we preventDefault in every branch can we pull that out?

gportela85: probably not because we don't call it on `default`.

gportela85: I'd extract this keydown into a separate function instead of making it line line like this and use `switch` as it is the way we handle keydown listeners like this (see `agRichSelect#1253`)

sdwvit: updated, maybe it looks clean now?

packages/ag-grid-community/src/pagination/pageSummaryComp.ts
```diff
@@ -249,6 +279,14 @@ export class PageSummaryComp extends Component {
             lbCurrentInput.setMax(pageCount);
             lbCurrentInput.getInputElement().style.width = `${Math.floor(Math.log10(pageCount) + 3)}ch`; // log10 returns number of digits (as an integer part + fraction) - 1
             lbCurrentInput.setValue(lbCurrentValue.toString());
+            const eInput = lbCurrentInput.getInputElement();
+            _setAriaLabel(
+                eInput,
+                `${localeTextFunc('page', 'Page')} ${localeTextFunc('number', 'number')}, ${lbCurrentValue} ${localeTextFunc('of', 'of')} ${lbTotalStr}`
```
StephenCooper, line 224: I wonder if this would be better expressed with `${variable}` syntax. So that each language can control the order of these as well as just swapping the language. Maybe one to investigate as a follow-up task.

packages/ag-grid-community/src/pagination/pageSummaryComp.ts
```diff
@@ -141,13 +141,40 @@ export class PageSummaryComp extends Component {
...
+                    if (key !== KeyCode.ENTER && key !== KeyCode.ESCAPE && key !== KeyCode.UP && key !== KeyCode.DOWN) {
+                        return;
+                    }
+                    e.preventDefault();
+                    switch (key) {
+                        case KeyCode.ENTER:
+                            this.commitPageInput();
+                            break;
+                        case KeyCode.ESCAPE:
+                            lbCurrentInput.setValue(String(this.pagination.getCurrentPage() + 1), true);
```
StephenCooper: Can we extra these to consts to improve readability and also reduce duplicated calls.

StephenCooper (approved):
Just a few small tweaks please.

packages/ag-grid-community/src/pagination/pageSummaryComp.ts
sdwvit: as in format fn ? not sure I understand, but please feel free to create a jira for it

PR #14043, opened 2026-06-11
refactor: consolidate number utilities into shared helpers

StephenCooper:
## Summary

Adds three shared utilities to `packages/ag-grid-community/src/utils/number.ts`, exported via `main-internal.ts`:

- `_isFiniteNumber(v)` — type guard; `typeof v === 'number' && Number.isFinite(v)`. Replaces scattered private helpers and inline expressions.
- `_toFiniteNumber(v)` — coerces via `Number()`, returns `null` for NaN / ±Infinity. Note: `Number('')` is `0`, so callers must pre-filter blank strings if needed.
- `_clamp(value, min, max)` — `Math.max(min, Math.min(value, max))`. Replaces 19 nested `Math.max/min` pairs.

No behavioural changes. The `pageSizeSelectorComp` and `chartMenuParamsFactory` sites previously used `isNaN` (which passes `Infinity`); `_toFiniteNumber` now rejects it, which is the correct behaviour for those fields.

## Test plan

- [x] `yarn nx build:types ag-grid-community`
- [x] `yarn nx build:types ag-grid-enterprise`
- [ ] `yarn nx test ag-grid-community`
- [ ] `yarn nx test ag-grid-enterprise`


packages/ag-grid-community/src/agWidgets/agList.ts
```diff
@@ -260,7 +260,7 @@ export class AgList<
         } else {
             const currentIdx = listItems.indexOf(highlightedItem);
             let nextPos = currentIdx + (isDown ? 1 : -1);
-            nextPos = Math.min(Math.max(nextPos, 0), listItems.length - 1);
+            nextPos = Math.max(0, Math.min(nextPos, listItems.length - 1));
```
Stephanemw, line 263: Should this also use _clamp?

Stephanemw (approved):

PR #14047, opened 2026-06-12
 [AG-6377] Fix(pagination): page input RTL and alignment fixes

sdwvit:
- Add LTR/RTL-aware padding to the pagination page number input (`paginationComp.css`)
- Mirror LTR/RTL padding fix to the legacy Sass theme (`_footer.scss`)
- Vertically centre pagination panel content in Theming API CSS and legacy Sass theme

Fix [AG-6377](https://ag-grid.atlassian.net/browse/AG-6377)

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_footer.scss
+++ b/community-modules/styles/src/internal/base/parts/_footer.scss
@@ -29,6 +29,18 @@
         color: var(--ag-disabled-foreground-color);
     }
 
+    .ag-ltr .ag-number-field.ag-paging-number .ag-input-field-input {
+        text-align: right;
+        padding-right: var(--ag-input-padding-start);
+        padding-left: 0;
+    }
+
+    .ag-rtl .ag-number-field.ag-paging-number .ag-input-field-input {
+        text-align: left;
+        padding-left: var(--ag-input-padding-start);
+        padding-right: 0;
+    }
+
     @include ag.keyboard-focus((ag-paging-button), 0px);
 
     .ag-paging-button,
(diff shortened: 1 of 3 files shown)


SalvatorePreviti (approved):

community-modules/styles/src/internal/base/parts/_footer.scss
```diff
@@ -29,6 +29,18 @@
         color: var(--ag-disabled-foreground-color);
     }
 
+    .ag-ltr .ag-number-field.ag-paging-number .ag-input-field-input {
```
StephenCooper: There is no need for the `.ag-ltr` this will be applied automatically . So we only need to write the ltr version and the rtl will be generated automatically.

packages/ag-grid-community/src/pagination/paginationComp.css
```diff
@@ -44,8 +45,16 @@
     display: inline-block;
 }
 
-:where(.ag-number-field.ag-paging-number) .ag-input-field-input {
+:where(.ag-ltr) .ag-number-field.ag-paging-number .ag-input-field-input {
```
StephenCooper: Same for this. Should be handled automatically.

StephenCooper (changes requested):

community-modules/styles/src/internal/base/parts/_footer.scss
sdwvit: I need it because specificity is not enough without it

StephenCooper (approved):

PR #14048, opened 2026-06-12
AG-14628 - [Tooltip] Inherit colDef tooltips in group columns

sdwvit:
This PR fixes tooltip inheritance for auto group columns.

When a column used for row grouping configured any of the following tooltip settings, those settings were previously ignored on the generated auto group column:

* `tooltipValueGetter`
* `tooltipField`
* `tooltipComponent`
* `tooltipComponentSelector`
* `headerTooltip`

As a result, group rows showed no tooltip even though the source column had one configured.

In `multipleColumns` mode, this is a straight copy: `createBaseColDef` now copies the five tooltip properties from the grouped column’s `colDef`, the same way it already copies `headerValueGetter`.

In `singleColumn` mode, the logic is more involved because one auto group column covers all group levels, and the relevant source column is only known per row at render time via `params.node.rowGroupColumn`.

To support this, a dynamic `tooltipValueGetter` and `tooltipComponentSelector` are installed on the auto group column `colDef`. These dispatch at render time:

* group rows look up the tooltip config from `params.node.rowGroupColumn.colDef`
* leaf rows fall back to whatever `autoGroupColumnDef` has set

The logic is extracted into a private helper, `applyDynamicSingleColumnTooltips`, to keep `createAutoColDef` readable.

One behavioural change to flag: `autoGroupColumnDef` tooltip settings now only apply to leaf rows. Previously, they applied to all rows; now group rows always inherit from the source column.

--

Changes:

**multipleColumns** mode (`createBaseColDef`): Copies `tooltipField`, `tooltipValueGetter`, `tooltipComponent`, `tooltipComponentParams`, `tooltipComponentSelector`, and `headerTooltip` from the underlying grouped `colDef` — same pattern as the existing `headerValueGetter` inheritance.

**singleColumn** mode (`applyDynamicSingleColumnTooltips`): A single auto group column spans all group levels, so the underlying column differs per row. A dynamic `tooltipValueGetter` and `tooltipComponentSelector` are installed that dispatch at render time:

- **Group rows** — resolve via `params.node.rowGroupColumn.colDef` (calls `tooltipValueGetter` / `tooltipField` / `tooltipComponentSelector` / `tooltipComponent` from the grouped column)
- **Leaf rows** — use `autoGroupColumnDef` tooltip settings

**Code quality:** Extracted the singleColumn tooltip logic from `createAutoColDef` into a focused private helper `applyDynamicSingleColumnTooltips`. Added `GridRows` / `GridColumns` snapshots to all tests for regression coverage.

--

**Breaking change:** `autoGroupColumnDef` tooltip properties (`tooltipValueGetter`, `tooltipField`, `tooltipComponent`, `tooltipComponentSelector`) now apply to leaf rows only. Group rows always inherit from their underlying grouped `colDef`. Previously these settings applied to all rows.

--

**Out of scope (separate PRs):** `groupDisplayType: 'groupRows'` (full-width group rows) — PR2; documentation and migration note — PR3.

Fix [AG-14628](https://ag-grid.atlassian.net/browse/AG-14628)

### DIFF
--- a/packages/ag-grid-community/src/entities/colDef.ts
+++ b/packages/ag-grid-community/src/entities/colDef.ts
@@ -360,7 +360,7 @@ export interface ColDef<TData = any, TValue = any> extends AbstractColDef<TData,
      * Callback to select which tooltip component to be used for a given row within the same column.
      * @agModule `TooltipModule`
      */
-    tooltipComponentSelector?: TooltipComponentSelectorFunc;
+    tooltipComponentSelector?: CellEditorSelectorFunc | CellRendererSelectorFunc;
 
     /**
      * @deprecated v32.2 Use the new selection API instead. See `GridOptions.rowSelection`
@@ -1226,10 +1226,6 @@ export type HeaderTooltipValueGetterFunc<TData = any, TValue = any, TContext = a
 export type TooltipValueGetterFunc<TData = any, TValue = any, TContext = any> = (
     params: ITooltipParams<TData, TValue, TContext>
 ) => string | any;
-
-export type TooltipComponentSelectorFunc<TData = any, TValue = any, TContext = any> = (
-    params: ITooltipParams<TData, TValue, TContext>
-) => CellRendererSelectorResult | undefined;
 
 export interface NewValueParams<TData = any, TValue = any, TContext = any> extends ChangedValueParams<
     TData,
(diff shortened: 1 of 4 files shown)


sdwvit:
quick localhost test https://plnkr.co/edit/FFTUed3Ba9uY7el4?open=main.js hovering over first row shows tooltip. In stable it doesn't show tooltip

SalvatorePreviti (approved):

PR #14081, opened 2026-06-15
Fix: Viewport RM leaves nodes at stale indexes, orphaning row ctrls

slow89:
## Bug

With `rowModelType="viewport"` + `getRowId`, a row id can end up referenced at two row indexes at once. When that happens the row renderer silently drops one of the two RowCtrls without destroying it. The dropped ctrl keeps its managed listeners registered (the global `modelUpdated` listener on the event service, and `dataChanged` on its rowNode, which is still live), so it is retained for the lifetime of the grid and keeps reacting to every data update.

This is a severe memory leak for long-lived grids on streaming data. In framework grids it is compounded: the orphaned ctrl keeps pushing render details into its (unmounted) framework cell components on every rowNode `dataChanged`; with React 18, every such `setState` on an unmounted fiber enqueues an update object that is never processed and never released. We traced a production renderer to 4GB this way: ~1,760 orphaned RowCtrls anchored solely by `allSyncListeners['modelUpdated']`, holding ~16,400 cell comps and 6.2M queued React state updates.

Affected: reproduced on 34.2.0; the relevant code is unchanged through 35.3.1.

## Root cause

1. `ViewportRowModel.setRowData` documents that omitted indexes are "left unchanged". When an id-matched node is found via `getRowId` and assigned a new index, nothing clears `rowNodesByIndex[previousIndex]`, so if the previous index is not covered by the same update, the **same RowNode object exists at two indexes**. This happens routinely with server-side sorted feeds (rows reshuffle every tick) combined with scrolling (range changes leave uncovered indexes).
2. `RowRenderer.getRowsToRecycle` then builds `ctrlsByIdMap[rowNode.id]` from `rowCtrlsByRowIndex`. Two indexes give two ctrls for one id: the second assignment overwrites the first, and the overwritten ctrl is never destroyed. It is no longer in any renderer map, `zombieRowCtrls`, or destroy queue, but its managed listeners keep it alive.

## Fix

- `viewportRowModel.setRowData`: when a node moved to a new index, delete the stale `rowNodesByIndex` entry at its previous index (root cause).
- `rowRenderer.getRowsToRecycle`: if two ctrls map to one row id, destroy the one that no longer matches the node's current rowIndex instead of silently overwriting it (defense in depth, protects any row model that produces a duplicate id).

## Repro / test

Added `testing/behavioural/src/rows/viewport-row-model-moved-nodes.test.ts`: seed a viewport grid, then apply a partial `setRowData` where an id moves to a new index while its old index is omitted. Without the fix, `api.forEachNode` visits the moved node twice (and the next recycle pass orphans a RowCtrl).

Manual repro of the leak itself: viewport grid on a feed that re-sorts ids across indexes, scroll up/down for a minute, then take a heap snapshot. `RowCtrl` instance count grows far past the rendered row count and never recovers, and in React it then grows unboundedly at idle (one retained state update per orphaned cell per data tick).

### DIFF
--- a/packages/ag-grid-community/src/rendering/rowRenderer.ts
+++ b/packages/ag-grid-community/src/rendering/rowRenderer.ts
@@ -1027,8 +1027,24 @@ export class RowRenderer extends BeanStub implements NamedBean {
 
         // then clear out rowCompsByIndex, but before that take a copy, but index by id, not rowIndex
         const ctrlsByIdMap: RowCtrlByRowNodeIdMap = {};
-        for (const rowCtrl of Object.values(this.rowCtrlsByRowIndex)) {
+        for (const index of Object.keys(this.rowCtrlsByRowIndex)) {
+            const rowCtrl = this.rowCtrlsByRowIndex[index as any];
             const rowNode = rowCtrl.rowNode;
+            const existingCtrl = ctrlsByIdMap[rowNode.id!];
+            if (existingCtrl && existingCtrl !== rowCtrl) {
+                // two ctrls can reference the same row id when the row model moved a node to a new
+                // index while a stale reference remained at the old one (e.g. viewport row model with
+                // partial setRowData coverage). overwriting the map entry would orphan one ctrl without
+                // destroying it, leaking the ctrl together with its event listeners and framework
+                // components. keep the ctrl matching the node's current row index and destroy the other.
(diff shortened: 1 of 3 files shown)


StephenCooper:
Thank you for reporting this has now been fixed.

PR #14087, opened 2026-06-16
[AG-5601] Feat(pivoting): add valueIndex for value column order state

sdwvit:
Fix [AG-5601](https://ag-grid.atlassian.net/browse/AG-5601)

### DIFF
--- a/packages/ag-grid-community/src/misc/state/stateUtils.ts
+++ b/packages/ag-grid-community/src/misc/state/stateUtils.ts
@@ -25,7 +25,7 @@ export function convertColumnState(
 > {
     const sortColumns: SortModelItem[] = [];
     const groupColIds: string[] = [];
-    const aggregationColumns: AggregationColumnState[] = [];
+    const aggregationColumns: (AggregationColumnState & { valueIndex: number | null | undefined })[] = [];
     const pivotColIds: string[] = [];
     const leftColIds: string[] = [];
     const rightColIds: string[] = [];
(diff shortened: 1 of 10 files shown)


packages/ag-grid-community/src/misc/state/stateUtils.ts
```diff
@@ -25,7 +25,7 @@ export function convertColumnState(
 > {
     const sortColumns: SortModelItem[] = [];
     const groupColIds: string[] = [];
-    const aggregationColumns: AggregationColumnState[] = [];
+    const aggregationColumns: (AggregationColumnState & { valueIndex: number | null | undefined })[] = [];
```
StephenCooper: Should this be its own type?

packages/ag-grid-community/src/misc/state/stateUtils.ts
```diff
@@ -96,6 +99,21 @@ function _removeEmptyValues<T>(array: T[]): T[] {
...
+    return columns
+        .map((column, index) => ({ column, index }))
+        .sort((a, b) => {
+            const aIdx = a.column.valueIndex ?? Infinity;
+            const bIdx = b.column.valueIndex ?? Infinity;
+            return aIdx !== bIdx ? aIdx - bIdx : a.index - b.index;
+        })
+        .map(({ column }) => ({ colId: column.colId, aggFunc: column.aggFunc }));
+}
+
```
StephenCooper: I think this can be shorter and more efficient. sort is stable by default since es2019 so don't need to do the manual index checking.

StephenCooper (approved):

PR #14095, opened 2026-06-16
[AG-17401] feat(pagination): object-based config for pageSize and rowSummary panels

sdwvit:
- Pagination panels now accept an object form for the pageSize entry, with panel-level paginationPageSize and paginationPageSizeSelector that override the equivalent grid-level options.                                                                                                                          
  - Removing a panel page-size override now preserves the user's selected page size instead of falling back to the grid option/default.                                                                                                                                                                              
  - Added behavioural tests covering object-based panel config and runtime paginationPanels changes.

Fixes [AG-17401](https://ag-grid.atlassian.net/browse/AG-17401)


documentation/ag-grid-docs/src/content/docs/row-pagination/index.mdoc
```diff
@@ -54,6 +54,41 @@ paginationPanels: [
 ]
 ```
 
+### Configuring the Page Size Panel
```
StephenCooper: I don't think we need to add a seperate docs section for this. I would just try to make a minimal change to the existing section https://grid-staging.ag-grid.com/javascript-data-grid/row-pagination/#pagination-panel-layout

I also wouldn't call out the priority handling as really people should not be doing both.

packages/ag-grid-community/src/pagination/paginationService.ts
```diff
@@ -216,6 +242,16 @@ export class PaginationService extends BeanStub implements NamedBean {
             case 'initialState':
                 this.pageSizeFromInitialState = size;
                 break;
+            case 'panel':
+                this.pageSizeFromPanel = size;
+                if (size !== undefined) {
```
StephenCooper: Should this mirror the behaviour in the gridOptions case? Why is there an undefined check here for that clearing.

testing/behavioural/src/pagination/pagination-panels.test.ts
```diff
@@ -713,6 +713,93 @@ describe('paginationPanels', () => {
...
+                paginationPageSize: 20,
+                paginationPanels: ['pageSize'],
+            });
+            expect(api.paginationGetPageSize()).toBe(20);
+
+            api.setGridOption('paginationPanels', [{ type: 'pageSize', paginationPageSize: 50 }]);
+
+            expect(api.paginationGetPageSize()).toBe(50);
+        });
+    });
```
StephenCooper, line 847: Do we need some tests to validate that changing the panels config causes the correct updates to the panel?

sdwvit: added some

packages/ag-grid-community/src/pagination/paginationService.ts
```diff
@@ -162,6 +187,7 @@ export class PaginationService extends BeanStub implements NamedBean {
             autoValue ??
             pageSizeFromPageSizeSelector ??
             pageSizeFromInitialState ??
+            pageSizeFromPanel ??
             pageSizeFromGridOptions ??
```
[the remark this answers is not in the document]
sdwvit: I think it's okay, user is in control here, and they shouldn't have a conflicting situation like this

StephenCooper (approved):

PR #14130, opened 2026-06-18
AG-17874 fix broken e2e tests

skewis6:


StephenCooper (approved):

documentation/ag-grid-docs/src/content/docs/page-verification.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { expect, test } from '@playwright/test';
 import type { Page } from '@playwright/test';
 
-const isCspIssue = (msg: string) => /Content-Security-Policy|Refused to (load|execute|connect)/i.test(msg);
+const isCspIssue = (msg: string) => /Content[- ]Security[- ]Policy|Refused to (load|execute|connect)/i.test(msg);
```
taktran: After my CSP work, we should probably make this error

taktran (approved):
LGTM, just a note for after my CSP work

documentation/ag-grid-docs/src/content/docs/page-verification.spec.ts
skewis6: will do. once your stuff is done let me know, i'll update the error handling and notifications

PR #14142, opened 2026-06-18
Ag 16590/bundle csrm and error overlay

StephenCooper:
Fix [AG-16590](https://ag-grid.atlassian.net/browse/AG-16590)

Improve getting started experience by including CSRM in core. 
Introduce a configuration Error overlay to make is quicker and more obvious when there is an error in the configuration.

Just starting with Module related issues, not all validation warnings.


StephenCooper:
Closing this approach.

PR #14151, opened 2026-06-19
AG-3390 - Create staging captcha sitekey

taktran:
## Summary

Introduces a separate reCAPTCHA site key per environment for the contact form, rather than sharing a single global key across staging and production.

- Replaces the global `RECAPTCHA_SITE_KEY` constant with a per-environment `captchaSiteKey` field on each form-data config (`CONTACT_FORM_DATA` and `STUDIO_FORM_DATA`, both `default`/staging and `production`).
- Staging/`default` now uses a dedicated staging site key (`6Ld_ro4s...`); `production` retains the existing key (`6LfvTQos...`).
- `ContactForm` now reads `captchaSiteKey` from the resolved environment form data instead of the global constant.

## Notes

The same change is being ported manually to the `ag-studio` repo (separate PR), rather than via an `ag-website-shared` sync.

### DIFF
--- a/external/ag-website-shared/src/components/contact-form/ContactForm.tsx
+++ b/external/ag-website-shared/src/components/contact-form/ContactForm.tsx
@@ -1,12 +1,6 @@
 import { initCaptcha } from '@ag-website-shared/components/contact-form/initCaptcha';
 import { Icon } from '@ag-website-shared/components/icon/Icon';
-import {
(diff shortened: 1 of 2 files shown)


skewis6 (approved):

markdurrant (approved):
LGTM 👍

PR #14159, opened 2026-06-19
[AG-17603] Fix(types): allow valueFormatter to return null | undefined

sdwvit:
This pr enables the valueFormatter return type to accept string | null | undefined without assertions, matching the behaviour the runtime already supports

Fix [AG-17603](https://ag-grid.atlassian.net/browse/AG-17603)

### DIFF
--- a/packages/ag-grid-community/src/interfaces/iSetFilter.ts
+++ b/packages/ag-grid-community/src/interfaces/iSetFilter.ts
@@ -1,6 +1,6 @@
 import type { AgPromise } from 'ag-stack';
 
-import type { ColDef, KeyCreatorParams, ValueFormatterParams } from '../entities/colDef';
+import type { ColDef, KeyCreatorParams, ValueFormatterFunc } from '../entities/colDef';
 import type { FilterUiChangedEvent } from '../events';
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/interfaces/iSetFilter.ts
```diff
@@ -196,7 +196,7 @@ export interface ISetFilterParams<TData = any, V = string> extends IProvidedFilt
      * If specified, this formats the value before it is displayed in the Filter List.
      * If a Key Creator is provided (see `keyCreator`), this must also be provided.
      */
-    valueFormatter?: (params: ValueFormatterParams) => string;
+    valueFormatter?: (params: ValueFormatterParams) => string | null | undefined;
```
StephenCooper: Should we use the `ValueFormatterFunc` instead of repeating the extended type all through the code base?

sdwvit: right, I can just use ValueFormatterFunc, thanks for spotting

StephenCooper (approved):

Issue #62, opened 2015-04-15
[pagination] Make pagination panel extractable


juanrapoport:
Hi!

It would be useful to make pagination panel extractable, that way you could place it anywhere in you app.

Thanks.

ceolter:
Excellent idea.

juanmav:
+1

juanrapoport:
@ceolter The extractable panel was your idea, just copy & paste from your chat :smile: .

ceolter:
Yea was trying to be funny!!

JimmyBoh:
One point of consideration I'd like to bring up is the ability to internationalize/localize (all) the text built in to the grid. Is there any support for this? Maybe instead of extracting the panel just expose a paging API, then we can create our own UIs (with text) that just call in to the grid to change page, etc. (And still have an extractable panel, for simplicity/common use cases.)

ceolter:
Re localisation, that's on my plan. Am surprised there hasn't been much chat requesting it!!!

chiptus:
:+1:

ceolter:
tracked by https://ag-grid.atlassian.net/browse/AG-42

gregbown:
Typical requests from art or UI design have pagination above and below the data grid.  It would be nice to have the option of more than one.  Currently, I would have to modify source to implement this. Even if it wasn't external it would be nice to have pagination north and pagination south options. 
Thank you

makinggoodsoftware:
Hi, 

Thank you for raising this issue for us, at this moment we are reviewing how we process Feature Requests/questions that come through from GitHub.

As I am sure you are aware, we offer two products:

ag-Grid - this is entirely free to use in line with our MIT license
ag-Grid Enterprise - our commercial version which funds the development of both versions.

We currently have a very full pipeline of work which can be viewed here: https://www.ag-grid.com/ag-grid-pipeline/

This will be our priority for the coming months and as a result, we don’t have capacity to take feature requests/questions from GitHub, but you might find that yours has already been raised in our pipeline.

We will do a periodic review of GitHub issues for any bugs raised and these can also be tracked on our pipeline page.

Thanks for your understanding.
ag-Grid Team.

PR #14185, opened 2026-06-22
RTI-3387 pivot filter state

SalvatorePreviti:
FIX RTI-3387

Fix filter restore on pivot result columns via `initialState`

Filters on pivot result columns (e.g. `2000 → sum(Gold)`) were not restored via `initialState`: all rows shown, filter not applied, `getState().filter` undefined. `setState` already worked.

**Cause:** filter state is applied at `rowCountReady`, before pivot result columns exist (they're created before `firstDataRendered`), so the filter is dropped (warning #62). `setState` works because it runs after the columns exist.

**Fix:** during init, if a filter targets a not-yet-created column, defer the filter state to `firstDataRendered` — mirroring how column state already defers via `setColumnPivotState`. Non-pivot grids and pivot grids without pivot-column filters are unaffected.

**Tests:** new `Pivot result column state` block — filter restore via `setState` + `initialState`, plus regression guards for sort, sizing, visibility, pinning and open column groups on pivot result columns.

### DIFF
--- a/packages/ag-grid-community/src/misc/state/stateService.ts
+++ b/packages/ag-grid-community/src/misc/state/stateService.ts
@@ -62,6 +62,8 @@ export class StateService extends BeanStub implements NamedBean {
     );
     private columnStates?: ColumnState[];
     private columnGroupStates?: { groupId: string; open: boolean | undefined }[];
+    /** Filter state held back until firstDataRendered, when pivot result columns exist. */
+    private deferredFilterState?: FilterState;
     private readonly staleStateKeys: Set<keyof GridState> = new Set();
 
     public postConstruct(): void {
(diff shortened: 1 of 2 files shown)


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -646,6 +654,23 @@ export class StateService extends BeanStub implements NamedBean {
...
 
+    /** Defers to firstDataRendered if any target column is missing (a pivot result column not yet created). */
+    private setFilterStateDeferringPivot(state: FilterState | undefined, source: 'gridInitializing' | 'api'): void {
+        const { colModel, pivotResultCols } = this.beans;
+        const filterModel = state?.filterModel;
+        if (filterModel && source === 'gridInitializing' && colModel.pivotMode && !pivotResultCols?.pivotCols) {
+            const colsById = colModel.colsById;
+            const colIds = Object.keys(filterModel);
+            for (let i = 0, len = colIds.length; i < len; ++i) {
+                if (colsById[colIds[i]] == null) {
```
StephenCooper, line 665: Does this correctly handle the case that we have pivot mode enabled but no column labels yet? Should that case also be deferred or can that he immediately applied?

StephenCooper (approved):

PR #14198, opened 2026-06-23
RTI-3399 Render example-runner ExampleStyle as an Astro component

taktran:
## Summary

Standalone example pages (e.g. `/archive/36.0.0/examples/getting-started/quick-start-example/reactFunctionalTs/`) rendered in the system UI font instead of **IBM Plex Sans** — a regression after the Astro 6 upgrade. [RTI-3399](https://ag-grid.atlassian.net/browse/RTI-3399).

## Root cause

`@astrojs/react@5` (required by Astro 6) post-processes React component output with a regex that strips `<link rel="stylesheet">` (and `preload`/`preconnect`/etc.) tags. `ExampleStyle` was a React component, so its Google Fonts `<link>` was silently erased from every built standalone example page. Its `<style>` survived (only `<link>` is stripped), so layout CSS still worked but the font fell back to system UI.

This is the same class of bug fixed earlier for `Styles`/`Extras` (commit `9de149418f7`), which converted them to Astro components — `ExampleStyle` was missed.

## Fix

- Convert `ExampleStyle.tsx` → **`ExampleStyle.astro`** so the font `<link>` routes through Astro's native pipeline. The inline `<style>` is preserved verbatim via `is:inline set:html`.
- Update the five framework templates (React/TS/JS/Angular/Vue) to the default Astro import.
- Add a regression test on the quick-start example asserting the IBM Plex Sans `<link>` is present.

## Verification

- `nx format`, `nx lint ag-grid-docs`, `nx build ag-grid-docs` all pass.
- Built `dist` HTML confirms exactly **1** font `<link>` + the inline `<style>` per page across all 5 frameworks (correct root selectors `#root`/`my-app`/`#app`).
- New Playwright regression test passes against the served build for **all 6 framework variants**.


skewis6 (approved):
LGTM

PR #14206, opened 2026-06-24
[RTI-3352] Docs(show-values-as): clarify enablement, headings and links

StephenCooper:
Fix [RTI-3352](https://ag-grid.atlassian.net/browse/RTI-3352)

## Summary

Documentation clarity pass on the **Aggregation – Show Values As** page and related references.

- Overview example now sets `enableRowGroup` on dimension columns and `enableValue`/`enableShowValuesAs` on value columns, so the tool panel / column menu no longer offers nonsensical combinations.
- Clarified where `enableShowValuesAs` takes effect (selective on `defaultColDef` vs forced on a single column).
- Shortened verbose headings to noun phrases: *Choosing a Mode from the Column Menu* → **Column Menu**, *Setting the Mode in Code* → **Setting the Mode** (cross-reference anchor updated).
- Linked `initialShowValuesAs` to the column-state docs section.
- Reworded the `showValuesAs` row in *column-updating-definitions* ("relative to another value").
- Converted the Client-Side Row Model restriction into a `note` callout.
- Used **parent row total** terminology in the Tree Data section.

## Notes

Docs/examples only — no runtime code changes. Base branch: `b36.0.0`.

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
@@ -13,13 +13,13 @@ A mode can be chosen by the user from the column menu, set on the column definit
 
 This example groups Olympic winners by country and year, then shows raw medal totals alongside the same values as a percentage of their parent group.
 
-## Enabling Show Values As
+{% note %}
+Show Values As requires the `ShowValuesAsModule` and is only supported with the [Client-Side Row Model](./row-models/#client-side).
+{% /note %}
 
-Show Values As requires the `ShowValuesAsModule` and only works with the [Client-Side Row Model](./row-models/#client-side).
+## Column Menu
 
-## Choosing a Mode from the Column Menu
-
-The Show Values As submenu is off by default. To let the user switch mode from the [Column Menu](./column-menu/), register the `ColumnMenuModule` and set `enableShowValuesAs: true` on the required columns. Switching mode only repaints the affected cells — it does not re-aggregate, re-sort or re-filter.
+The Show Values As submenu is off by default. To let the user switch mode from the [Column Menu](./column-menu/), register the `ColumnMenuModule` and set `enableShowValuesAs: true` on the required columns. Switching mode only updates the affected column — it does not re-aggregate, re-sort or re-filter.
 
 The effect depends on where you set it:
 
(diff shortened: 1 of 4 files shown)


documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
```diff
@@ -51,11 +51,11 @@ A numeric column that is not yet aggregated is promoted to a value column when a
 
 `percentOfGrandTotal` and `percentOfColumnTotal` give the same result outside [Pivot](./pivoting/) mode, where there is a single column total. They diverge only while pivoting: `percentOfGrandTotal` keeps the whole column's grand total as the denominator, whereas `percentOfColumnTotal` uses each pivot column's own total, so every pivot column sums to 100%.
 
-A mode that is not meaningful in the current view — for example `percentOfParentRowTotal` on a grid with no grouping — shows `#N/A` rather than a transformed value. In the column menu it appears greyed and cannot be selected. If a mode was already active when the grid left the view that supports it, the selection is kept — its cells show `#N/A`.
+If a mode was already active when the grid left the view that supports it, the selection is kept — its cells show `#N/A`. In the column menu it appears greyed and cannot be selected.
```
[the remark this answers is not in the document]
StephenCooper: intentional

sdwvit, line 54: Can you confirm that disabled and selected have different styles?

documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
```diff
@@ -141,7 +139,7 @@ Show Values As does not require grouping. On a flat grid each row can be shown a
 
 ## Tree Data
 
-With [Tree Data](./tree-data/), aggregates are populated on parent nodes, so each node's value can be shown as a share of the folder that contains it.
+With [Tree Data](./tree-data/), aggregates are populated on parent nodes, so each node's value can be shown as a percentage of its parent row total.
```
sdwvit, line 142: is 'total' accurate here? in case I have avg() instead of sum() aggregation

sdwvit:
few questions, rest lgtm

documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
StephenCooper: The selected state is shown by the Check mark in the menu. This is still shown for disabled options that were selected before.

documentation/ag-grid-docs/src/content/docs/aggregation-show-values-as/index.mdoc
StephenCooper: The total name here is in line with the options available currently and the way that Show Value As works. We will need to update this when we include the other options to compare against.

sdwvit (approved):

PR #14215, opened 2026-06-24
AG-15464 Convert cell feature classes to functions

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-15464

Converts four per-cell `CellCtrl` feature classes into stateless module-level functions to reduce per-cell allocations (~9% fewer allocations measured), while preserving behaviour and tree-shaking.

- `CellKeyboardListenerFeature` → `_onCellKeyDown` / `_processCellCharacter`
- `CellMouseListenerFeature` → `_onCellMouseEvent` / `_onCellDoubleClicked`
- `CellCustomStyleFeature` → `_setupCellCustomStyle` / `_applyCell*`
- `CellEditStyleFeature` → `_applyCellEditStyles`

The small amount of per-cell state these classes held is relocated onto `CellCtrl` (`customStyleStaticClasses`, `customStyleClassRules`, `lastIPadMouseClickEvent`). `CellStyleService` and `EditService` expose thin wrappers so optional-bean call sites still route through them (preserving tree-shaking of the custom-style code). The now-unused `ICellStyleFeature` interface is deleted.

No behavioural change intended.

Fix [AG-15464](https://ag-grid.atlassian.net/browse/AG-15464)

PR #10772, opened 2025-06-12
AG-9780 TC114 pending pinned rows

StephM-W:

PR #14222, opened 2026-06-25
AG-15464 Convert cell position & row edit style features to functions

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-15464

Second pass of the AG-15464 feature-class→function conversion (follows #14215). Converts two more per-instance feature helpers in `ag-grid-community` from classes to stateless module-level functions, relocating their per-instance state onto the controller:

- **`RowEditStyleFeature`** → `_applyRowEditStyles(beans, rowCtrl)`, exposed via `EditService.applyRowEditStyles` (mirrors the existing `applyCellEditStyles`). `IRowStyleFeature` removed.
- **`CellPositionFeature`** → module functions (`_setupCellPosition`, `_initCellPosition`, `_getColSpanningList`, `_onCellLeftChanged`, `_onCellWidthChanged`). `colsSpanning`/`rowSpan` state moved onto `CellCtrl`; the position event-listeners now register via `cellCtrl.addManagedListeners` — identical lifetime, they die with the cell. Setup stays in the `CellCtrl` constructor (commented) to preserve the React invariant that `getColSpanningList()` is queryable before first render.

Removes one allocation per rendered cell/row (memory) plus the class shells (bundle). Behaviour-preserving, with one deliberate exception below.

### Intentional incidental fix
Removes a redundant `rowEditStyleFeature` teardown bundled into `addRowDraggerToRow`'s `compBean` destroy callback (added incidentally in #10772 — empty PR body, no rationale in PR/JIRA). A second canonical teardown already exists on `RowCtrl`'s own destroy. In the old code that callback could null the feature on a draggable row whose GUI was torn down while the `RowCtrl` survived, permanently silencing its row-edit styling — but that state isn't reachable via public API in vanilla (the recycle path either destroys the ctrl or stops editing first), so this is dead/redundant teardown, not a user-facing behaviour change. The function form simply has no object to null.

### Performance
Incremental A/B (parent vs this branch, 60k unvirtualised cells): allocated bytes/cycle **−3.2%**, live JS heap **−2.0%** (both tight sd); JS script time trended down (within noise). Bundle size also reduced.

### Test plan
- `yarn nx build:types ag-grid-community`, `yarn nx lint ag-grid-community` — green
- Behavioural: spanning (56), columns (1230), navigation (134), pinning/RTL (2609), cell-editing (777), full-row + batch-edit styles (32) — all green

Fix [AG-15464](https://ag-grid.atlassian.net/browse/AG-15464)

### DIFF
--- a/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
+++ b/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
@@ -607,7 +607,6 @@ export class CellCtrl extends BeanStub {
     // + rowRenderer: api softRefreshView() {}
     public refreshCell(params?: RefreshCellsParams & { newData?: boolean }): void {
         const {
-            rowCtrl: { rowEditStyleFeature },
             beans: { cellFlashSvc, filterManager, cellStyles },
             column,
             comp,
@@ -668,7 +667,7 @@ export class CellCtrl extends BeanStub {
             this.editSvc?.applyCellEditStyles(this);
             cellStyles?.applyCellUserStyles(this);
             cellStyles?.applyCellClassesFromColDef(this);
-            rowEditStyleFeature?.applyRowStyles();
+            this.editSvc?.applyRowEditStyles(this.rowCtrl);
 
             this.checkFormulaError();
         }
--- a/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
+++ b/packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
@@ -53,7 +53,13 @@ import type { CellSpan } from '../spanning/rowSpanCache';
 import { _createCellEvent } from './cellEvent';
 import { _onCellKeyDown, _processCellCharacter } from './cellKeyboardListenerFeature';
 import { _onCellMouseEvent } from './cellMouseListenerFeature';
-import { CellPositionFeature } from './cellPositionFeature';
+import {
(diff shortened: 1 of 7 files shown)


packages/ag-grid-community/src/rendering/cell/cellCtrl.ts
```diff
@@ -116,10 +122,13 @@ export class CellCtrl extends BeanStub {
 
     public lastIPadMouseClickEvent = 0;
 
+    // per-cell positioning state, owned by the cell position functions (rendering/cell/cellPositionFeature)
+    public colsSpanning?: AgColumn[];
+    public rowSpan = 1;
```
StephenCooper: Is this a new default value that we did not have before?

alistairjcbrown: @StephenCooper Sort of -- it's a new initialised value, rather than being `undefined` (to satisfy the `number` type). But nothing ever reads that default value; `setupRowSpan` sets it before any reads. 
~~`public rowSpan: number;` is a little looser against the type, but I'll set it back to match the original :+1:~~

Update: Chatted offline -- we'll keep the initialised value + type ailgnment

alistairjcbrown:
@StephenCooper Updated for your comment 👍

StephenCooper (approved):

PR #14243, opened 2026-06-26
[AG-9664] Feat(pivoting): interactive pivot column sorting

sdwvit:
Adds `colDef.pivotSort` so pivot columns can be sorted from their pivot-panel / Column Tool Panel pills (asc → desc → none), isolated from `colDef.sort`. On by default; `pivotPanelSuppressSort` disables it. Client-Side Row Model only (SSRM out of scope).

fix [AG-9664](https://ag-grid.atlassian.net/browse/AG-9664)


packages/ag-grid-community/src/columns/columnStateUtils.ts
```diff
@@ -788,6 +808,7 @@ export function getColumnStateFromColDef(beans: BeanCollection, column: AgColumn
         rowGroupIndex,
```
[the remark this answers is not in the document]
sdwvit: proposed fix (seed initialPivotSort only at creation, drop it from this helper) would make pivotSort inconsistent with every other initial* property and would break resetColumnState for pivot sort.

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -521,28 +535,45 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
-            depth: number
+            depth: number,
+            path: string[]
         ): ColDef | ColGroupDef => {
             const children: (ColDef | ColGroupDef)[] = [];
-            for (const [key, item] of uniqueValues) {
-                const child = uniqueValuesToGroups(`${id}${this.fieldSeparator}${key}`, key, item, depth + 1);
+            for (const [childKey, item] of uniqueValues) {
+                const child = uniqueValuesToGroups(
+                    `${id}${this.fieldSeparator}${childKey}`,
```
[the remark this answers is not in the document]
sdwvit: that's ok

packages/ag-grid-enterprise/src/columnToolPanel/updates/columnStateUpdateExecutionStrategy.ts
```diff
@@ -229,6 +257,17 @@ class SynchronousColumnStateUpdateStrategy implements ColumnStateConcreteUpdateS
     public getSortDef(column: AgColumn): SortDef | null {
         return column.getSortDef();
     }
+
+    public progressPivotSortFromEvent(column: AgColumn): void {
+        column.setPivotSort(getNextPivotSort(column.getPivotSort()));
+        // Pivot membership is unchanged, so applyColumnState wouldn't fire this - dispatch it to trigger
+        // the pivot refresh (columnPivotChanged → refreshModel step 'pivot') that re-derives column order.
```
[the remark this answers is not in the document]
sdwvit: it's intended

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -320,6 +323,11 @@ export class ColumnModel extends BeanStub implements NamedBean {
         }
         const beans = this.beans;
         const gos = this.gos;
+        // Pivot-sort reorders existing columns, so reuse the column-move animation to slide them.
+        const animatePivotSort = this.isPivotSortReorder();
+        if (animatePivotSort) {
```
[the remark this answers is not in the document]
sdwvit: that's ok, we don't do it anywhere

documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
```diff
@@ -82,6 +82,34 @@ const gridOptions = {
...
 
+## Sorting Pivot Columns
+
+End users can sort the pivot columns by clicking a pivot column's pill in the pivot panel or the Column Tool Panel,
+the same way row group columns are sorted from their pills.
+
+{% apiDocumentation source="column-properties/properties.json" section="pivoting" names=["pivotSort"] /%}
+
+`pivotSort` is independent of `sort`: it controls the order of a pivoted column's result columns only, and neither
+direction flows to or from the column's own sort. Pivot columns are sorted ascending by default. Clicking a pill
```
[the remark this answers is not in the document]
sdwvit: that's ok

documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/_examples/sort-pivot-columns/main.ts
```diff
@@ -0,0 +1,57 @@
...
+        { field: 'silver' },
+        { field: 'bronze' },
+    ],
+    defaultColDef: {
+        flex: 1,
+        minWidth: 130,
+        enableRowGroup: true,
+        enablePivot: true,
+        enableValue: true,
+    },
```
StephenCooper, line 33: Can we move these to the relevant columnDefs instead. Prevents the user from doing bad configs like grouping by silver.

documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/_examples/sort-pivot-columns/main.ts
```diff
@@ -0,0 +1,57 @@
...
+    autoGroupColumnDef: {
+        minWidth: 200,
+    },
+    pivotMode: true,
+    sideBar: {
+        toolPanels: ['columns'],
+        defaultToolPanel: '',
```
StephenCooper: I have noticed examples have been using defaultToolPanel: '', to not show it by default. But I believe that just not including it and only having the toolPanels will give the same behaviour.

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -948,7 +959,9 @@ type SortDefOverride = () => SortDef | null | undefined;
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
 export const _getDisplaySortForColumn = (column: AgColumn, beans: BeanCollection, override?: SortDefOverride) => {
     const overrideSortDef = override?.();
-    const sortDef = overrideSortDef ?? beans.sortSvc?.getDisplaySort(column);
+    // An override returning `null` means "no sort, show nothing"; only an absent override (`undefined`)
+    // falls back to the column's own sort.
+    const sortDef = overrideSortDef !== undefined ? overrideSortDef : beans.sortSvc?.getDisplaySort(column);
```
StephenCooper, line 956: Is this not a change in behaviour for the standard sort? Or is this ok because of the way overrideSortDef is provided?

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -531,6 +534,14 @@ export class AgColumn<TValue = any>
         this.sortDef = sortDef;
     }
 
+    public getPivotSort(): SortDirection | undefined {
+        return this.pivotSort;
+    }
+
+    public setPivotSort(pivotSort: SortDirection | undefined): void {
+        this.pivotSort = pivotSort;
+    }
```
StephenCooper: Are these required if pivotSort is public?

packages/ag-grid-community/src/entities/gridOptions.ts
```diff
@@ -1635,6 +1635,13 @@ export interface GridOptions<TData = any> {
      */
     rowGroupPanelSuppressSort?: boolean;
 
+    /**
+     * Set to `true` to suppress sort indicators and actions from the pivot panel and the column tool panel pivot pills.
+     * @default false
+     * @agModule `RowGroupingPanelModule`
```
StephenCooper, line 1641: Should this not be the `PivotModule`?

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -521,28 +543,45 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
+                    item,
+                    depth + 1,
+                    [...path, childKey]
+                );
                 children.push(child);
             }
 
+            // Children are the next pivot level; the measure level (>= pivot column count) keeps server order.
+            const childLevel = depth + 1;
+            if (childLevel < pivotColumns.length) {
```
StephenCooper, line 566: Can we validate the benchmarks around pivoting as this looks like we could have introduced extra sort calls on every iteration?

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
```diff
@@ -268,10 +268,22 @@ export class PivotStage extends BeanStub implements NamedBean, _IRowNodePivotSta
  * function reference or source equality.
  */
```
StephenCooper, line 270: JsDoc has moved to the wrong method. 

Also do we need this manual ascendingStringComparator?

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/dropZoneColumnComp.ts
```diff
@@ -112,7 +111,11 @@ export class DropZoneColumnComp extends PillDragComp<AgColumn> {
             ariaInstructions.push(aggregationMenuAria);
         }
 
-        if (this.isGroupingZone() && this.column.isSortable() && !isSortSuppressed) {
+        const isGroupSortable =
+            this.isGroupingZone() && this.column.isSortable() && !this.gos.get('rowGroupPanelSuppressSort');
```
StephenCooper: Lets reduce the duplication of this.column.isSortable()

packages/ag-grid-community/src/entities/gridOptions.ts
sdwvit: I checked. pivotPanelShow and rowGroupPanelSuppressSort are both tagged RowGroupingPanelModule. pivotPanelSuppressSort is also read in dropZoneColumnComp, which is part of RowGroupingPanelModule. but I agree it is weird that pivot panel related stuff lives in row grouping. perhaps renaming of the module is needed.

packages/ag-grid-enterprise/src/rowGrouping/columnDropZones/dropZoneColumnComp.ts
```diff
@@ -172,29 +175,46 @@ export class DropZoneColumnComp extends PillDragComp<AgColumn> {
     }
 
     private setupSort(): void {
-        if (!this.column.isSortable() || !this.isGroupingZone()) {
+        const { gos, column, eSortIndicator } = this;
```
StephenCooper: I wonder if this method would be cleaner if we have two separate paths instead of multiple group / pivot branches.

packages/ag-grid-community/src/entities/agColumn.ts
sdwvit: not a change in behaviour, overridesortdef is provided by clicking pivotcol pill, ei user interaction

testing/behavioural/src/sorting/pivot-column-sort-animation.test.ts
```diff
@@ -0,0 +1,65 @@
+import type { GridApi, GridOptions } from 'ag-grid-community';
+import { ClientSideRowModelModule } from 'ag-grid-community';
+import { PivotModule, RowGroupingModule } from 'ag-grid-enterprise';
+
+import { TestGridsManager, applyTransactionChecked, asyncSetTimeout } from '../test-utils';
+
+// AG-9664: toggling pivotSort reorders existing pivot columns, so it reuses the column-move animation
```
StephenCooper, line 7: Does this respect suppressColumnAnimation flag?

testing/behavioural/src/sorting/pivot-column-sort.test.ts
```diff
@@ -0,0 +1,278 @@
+import type { GridApi, GridOptions } from 'ag-grid-community';
+import { ClientSideRowModelModule } from 'ag-grid-community';
+import { PivotModule, RowGroupingModule, RowGroupingPanelModule } from 'ag-grid-enterprise';
+
+import { getColumnOrder } from '../columns/column-test-utils';
+import { TestGridsManager, applyTransactionChecked, asyncSetTimeout } from '../test-utils';
+
```
StephenCooper, line 7: Can we add tests for the deferred panel case in addition to these tests.

testing/behavioural/src/sorting/pivot-column-sort-animation.test.ts
sdwvit: Yes, colAnimation.start() returns early when suppressColumnMoveAnimation is set. pivot sort reuses the service , so goes through the same route

sdwvit: i'll add a test

packages/ag-grid-community/src/entities/gridOptions.ts
StephenCooper: Right ok. Let's leave for now but can you update the JsDoc for the `RowGroupingPanelModule` to include both grouping and pivoting references.

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -620,6 +648,34 @@ export class ColumnModel extends BeanStub implements NamedBean {
     }
 }
 
+const pivotGroupKey = (col: AgColumn): string => (col.getColDef().pivotKeys ?? []).join('');
+
+/** Re-rank `stickyOrder` so its pivot groups follow the freshly-sorted `defColsList` group order, while
+ *  keeping each group's within-group (user) column order. Stable: equal group rank keeps sticky order. */
+const reRankByPivotGroupOrder = (
```
SalvatorePreviti: ok! this can be optimized and should be, and not sure if there is a better more optimal way of doing this as we have already logic in "restore column order" that should be responsible for this - we need to run the benchmarks here and compare latest vs this to see wether pivot has lost performance and how slow this is when enabled. To run the benchmarks comparison clone in "../ag-grid2" the repo, npm run bootstrap, run `./benches.sh --bench-compare all` to run both comparison, and to update the comparison only on your branch `./benches.sh --bench-compare test && `./benches.sh --bench-compare compare` and it will tell you where the output is generated

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -496,7 +499,26 @@ export class PivotColDefService extends BeanStub implements NamedBean, IPivotCol
...
+        }
+        const pivotComparator = primaryColumn.colDef.pivotComparator;
+        const baseComparator = pivotComparator ? convertToHeaderNameComparator(pivotComparator) : headerNameComparator;
+        if (pivotSort === 'desc') {
+            return (a, b) => baseComparator(b, a);
+        }
+        return baseComparator;
+    }
+
     public createColDefsFromFields(fields: string[]): (ColDef | ColGroupDef)[] {
```
SalvatorePreviti, line 520: this method became way more complicated and looks slower, maybe we need some optimization and redesign and refactoring of the whole flow here

testing/behavioural/src/sorting/pivot-column-sort-animation.test.ts
sdwvit: https://github.com/ag-grid/ag-grid/pull/14243/changes#diff-6b8b62d9adef35b46a9ea4db1af9b0507fd54bb69d93ef98d5d30bfb5c78281fR57

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
sdwvit: Extra calls are needed, but some caching can be done. It was rebuilding the comparator for every group node when it only depends on the level. Now it's built once per level and reused. I've checked pivot-transaction.bench.ts: pivot on with 240 cols holds ~288 iter/s (3.47ms), no regression.

packages/ag-grid-enterprise/src/pivot/pivotStage.ts
sdwvit: Good spot on the JSDoc. I've fixed it so it now describes what the function actually does.

On the manual sort: yeah, I think we need it. `computePivotOrder` is not building rendered columns - it just produces a throwaway key-order snapshot that we compare with the previous one to decide whether pivot columns need rebuilding, for example after a sort toggle or comparator mutation. So it sorts the raw keys directly.

I did try reusing `getPivotGroupComparator` to remove the duplication, but that sorts by header name and passes header names into a custom `pivotComparator`. Reusing it would change the ordering when `refData` is involved, so the same toggle could produce different behaviour.

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
sdwvit: The comparator is resolved once per level and then reused for each node, so this is not slower than before. The only added cost is a path array per node and a no-op sort when using natural order. Both are negligible and this is not a measured hot path.

I would rather not redesign the flow speculatively. If benchmarks show a real regression, we can revisit it.

packages/ag-grid-community/src/columns/columnModel.ts
sdwvit: ✅ Benchmarks done — bench-compare all, latest (0de470a05e3) vs ag-9664 (9afde066662), real Chromium, 1 run/side, 8m33s (AMD 5950X).

Verdict: pivot did not regress. The new pivot-sort re-rank is gated behind hasInteractivePivotSort(), so the pivot benches (which don't trigger interactive sort) only pay one boolean check:pivot ON transaction (240 cols): 1.01x slower — within ±1.1% noisepivot ON + rowTotals: 1.04x faster · pivot OFF: 1.02x fasterpivot-aggregation (incl. 999-col high-cardinality, ± comparator): flat→slightly faster, all noisycolumn-refresh "pivot mode on/off": 1.01x faster
Caveat: the "Notable" regressions are all in unrelated paths — grouping-pipelines RowsPath (1.05–1.13x slower) and formulas (1.10x slower). Neither touches the PR diff (pivot column-sort), and they're single-run on a busy box (e.g. immutable RowsPath showed ±4.2% rme). Almost certainly thermal/run-to-run noise.

No regression.

testing/behavioural/src/sorting/pivot-column-sort.test.ts
sdwvit: added

packages/ag-grid-community/src/columns/columnModel.ts
```diff
@@ -421,9 +434,15 @@ export class ColumnModel extends BeanStub implements NamedBean {
...
         }
+        // An active interactive pivotSort forces strict pivot column order, overriding sticky-order preservation.
         const restoreOrder = !newColDefs || _shouldMaintainColumnOrder(gos, showingPivotResult);
         const lastOrder = showingPivotResult ? this.lastPivotOrder : this.lastOrder;
-        const prevOrder = restoreOrder ? lastOrder : null;
+        let prevOrder = restoreOrder ? lastOrder : null;
+        // pivotSort reorders the groups but keeps the user's within-group order and widths: re-rank the
+        // preserved order by the freshly-sorted group order rather than discarding it.
+        if (prevOrder != null && showingPivotResult && !!beans.pivotColsSvc?.hasInteractivePivotSort()) {
+            prevOrder = reRankByPivotGroupOrder(colsList, prevOrder, colsById);
```
StephenCooper: Lets move the reRankByPivotGroupOrder to the pivotColsService so that Core is not paying the cost of this method when pivoting is not required.

StephenCooper (approved):

PR #14259, opened 2026-06-29
Add glob-scoped agent rule: shared controllers vs framework views

StephenCooper:
Adds a glob-scoped agent rule (`.rulesync/rules/framework-view-layers.md`) that auto-injects when a shared `*Ctrl`/`*Comp` (community/enterprise) or a React/Angular/Vue view source file is edited.

**Why:** In AG-17663 a renderer-tooltip teardown was implemented only in the vanilla `cellComp.ts` and missed the React/Angular/Vue `CellComp` re-implementations, which share `CellCtrl`. Vanilla behavioural tests passed, giving false confidence. The rule captures the lesson: framework-shared behaviour belongs in the controller, a fix in one view doesn't reach the others, and vanilla tests don't cover the framework views.

**Mechanism:** local `.rulesync/` tier (the controller/view-per-framework split is grid-specific). Tracked via an allowlist entry in `.rulesync/.gitignore`; the generated `.claude/rules/` output is reproduced by `setup-prompts.sh` (not committed). `verify-rulesync.sh` passes.

### DIFF
--- a/.rulesync/.gitignore
+++ b/.rulesync/.gitignore
@@ -31,6 +31,7 @@ mcp.json
 !/rules/benchmarks.md
 !/rules/docs-pages.md
 !/rules/examples.md
+!/rules/framework-view-layers.md
 !/rules/testing.md
 # docs-review-testing.md ships as a sibling asset of the ag-prodeng docs-review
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #14269, opened 2026-06-30
Exempt test files from no-restricted-properties lint rule

StephenCooper:
Test files were being flagged by the `no-restricted-properties` ESLint rule (e.g. `document.createElement`), which exists to steer runtime code toward the `_createElement` helper. That guidance is irrelevant to tests, so test files are now exempt.

- Add a `**/*.test.ts` / `**/*.test.tsx` override turning off `no-restricted-properties` in the community and enterprise ESLint configs.
- Remove the now-redundant `eslint-disable no-restricted-properties` directives from `errorText.test.ts` and `agRichSelect.test.ts`.
- The rule is now enforced as `error` (was `warn`); the few non-test violations this surfaced are replaced with the `_createElement` helper in `autoWidthCalculator`, `chartSettingsPanel`, and `miniChartsContainer`.

### DIFF
--- a/packages/ag-grid-community/eslint.config.mjs
+++ b/packages/ag-grid-community/eslint.config.mjs
@@ -49,7 +49,7 @@ export default [
                 },
             ],
             'no-restricted-properties': [
-                'warn',
+                'error',
                 { property: 'innerText', message: 'Prefer textContent where possible' },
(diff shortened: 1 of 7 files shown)


alistairjcbrown (approved):
LGTM :shipit:

sdwvit (approved):

PR #14281, opened 2026-06-30
AG-17689 tree data is selectable

SalvatorePreviti:
FIX AG-17689

Makes `isRowSelectable` work correctly for tree data, and reworks how selection is
maintained when rows are added, removed or regrouped so a single, correctly-sourced
`selectionChanged` is emitted per refresh.

### Selection cleanup is centralised
- New event-free `removeFromSelection(node, source)` replaces `filterFromSelection`. It only
  mutates the selection map, so callers can drop a node inline as they tear it down.
- The old per-site "collect `nodesToUnselect[]` → `setNodesSelected({newValue:false})` → dispatch"
  blocks are removed from `clientSideNodeManager`, `groupStrategy` and `treeGroupStrategy`
  (incl. the `deselect` / `deselectHiddenNodes` helpers).

### Single coalesced `selectionChanged` per refresh
- Added a selection batch (depth counter): while open, `selectionChanged` is coalesced into one
  pending event carrying the first source seen, flushed when the outermost batch closes.
- `updateSelectableAfterGrouping` is now one post-refresh pass for both flat and hierarchical
  grids. A flat-delta fast path recomputes only added/updated rows (`changedRowNodes`) instead
  of every node.
- `flushPendingSelectionChanged` covers the deferred-refresh case so a later unrelated change
  can't emit the pending event with the wrong source.

### `selectedNodes` is a derived index
- The map is now maintained in one place via the `selectRowNode` chokepoint, replacing the
  reactive `rowSelected` listener (`onRowSelected`) and the `syncInOld/NewRowNode` split.

### `isRowSelectable` applied at the right points
- `updateRowSelectable` is invoked where nodes become fully formed: `blockUtils` (SSRM),
  `infiniteBlock`, `viewportRowModel`; removed from `rowGroupingUtils`.
- CSRM exposes `refreshingData` so the base service skips per-node recompute mid bulk refresh
  (the single pass handles it).

### Misc
- Hot selection loops converted from functional methods to indexed loops; `csrm` cached;
  repeated field reads hoisted, less temporary allocations.

files:         0 added, 0 removed, 14 changed (14 total)
lines added:   +344
lines removed: -309
net delta:     +35


packages/ag-grid-community/src/selection/selectionService.ts
```diff
@@ -636,101 +642,145 @@ export class SelectionService extends BaseSelectionService implements NamedBean,
...
         if (clearSelection) {
             this.resetNodes();
         }
-        this.setNodesSelected({
-            newValue: true,
-            nodes,
-            source,
-        });
+        this.setNodesSelected({ newValue: true, nodes, source });
     }
```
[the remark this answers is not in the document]
SalvatorePreviti: false positive and there is a test to prove it

packages/ag-grid-community/src/selection/baseSelectionService.ts
```diff
@@ -47,9 +47,14 @@ export abstract class BaseSelectionService extends BeanStub {
 
         this.isRowSelectable = _getIsRowSelectable(gos);
 
+        const rowModel = beans.rowModel;
         this.addManagedEventListeners({
             cellValueChanged: (e) => this.updateRowSelectable(e.node as RowNode),
-            rowNodeDataChanged: (e) => this.updateRowSelectable(e.node),
+            rowNodeDataChanged: (e) => {
+                if (!rowModel.refreshingData) {
```
StephenCooper, line 54: Does this not break the other row models as now they do not respond to this event?

SalvatorePreviti: Ah no as the flag will be undefined for the other models so it will be a pass through, always executed

StephenCooper (approved):

PR #14291, opened 2026-07-01
[AG-17118] Fix(pagination): keyboard focus and focus outline for page numbers

sdwvit:
Fix [AG-17118](https://ag-grid.atlassian.net/browse/AG-17118)

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_footer.scss
+++ b/community-modules/styles/src/internal/base/parts/_footer.scss
@@ -64,6 +64,8 @@
     }
 
     .ag-paging-page-number {
+        // relative so the keyboard-focus ::after focus ring positions over the button
+        position: relative;
         display: flex;
         align-items: center;
         justify-content: center;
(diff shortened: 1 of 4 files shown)


community-modules/styles/src/internal/base/parts/_footer.scss
```diff
@@ -64,6 +64,8 @@
     }
 
     .ag-paging-page-number {
+        // relative so the keyboard-focus ::after focus ring positions over the button
```
alistairjcbrown, line 67: ❓ Where is the `::after` style set? I can only see `:focus-visible` below in `pageNumbersComp.css`

alistairjcbrown (approved):
Just one question about the footer style change, but LGTM 👍

community-modules/styles/src/internal/base/parts/_footer.scss
sdwvit: The ::after comes from the keyboard-focus mixin on line 42 for legacy themes. 
The button needs position: relative for the ::after component to sit over it and not render elsewhere.

PR #14296, opened 2026-07-01
AG-17729 Add site-wide announcement banner config

skewis6:
## Summary
- Extract the announcement banner's enabled flag and content into a dedicated `announcementBanner.config.ts` so it can be toggled/updated in one place without touching `Layout.astro` (previously the "Beyond the Prompt" content was hardcoded there).
- Enable the banner with the AG Grid & AG Charts webinar announcement, linking out to the Livestorm registration page in a new tab.
- Reuses the existing `AnnouncementBanner` component/styling as-is (same visual design as the previous "Beyond the Prompt" banner), including localStorage-based permanent dismissal and the archived-docs/per-page opt-outs.

## Test plan
- [x] `yarn nx format --sort-root-tsconfig-paths=false`
- [x] `yarn nx lint ag-grid-docs`
- [x] Verified via dev server + server-rendered HTML that the banner renders above the site header when enabled, with correct title/description/CTA/link (opens in new tab)
- [x] Confirmed dismiss-state CSS/script (localStorage persistence) and no-layout-shift behaviour are unchanged from the original banner
- [ ] Fill in the real date/time/timezone in the description before this goes live (currently placeholder `DAY TIME TIMEZONE`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


documentation/ag-grid-docs/src/components/announcement-banner/announcementBanner.config.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Single place to enable/disable the site-wide announcement banner and set its content.
+ * Flip `enabled` to `true` to show the banner above the header on every page (archived
+ * doc versions never show it, and individual pages can still opt out via
+ * `showAnnouncementBanner={false}` on `<Layout>`).
+ */
+export const announcementBannerConfig = {
```
taktran: The more astro-native way to do this is to put this file in `documentation/ag-grid-docs/src/content`, add a `documentation/ag-grid-docs/src/content.config.ts` schema validation and use `await getEntry` to get the data. This will fail loudly if the content is in the wrong format

taktran (changes requested):

taktran (approved):
Looks good

PR #14305, opened 2026-07-02
AG-17736-ssrm-is-explandable-pivot

SalvatorePreviti:
FIX [AG-17736](https://ag-grid.atlassian.net/browse/AG-17736)

## Problem

In pivot mode the deepest row-group level is a leaf group: `isExpandable()` is `false` and no chevron shows. The SSRM write paths (`setExpanded`, `expandAll`) ignored this, so a leaf group could be marked expanded and its id leaked into the serialized expansion state — diverging from CSRM. The same gap affected other permanently non-expandable rows (footers, detail rows, pivot master rows).

## Change

- **Centralised the expandability rule.** `BaseExpansionService.checkExpandable(node, hasChildren)` now holds the footer / pivot / leaf-group / master logic; `isExpandable()` passes `hasChildren()` ("expandable now") and the SSRM write path passes the structural `group` flag ("could ever expand, once children load"). Removes the rule being copied three times.
- **`setExpanded` (SSRM):** no-op for rows that can never open (footer / detail / pivot leaf-group / pivot master); for a not-yet-loaded group, record intent only — `checkOpenByDefault` applies it on load.
- **`expandAll` (default strategy):** skip non-expandable nodes on expand; on collapse, drop stale expand-intent without recording a collapse, so no dead id leaks into the state.
- **`isExpanded` / `isNodeExpanded` (SSRM):** a non-expandable node reports collapsed, and the lazy default isn't resolved/cached while non-expandable, so it resolves correctly if the node later becomes expandable.
- **`CsrmExpansionService.isExpanded`:** delegates to `checkExpandable`, aligning CSRM with `isExpandable()` — a pivot-mode master row is now consistently non-expandable (previously `isExpanded` reported it expandable while `isExpandable()` did not).

## Result

Expansion behaves identically for manual clicks, `setExpanded()` and `expandAll()`, matches CSRM↔SSRM, and keeps the serialized state free of dead ids for every kind of non-expandable row.

## Tests

- Unit `serverSideExpansionService.test.ts`: never-expandable (pivot leaf group, detail row, pivot master) vs. not-yet-expandable (unloaded group stub), and collapseAll clearing stale intent.
- Behavioural `pivot-leaf-group-expansion.test.ts`: CSRM + SSRM, both `ssrmExpandAllAffectsAllRows` values.


packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -108,16 +116,33 @@ export class ServerSideExpansionService
     }
 
     public isNodeExpanded(node: RowNode): boolean {
-        return this.strategy.isRowExpanded(node);
+        return this.strategy.isRowExpanded(node) && node.isExpandable();
```
sdwvit, line 115: why do you need isExpandable() here?

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -108,16 +116,33 @@ export class ServerSideExpansionService
...
     }
 
     public override setExpanded(node: RowNode, expanded: boolean, e?: MouseEvent | KeyboardEvent, _?: boolean): void {
+        // A pivot leaf group / footer can never open, so this is a true no-op that keeps the serialized state clean.
+        if (this.isNeverExpandable(node)) {
```
sdwvit: do we really need to distinguish isExpandable() === false and isNeverExpandable() === true?

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -108,16 +116,33 @@ export class ServerSideExpansionService
...
+            this.dispatchStateUpdatedEvent();
+            return;
+        }
         super.setExpanded(node, expanded, e);
         this.dispatchStateUpdatedEvent();
         this.updateExpandedState(node);
     }
 
+    /** Non-expandable by construction (pivot leaf group / footer), as opposed to a group whose children aren't loaded yet. */
+    private isNeverExpandable(node: RowNode): boolean {
```
sdwvit: I'd prefer `isEverExpandable`, however I have doubts if we need this method. But don't have a specific arguments yet.

SalvatorePreviti: refactored the code - the logic has been properly centralized and moved, this method does not exist anymore

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
```diff
@@ -108,16 +116,33 @@ export class ServerSideExpansionService
...
 
     public override setExpanded(node: RowNode, expanded: boolean, e?: MouseEvent | KeyboardEvent, _?: boolean): void {
+        // A pivot leaf group / footer can never open, so this is a true no-op that keeps the serialized state clean.
+        if (this.isNeverExpandable(node)) {
+            return;
+        }
+
+        // Record intent even while not yet expandable, so an unloaded group applies it once its children arrive.
         this.strategy.setRowExpanded(node, expanded);
+
```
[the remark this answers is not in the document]
SalvatorePreviti: false positive - is correct as is

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
SalvatorePreviti: Recorded expand-intent short-circuits before the strategy's expandability check, so the && isExpandable() stops a not-yet-loaded node from being opened before it can.

packages/ag-grid-enterprise/src/serverSideRowModel/services/serverSideExpansionService.ts
SalvatorePreviti: refactored the code - now we don't need aymore isEverExpandable

sdwvit (approved):
pls confirm the fix after refactoring. Otherwise Code looks fine, approved.

PR #14309, opened 2026-07-02
AG-17641 Add Development Validation docs and error pages

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-17641

Fix [AG-17641](https://ag-grid.atlassian.net/browse/AG-17641)

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/dev-validation/_examples/overlay/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/dev-validation/_examples/overlay/main.ts
@@ -1,10 +1,12 @@
 import type { ColDef, GridApi, GridOptions } from 'ag-grid-community';
-import { AllCommunityModule, ModuleRegistry, ValidationModule, createGrid } from 'ag-grid-community';
+import { AllCommunityModule, ModuleRegistry, createGrid, enableDevValidations } from 'ag-grid-community';
 
-// Register the ValidationModule (with the overlay enabled) alongside the community modules. Registering
-// it in the modules array — rather than via enableDevValidations — is what lets the framework example
-// generators carry it into the modules prop.
-ModuleRegistry.registerModules([AllCommunityModule, ValidationModule.with({ overlay: 'all' })]);
+// Enable development validations so captured diagnostics surface in an overlay over the grid.
+if (process.env.NODE_ENV !== 'production') {
+    enableDevValidations();
+}
+
+ModuleRegistry.registerModules([AllCommunityModule]);
 
 const gridOptions: GridOptions = {
     columnDefs: [
--- a/documentation/ag-grid-docs/src/content/docs/dev-validation/_examples/overlay/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/dev-validation/_examples/overlay/main.ts
@@ -1,0 +1,26 @@
+import type { ColDef, GridApi, GridOptions } from 'ag-grid-community';
+import { AllCommunityModule, ModuleRegistry, ValidationModule, createGrid } from 'ag-grid-community';
+
+// Register the ValidationModule (with the overlay enabled) alongside the community modules. Registering
+// it in the modules array — rather than via enableDevValidations — is what lets the framework example
+// generators carry it into the modules prop.
+ModuleRegistry.registerModules([AllCommunityModule, ValidationModule.with({ overlay: 'all' })]);
+
+const gridOptions: GridOptions = {
+    columnDefs: [
+        // A stray property the grid does not recognise (as a prop-spreading wrapper might add),
+        // surfacing warnings #307 (per property) and #310 (summary) in the overlay.
+        { field: 'make', notAColumnProperty: true } as ColDef,
+        { field: 'model' },
+        { field: 'price' },
+    ],
+    rowData: [
+        { make: 'Tesla', model: 'Model Y', price: 64950 },
+        { make: 'Ford', model: 'F-Series', price: 33850 },
+        { make: 'Toyota', model: 'Corolla', price: 29600 },
+    ],
+};
+
+let api: GridApi;
+
+api = createGrid(document.querySelector<HTMLElement>('#myGrid')!, gridOptions);
--- a/documentation/ag-grid-docs/src/content/docs/dev-validation/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/dev-validation/index.mdoc
@@ -1,0 +1,86 @@
+---
+title: "Development Validation"
+description: "Enable the AG Grid ValidationModule for development-time diagnostics: full console messages, an on-grid error overlay, throwing on problems, and suppressing accepted ones."
+---
+
+The `ValidationModule` adds development-time diagnostics that help you catch misconfiguration early. It provides full-text console messages, an overlay that surfaces problems on the grid, and options to fail fast on issues or ignore ones you have accepted.
+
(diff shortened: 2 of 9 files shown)


documentation/ag-grid-docs/src/content/errors/307.mdoc
```diff
@@ -0,0 +1,18 @@
...
+- [Grid Options](./grid-options/) for grid-level options
+- [Column Properties](./column-properties/) for column definitions
+
+### Properties added by a wrapper
+
+A shared component that spreads unrelated props onto your grid options or column definitions surfaces every extra property as this warning. Where you can, remove the unrelated properties at the source.
+
+### Storing your own data
+
+To attach your own data to the grid options or a column, do not add it as a top-level property. Use the [`context`](./context/) property, which is provided on both grid options and column definitions for exactly this.
```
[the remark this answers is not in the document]
alistairjcbrown: `ColDef` does support `context` — it is declared in `colDef.ts` ("Context property that can be used to associate arbitrary application data with this column definition"), the same as on `GridOptions`. So the guidance points at a supported property, not an invalid one.

Concretely, the fix for this page's scenario is:

```js
// triggers #307 — `notAColumnProperty` is an unrecognised top-level colDef key
{ field: 'make', notAColumnProperty: true }

// clean — `context` is a recognised colDef property; its contents are not validated
{ field: 'make', context: { notAColumnProperty: true } }
```

The validator only checks the top-level property names of the colDef against the known set (`context` is in that set); it does not recurse into `context`, which is typed `any` by design. It also only ever suggests `.context` when the validated object actually has a valid `context` property (`hasContext: allValidNames.has('context')`), so it will not send users back into the same warning. Leaving the docs as-is.

documentation/ag-grid-docs/src/content/errors/310.mdoc
```diff
@@ -0,0 +1,9 @@
+## Unrecognised Properties
+
+AG Grid does not recognise a property on the object shown in the message. Usually this is on your grid options or a column definition. Each unrecognised property is also reported on its own as warning #307 with a suggested correction. This message is the summary.
+
+### Resolving
+
+Check each flagged property against the reference for the object it was set on ([Grid Options](./grid-options/) or [Column Properties](./column-properties/)) and correct any typos or misplaced properties.
+
+If the properties come from a shared wrapper that spreads unrelated props, remove them at the source. To attach your own data deliberately, use the [`context`](./context/) property rather than a top-level property.
```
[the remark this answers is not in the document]
alistairjcbrown: Same as the #307 thread: `context` is a valid column-definition property (`ColDef.context`, declared in `colDef.ts` for "arbitrary application data"), not just a grid option. Moving your own data into it is a supported pattern and clears the warning:

```js
{ field: 'make', notAColumnProperty: true }              // ⚠️ #307/#310
{ field: 'make', context: { notAColumnProperty: true } } // ✅ clean
```

Validation checks only the top-level colDef keys (where `context` is recognised) and does not descend into `context`'s value, so this does not re-trigger the warning. Leaving the docs as-is.

alistairjcbrown:
Some things to consider:
* If the error is a deprecation, could we link to the AI update skill
* Are there any dev skills we can link into?

documentation/ag-grid-docs/src/content/docs/dev-validation/_examples/overlay/main.ts
```diff
@@ -0,0 +1,26 @@
+import type { ColDef, GridApi, GridOptions } from 'ag-grid-community';
+import { AllCommunityModule, ModuleRegistry, ValidationModule, createGrid } from 'ag-grid-community';
+
+// Register the ValidationModule (with the overlay enabled) alongside the community modules. Registering
+// it in the modules array — rather than via enableDevValidations — is what lets the framework example
```
StephenCooper: All the examples are now using `enableDevValidations()` instead of the validation module.

documentation/ag-grid-docs/src/content/docs/modules/index.mdoc
```diff
@@ -117,7 +117,7 @@ To work out which modules are required, select the features of the grid that you
 
 ## Development Validations
 
-The `ValidationModule` adds console warnings and errors that help identify misconfiguration during development. It is not included in the `AllCommunityModule`/`AllEnterpriseModule` bundles, so opt into it yourself — and we recommend doing so only in development builds to keep production bundles small.
+The `ValidationModule` provides development-time diagnostics that help identify misconfiguration early: full-text console messages, an on-grid overlay of captured problems (shown by default), and options to throw on problems or suppress ones you have accepted. It is not included in the `AllCommunityModule`/`AllEnterpriseModule` bundles, so opt into it yourself — and we recommend doing so only in development builds to keep production bundles small.
```
StephenCooper: I think we should re-phrase this without calling out the ValidationModule like we currently do.

documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro
```diff
@@ -123,7 +123,12 @@ const migrationUrl = pathJoin(PRODUCTION_GRID_SITE_URL, FRAMEWORK_REDIRECT_PATH,
                 </div>
                 <div class="full-error-message">
                     <pre id="errorCodeText">{defaultErrorText}</pre>
-                    <Note>To show this error text in the dev console, import the <code>ValidationModule</code>.</Note>
+                    <Note>
+                        To show this error text in the dev console, <a
+                            href={urlWithPrefix({ framework, url: './dev-validation' })}
+                            >import the <code>ValidationModule</code></a
```
Comment (StephenCooper) [inline documentation/ag-grid-docs/src/pages/[framework]-data-grid/errors/[code].astro]: Lets update this too away from explicit ValidationModule

documentation/ag-grid-docs/src/content/docs/dev-validation/index.mdoc
```diff
@@ -0,0 +1,86 @@
+---
+title: "Development Validation"
+description: "Enable the AG Grid ValidationModule for development-time diagnostics: full console messages, an on-grid error overlay, throwing on problems, and suppressing accepted ones."
```
StephenCooper: I would make this less verbose and just give a high level overview of the page instead of calling out every feature.

documentation/ag-grid-docs/src/content/docs/dev-validation/index.mdoc
```diff
@@ -0,0 +1,86 @@
...
+- `'error'` — throws on errors
+- `'warning'` — throws on warnings and errors
+- `'deprecation'` — throws on deprecations, warnings and errors
+- `false` (default) — never throws
+
+`throwOn` throws synchronously wherever a matching diagnostic is raised. That is not only at start-up, it also happens during API calls, data updates, and rendering. A throw can interrupt an operation part-way and leave the grid in an inconsistent state. Use `throwOn` with harnesses that recreate the grid on failure rather than carrying on with the same instance, and never in production.
+
+### Suppressing Diagnostics
+
+`suppress` takes the error ids to ignore, for diagnostics you have reviewed and accepted. A suppressed id is kept out of the overlay and is never thrown by `throwOn`, but is still logged to the console once. Suppression is by id, so it silences every diagnostic with that number — not one specific occurrence.
```
StephenCooper, line 78: We probably should add a warning box that we do not recommend suppressing any issues but this is provided as an escape hatch.

StephenCooper (approved):

PR #14310, opened 2026-07-02
[AG-17385] Fix(column-tool-panel): regenerate value column when re-added in deferred pivot mode

sdwvit:
Re-adding a value column in deferred pivot mode (pivot mode active, no active pivot column) updated the value-column model but left the table stale - the aggregation never appeared after Apply.

On commit, `aggFuncs` was applied before `aggregation`; `setColumnAggFunc` added the value column early, turning the same-seq `aggregation` `setColumns` into a no-op that skipped the pivot result-column refresh. Ordering `aggregation` before `aggFuncs` fixes it.

Follow-up to #14276, which enabled the Apply button but not the actual re-add. 

Fix [AG-17385](https://ag-grid.atlassian.net/browse/AG-17385)


alistairjcbrown (approved):
LGTM :shipit:

StephenCooper (approved):

PR #14313, opened 2026-07-02
AG-17741 Fix viewport row model stale index leak and add test coverage

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-17741

Fix [AG-17741](https://ag-grid.atlassian.net/browse/AG-17741)


sdwvit:
i'm glad viewport gets some love!

sdwvit (approved):

PR #14329, opened 2026-07-03
AG-17719 Support Calculated Columns and Show Values As in AI Toolkit

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-17719

Fix [AG-17719](https://ag-grid.atlassian.net/browse/AG-17719)

### DIFF
--- a/documentation/ag-grid-docs/astro.config.mjs
+++ b/documentation/ag-grid-docs/astro.config.mjs
@@ -131,6 +131,25 @@ const {
      * Studio robots.txt disallow json url to merge
      */
     STUDIO_ROBOTS_DISALLOW_JSON_URL,
+
+    /**
+     * Upstream AI Toolkit API, proxied in dev to avoid browser CORS on localhost.
+     * Only used when an example points `AI_API_URL` at the local `/ai-proxy` path.
+     */
+    AG_AI_API_UPSTREAM_URL = 'https://ai-api.ag-grid.com/api/openai/v1',
+
+    /**
+     * Dev token for the AI Toolkit proxy. When set, injected server-side by the dev
+     * proxy so the token is never exposed to example browser code.
+     */
+    AG_AI_API_DEV_TOKEN,
+
+    /**
+     * Origin/Referer the dev proxy presents upstream. The AI proxy's access control is
+     * primarily an origin allow-list, so set this to an allowed origin (e.g. the prod
+     * site) if the upstream rejects the proxied request.
+     */
+    AG_AI_API_DEV_ORIGIN,
 } = dotenvExpand.expand(dotenv).parsed;
 console.log(
     'Astro configuration',
@@ -208,6 +227,30 @@ export default defineConfig({
         plugins,
         server: {
             https: httpsEnabled,
+            /**
+             * Dev-only proxy for the AI Toolkit examples. The public example calls an absolute
+             * `ai-api.ag-grid.com` URL, which browsers block from `localhost` on CORS. Point an
+             * example's `AI_API_URL` at the same-origin `/ai-proxy` path (via `.env.local`) and
+             * this forwards it upstream, adding the dev token server-side if one is configured.
+             */
+            proxy: {
+                '/ai-proxy': {
+                    target: AG_AI_API_UPSTREAM_URL,
+                    changeOrigin: true,
+                    rewrite: (path) => path.replace(/^\/ai-proxy/, ''),
+                    configure: (proxy) => {
+                        proxy.on('proxyReq', (proxyReq) => {
+                            if (AG_AI_API_DEV_TOKEN) {
+                                proxyReq.setHeader('Authorization', `Bearer ${AG_AI_API_DEV_TOKEN}`);
+                            }
+                            if (AG_AI_API_DEV_ORIGIN) {
+                                proxyReq.setHeader('Origin', AG_AI_API_DEV_ORIGIN);
+                                proxyReq.setHeader('Referer', `${AG_AI_API_DEV_ORIGIN}/`);
+                            }
+                        });
+                    },
+                },
+            },
             cors: {
                 /**
                  * CORS allow list for opening examples on external sites
--- a/documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/natural-language-grid-state/chatgptApi.ts
+++ b/documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/natural-language-grid-state/chatgptApi.ts
@@ -109,6 +109,8 @@ Respond with only the necessary state changes and column definition operations, 
 
 Any unchanged properties that are present in the current state must be included in \`propertiesToIgnore\`. Otherwise they will be removed from the state.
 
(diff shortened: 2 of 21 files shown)


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -462,15 +462,17 @@ export class StateService extends BeanStub implements NamedBean {
         if (shouldSetAggregationState && aggregationState) {
             const aggregationModel = aggregationState.aggregationModel;
             for (let i = 0, len = aggregationModel.length; i < len; ++i) {
-                const { colId, aggFunc } = aggregationModel[i];
+                const { colId, aggFunc, showValuesAs } = aggregationModel[i];
                 const columnState = getColumnState(colId);
                 columnState.aggFunc = aggFunc;
```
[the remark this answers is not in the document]
alistairjcbrown: It's already handled. When aggregation state is applied, the apply path also sets `defaultState.showValuesAs = null`, and `syncColState` uses `defaultState` as the fallback for an `undefined` state value on listed columns too (not just unlisted). So an omitted `showValuesAs` falls through to `null` and clears the active mode — no explicit `null` required from the caller.

I've added a regression test that sets a mode, then calls `setState` with an aggregation entry that omits `showValuesAs`, and asserts the mode is cleared. It passes ✅

documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/tool-panel-chat-assistant/systemPrompt.ts
```diff
@@ -14,11 +14,14 @@ The schema provided can be used to manipulate multiple features of the table to
...
 
-Respond with only the necessary state changes, not the complete state. Provide a clear explanation of what you changed.
+Respond with only the necessary state changes and column definition operations, not the complete state. Provide a clear explanation of what you changed.
 
 Any unchanged properties that are present in the current state must be included in \`propertiesToIgnore\`. Otherwise they will be removed from the state.
 
-You are not able to make any changes to the grids configuration, e.g. enabling features, you are only able to modify state.
+When you include a column in \`aggregationModel\`, carry over its current \`showValuesAs\` from the grid state unless the user asked to change it. Nulling it clears a mode the user had set.
+
+Use \`columnDefOperations\` for Calculated Columns. Calculated Columns use formula syntax: reference other columns with bracketed column IDs, for example \`[amount] * 1.2\`, and wrap literal text in double quotes, for example \`"Hello"\`. Do not use valueGetter syntax for Calculated Columns.
```
alistairjcbrown, line 23: Calculated Columns aren't grid state, so they can't go through `getStructuredSchema()`/`setState()` — we bridge them example-side via `columnDefOperations`, applied before `setState()`. Kept at the example level deliberately, to avoid changing the toolkit's `setState` contract.

📣  Open question: what do folks think of this approach, vs. teaching the toolkit to handle this so a single `setState()` creates the column? Feedback welcome!

documentation/ag-grid-docs/astro.config.mjs
```diff
@@ -208,6 +227,30 @@ export default defineConfig({
...
         server: {
             https: httpsEnabled,
+            /**
+             * Dev-only proxy for the AI Toolkit examples. The public example calls an absolute
+             * `ai-api.ag-grid.com` URL, which browsers block from `localhost` on CORS. Point an
+             * example's `AI_API_URL` at the same-origin `/ai-proxy` path (via `.env.local`) and
+             * this forwards it upstream, adding the dev token server-side if one is configured.
+             */
+            proxy: {
+                '/ai-proxy': {
```
alistairjcbrown, line 237: Dev-only proxy: the examples call `ai-api.ag-grid.com`, which the browser CORS-blocks from `localhost`. Point an example's `AI_API_URL` at `/ai-proxy` via `.env.local` to test locally. `vite.server.*` isn't in the production build, so no prod footprint.

packages/ag-grid-community/src/misc/state/stateUtils.ts
```diff
@@ -59,7 +60,7 @@ export function convertColumnState(
             groupColIds[rowGroupIndex ?? 0] = colId;
         }
         if (typeof aggFunc === 'string') {
-            aggregationColumns.push({ colId, aggFunc, valueIndex });
+            aggregationColumns.push({ colId, aggFunc, showValuesAs, valueIndex });
```
alistairjcbrown, line 63: `showValuesAs` has lived in `ColumnState` since it shipped but was never wired into `GridState` — so `getState()` dropped it and `setState()` couldn't restore it.  Follow-up AG-17765 planned which adds a guard against this (`pivotSort` is a second instance).

documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/natural-language-grid-state/chatgptApi.ts
```diff
@@ -36,12 +36,65 @@ export async function callChatGPT(userRequest: string, currentState: any, gridAp
...
+                                    description: 'Stable unique column ID, for example weightedTotal',
+                                },
+                                headerName: {
+                                    type: 'string',
+                                    description: 'Column header shown to users',
+                                },
+                                calculatedExpression: {
+                                    type: 'string',
+                                    description:
+                                        'Calculated Column formula. Reference other columns with bracketed column IDs, for example [gold] * 3 + [silver] * 2 + [bronze], and wrap literal text in double quotes, for example "Hello"',
```
AlpineJosh, line 62: You may need more detail here as to valid operators and functions. It can also be provided as top level context.

documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/natural-language-grid-state/chatgptApi.ts
```diff
@@ -36,12 +36,65 @@ export async function callChatGPT(userRequest: string, currentState: any, gridAp
...
+                                    type: 'number',
+                                    minimum: 20,
+                                    description: 'Column width in pixels',
+                                },
+                                hide: {
+                                    type: 'boolean',
+                                    description: 'Whether the new column should be hidden',
+                                },
+                            },
+                            required: ['operation', 'colId', 'headerName', 'calculatedExpression'],
```
AlpineJosh, line 84: Annoyingly ChatGPT requires that all fields are marked as required. The only way to make them optional is to union with 'null'. See https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas

AlpineJosh (commented):
Overall looks good. A few minor points to clear up the schema for OpenAI.

I do think it might be worth us splitting up structured schema into individual tool schemas. That way we could provided the calculated column schema ourselves rather than requiring devs to manually type it.

documentation/ag-grid-docs/src/content/docs/ai-toolkit/_examples/tool-panel-chat-assistant/systemPrompt.ts
alistairjcbrown: Confirmed, this is the correct approach

PR #14334, opened 2026-07-03
AG-17726: Replace header magic-number breakpoints with a container-query manifest

skewis6:
## Summary

Replaces the shared site-header's hardcoded pixel breakpoints (duplicated/drift-prone across ag-grid, ag-charts, ag-studio) with:

- **Container queries** (`@container site-header`) instead of `@media`, so the header responds to its own rendered width rather than the viewport — it can't tell the difference today, but should, since it's dropped into different page chrome per product.
- **A single manifest** (`site-header/_breakpoints.scss`) naming every header threshold by what it controls, with a comment on every consumer — replacing a flat, unnamed-by-purpose, 90-line sitewide breakpoint list that had already caused two real bugs (see below).
- **A working per-repo override seam**: each of the four header-exclusive tokens (`$nav-collapse`, `$docs-search-inline`, `$switcher-popup-at`, `$search-box-cap-at`) now resolves via a `_site-header-tokens.scss` module found through Sass `loadPaths` — a repo can override any of them by dropping its own file in `documentation/ag-grid-docs/src/styles/header-tokens/`, without forking the shared package. (An earlier `!default`-based version of this didn't actually work — see the last commit — this codebase's Dart Sass `@use`/`@forward` module system doesn't support "declare before your first `@use`", and each `.module.scss` compiles independently anyway.)
- **CSS Grid** for `.headerInner` with named `grid-area`s, replacing flex `order`/margin hacks.
- **`clamp()`** on nav-link padding so the nav compresses gracefully instead of hitting a hard pixel cliff.

Two real, pre-existing bugs fixed along the way:
- The product/site switcher had a visibility gap between 1100–1220px (three independent, uncoordinated breakpoints for switcher-show / nav-collapse / docs-search-inline).
- `Search.module.scss` had three overlapping `max-width` rules where one (the intended 512px cap) was dead code, always overridden by a later rule for every width it could match.

`$nav-collapse` — the number the original problem was actually about, since nav item count differs by product (AG Grid 9 items including GitHub, AG Studio 8, AG Charts in between) — is decoupled from the shared `$breakpoint-site-header-small` token specifically so it can be tuned per repo without moving that value for unrelated consumers (`AnnouncementBanner`, among others).

## Verification

- Compiled every touched `.scss` file directly with the `sass` compiler (not just read) to confirm the module resolution chain works.
- Proved the override mechanism concretely: with no repo override present, the shared default compiles through; with a repo-local `_site-header-tokens.scss` overriding `$nav-collapse`, the override wins completely and `AnnouncementBanner` (an unrelated consumer of the token `$nav-collapse` used to alias) is unaffected.
- Ran a live dev server + Playwright against `ag-grid-docs`: confirmed `container-type: inline-size` is active, nav/switcher/hamburger visibility flips exactly at the documented thresholds, no overlapping or misaligned elements across 400–1440px in both light and dark mode, dark-mode toggling renders correctly on docs pages.

## Not done in this PR

- Not yet pulled into `ag-charts` / `ag-studio` via `git subrepo pull` to confirm the change round-trips cleanly across all three repos.
- `$sticky-at` (header sticky-on-scroll) intentionally left as a plain shared breakpoint — a container can't query its own size for its own position, and it's a uniform product decision, not a per-site number.

## Test plan

- [ ] Visually check the header on `ag-grid-docs` across a range of widths (mobile through wide desktop), light and dark mode
- [ ] Confirm the product switcher shows/hides correctly with no gap around 960–1220px
- [ ] Confirm the search box's max-width behavior is unchanged from current production
- [ ] Pull into `ag-charts`/`ag-studio` and confirm no visual regression before relying on the per-repo override capability


markdurrant (commented):
Generally looking good. A few things I'd like to review/update though...

- I like having the `src/styles/header-tokens` at the repo level. However I think it would be good to have an actual or dummy file here as well as the Readme to make it super-duper clear how to use overrides. 
- `[_site-header-tokens.scss](https://github.com/ag-grid/ag-grid/pull/14334/changes#diff-f5910144d09d9bd9c54002238ba783068430af9530fa22c5400f90d32d38ab5a)` living in website-shared `/src/` seems a bit low level? Would it not be in `/design-system/`? 

Then a nit. I like the extra explanation comments a lot. But they seem very verbose right now. I think I'm becoming blind too these paragraph long comments. Maybe shorten up?

markdurrant (approved):
LGTM 👍

PR #14337, opened 2026-07-06
AG-17779 Fix React pinned rows not rendering on state update

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-17779

Fix [AG-17779](https://ag-grid.atlassian.net/browse/AG-17779)


alistairjcbrown (approved):
May be worth having someone with more context review, but the fix makes sense to me and nice tests to confirm -- LGTM :shipit:

PR #14351, opened 2026-07-07
AG-17727 Move Products dropdown left of Demos and add Contact Us to top nav

markdurrant:
## Summary

Implements [AG-17727](https://ag-grid.atlassian.net/browse/AG-17727).

- **Move the Products dropdown** so it sits directly to the left of **Demos** in the top navigation. Achieved by reordering the `site-header` grid template (`logo docs-search switcher nav`) so the product/site switcher region sits immediately before the nav. On docs pages the search cluster takes the freed flexible column next to the logo.
- **Add a Contact Us item** to the top navigation, positioned to the right of **Pricing**, linking to `/about/#contact-section` (matching the existing footer/demo links to the same anchor).

## Acceptance criteria

- **AC1** — Products dropdown appears immediately to the left of Demos ✅
- **AC2** — Contact Us appears to the right of Pricing ✅
- **AC3** — Contact Us links to `about/#contact-section` ✅

## Verification

Verified in the dev server on both the homepage and a docs page (React Quick Start) — DOM ordering and rendered layout confirm Products → Demos adjacency and Pricing → Contact Us adjacency, with the correct href.

Note: the grid reorder lives in the shared `external/ag-website-shared` site-header, so the same layout change applies to other AG sites consuming it.

### DIFF
--- a/documentation/ag-grid-docs/src/content/site-header/header.json
+++ b/documentation/ag-grid-docs/src/content/site-header/header.json
@@ -36,6 +36,10 @@
                 "path": "/license-pricing"
             },
             {
+                "title": "Contact Us",
+                "path": "/about/#contact-section"
+            },
+            {
                 "title": "GitHub",
                 "url": "https://github.com/ag-grid/ag-grid",
                 "icon": "github"
--- a/documentation/ag-grid-docs/src/styles/header-tokens/_site-header-tokens.scss
+++ b/documentation/ag-grid-docs/src/styles/header-tokens/_site-header-tokens.scss
@@ -1,7 +1,15 @@
-// No-op — forwards every header token unchanged. To override one or more for this
-// repo, delete the line below and uncomment the example beneath it.
-@forward '../../../../../external/ag-website-shared/src/design-system/site-header-tokens';
-
-// @forward '../../../../../external/ag-website-shared/src/design-system/site-header-tokens' with (
-//     $nav-collapse: 900px
-// );
+// AG Grid overrides for the header's per-repo-tunable breakpoints (see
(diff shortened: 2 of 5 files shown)


markdurrant:
### Follow-up: breakpoint fixes

Adding the 7th nav item ("Contact Us") pushed the header past the widths the layout was tuned for, causing two malformations:

- **Flex band (~960–1180px):** the wrapping row + `justify-content: space-between` stranded the Products switcher alone on the logo's line, detached from the nav.
- **Grid mode (~1180–1280px):** the extra item squeezed the inline search box until it truncated ("Searc⌘K").

Fixed by:
1. Raising the **Grid-only** header token overrides (`$nav-collapse` 960→1120, `$docs-search-inline` 1180→1300) so the header collapses to the mobile hamburger while the full nav + switcher can't share one row, and only goes fully inline once search has usable width.
2. Anchoring the switcher to the nav with `margin-left: auto` (was `margin-right: auto`) so, while the header is a wrapping flex row, the logo sits left and switcher + nav form one right-aligned group — the switcher can never be stranded mid-row or on its own line. Inert in grid mode (grid-area handles placement).

**Verified** in-browser at 1000 / 1130 / 1150 / 1200 / 1320 / 1350 / 1440px on both docs and homepage, plus the mobile menu (opens with all items incl. Contact Us).

⚠️ **Shared-component note:** change #2 is in `external/ag-website-shared` (`ProductDropdown.module.scss`), so it also affects AG Charts/Studio — the switcher moves from logo-adjacent to nav-adjacent there too, consistent with this ticket's intent (ticket lists both Charts & Grid). Change #1 is Grid-only (token override). Worth a quick visual check on Charts when syncing.

kylerphillips (approved):

PR #14360, opened 2026-07-07
AG-17641 Dev-validation overlay: severity sections + grid attribution

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-17641

Fix [AG-17641](https://ag-grid.atlassian.net/browse/AG-17641)

### DIFF
--- a/packages/ag-grid-community/src/main-internal.ts
+++ b/packages/ag-grid-community/src/main-internal.ts
@@ -376,7 +376,16 @@ export { _clamp, _formatNumberCommas, _isFiniteNumber, _toFiniteNumber } from '.
 export { _selectAllCells } from './utils/selection';
 export { _getValueUsingDotPath } from './utils/value';
 export type { LogService } from './validation/logService';
-export { _deprecated, _errMsg, _error, _logPreInitWarn, _preInitErrMsg, _warn } from './validation/logging';
+export {
+    _deprecated,
+    _errMsg,
+    _error,
+    _errorForGrid,
+    _logPreInitWarn,
+    _preInitErrMsg,
+    _warn,
+    _warnForGrid,
+} from './validation/logging';
 export { _emitValidationWarning, _createValidationWarning } from './validation/validationTypes';
 export type { ValidationWarning } from './validation/validationTypes';
 export type { ExpressionService } from './valueService/expressionService';
--- a/packages/ag-grid-community/src/main-internal.ts
+++ b/packages/ag-grid-community/src/main-internal.ts
@@ -377,14 +377,13 @@ export { _selectAllCells } from './utils/selection';
 export { _getValueUsingDotPath } from './utils/value';
 export type { LogService } from './validation/logService';
 export {
-    _deprecated,
     _errMsg,
-    _error,
     _errorForGrid,
+    _errorWithoutAttribution,
     _logPreInitWarn,
     _preInitErrMsg,
-    _warn,
     _warnForGrid,
+    _warnWithoutAttribution,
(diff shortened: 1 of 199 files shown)


packages/ag-grid-community/src/main-internal.ts
```diff
@@ -375,7 +375,17 @@ export { _isProtoPollutionKey, _mergeDeep, _mergedEqual } from './utils/mergeDee
...
-export { _deprecated, _errMsg, _error, _logPreInitWarn, _preInitErrMsg, _warn } from './validation/logging';
+export type { LogService } from './validation/logService';
+export {
+    _deprecated,
+    _errMsg,
+    _error,
+    _errorForGrid,
+    _logPreInitWarn,
+    _preInitErrMsg,
+    _warn,
```
StephenCooper: So are there still valid uses of `_warn`, `_error` .... that are not using the bean approach?

packages/ag-grid-enterprise/src/filterToolPanel/newFilterToolPanel/wrapperToolPanel.ts
```diff
@@ -24,7 +24,7 @@ export class WrapperToolPanel extends Component implements INewFiltersToolPanel,
 
     public init(params: ToolPanelNewFiltersCompParams): void {
         if (!this.gos.get('enableFilterHandlers')) {
-            _warn(282);
+            this.beans.log.warn(282);
```
StephenCooper, line 27: Maybe we should also add the methods to the base grid component too?

alistairjcbrown: I went down this route and had to backtrack. `Component` extends `AgComponentStub` from `ag-stack`, and `ag-stack` provides a bunch of `AgComponentStub` based widgets. In the grid, `Component` can therefore be used as both a base to extend, and a parameter type which still works with `ag-stack` widgets.

If we add methods to the base type, then that parameter type check starts failing in a bunch of places (`ag-stack` widgets are no longer compatible with `Component` type). I can dig into it, but it didn't end up as a clean addition like adding to `BeanStub` did.

packages/ag-grid-community/src/main-internal.ts
alistairjcbrown: They're used in a select few places when a bean doesn't exist:
E.g.
* `_error(11);` triggered when "No gridOptions provided, abort creating the grid"
* `_warn(26, { fnName, preDestroyLink });` triggered when " No grid context (grid destroyed)"

packages/ag-grid-vue3/src/components/AgGridVue.vue
```diff
@@ -97,12 +96,6 @@ const updateModelIfUsed = (eventType: string) => {
 };
 // v-model code end
 
-const checkForBindingConflicts = () => {
-    if ((props.rowData || props.gridOptions.rowData) && rowDataModel.value) {
-        _warn(232);
-    }
-};
-
```
alistairjcbrown, line 105: This appears to be dead code

packages/ag-grid-enterprise/src/filterToolPanel/newFilterToolPanel/wrapperToolPanel.ts
StephenCooper: Ok, no need to do this.

StephenCooper (approved):
Looks good.

PR #14366, opened 2026-07-08
AG-16527 Contract header height when auto-header-height is toggled off

StephenCooper:
Fix [AG-16527](https://ag-grid.atlassian.net/browse/AG-16527) — Header height does not contract when auto-header-height is toggled off

### Problem
Toggling `autoHeaderHeight` off (via `setGridOption('columnDefs', ...)`) left the header row at its previously-measured (grown) height. It never contracted back to the default single-line height.

### Root cause
When auto-height is switched off, `checkMeasuring()` calls `stopMeasuring()` in `abstractHeaderCellCtrl.ts`, which tore down the ResizeObserver but **never cleared the column's stored `autoHeaderHeight` nor dispatched `columnHeaderHeightChanged`**. With no recompute trigger, the header row kept the stale grown height.

### Fix
`stopMeasuring()` now clears the stored height (`setColHeaderHeight(column, null)`), which dispatches `columnHeaderHeightChanged` and forces a recompute back to the default height.

The clear is **guarded** on `!column.isAutoHeaderHeight()` so it only runs when auto-height is genuinely off. `stopMeasuring()` also fires on header-cell teardown (e.g. hiding a column) while the column is *still* auto-height — an unconditional clear there would wipe the measured height and emit a spurious `columnHeaderHeightChanged`, causing a re-measure and shrink→regrow flicker on show. The guard prevents that.

`setAutoHeaderHeight` on `AgColumn`/`AgColumnGroup` and `setColHeaderHeight` were widened to accept `number | null` (the backing fields were already `number | null`).

### Tests
- **Behavioural — toggle off clears** (`column-features.test.ts`): measure with auto-header on, remove `autoHeaderHeight` via `columnDefs`, assert stored height returns to `null`, `isAutoHeaderHeight()` is `false`, and `columnHeaderHeightChanged` fires. Red without the fix.
- **Behavioural — hide retains** (`column-features.test.ts`): hiding a still-auto column keeps its measured height and emits no event. Red without the guard.
- **React** (`auto-header-height-toggle-react.test.tsx`): regression guard for the twinned `*Ctrl` under React's async mount. Red without the fix.

Column-group auto-height was investigated and needs no test: group instances are rebuilt from scratch on a `columnDefs` change, so they never carried a stale height.

### Verification
- `yarn nx format`, `nx lint ag-grid-community`, `nx build:types ag-grid-community`
- `nx test ag-grid-community` (1101), `column-features` (41), `columns/headers` (12) — all green


sdwvit (approved):

.rulesync/rules/testing.md
```diff
@@ -63,6 +63,8 @@ Behavioural tests in `testing/behavioural/` are the primary test suite for verif
 ./behave.sh --watch
 ```
 
+> `./behave.sh` does not type-check (Vitest strips types via esbuild). Before committing, run `yarn nx run ag-behavioural-testing:build:test` to type-check.
```
sdwvit, line 66: should it be added to ag prompts

StephenCooper: This note is grid-specific — it points at the `ag-behavioural-testing:build:test` Nx target, which only exists in this repo. The shared ag-prompts testing guidance is product-agnostic, so I've kept the behave.sh type-check step in the local `.rulesync` override rather than promoting it upstream.

PR #14380, opened 2026-07-09
AG-17820 Add container-based image-snapshot testing for Integrated Charts

Stephanemw:
https://ag-grid.atlassian.net/browse/AG-17820

Renders Integrated Charts under jsdom via a real Skia rasterizer (OffscreenCanvas/requestAnimationFrame polyfills in `canvasPolyfill.ts`) and pixel-diffs the output with `jest-image-snapshot`.

Skia's font/AA rendering isn't identical across host OSes, so the new `test-chart-snapshots` nx target and its dedicated `vitest.chart-snapshots.config.ts` only ever run inside a container pinned by image digest — excluded from the regular cross-platform `test` target via `vitest.config.ts`.

`testing/behavioural/docker/run-chart-snapshots.sh` wraps the container invocation: bind-mounts the repo read-write, shadows every `node_modules` directory with its own named volume so Linux-native binaries never touch the host install, and mounts `yarn.lock` read-only.

`vitest.source-aliases.ts` extracts the source-aliasing logic shared between the two vitest configs.

CI wiring is not part of this PR — tracked as follow-up on the ticket.

Fix [AG-17820](https://ag-grid.atlassian.net/browse/AG-17820)


StephenCooper (approved):
Happy to see how this works going forward. Thanks!

PR #14393, opened 2026-07-10
Fix flaky SSRM blockLoadDebounceMillis characterization test

sdwvit:
## What

Fixes an intermittent CI failure in `testing/behavioural/src/row-data/ssrm-cache-config.test.ts` — the `blockLoadDebounceMillis coalesces rapid scroll-driven block loads within the debounce window` test.

Reported flaky on `latest`: `AssertionError: expected 0 to be greater than 0` at line 161.

## Why

The test used a fixed `asyncSetTimeout(400)` to wait for the 200ms-debounced block load to fire, then asserted a request had happened. On a congested CI box, real `setTimeout` callbacks get delayed and the block-load pipeline (debounce timer → row-render refresh → `getRows`) can still be in flight when the fixed sample fires — so `requestsAfterWindow` reads 0 and the assertion fails.

## Fix

Replace the fixed post-debounce sleep with `await waitForEvent('modelUpdated', api)`, the same deterministic pattern already used by the sibling `ssrm-block-loading.test.ts`. This waits exactly until the debounced load completes, regardless of CI speed. The 50ms in-window sample stays (timers fire in expiry order, so it always samples before the 200ms debounce — that assertion isn't timing-fragile).

Verified: passes 3× in isolation and the full file (5/5) passes.

### DIFF
--- a/testing/behavioural/src/row-data/ssrm-cache-config.test.ts
+++ b/testing/behavioural/src/row-data/ssrm-cache-config.test.ts
@@ -155,8 +155,10 @@ describe('SSRM cache config', () => {
         const requestsDuringWindow = requests.length - requestsAfterInitial;
         expect(requestsDuringWindow).toBe(0);
 
-        // After the debounce window elapses the (final) block load fires.
-        await asyncSetTimeout(400);
+        // After the debounce window elapses the (final) block load fires. Wait for
+        // the load to actually complete (modelUpdated) rather than a fixed sleep, so
+        // a congested CI box cannot sample before the debounced fetch has fired.
+        await waitForEvent('modelUpdated', api);
         const requestsAfterWindow = requests.length - requestsAfterInitial;
         expect(requestsAfterWindow).toBeGreaterThan(0);
         // The debounce coalesced the three rapid hops into far fewer than three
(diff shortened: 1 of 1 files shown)


testing/behavioural/src/row-data/ssrm-cache-config.test.ts
alistairjcbrown, line 154: ❓ Just confirming, this `asyncSetTimeout` is ok to stay? Do we know we'll be "inside the debounce window" 50ms later?
(This feels like it works as expected, and doesn't hit the same issue as below because we' want to trigger before the update, not wait for the update the complete)

alistairjcbrown (approved):
One question, but this LGTM :shipit:  -- much nicer waiting on the event than on an arbitrary timeout 👏

testing/behavioural/src/row-data/ssrm-cache-config.test.ts
sdwvit: I also checked whether setTimeout(0) would make sense here.

In SSRM with the lazy store, getRows is reached through two nested setTimeout callbacks:

queueLoadCheck → setTimeout(0) → queueLoadAction → setTimeout(blockLoadDebounceMillis) → getRows

For this test, the intermediate assertion must run after the first timeout has set up the debounce, but before the 200 ms debounce fires.

setTimeout(50) gives enough margin on both sides, so it reliably verifies that no load happens during the debounce window.

setTimeout(0) is too close to the lower boundary. If ensureIndexVisible is ever deferred, the assertion could run before the debounce is even scheduled, making the test pass without actually verifying the intended behaviour.

PR #14395, opened 2026-07-10
Speed up documentation tests CI workflow

StephenCooper:
Reduces per-job setup time in the Documentation Tests workflow from ~4.5 min to ~50s, and shards the two slowest (all-browser) jobs.

The framework example tests run Playwright against a remote deployed site, so they don't need the example generators, `setup-prompts`, or a local grid build beyond the `ag-grid-community` package the test harness imports at runtime.

**Changes**

- Add an `initialise` job that warms the `node_modules` and Nx caches once (rw) and pre-builds `ag-grid-community`. The framework test jobs `needs: initialise` and consume the caches read-only, so the expensive install happens once instead of racing across all jobs.
- Skip `postinstall` in the test jobs (no generators / `setup-prompts` — the latter was failing in CI anyway), building only `ag-grid-community` (served from the warmed Nx cache) for the harness import.
- Run Playwright via `./docs-e2e.sh`, bypassing the Nx target's grid-build `dependsOn`.
- Cache Playwright browser binaries keyed on version + browser set, so downloads become a near-noop.
- Shard `vanilla` into 2 and `reactFunctionalTs` into 3 to cut wall-clock on the all-browser jobs. (They both run all browsers instead of just Chromium)

Validated with a typescript dispatch run: `node_modules` and Nx caches hit on the exact key, community build was a 2s cache hit, setup dropped to ~50s, and 1555 tests passed (the single failure is a pre-existing content assertion unrelated to this change).

### DIFF
--- a/.github/actions/test-framework-examples/action.yml
+++ b/.github/actions/test-framework-examples/action.yml
@@ -27,18 +27,60 @@ runs:
       with:
         fetch-depth: 1
 
+    # These tests run Playwright against a remote deployed site (base_url), so they need
+    # node_modules for Playwright + the test specs, but NOT the example generators or
+    # setup-prompts. The `initialise` job warms the node_modules cache (rw); here we consume
+    # it read-only and skip postinstall. nx_restore stays on so the ag-grid-community build
+    # below is served from the Nx cache. The test harness (src/utils/grid/test-utils.ts)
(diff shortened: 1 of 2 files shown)


StephenCooper:
Reopening on a fresh branch (sc/doc-tests-ci-speedup); this branch was reused from merged PR #14394.

PR #14399, opened 2026-07-10
Replace placeholder SSRM example specs with behavioural tests

sdwvit:
## What

All 64 Server-Side Row Model (SSRM) documentation examples shipped an auto-generated placeholder `example.spec.ts` that only verified the grid mounted (`ensureGridReady` + `waitForGridContent` + `clickAllButtons`) and asserted nothing about the feature the example demonstrates.

This replaces every one with a real Playwright test that asserts the documented behaviour, grounded in each page's `index.mdoc` prose and the example source:

- **datasource / configuration** — on-demand block loading, cache block size/limits, block-load debounce, initial scroll position, `applyServerSideRowData`
- **sorting / filtering** — server-side sort reload, client-side sort, text/set/advanced filters, batched global apply
- **grouping** — lazy child loading, child counts, complex-object keys, grand totals, group footers, hide-open-parents, open-by-default, `ssrmExpandAllAffectsAllRows`, unbalanced groups
- **pivoting / pagination / selection / row height / retry / master-detail / tree data**
- **updating** — refresh (grid & group routes), single-row set/update, sync & async transactions

## Why

The placeholders gave the appearance of coverage while verifying none of the SSRM behaviour the examples exist to demonstrate; a regression in any of these features would not have failed a test.

## Testing

All 64 specs pass on chromium across **typescript, react, angular and vue3** (`FRAMEWORK=<fw> npx playwright test --project=chromium "server-side-model"`), verified with `retries=0`. Two specs (`resetting-row-height`, `transactions-tree-data`) and `expand-all-affects-all-rows` were hardened for cross-framework/timing determinism.

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-datasource/_examples/simple/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-datasource/_examples/simple/example.spec.ts
@@ -1,5 +1,9 @@
 import { expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
+// A row index comfortably past the first 100-row block, so reaching it proves a further block was
+// fetched on demand while scrolling. Kept low enough to reach reliably within the test budget.
+const DEEP_INDEX = 60;
+
 test.agExample(import.meta, () => {
     test.eachFramework('server-side datasource loads the first block and more on scroll', async ({ page }) => {
         await waitForGridContent(page);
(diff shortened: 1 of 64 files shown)


sdwvit:
> /pr-review

documentation/ag-grid-docs/src/content/docs/server-side-model-datasource/_examples/simple/example.spec.ts
```diff
@@ -1,11 +1,61 @@
-import { clickAllButtons, ensureGridReady, test, waitForGridContent } from '@utils/grid/test-utils';
+import { expect, test, waitForGridContent } from '@utils/grid/test-utils';
+
+// A row index comfortably past the first 100-row block, so reaching it proves a further block was
+// fetched on demand while scrolling. Kept low enough to reach reliably within the test budget.
+const DEEP_INDEX = 60;
```
StephenCooper: The comment does not seem to match up with the value of 60 being less than 100.

sdwvit: addressing , ty

StephenCooper (approved):

PR #14407, opened 2026-07-13
SE-67 - Remove link tag from localhost links

taktran:
For better SEO, as localhost links don't go anywhere for crawlers.


markdurrant (approved):
LGTM 👍

PR #14410, opened 2026-07-13
fix(ci): tolerate network-cache read race in docs e2e tests

StephenCooper:
## Problem

The `Documentation Tests` workflow has been red on `latest` (runs [29222416499](https://github.com/ag-grid/ag-grid/actions/runs/29222416499) and prior). The failures are `Error: ENOENT: no such file or directory, open '.playwright-network-cache/.../body.json'` plus a cascade of `toBeVisible` / `toContainText` / click **timeouts** across otherwise-healthy examples.

## Root cause

The docs example tests cache third-party CDN responses via `playwright-network-cache` to a per-job directory. With `workers: 4` against a **cold** cache, workers race to populate the same URL:

- `saveResponse` writes `headers.json` **then** `body.json` as two separate steps.
- A concurrent worker whose `isExpired()` check sees the freshly-written `headers.json` mtime decides the entry is cached, reads it, and hits **`ENOENT`** on the not-yet-written `body.json`.

That aborts the request, so the grid never receives its CDN asset/data → the downstream render assertions time out on unrelated examples. The examples themselves are fine (verified rendering correctly on staging).

## Fix

Patch `playwright-network-cache`'s `fetchFromCache` to fall back to a live server fetch when the cache files are missing (`ENOENT`), instead of throwing. Any other error still propagates. On a miss the worker simply does what it would on a cold cache — a live fetch — so behaviour is unchanged apart from the race no longer failing the run.

Upstream is at its latest release (`0.3.0`) with the race unfixed; issue #6 (already referenced in `test-utils.ts`) is the same class of concurrency limitation, so a version bump isn't available.

## Validation

- Red/green reproduction of the exact race (headers present, body absent, fresh mtime): pristine `0.3.0` throws `ENOENT ... body.json`; patched build falls through to a live fetch.
- Applies cleanly via the repo `patch-package` flow (`playwright-network-cache@0.3.0 ✔`).
- Cut-down `Documentation Tests` run on this branch (vanilla, `aggregation-columns`): [29250419858](https://github.com/ag-grid/ag-grid/actions/runs/29250419858).

## Notes

A pre-warm approach (populate the cache once before the parallel workers) was prototyped but abandoned: `www.ag-grid.com`'s WAF 403s the replayed warm fetches on CI runners, so it couldn't cover the actual failure surface. This patch is host-agnostic and covers it directly.

### DIFF
--- a/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
+++ b/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
@@ -105,6 +105,9 @@ const excludeErrors = [
     'Layout was forced before the page was fully loaded. If stylesheets are not yet loaded this may cause a flash of unstyled content.',
     'Request to access cookie or storage on “<URL>” was blocked because it came from a tracker and Enhanced Tracking Protection is enabled.',
     'This site appears to use a scroll-linked positioning effect.',
+    // Timing-dependent browser warning: a preloaded font occasionally isn't consumed within the
+    // browser's few-second window (e.g. under CI load), emitting a benign warning unrelated to the grid.
+    'preloaded with link preload was not used within a few seconds',
     // AG-17134: staging serves a Content-Security-Policy-Report-Only header with no report endpoint
     // (violations are captured via the securitypolicyviolation JS listener, not report-uri/report-to).
     // Browsers warn that such a policy can't report — benign for our validation window.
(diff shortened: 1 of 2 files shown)


alistairjcbrown:
Should we put an issue upstream -- i.e. only consider a cache hit when both header and body fails are available?

I'm assuming for our setup we don't want to or can't give each worker its own cache location via `baseDir`? -- saves us patching at the expense of ~4x more network requests

SalvatorePreviti (approved):

sdwvit (approved):

PR #14411, opened 2026-07-13
[AG-115] Feat(column-header): editable column header name via UI

sdwvit:
Enterprise feature: rename a column's header from the UI.

### Implemented
- **`editableHeaderName` colDef option** (community type + validation registry).
- **Header-name override** on `AgColumn` (`setHeaderNameOverride`), preferred by `ColumnNameService` and persisted in **column state** (`ColumnStateParams.headerName`, cleared on reset).
- **`ColumnHeaderEditModule`** (enterprise): `ColumnHeaderEditService`, `setColumnHeaderName` API, and a modal editor popup — Enter commits, Esc cancels, empty reverts to the colDef value.
- **Two entry points**: the **Edit Column Name** column-menu item (own section, edit icon) and the columns tool panel right-click context menu.
- **Locales**: new keys (`editColumnName`, `ariaColumnHeaderNameEditor`) added to en-US and translated across all locale files (key parity verified).
- **Docs**: `column-headers` page section + runnable example (`editable-header-name`) with an `example.spec.ts` covering rename-via-menu and the non-editable case.
 

Fix [AG-115](https://ag-grid.atlassian.net/browse/AG-115)


packages/ag-grid-enterprise/src/columnHeaderEdit/columnHeaderEditPopup.ts
```diff
@@ -0,0 +1,125 @@
...
+    public setValue(value: string): void {
+        this.eEditor.setValue(value, true);
+    }
+}
+
+export class ColumnHeaderEditPopup extends BeanStub {
+    private dialog?: Dialog;
+    private contentComp?: ColumnHeaderEditContent;
+    private saveOnClose = true;
+    private closed = false;
```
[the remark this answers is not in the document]
sdwvit: addressed, closing the dialog now discards edit

packages/ag-grid-community/src/api/gridApi.ts
```diff
@@ -1201,6 +1201,16 @@ export interface _NotesGridApi {
...
 }
 
+/** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
+export interface _ColumnHeaderEditGridApi {
+    /**
+     * Set (or clear, by passing `null`) the edited header name for a column.
+     * The value is persisted as part of column state.
+     * @agModule `ColumnHeaderEditModule`
+     */
+    setColumnHeaderName(key: string | Column, headerName: string | null): void;
```
StephenCooper: Do we need this api level function when it is possible to update the colDef via existing methods.

documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
```diff
@@ -139,6 +139,12 @@
             }
         },
         "headerName": {},
+        "editableHeaderName": {
```
StephenCooper: I would suggest headerNameEditable is more consistent with existing properties, for example groupRowEditable,

packages/ag-grid-community/src/api/gridApi.ts
sdwvit: extra abstraction, but I am not opposed to dropping it. I stay corrected on the earlier statement that this is for cross community-enterprise comms, which is wrong.

documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
sdwvit: makes more sense, ty

packages/ag-grid-community/src/columns/columnStateUtils.ts
```diff
@@ -61,6 +61,8 @@ export interface ColumnStateParams {
     /** The active "Show Values As" selection: the mode name, or the object form (with `params` / `precision`)
      *  for modes that take input. `null` clears it. */
     showValuesAs?: ShowValuesAsStateValue;
+    /** User-edited header name overriding `colDef.headerName`. `null` reverts to the colDef value. */
```
StephenCooper, line 64: Do you need to update the column state docs page to include this new property?

packages/ag-grid-community/src/entities/colDef.ts
```diff
@@ -1261,6 +1268,7 @@ export type ValueGetterFunc<TData = any, TValue = any, TContext = any> = (
 export type HeaderLocation =
     | 'chart'
     | 'columnDrop'
+    | 'columnHeaderEdit'
```
StephenCooper: Why is this its own location?

packages/ag-grid-community/src/columns/columnStateUtils.ts
sdwvit: it's autogenerated from jsdocs luckly https://localhost:4610/javascript-data-grid/column-state/#reference-ColumnStateParams-headerName

packages/ag-grid-community/src/entities/colDef.ts
sdwvit: because in case there is a way to edit header name externally, I want to also update header name value in the popup. so 2 way data bind

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelColumnComp.ts
```diff
@@ -206,6 +207,13 @@ export class ToolPanelColumnComp extends Component {
         setAllColumns(this.beans, [this.column], nextState, 'toolPanelUi', this.params);
     }
 
+    private onColDefChanged(): void {
```
StephenCooper, line 211: I think we might want a new event that is more specific to the headerName changing. This event being triggered is going to cause other parts of the code base to fire which may not be required. 

But you may have to validate where this event is handled and if we would need the headerName event to be added.

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelContextMenu.ts
```diff
@@ -262,12 +265,29 @@ export class ToolPanelContextMenu extends Component {
...
+            this.allowEditHeaderName
+        );
     }
 
     private getMappedMenuItems(): MenuItemDef[] {
         const ret: MenuItemDef[] = [];
         const { menuItemMap, columns, displayName, beans } = this;
+
+        if (this.allowEditHeaderName) {
+            const localeTextFunc = this.getLocaleTextFunc();
```
StephenCooper: I would add a method to the column header svc to provide this. Keeps that isolated to the module.

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -538,6 +539,16 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                               action: () => calculatedColsSvc.removeCalculatedColumn(column),
                           }
                         : null;
+                case 'editColumnName': {
+                    if (!colHeaderEditSvc || !column?.colDef.headerNameEditable) {
+                        return null;
+                    }
+                    return {
+                        name: localeTextFunc('editColumnName', 'Edit Column Name'),
```
StephenCooper: Re-use the same helper method on the svc.

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -838,6 +841,15 @@ export class AgColumn<TValue = any>
         return this.allowFormula;
     }
 
+    /** Override the displayed header name. Pass `null` to revert to the `colDef` value. */
+    public setHeaderNameOverride(headerName: string | null, source: ColumnEventType): void {
+        if (this.headerNameOverride === headerName) {
+            return;
+        }
+        this.headerNameOverride = headerName;
+        this.dispatchColEvent('colDefChanged', source);
```
StephenCooper: Is colDefChanged the right event to be raised here?

packages/ag-grid-community/src/entities/colDef.ts
StephenCooper: This would be a breaking change for strict typescript users.

documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
```diff
@@ -37,6 +37,27 @@ The parameters for `headerValueGetter` differ from a [Cell Value Getter](./value
...
+## Editable Header Name {% enterpriseIcon /%}
+
+Set `headerNameEditable: true` on a `ColDef` to let users rename that column's header from the UI. This is an AG Grid Enterprise feature.
+
+{% apiDocumentation source="column-properties/properties.json" section="header" names=["headerNameEditable"] /%}
+
+Editable columns can be renamed via:
+
+- The **Edit Column Name** item in the [Column Menu](./column-menu/).
+- Right-clicking the column in the [Columns Tool Panel](./tool-panel-columns/) and choosing **Edit Column Name**.
```
StephenCooper, line 49: So the Edit Column Name menu item is being added to the default items for the Columns Tool Panel? That currently has a much smaller set of default items. Let's confirm with product if not already done so.

packages/ag-grid-community/src/events.ts
```diff
@@ -1042,7 +1042,8 @@ export type ColumnEventType =
     | 'cellDataTypeInferred'
     | 'rowNumbersService'
     | 'viewportSizeFeature'
-    | 'calculatedColumn';
+    | 'calculatedColumn'
+    | 'uiColumnHeaderEdit';
```
StephenCooper: Nit: group with the other `ui*` events.

packages/ag-grid-community/src/entities/colDef.ts
sdwvit: reverted!

documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
sdwvit: yes, that's fine, it is gated by the feature flag

documentation/ag-grid-docs/src/content/docs/column-headers/_examples/editable-header-name/main.ts
```diff
@@ -0,0 +1,45 @@
...
+];
+
+let gridApi: GridApi<IOlympicData>;
+
+const gridOptions: GridOptions<IOlympicData> = {
+    columnDefs: columnDefs,
+    defaultColDef: {
+        width: 170,
+    },
+    sideBar: 'columns',
```
StephenCooper, line 55: Sidebar enables the row groups and pivot. We should either enable the columns for those or restrict the sideBar to only show the columns. Currently the example feels a bit restricted.

StephenCooper: Oh this is only the case for Javascript due to it having all the modules. I see in Typescript that those sections are automatically hidden based on the available modules.

documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
```diff
@@ -37,6 +37,27 @@ The parameters for `headerValueGetter` differ from a [Cell Value Getter](./value
...
+
+Editable column groups can be renamed via the **Edit Column Name** item in the group header right-click menu or the Columns Tool Panel context menu.
+
+Both open the same editor. Press `Enter` to commit or `Escape` to cancel. Committing an empty value sets an empty header name; the header reverts to its Column Definition default only when the edit is cleared programmatically, such as `resetColumnState()`. Edited column names are persisted as part of [Column State](./column-state/) and [Grid State](./grid-state/); edited group names are persisted as part of Grid State. Both survive save and restore.
+
+{% note %}
+An edited name takes priority over any `headerValueGetter` on the column. Once the user has provided a custom header name, the `headerValueGetter` is no longer called for that column.
```
StephenCooper, line 56: I am wondering if we should have this example support save / restore state seeing as we call it out. 

Potentially also the headerValueGetter behaviour too?

I would say we should definitely add Column Groups into the example too.

documentation/ag-grid-docs/src/content/docs/column-headers/index.mdoc
```diff
@@ -37,6 +37,27 @@ The parameters for `headerValueGetter` differ from a [Cell Value Getter](./value
...
+{% apiDocumentation source="column-properties/properties.json" section="header" names=["headerNameEditable"] /%}
+
+Editable columns can be renamed via:
+
+- The **Edit Column Name** item in the [Column Menu](./column-menu/).
+- Right-clicking the column in the [Columns Tool Panel](./tool-panel-columns/) and choosing **Edit Column Name**.
+
+Editable column groups can be renamed via the **Edit Column Name** item in the group header right-click menu or the Columns Tool Panel context menu.
+
+Both open the same editor. Press `Enter` to commit or `Escape` to cancel. Committing an empty value sets an empty header name; the header reverts to its Column Definition default only when the edit is cleared programmatically, such as `resetColumnState()`. Edited column names are persisted as part of [Column State](./column-state/) and [Grid State](./grid-state/); edited group names are persisted as part of Grid State. Both survive save and restore.
```
StephenCooper: I think we should update at least one of the column-state examples to include column name editing. This will both keep it tested and also show users about how the state works if they have followed the link from here. 

The column events one would validate the header name changed event is fired correctly.

packages/ag-grid-enterprise/src/columnHeaderEdit/columnHeaderEditPopup.ts
```diff
@@ -0,0 +1,153 @@
...
+                alwaysOnTop: true,
+                cssIdentifier: 'column-header-edit',
+                closedCallback: () => this.onDialogClosed(),
+            })
+        );
+        this.dialog = dialog;
+
+        this.addManagedElementListeners(dialog.getGui(), {
+            keydown: (event: KeyboardEvent) => {
+                if (event.key === KeyCode.ENTER) {
```
StephenCooper, line 146: I think this behaviour of saving only on ENTER is not an existing pattern?

Especially when you compare this to CalculatedColumns. The default is live update of the header name. I think we should consider this approach for edit name too. 

Calculated columns has a configurable apply buttons. We may not need to do that now but it might be good to consider if our configuration approach will enable this if required.

packages/ag-grid-community/src/columns/columnGroups/columnGroupState.ts
```diff
@@ -3,14 +3,17 @@ import type { AgProvidedColumnGroup } from '../../entities/agProvidedColumnGroup
 import { isProvidedColumnGroup } from '../../entities/agProvidedColumnGroup';
 import type { ColumnEventType } from '../../events';
 
-export const _getColGroupState = (beans: BeanCollection): { groupId: string; open: boolean }[] => {
+export const _getColGroupState = (
+    beans: BeanCollection
+): { groupId: string; open: boolean; headerName: string | null }[] => {
```
StephenCooper: Can we extract the return type into a Type please.

testing/behavioural/src/columnHeaderEdit/editable-group-header-name-react.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        gridApi!.setState({
+            columnGroup: {
+                openColumnGroupIds: [],
+                headerNames: [{ groupId: 'athleteGroup', headerName: 'Renamed' }],
+            },
+        });
+
+        await waitFor(() => expect(groupText()).toBe('Renamed'));
+    });
+});
```
StephenCooper, line 57: ---

**Missing coverage: edited group name through split/merge.** `groupHeaderNameOverrides` is keyed by `groupId` so a rename survives instance rebuilds and is shared across split instances — but no test locks this in (a per-`instanceId` refactor would pass the whole suite).

Suggested test (split setup from `column-groups.test.ts:928` + editable setup from `editable-header-name.test.ts`):
1. Rename a group with `left1`/`center1`/`right1` children → all 3 split instances show the new name.
2. Rename via one split instance → name updates across all instances.
3. Re-merge the columns → merged instance keeps the edited name.

StephenCooper: Opens up the question about how this should behave and future proof in case we get user requests to be able to rename them to different groups.

packages/ag-grid-community/src/columns/columnStateUtils.ts
```diff
@@ -486,6 +493,13 @@ export function _resetColumnState(beans: BeanCollection, source: ColumnEventType
         orderState[orderIdx++] = { colId: col.colId };
     });
 
+    // Group header names live outside column state, so clear their overrides here too.
+    const groupOverrides = colModel.groupHeaderNameOverrides;
+    if (groupOverrides.size) {
+        groupOverrides.clear();
+        eventSvc.dispatchEvent({ type: 'columnHeaderNameChanged' });
```
StephenCooper: Can you check what our approach is for events that can be fired for both groups and columns. Do they all share the same event or do we have separate events for columns and groups.

packages/ag-grid-community/src/headerRendering/cells/columnGroup/headerGroupCellComp.ts
```diff
@@ -64,6 +64,14 @@ export class HeaderGroupCellComp extends AbstractHeaderCellComp<HeaderGroupCellC
         const eGui = this.getGui();
         const eHeaderGroupGui = headerGroupComp.getGui();
 
+        // Replace any previously-rendered component (e.g. when the header name is edited) rather than
+        // appending a second one alongside it.
+        const previousComp = this.headerGroupComp;
+        if (previousComp) {
+            this.eHeaderCompWrapper.removeChild(previousComp.getGui());
```
StephenCooper: Prefer `childNode.remove()` over `parentNode.removeChild(childNode)`.

packages/ag-grid-community/src/interfaces/gridState.ts
```diff
@@ -155,6 +155,13 @@ export interface ColumnOrderState {
 
 export interface ColumnGroupState {
     openColumnGroupIds: string[];
+    /** User-edited group header names, keyed by group id. */
+    headerNames?: { groupId: string; headerName: string }[];
+}
+
+export interface ColumnHeaderNameState {
+    /** User-edited column header names, keyed by column id. */
+    columnHeaderNames: { colId: string; headerName: string }[];
```
StephenCooper: Can we extract this object type and export it to be consistent with all the other state items.

packages/ag-grid-community/src/headerRendering/cells/columnGroup/headerGroupCellComp.ts
```diff
@@ -64,6 +64,14 @@ export class HeaderGroupCellComp extends AbstractHeaderCellComp<HeaderGroupCellC
         const eGui = this.getGui();
         const eHeaderGroupGui = headerGroupComp.getGui();
 
+        // Replace any previously-rendered component (e.g. when the header name is edited) rather than
+        // appending a second one alongside it.
+        const previousComp = this.headerGroupComp;
+        if (previousComp) {
```
StephenCooper, line 70: Does this mean that we are always replacing the component and not doing an optimised just name change?

With a custom header component is `refresh` called or is it always re-created?

sdwvit: addressed, ty

packages/ag-grid-community/src/columns/columnGroups/columnGroupState.ts
```diff
@@ -27,7 +34,7 @@ export const _setColGroupOpen = (
 
 export const _setColGroupState = (
     beans: BeanCollection,
-    stateItems: { groupId: string; open: boolean | undefined }[],
+    stateItems: { groupId: string; open: boolean | undefined; headerName?: string | null }[],
```
StephenCooper: Should this also be using `ColGroupState` type?

packages/ag-grid-community/src/entities/colDef.ts
```diff
@@ -139,6 +139,14 @@ export interface ColGroupDef<TData = any> extends AbstractColDef<TData> {
      */
     suppressStickyLabel?: boolean;
 
+    /**
+     * Set to `true` to allow the user to edit this column group's header name from the UI.
+     * The edited value is persisted as part of grid state.
+     * @default false
+     * @agModule `ColumnHeaderEditModule`
+     */
+    headerNameEditable?: boolean;
```
StephenCooper: I think this should be defined on AbstractColDef next to headerName. Then just update the JsDoc to include column / column Group

packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -838,6 +841,17 @@ export class AgColumn<TValue = any>
...
 
+    /** Override the displayed header name. Pass `null` to revert to the `colDef` value. */
+    public setHeaderNameOverride(headerName: string | null): void {
+        if (this.headerNameOverride === headerName) {
+            return;
+        }
+        this.headerNameOverride = headerName;
+        // Grid-level event, matching column groups, so header cells, the tool panel and the state
+        // service all refresh from a single event keyed by colId.
+        this.beans.eventSvc.dispatchEvent({ type: 'columnHeaderNameChanged', colId: this.colId });
```
StephenCooper, line 854: I am wondering if this firing the global event is the right approach. Seems like we should follow the pattern of firing the colEvent. Then all the column only update paths do not need to check the colId, they will just already be only firing for the column. 

If we do that then you might just need to also use the existing dispatchStateUpdatedEvent method for the column headers?

packages/ag-grid-community/src/columns/columnGroups/columnGroupState.ts
sdwvit: Done — the group-state setter/reset now use the `ColGroupState` type instead of inline literals (kept file-local; `ColumnGroupState` remains the public state type).

packages/ag-grid-community/src/entities/colDef.ts
sdwvit: Done — moved `headerNameEditable` onto `AbstractColDef` next to `headerName` and removed the duplicates on `ColDef`/`ColGroupDef`; JSDoc now covers column and column group.

packages/ag-grid-community/src/entities/agColumn.ts
sdwvit: Switched to a column-scoped `headerNameChanged` colEvent — `headerCellCtrl` and `toolPanelColumnComp` now listen via `addManagedListeners(column, …)` with no `colId` checks. I kept the grid-level `columnHeaderNameChanged` firing alongside (same dual-dispatch as `setActualWidth`) because the state service and column groups still need a grid-level signal — groups have no per-column event bus. Bulk state restore goes through `setHeaderNameOverride` per column, so header cells still refresh.

StephenCooper (approved):

PR #14429, opened 2026-07-14
AG-12609 Add getColumnMenuItems callback for column menu surfaces

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-12609

Fix [AG-12609](https://ag-grid.atlassian.net/browse/AG-12609)

### DIFF
--- a/documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
+++ b/documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
@@ -181,8 +181,8 @@
         },
         "columnMenuItems": {
             "more": {
-                "name": "Customising Across Menus",
-                "url": "./column-menu/#customising-across-menus"
+                "name": "Customising the Menu Items",
+                "url": "./column-menu/#customising-the-menu-items"
             }
         },
         "suppressHeaderMenuButton": {
--- a/documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
+++ b/documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
@@ -179,6 +179,12 @@
                 "url": "./column-menu/#customising-the-menu-items"
             }
         },
+        "columnMenuItems": {
+            "more": {
+                "name": "Customising Across Menus",
+                "url": "./column-menu/#customising-across-menus"
+            }
+        },
         "suppressHeaderMenuButton": {
             "more": {
                 "name": "Customising the Column Menu",
--- a/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
+++ b/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
@@ -68,6 +68,12 @@
                 "url": "./column-menu/#customising-the-menu-items"
             }
         },
+        "getColumnMenuItems": {
+            "more": {
+                "name": "Column Menu",
+                "url": "./column-menu/#customising-across-menus"
+            }
+        },
         "columnMenu": {
             "more": {
                 "name": "Column Menu",
--- a/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
+++ b/documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
@@ -71,7 +71,7 @@
         "getColumnMenuItems": {
             "more": {
                 "name": "Column Menu",
-                "url": "./column-menu/#customising-across-menus"
+                "url": "./column-menu/#customising-the-menu-items"
             }
         },
         "columnMenu": {
--- a/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
@@ -54,6 +54,16 @@ The menu item list should be a list with each item either a) a string or b) a `M
 
 {% apiDocumentation source="grid-options/properties.json" section="accessories" names=["getMainMenuItems"] /%}
 
+### Customising Across Menus
+
+The `mainMenuItems` and `getMainMenuItems()` properties customise the column menu only. To customise the menu items across the column menu, the [Columns Tool Panel](./tool-panel-columns/#customising-the-right-click-menu) right-click menu, and the [Column Chooser](#customising-the-column-chooser) from a single place, use `colDef.columnMenuItems` and the grid option `getColumnMenuItems()` instead.
+
+These mirror `mainMenuItems` / `getMainMenuItems()`, but their callback params include a `source` (`'columnMenu'`, `'columnsToolPanel'` or `'columnChooser'`) so you can tailor the items per menu. When set, they take priority over `mainMenuItems` / `getMainMenuItems()`.
+
+{% apiDocumentation source="column-properties/properties.json" section="header" names=["columnMenuItems"] /%}
+
+{% apiDocumentation source="grid-options/properties.json" section="accessories" names=["getColumnMenuItems"] /%}
+
 ### Built-In Menu Items
 
 The following is a list of all the default built-in menu items with the rules about when they are shown.
--- a/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
@@ -109,7 +109,7 @@ For more details, refer to the section: [Custom Menu Item Components](./componen
 The following example demonstrates the `colDef.columnMenuItems` property:
 
 - The **Athlete** column shows the list of built-in items.
-- The **Age** column appends custom items to the list of built-in items.
+- The **Age** column leads with the Columns Tool Panel tokens `scrollIntoView` and `value` - showing they resolve on the column menu too - then appends custom items. The grid is grouped by **Sport**, so **Add Age to values** shows Age aggregated in the group rows.
 - The **Country** column provides custom items and adds one built-in default item. Clicking on the custom items logs to the developer console.
 - The **Year** column trims down the default items by removing values.
 
--- a/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
@@ -48,7 +48,7 @@ The menu shows a default set of items. You can add your own items, or change whi
 
 When both are set, `colDef.columnMenuItems` takes priority over `getColumnMenuItems()`.
 
-The `source` param is one of `'columnMenu'`, `'columnsToolPanel'` or `'columnChooser'`, so a single callback can tailor the items for the column menu, the [Columns Tool Panel](./tool-panel-columns/#right-click-menu) right-click menu, and the [Column Chooser](#customising-the-column-chooser).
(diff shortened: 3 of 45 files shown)


packages/ag-grid-community/src/interfaces/menuItem.ts
```diff
@@ -190,3 +190,10 @@ export type DefaultMenuItem =
     | 'sortAbsoluteAscending'
     | 'sortAbsoluteDescending'
     | 'sortUnSort';
+
+/**
+ * Built-in menu items for the Columns Tool Panel and Column Chooser right-click menu.
+ * The concrete item shown for a token depends on the column's current state (e.g. `rowGroup`
+ * renders "Group by" or "Un-Group by").
+ */
+export type DefaultToolPanelItem = 'scrollIntoView' | 'rowGroup' | 'value' | 'pivot';
```
alistairjcbrown, line 199: ❓ Can we use context menu items from Column heading on the Column Tool Panel, and vice versa?

StephenCooper: Yes, I would expect that we should support the same set of events under a single combined type.

alistairjcbrown: Combined type created to support the same set of event (where it makes sense to) -- e.g. there's no opportunity to use 'pivot' on main columns AFAICT, so it's a noop

documentation/ag-grid-docs/src/content/docs/column-menu/index.mdoc
```diff
@@ -94,10 +108,10 @@ For more details, refer to the section: [Custom Menu Item Components](./componen
 
 ### Example: Customising the Menu Items
 
-The following example demonstrates the `colDef.mainMenuItems` property:
+The following example demonstrates the `colDef.columnMenuItems` property:
 
 - The **Athlete** column shows the list of built-in items.
-- The **Age** column appends custom items to the list of built-in items.
+- The **Age** column leads with the Columns Tool Panel tokens `scrollIntoView` and `value` - showing they resolve on the column menu too - then appends custom items. The grid is grouped by **Sport**, so **Add Age to values** shows Age aggregated in the group rows.
```
StephenCooper: I don't think our docs should make a point of the fact you can add Columns Tool Panel to the column menu. Especially when the scrollIntoView does not really make sense in the column menu. 

I would play it down on this section.

documentation/ag-grid-docs/src/content/api-documentation/column-properties/properties.json
```diff
@@ -179,6 +179,12 @@
                 "url": "./column-menu/#customising-the-menu-items"
             }
         },
+        "columnMenuItems": {
+            "more": {
+                "name": "Customising Across Menus",
+                "url": "./column-menu/#customising-across-menus"
```
StephenCooper: Does this link work? Should it be the same as the item above?

documentation/ag-grid-docs/src/content/api-documentation/grid-options/properties.json
```diff
@@ -68,6 +68,12 @@
                 "url": "./column-menu/#customising-the-menu-items"
             }
         },
+        "getColumnMenuItems": {
+            "more": {
+                "name": "Column Menu",
+                "url": "./column-menu/#customising-across-menus"
```
StephenCooper: Is this link correct?

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/_examples/customising-tool-panel-menu/main.ts
```diff
@@ -0,0 +1,91 @@
...
+
+// Enable extended validations only for development
+if (process.env.NODE_ENV !== 'production') {
+    enableDevValidations();
+}
+
+ModuleRegistry.registerModules([
+    ClientSideRowModelModule,
+    ColumnsToolPanelModule,
+    // ColumnMenuModule provides the stock menu items (e.g. pinning) used in the tool panel menu below
```
StephenCooper: We don't do this in other examples, as in explaining each modules use.

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
```diff
@@ -162,6 +162,38 @@ The example below demonstrates the Columns Tool Panel using a mixture of items e
...
+1. Set `colDef.columnMenuItems`. This can either be a list of menu items, or a callback which is passed the list of default menu items.
+1. Set the grid option `getColumnMenuItems()`. This callback is passed the list of default menu items, the column, and a `source` indicating which menu is being shown.
+
+The `source` is `'columnsToolPanel'` for the Columns Tool Panel, `'columnChooser'` for the [Column Chooser](./column-menu/#customising-the-column-chooser), and `'columnMenu'` for the column header menu. Branch on it to tailor the items per menu.
+
+{% apiDocumentation source="column-properties/properties.json" section="header" names=["columnMenuItems"] /%}
+
+{% apiDocumentation source="grid-options/properties.json" section="accessories" names=["getColumnMenuItems"] /%}
+
+`params.defaultItems` is a list of string tokens, so you can filter and reorder it. All built-in tokens share one type, `DefaultColumnMenuItem`, and can be returned for any surface; each is shown only where it applies to the column and grid state. The tokens shown by default in the tool panel menu are:
```
StephenCooper: "for any surface" could be clearer.

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
```diff
@@ -162,6 +158,38 @@ The example below demonstrates the Columns Tool Panel using a mixture of items e
 
 {% gridExampleRunner title="Tool Panel Styling" name="styling"  exampleHeight=610 /%}
 
+## Customising the Right-Click Menu
```
StephenCooper: I think we need to have a new section the first documents the Column Menu items on right click. Then the Customising of that menu would be a natural sub section. We are currently missing documentation for those default items.

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
```diff
@@ -162,6 +158,38 @@ The example below demonstrates the Columns Tool Panel using a mixture of items e
...
+
+{% note %}
+Resolving column menu tokens (such as `'pinSubMenu'`) in the tool panel menu requires the `ColumnMenuModule` to be registered. Pin items are only available for individual columns, not column groups.
+{% /note %}
+
+When `columnMenuItems` or `getColumnMenuItems()` is configured, the tool panel and Column Chooser menu opens even when [`functionsReadOnly`](#read-only-functions) is enabled - where it is otherwise suppressed. The built-in group / value / pivot items are hidden in this case, but the items you return are still shown. Branch on `source` if you only want to customise the column menu.
+
+The following example demonstrates customising the tool panel menu. Right-click a column in the Columns Tool Panel to see an optional pinning sub-menu and a custom **Highlight Column** item, which toggles a highlight style on the column:
+
+{% gridExampleRunner title="Customising the Tool Panel Menu" name="customising-tool-panel-menu" exampleHeight=670 /%}
```
StephenCooper, line 190: We can still keep one example and just point to it for the default and custom case.

documentation/ag-grid-docs/src/content/docs/tool-panel-columns/index.mdoc
```diff
@@ -162,6 +158,38 @@ The example below demonstrates the Columns Tool Panel using a mixture of items e
...
+## Customising the Right-Click Menu
+
+Right-clicking a column in the Columns Tool Panel opens a menu. By default this shows scroll, group, aggregation and pivot items based on the column's configuration. You can customise these items, including adding your own, using the same callbacks as the [Column Menu](./column-menu/#customising-the-menu-items):
+
+1. Set `colDef.columnMenuItems`. This can either be a list of menu items, or a callback which is passed the list of default menu items.
+1. Set the grid option `getColumnMenuItems()`. This callback is passed the list of default menu items, the column, and a `source` indicating which menu is being shown.
+
+The `source` is `'columnsToolPanel'` for the Columns Tool Panel, `'columnChooser'` for the [Column Chooser](./column-menu/#customising-the-column-chooser), and `'columnMenu'` for the column header menu. Branch on it to tailor the items per menu.
+
+{% apiDocumentation source="column-properties/properties.json" section="header" names=["columnMenuItems"] /%}
```
StephenCooper: I would be tempted to just point to the column menu doc page instead of having this all repeated here too. Just trying to think about keeping this page from growing too long.

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
```diff
@@ -1998,6 +1999,20 @@ export class AgGridAngular<TData = any, TColDef extends ColDef<TData> = ColDef<a
      * @agModule `ColumnMenuModule`
      */
     @Input() public getMainMenuItems: GetMainMenuItems<TData> | undefined = undefined;
+    /** For customising the menu items shown for a column across the column menu, the Columns Tool Panel
+     * right-click menu, and the Column Chooser. The `source` param indicates which surface the menu is for;
+     * branch on it to target a single surface. Takes precedence over `getMainMenuItems` for the column menu.
+     *
+     * Note: when this is set, the Columns Tool Panel / Column Chooser right-click menu opens even under
```
StephenCooper: I think the functionReadonly is too detailed for this top level property. 

Just list the multiple modules in the agModule tag.

Apply to the colDef and GridOptions files as they are the source.

packages/ag-grid-enterprise/src/menu/columnMenuItemsResolver.ts
```diff
@@ -0,0 +1,76 @@
...
+type ColumnMenuItems = (DefaultColumnMenuItem | MenuItemDef)[];
+
+/**
+ * Resolves the final menu item list for a column-scoped menu on any surface (column menu,
+ * Columns Tool Panel, Column Chooser), applying user customisation with this precedence:
+ * `columnMenuItems` (col/group) -> `getColumnMenuItems` (grid) -> [`source: 'columnMenu'` only]
+ * legacy `mainMenuItems` (col/group) -> `getMainMenuItems` (grid) -> `defaultItems`.
+ * @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time.
+ */
+export function _resolveColumnMenuItems(
```
StephenCooper, line 22: If these functions are only used in Enterprise lets move them into Enterprise.

packages/ag-grid-enterprise/src/menu/menuItemMapper.ts
```diff
@@ -335,6 +311,53 @@ export class MenuItemMapper extends BeanStub implements NamedBean {
                         return null;
                     }
                 }
+                case 'scrollIntoView': {
```
StephenCooper, line 315: Not sure if its possible based on the differences but it seems like we have duplication here with the toolPanelContextMenu. I would be interested to know if we can share code.

alistairjcbrown: Both menus render the same items, but they drive different backends; the header menu acts on the column directly, while the tool-panel menu routes through its own update strategy, so a single shared action function would couple the two. Instead we've unified the tokens under one type, resolved them through the same mapper, and pulled the shared labels/icons into `columnMenuTokenLabels` to stop them drifting. Each menu keeps its own action logic.

StephenCooper (approved):

PR #14435, opened 2026-07-15
Advanced filter testing

SalvatorePreviti:
# Advanced Filter regression coverage + two bug fixes

## Bug fixes

1. **Crash on unsupported `filterOptions` keys** — `getEntries` read `operators[key].displayValue` for every `filterOptions` key, so a key with no Advanced Filter operator (relative-date preset, custom or simple-filter-only option) threw. Now only own operators produce an entry, so unmapped and inherited names (`toString`) are skipped.
2. **Quoted operand values didn't round-trip** — the formatter always wrapped operands in `"`, so a value containing `"` produced malformed text that wouldn't re-parse. It now quotes with the character the value lacks (`a"b` → `'a"b'`), no parser change. Values with _both_ quote kinds still fail safe (rejected, not mis-filtered);

## Tests

Behavioural, via the public API and real DOM — no bean/internal access, no mocking. New `FilterDom` panel-snapshot serializer alongside `GridRows` / `GridColumns`.

- **Advanced Filter regression** — 138 tests / 10 files: parser & quoting edges, autocomplete/editing, builder (drag, keyboard), dates, model ↔ text round-trip, `filterOptions` reuse, validation, complex nested expressions.
- **Filter behaviour** — 153 tests / 16 files: text/number/date conditions, set filter (values, tree/excel, API), group & multi filters, floating-filter editing, quick filter, filters tool panel, filter-manager API, state & events, and filter × sort/pagination/grouping/tree-data.

~290 new tests, plus `GridRows` hardening of 3 existing files and the `test-utils/filters/` harnesses (`ColumnFilterHarness`, `AdvancedFilterHarness`, builder harness, `installFilterLayoutMock`).

Three shared-code behaviours are locked as current rather than changed (product decisions): empty-value filter reported present yet hides all rows; `defaultToNothingSelected` Select-All emits all-values instead of `null`; both-quote-kind values can't be expressed in text. Also fixes a few unrelated pre-existing tests leaking warnings (ssrm, overlays, viewport, react).


packages/ag-grid-enterprise/src/advancedFilter/advancedFilterExpressionService.ts
```diff
@@ -165,7 +165,10 @@ export class AdvancedFilterExpressionService extends BeanStub implements NamedBe
         if (filterType !== 'number' && filterType !== 'bigint') {
             operand1 ??= _toStringOrNull(filter) ?? '';
             if (!skipFormatting) {
-                operand1 = `"${operand1}"`;
+                // Quote with the char the value does not contain so a value holding one quote kind
+                // still round-trips (the parser accepts either quote); a value with both fails safe.
+                const quote = operand1.includes('"') && !operand1.includes(`'`) ? `'` : `"`;
+                operand1 = `${quote}${operand1}${quote}`;
             }
```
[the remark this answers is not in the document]
SalvatorePreviti: known limitation - finding a quote escaping mechanism would be a behavioural change and need a specific ticket - out of scope for this

testing/behavioural/src/test-utils/polyfills/filterLayoutMock.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * jsdom gives virtual-list viewports 0 height, so filter/rich-select VirtualLists render 0 rows.
+ * Wraps the active `getBoundingClientRect` to force a tall height so tests can query/click rows.
+ * Install in `beforeAll`, uninstall in `afterAll`.
+ */
+
+const DEFAULT_VIEWPORT_HEIGHT = 400;
+
+/** Viewport elements whose height must be forced so their VirtualList renders rows. */
+const TALL_VIEWPORT_SELECTORS = [
```
StephenCooper, line 10: Is this approach more brittle then enabling the useRealMeasurements option that Gil had introduced?

StephenCooper (approved):
LGTM!

testing/behavioural/src/test-utils/polyfills/filterLayoutMock.ts
SalvatorePreviti: mhhh yes you are right, we can return to it later when there is the need, my approach with harness and tools so far is : implement only what's strictly needed and extend when needed

PR #14436, opened 2026-07-15
AG-17846 Correct example value formatter to deal with undefined value correctly

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-17846

Fixes AG-17846

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/printing/_examples/for-print-complex/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/printing/_examples/for-print-complex/main.ts
@@ -1,4 +1,4 @@
-import type { ColDef, FirstDataRenderedEvent, GridApi, GridOptions } from 'ag-grid-community';
+import type { ColDef, FirstDataRenderedEvent, GridApi, GridOptions, ValueFormatterParams } from 'ag-grid-community';
 import {
     ClientSideRowModelApiModule,
     ClientSideRowModelModule,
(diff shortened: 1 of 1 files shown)


documentation/ag-grid-docs/src/content/docs/printing/_examples/for-print-complex/main.ts
```diff
@@ -25,7 +25,7 @@ const columnDefs: ColDef[] = [
     { field: 'color', width: 100 },
     {
         field: 'price',
-        valueFormatter: "'$' + value.toLocaleString()",
+        valueFormatter: "value ? '$' + value?.toLocaleString() : ''",
```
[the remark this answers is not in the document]
alistairjcbrown: This is fine for this example where we know the input data is > 0

sdwvit: `value ?` would already prevent value being `falsy`, so `value?.toLocaleString()` should become `value.toLocaleString?.()` maybe?

sdwvit: oh it's a string, I am even more confused :D

SalvatorePreviti: yes :) is a old very unrecommended feature - we use eval for expressions - and yes for toLocaleString?.() i am not sure what would be the correct thing here. But would not be better to use a () =>  so we start convincing users to not use string expressions

SalvatorePreviti (approved):

sdwvit (approved):

documentation/ag-grid-docs/src/content/docs/printing/_examples/for-print-complex/main.ts
alistairjcbrown: `.?` was my first pass to stop the errors that I forgot to remove 🤦 

> oh it's a string,

Lol, yea -- a lovely little eval sitting in the codebase 😈  

> But would not be better to use a () => so we start convincing users to not use string expressions

Good point -- happy to change this to a better example 👍

alistairjcbrown:
Updated to use callback instead of evaled string 👍

PR #14439, opened 2026-07-15
AG-17624: show Cmd instead of Ctrl in kbd shortcuts for Mac visitors

skewis6:
## Summary

Fix AG-17624

Documentation currently shows `Ctrl` in `{% kbd %}` keyboard-shortcut tags regardless of the visitor's OS. Since docs pages are static-generated at build time, there's no build-time or pure-CSS way to know the visitor's OS, so this uses a small client-side script:

- `{% kbd %}` now renders both a default and a Mac-specific label where they differ (e.g. `^ Ctrl` → also renders `⌘ Command`), hidden/shown via CSS.
- A tiny inline script (same pattern as the existing dark-mode bootstrap) sets `html[data-os="mac"]` client-side, feature-detecting `navigator.userAgentData.platform` (Chromium) with a `navigator.platform` fallback. Runs render-blocking at the top of `<body>` to avoid a flash of the wrong label.
- With JS disabled, pages still show `Ctrl` (today's behaviour) — no regression.
- Cleaned up a few docs pages that previously spelled out both platforms manually (e.g. "Ctrl (or Command on Apple)") now that it's automatic. Left the Undo/Redo page's Redo shortcut as manual dual-text since it's a genuinely different key combo on Mac (`Cmd+Shift+Z`, not just `Cmd+Y`), not just a relabel.
- Registered the new script's hash in the CSP allowlist and updated the affected snapshot/unit tests.

## Test plan

- [ ] Open a page with keyboard shortcuts (e.g. `/react-data-grid/keyboard-navigation/` or `/react-data-grid/undo-redo-edits/`)
- [ ] In a browser confirm Ctrl is shown (if on windows) or Command is shown (if on Mac)
- [ ] Spoof a user agent (e.g. Chrome DevTools → More tools → Network conditions → User agent) and reload — confirm the same tags now show the oppisite (Ctrl for windows and Command for Mac)
- [ ] Disable JavaScript and reload — confirm the page still falls back to `Ctrl` with no layout shift or console errors


StephenCooper (approved):
Nice!

taktran (approved):
LGTM

PR #14440, opened 2026-07-15
AS-942: extract grid theme builder (including composition root) into ag-website-shared

skewis6:
## Goal

AS-942 is about unifying the AG Grid and AG Studio theme builders, which have grown independently on the same underlying theming API but with duplicated (and, on the Studio side, partially reinvented) logic and UI. This PR extracts Grid's theme builder — the more mature, "reference" implementation — into `ag-website-shared`, the repo already shared across ag-grid, ag-charts and ag-studio's doc sites. This is the **complete grid-side migration**, including the composition root itself — a later PR will point AG Studio's theme builder at this same shared code/components instead of its own hand-rolled equivalents.

`ag-website-shared` was chosen over a new published package deliberately: it's already wired into all three doc sites via the existing `@ag-website-shared/*` alias, with no extra publish/versioning step, and there's precedent for shared code there importing host packages directly (e.g. `components/grid/Grid.tsx` already imports `ag-grid-react`).

## What this does

**Model layer** — `external/ag-website-shared/src/theming/`: the theme param/part introspection model (`ParamModel`, `PartModel`), the jotai-backed store/persistence layer, the theme-code import/export parser, `RGBAColor`, and their supporting utils. `ParamModel`'s per-param docs and `parseThemeCode`'s recognised-param check / swappable-parts extraction are all pluggable (`setParamDocsProvider`, `{ isRecognizedParam, extractParts? }`) so the parser has no built-in knowledge of grid's model types. `Preset` (the type) and `applyPreset()` (writes a preset's params/parts into the store) also live here now — both were already fully generic, just stranded in grid's preset data file.

**Full editor/sidebar/chrome layer** — `external/ag-website-shared/src/components/theme-builder/`: essentially the entire theme-builder UI is shared - every color/length/scale/border/font/css/image param editor, `ParamEditor` (the dispatcher), `ParamSearchSelector`, `CollapsibleSection`, `PartEditor`, `Checkbox`/`Tooltip`/`FormField`/`ErrorBoundary`/`Card`/`UIPopupButton`/`ResetChangesModal`, and the preset-scroller layout (`PresetScroller`/`PresetButton`). Two components (`FontFamilyValueEditor`, `ImageValueEditor`) expose a configure-once hook for their per-host curated content (font list, docs URL), mirroring `ParamModel`'s `setParamDocsProvider` pattern.

**Composition root** — `external/ag-website-shared/src/components/theme-builder/ThemeBuilderProvider.tsx`: following review feedback that the grid `ThemeBuilder.tsx` root (jotai store init, initial preset apply, change-tracking subscriptions, `<Provider>`) was stuck in `ag-grid-docs` — meaning a Studio implementation would have had to duplicate this non-trivial setup wholesale rather than reuse it — this bootstrap sequence is now a shared `ThemeBuilderProvider` taking `initialPreset` + `children`. Each host's own `ThemeBuilder.tsx` is now a few lines: pick the initial preset, wrap children in the shared provider.

**Grid-specific content, relocated but namespaced** — `external/ag-website-shared/src/components/theme-builder-grid/`: everything that's genuinely grid content rather than generic mechanism — `RootContainer` (two-pane shell), `EditorPanel`/`BordersEditor`/`AdvancedParamSelector` (curated sidebar layout — which params, groups, labels, order), `presets.ts`/`PresetRender`/`PresetSelector` (grid's 8 preset definitions + live-grid thumbnail), `grid-config/*` (kitchen-sink demo scaffolding), `ThemeImportExportDialog`/`GetTheme` (code-gen + import/export UI — Studio's own import/export is a deliberately different popover design, so unifying this is bigger work than the rest), `EmptyState` (mobile marketing splash), and the two host-registration files (`registerParamDocs.ts`, `registerEditorConfig.ts`) — now live in this grid-owned namespace instead of `ag-grid-docs`, mirroring where a future `theme-builder-studio/` would sit alongside it. `ag-grid-docs/src/pages/theme-builder.astro` shrinks to a thin import of `ThemeBuilder`/`EmptyState` from the shared namespace; nothing else about the page (SEO, `Layout`, nav chrome) changed.

Two per-host seams were needed to make this relocation clean:
- `ThemeImportExportDialog`'s docs link now goes through a `setThemeBuilderDocsUrl()` configure-once hook instead of importing grid's `urlWithBaseUrl` directly (grid registers it via `urlWithBaseUrl(...)` in `registerEditorConfig.ts`).
- `registerParamDocs.ts`'s import of the generated `theming-api.AUTO.json` was a 6-level relative path reaching into the ag-grid monorepo's `dist/` folder — not portable once physically relocated. Replaced with a new `@generated-reference/*` tsconfig alias, resolved the same way `@utils/*`/`@components/*` already are.

`ShadowDom.tsx` (a generic hydration-safe shadow-DOM wrapper with zero grid-specific logic) also moved into shared, since `GridPreview` needed it and it had no reason to live in `ag-grid-docs`.

Every consumer across the theme-builder is rewired onto the shared code, and every superseded original was deleted once nothing referenced it.

## Verification

- Full `ag-grid-docs` vitest suite passes (447/447) throughout.
- `tsc --noEmit` introduces no new type errors versus the `latest` baseline at every step (checked by diffing filtered output against an unmodified checkout) — remaining errors are pre-existing and just relocate when a file moves.
- `eslint`/`prettier` clean on all changed files at every step (including stripping `react-hooks/*` disable comments that don't apply under `ag-website-shared`'s eslint config).
- Manually verified in a running dev server at each step: live grid thumbnails in the preset selector, the full import/export flow (including the new docs-URL seam resolving correctly), the borders dropdown / icon-set `PartEditor` / grid-features checkboxes, the font-family dropdown with Google Fonts loading, accent-color changes propagating to live previews, shadow-type params dispatching correctly via search, and the reset-confirmation modal correctly triggering when switching presets after making changes.

## Test plan

- [ ] Run the theme builder in `ag-grid-docs` locally and confirm editing every param type, switching presets, and import/export still work exactly as before
- [ ] Confirm docs build succeeds (generates `theming-api.AUTO.json` and picks it up via `registerParamDocs.ts` through the new `@generated-reference/*` alias)


BernieSumption (approved):
LGTM as long as you're happy that this creates a requirement for studio's theme builder to start using the same version of jotai (Theme Builder's React state management library).

external/ag-website-shared/src/theming/advanced-params.ts
```diff
@@ -1,6 +1,6 @@
-import { atomWithJSONStorage } from '@components/theme-builder/model/JSONStorage';
 import { atom, useAtom, useAtomValue } from 'jotai';
```
taktran, line 1: You might be able to replace jotai with nanostore, which we already include in the website deps. It's got a similar API

external/ag-website-shared/src/theming/utils.ts
```diff
@@ -123,8 +125,8 @@ export const unescapeStringLiteral = (literal: string): string => {
                 }
                 break;
             case '\n':
-            case '\u2028':
-            case '\u2029':
+            case ' ':
```
taktran, line 128: It seems easier to read and debug, with the unicode

taktran (approved):
LGTM - just some comments

PR #14446, opened 2026-07-15
[AG-11476] feat(master-detail): allow master rows to be open by default

sdwvit:
Fixes [AG-11476](https://ag-grid.atlassian.net/browse/AG-11476)

Adds two new grid options so master rows can be expanded by default, following Salvatore's non-breaking approach (existing `isGroupOpenByDefault` behaviour is unchanged):

- `masterDefaultExpanded` — number of master-row levels to expand (`-1` for all). Falls back to `groupDefaultExpanded` when unset.
- `isMasterOpenByDefault` — callback returning whether a master row is open by default.

Master rows that are also group rows continue to use `isGroupOpenByDefault`/`groupDefaultExpanded`.

Includes behavioural tests and docs.


packages/ag-grid-enterprise/src/rowHierarchy/baseExpansionService.ts
```diff
@@ -50,7 +56,23 @@ export abstract class BaseExpansionService extends BeanStub {
...
+                    data: rowNode.data,
+                    level,
+                });
+                return !!isMasterOpenByDefault(params);
+            }
+            const masterDefaultExpanded = gos.get('masterDefaultExpanded') ?? gos.get('groupDefaultExpanded');
+            return masterDefaultExpanded === -1 || level < masterDefaultExpanded;
+        }
+
         // We call isGroupOpenByDefault only for group nodes and not for master/detail leafs
```
SalvatorePreviti, line 76: Minor performance nit - cache rownode.group in a local variable to avoid double access to a field

SalvatorePreviti (approved):

packages/ag-grid-enterprise/src/rowHierarchy/baseExpansionService.ts
sdwvit: what's the perf implication if it's not a getter?

sdwvit: will do a follow up if it is a concern :)

PR #14463, opened 2026-07-16
[AG-17857] Test(integrated-charts): behavioural coverage for example specs

sdwvit:
## Summary

Part of the AG-17857 enterprise test-coverage effort. Converts **45 placeholder `example.spec.ts` files** across the Integrated Charts documentation pages into real Playwright tests. The audit found Integrated Charts at ~1 real E2E example test (45 placeholders); these are now meaningful behavioural tests.

Integrated charts render series to a `<canvas>`, so where a value isn't in the DOM the tests assert **observable proxies** rather than pixels:
- `getChartModels()` — chart type, series, `seriesChartTypes`, cellRange, theme name
- the chart's accessible proxy DOM (`[role="figure"]` `aria-label="chart, N series"`, proxy legend toolbar)
- chart menu / chart tool-panel state (tabs, mini-chart selection, group visibility)
- chart lifecycle & standalone events (captured via console)
- grid setup: headers, cell values, range-highlight classes, filter model

### Pages covered (45 specs, 47 tests)
`chart-types` (10), `api-cross-filter-chart` (5), `customisation` (5), `chart-tool-panels` (6), `range-chart` (4), `api-range-chart` (3), `events` (3), `time-series` (3), `menu` (2), plus `api-downloading-image`, `api-pivot-chart`, `application-created`, `container` (1 each).

## Testing

## Notes
- **Test specs only** — no product source changed.
- No network shims; examples fetch data as all other docs examples do. Canvas-only aspects (theme colours, axis tick text) are noted in-file as render-level where not DOM-assertable.

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/integrated-charts-api-cross-filter-chart/_examples/custom-theme/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/integrated-charts-api-cross-filter-chart/_examples/custom-theme/example.spec.ts
@@ -1,11 +1,41 @@
-import { clickAllButtons, ensureGridReady, test, waitForGridContent } from '@utils/grid/test-utils';
+import { ensureGridReady, expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
 test.agExample(import.meta, () => {
-    test.eachFramework('Example', async ({ page }) => {
-        // PLACEHOLDER - MINIMAL TEST TO ENSURE GRID LOADS WITHOUT ERRORS
+    // Custom Theme creates four cross-filter charts on firstDataRendered, all styled by the
+    // registered custom chart themes ('my-custom-theme-light' / 'my-custom-theme-dark'):
+    //   - a 'column' chart (Quarterly Sales) in #columnChart
+    //   - a 'pie' chart (Sales by Representative) in #pieChart
+    //   - an 'area' chart (Handsets Sold) in #areaChart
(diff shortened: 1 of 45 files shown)


sdwvit:
Review comments addressed (both verified valid) + full cross-framework local validation:
- **[P1] Canvas coordinate scanning makes the event test unreliable** (standalone-events) — fixed: replaced pixel-scan clicks with deterministic accessible-DOM interactions — keyboard series selection on `.ag-charts-swapchain[tabindex=0]` (ArrowRight+Enter → `seriesNodeClick`) and the legend proxy `[role=switch]` button (→ `legendItemClick`); console-log waits now use `expect(...).toPass()` instead of racy scanning.
- **[P2] Fixed delay does not reliably await chart updates** — fixed: removed all fixed `waitForTimeout` before reading chart state (column-bar/funnel/hierarchical/line-area/pie-donut); replaced with `expect(async()=>{…getChartModels()…}).toPass()` polling.

Cross-framework validation also fixed two genuine quirks: combo now asserts `seriesChartTypes` (deterministic; react/angular report `customCombo` vs `columnLineCombo` for top-level type); provided-container uses framework-common selectors and skips only the Angular destroy round-trip (grid-API chart creation runs outside Angular's change-detection zone, so the `@if` button never re-renders) while still asserting container placement.

Validated locally across all frameworks: **282 passed**, 0 failed.

StephenCooper (approved):

PR #14464, opened 2026-07-16
[AG-17857] Test(pivoting): behavioural coverage for example specs

sdwvit:
## Summary

Part of the AG-17857 enterprise test-coverage effort. Converts **19 placeholder `example.spec.ts` files** across the Pivoting documentation pages into real Playwright tests. The audit found Pivoting at 2 real E2E example tests (19 placeholders).

### Pages covered (19 specs, 24 tests)
| Page | Specs |
|------|-------|
| pivoting | 5 |
| pivoting-column-groups | 9 |
| pivoting-result-columns | 3 |
| pivoting-totals | 2 |

- Pivot mode produces pivot result columns/header groups (`pivot_*` / `pivotGroup_*`) keyed by the pivot field, with **aggregated values computed from the source data** (e.g. US Alpine Skiing gold = 4, Athletics = 69, total gold = 552)
- Enabling pivot via API, tool-panel drag, and the pivot panel
- Pivot column-group ordering (`pivotComparator`), sorting, filtering pivoted columns, default-expanded groups, `suppressExpandablePivotGroups`, hidden single-value header
- Result-column `cellStyle` inheritance + `processPivotResultColDef`, `pivotMaxGeneratedColumns` / `onPivotMaxColumnsExceeded`, floating-filter on result columns
- Row totals (`pivotRowTotals: 'before'`) and expandable column-group totals, with breakdowns summing to the shown total

## Testing

## Notes

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/_examples/column-group-definitions-example/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/_examples/column-group-definitions-example/example.spec.ts
@@ -1,11 +1,27 @@
-import { clickAllButtons, ensureGridReady, test, waitForGridContent } from '@utils/grid/test-utils';
+import { ensureGridReady, expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
 test.agExample(import.meta, () => {
-    test.eachFramework('Example', async ({ page }) => {
-        // PLACEHOLDER - MINIMAL TEST TO ENSURE GRID LOADS WITHOUT ERRORS
-        await ensureGridReady(page);
-        await waitForGridContent(page);
-        await clickAllButtons(page);
-        // END PLACEHOLDER
-    });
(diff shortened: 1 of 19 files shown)


sdwvit:
All four review comments addressed (verified against runtime DOM):
- **[P2] DOM order does not prove leftmost / 'before' positioning** (order-pivot-groups, row-totals) — now assert visual order via `getBoundingClientRect().left`, not DOM/child index.
- **[P2] Virtualisation-dependent absence assertion** (order-pivot-groups) — removed; the test now asserts ordering of the rendered groups (robust to viewport width).
- **[P2] Strict ordering never verified** (strict-column-order) — now deterministically verifies append-vs-resort using the example's known append sequence (6,1,3,2,5,4): '6 left of 1' with the option off, '1 left of 6' with it on.

Validated locally across all frameworks: **24 passed**, 0 failed.

StephenCooper (approved):

PR #14465, opened 2026-07-16
[AG-17857] Test(row-grouping): behavioural coverage for example specs

sdwvit:
## Summary

Part of the AG-17857 enterprise test-coverage effort. Converts **47 placeholder `example.spec.ts` files** across the Row Grouping documentation pages into real Playwright tests. The audit found Row Grouping at 10 real E2E example tests (30 stubs).

### Pages covered (47 specs, 53 tests)
`grouping-data` (6), `grouping-group-rows` (5), `grouping-opening-groups` (5), `grouping-sorting` (4), `grouping-multiple-group-columns` (9), `grouping-single-group-column` (9), `grouping-custom-group-columns` (2), `grouping-edit` (2), `grouping-row-selection` (2), plus `grouping`, `grouping-display-types`, `grouping-row-dragging` (1 each).

- Grouping by field / `rowGroupIndex` / object data (`keyCreator`) / custom hierarchy; aggregated counts and values; expand/collapse; `(Blanks)` and `groupAllowUnbalanced`; single-child removal
- `groupDisplayType` single vs multiple vs group-rows vs custom; per-field auto group columns; `showRowGroup`
- Group / inner / fully-custom cell renderers; `suppressCount`; checkbox location; select-all; `groupSelects: 'descendants'`
- Group sorting, `groupMaintainOrder`, custom / initial-order comparators
- Filtering grouped columns (`agGroupColumnFilter`, set/tree-list); group editing distributing to children; cross-group row dragging with before/after counts

## Testing

## Notes

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/grouping-custom-group-columns/_examples/custom-grouping-many-group-columns/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/grouping-custom-group-columns/_examples/custom-grouping-many-group-columns/example.spec.ts
@@ -1,11 +1,37 @@
-import { clickAllButtons, ensureGridReady, test, waitForGridContent } from '@utils/grid/test-utils';
+import { expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
 test.agExample(import.meta, () => {
-    test.eachFramework('Example', async ({ page }) => {
-        // PLACEHOLDER - MINIMAL TEST TO ENSURE GRID LOADS WITHOUT ERRORS
-        await ensureGridReady(page);
+    test.eachFramework('Example', async ({ page, agIdFor }) => {
+        // groupDisplayType 'custom' with two showRowGroup columns: col-id '0' shows the
+        // 'country' group level, col-id '1' shows the 'year' group level.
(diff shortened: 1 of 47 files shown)


sdwvit:
Both review comments addressed (verified valid, fixed):
- **[P2] Test does not verify the opened group is scrolled into view** (row-group-scroll) — now expands the *last visible* group and asserts the viewport `scrollTop` advances 0 → >0 (the old assertion passed with or without `ensureIndexVisible`).
- **[P2] Sticky assertion lacks proof scrolling reached a sticky state** (suppress-sticky-groups) — now captures the group row's start position, asserts it scrolls above the viewport top (the sticky precondition), then asserts the sticky container is empty — so the assertion is attributable to `suppressGroupRowsSticky`.

Full cross-framework local validation also surfaced/fixed: `renderer-config-custom` (3 pages) used TS-only renderer class names → switched to framework-agnostic selectors; `grouping-group-rows/renderer-config-inner` skipped on vue3 (Vue renders an auto group column instead of full-width group rows for that config). **All frameworks green.**

StephenCooper (approved):

PR #14466, opened 2026-07-16
[AG-17857] Test(tool-panels): behavioural coverage for example specs

sdwvit:
## Summary

Part of the AG-17857 enterprise test-coverage effort. Converts **23 placeholder `example.spec.ts` files** across the Tool Panels & Side Bar documentation pages into real Playwright tests. The audit found Tool Panels & Side Bar at 3 real E2E example tests (23 stubs).

### Pages covered (23 specs, 32 tests)
| Page | Specs |
|------|-------|
| tool-panel-columns | 10 |
| side-bar | 7 |
| tool-panel-filters | 5 |
| tool-panel-filters-new | 1 |

- **Columns tool panel:** deferred apply/cancel, `functionsReadOnly`, section visibility APIs, `setColumnLayout` custom/group layouts, `expandColumnGroups`/`collapseColumnGroups`, `toolPanelClass` styling, custom drag-and-drop image
- **Filters tool panel:** list contents (respecting `filter: false` / `suppressFiltersToolPanel`), expand/collapse groups & filters, filtering drives the grid (incl. "No Matching Rows"), suppress search/expand-all, custom layouts; new filters panel add-filter list + type dropdowns + applying a filter
- **Side bar:** boolean / string / `sideBarDef` config, positions, custom labels, custom tool-panel `parent` containers, `hiddenByDefault`, and the full side-bar API (`setSideBarVisible`, `openToolPanel`, `closeToolPanel`, `setSideBarPosition`, `setSideBar`)

## Testing

## Notes

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/side-bar/_examples/animation/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/side-bar/_examples/animation/example.spec.ts
@@ -1,10 +1,27 @@
-import { ensureGridReady, test, waitForGridContent } from '@utils/grid/test-utils';
+import { ensureGridReady, expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
 test.agExample(import.meta, () => {
-    test.eachFramework('Example', async ({ page }) => {
-        // PLACEHOLDER - MINIMAL TEST TO ENSURE GRID LOADS WITHOUT ERRORS
+    test.eachFramework('Side bar shows Columns and Filters, Columns open by default', async ({ page }) => {
         await ensureGridReady(page);
         await waitForGridContent(page);
-        // END PLACEHOLDER
+
+        // sideBar: ['columns', 'filters-new']
+        await expect(page.locator('.ag-side-button')).toHaveCount(2);
+        await expect(page.locator('.ag-side-button').filter({ hasText: 'Columns' })).toBeVisible();
+        await expect(page.locator('.ag-side-button').filter({ hasText: 'Filters' })).toBeVisible();
+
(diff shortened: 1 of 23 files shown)


sdwvit:
Review comments addressed — each verified against grid source (`toolPanelColumnComp.ts`/`toolPanelColumnGroupComp.ts`) and runtime DOM dumps:

- **[P2] Date suppression case-sensitive/ineffective** (section-visibility) — **fixed**: matched `Date` (correct case) so the `suppressColumnsToolPanel` assertion is actually effective (the lowercase match was vacuous).
- **[P1] 'assertions target leaf-column labels' / 'group lookup uses leaf-label class' / 'group-name checks search only leaf labels' / 'expected order includes excluded groups' / 'Year assertion uses grid-header selector'** — **verified false positives.** `.ag-column-select-column-group-label` does not exist in the codebase; both group headers and leaf columns render their label as `.ag-column-select-column-label`, and the specs correctly scope group labels via the `.ag-column-select-column-group` wrapper. Filter-panel filter names do render in `.ag-header-cell-text` (`data-ref=eFilterName`) inside `.ag-filter-toolpanel`, so those assertions match the intended elements. Confirmed via DOM dumps.

Also: `side-bar/tool-panel-parent` popup test is skipped on vue3 only — the example resolves `ToolPanelDef.parent` at module-eval before `#popup` is mounted in vue3 (genuine example limitation); all other tests run on vue3.

Validated locally across all frameworks: **191 passed, 1 vue3-skipped**, 0 failed.

StephenCooper (approved):

PR #14474, opened 2026-07-16
[AG-17857] Fix(examples): tool-panel custom parent (vue3) & chart container destroy (angular)

sdwvit:
## Summary

Two documentation examples did not behave as documented on specific frameworks. Surfaced while adding cross-framework example-spec coverage (AG-17857) — the specs on those PRs currently `test.skip` the affected framework; once this merges those skips can be removed.

### 1. `side-bar/tool-panel-parent` — vue3
The columns tool panel rendered in the **default side bar** instead of the custom `#popup` container. The vue3 variant can't resolve the popup element at config time, and mutating `ToolPanelDef.parent` in `onMounted` (after grid init) doesn't re-wire the already-created panel.
**Fix:** pass the resolved content element to `openToolPanel(id, parent)` in `openPopup` — mirroring the Filters/drawer button in the same example, which already works.

### 2. `integrated-charts-container/provided-container` — angular
The grid invokes `createChartContainer` **outside Angular's zone**, so assigning `this.chartRef` didn't trigger change detection — the `@if (chartRef)` block never rendered, so the "Destroy Chart" button never appeared (the chart itself was appended imperatively, so it looked fine).
**Fix:** wrap the state update in `NgZone.run(...)`.

## Testing
Verified on the affected frameworks against the dev server:
- vue3: columns panel now renders inside `#popup` (`#popup .ag-column-panel` visible).
- angular: "Destroy Chart" button now renders after chart creation, and the destroy round-trip works.

## Notes
- Examples/source only — no library code changed.
- Follow-up (after merge): drop the `vue3`/`angular` `test.skip` guards in the AG-17857 example-spec PRs (#14463 provided-container, #14466 tool-panel-parent) once those branches rebase on this.

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/integrated-charts-container/_examples/provided-container/provided/angular/app.component.ts
+++ b/documentation/ag-grid-docs/src/content/docs/integrated-charts-container/_examples/provided-container/provided/angular/app.component.ts
@@ -1,6 +1,6 @@
 import { HttpClient } from '@angular/common/http';
 import type { ElementRef } from '@angular/core';
-import { Component, NgZone, ViewChild } from '@angular/core';
+import { Component, ViewChild, signal } from '@angular/core';
 import { AgChartsEnterpriseModule } from 'ag-charts-enterprise';
 
 import { AgGridAngular } from 'ag-grid-angular';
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/src/content/docs/integrated-charts-container/_examples/provided-container/provided/angular/app.component.ts
```diff
@@ -87,9 +90,13 @@ export class AppComponent {
...
     }
 
-    // Arrow function used to correctly bind this to the component
+    // Arrow function used to correctly bind this to the component.
+    // The grid invokes this outside Angular's zone, so run the state update inside
+    // it to trigger change detection (otherwise the chart-wrapper template never updates).
     createChartContainer = (chartRef: ChartRef) => {
-        this.updateChart(chartRef);
-        this.chartParent?.nativeElement.appendChild(chartRef.chartElement);
+        this.ngZone.run(() => {
```
StephenCooper: I think we maybe need to check if this Angular example needs updating to not rely on ngZone. Maybe there are other changes that we can do to avoid this.

SalvatorePreviti (approved):

StephenCooper (approved):

PR #14506, opened 2026-07-20
test(docs-e2e): meaningful example specs for grid state and lifecycle

sdwvit:
## What

Replaces placeholder Playwright `example.spec.ts` tests (grid-load-only, `clickAllButtons`) with real, assertion-bearing cross-framework specs for the **Grid State & Lifecycle** documentation examples.

## Examples covered

**grid-state**
- `grid-state` — sorting fires `onStateUpdated` ('State updated' log); Print State logs `getState()`; recreating the grid restores the applied sort from the previous state (aria-sort survives destroy/recreate).
- `set-state` — sort → Save State → Recreate with No State (sort cleared, aria-sort `none`) → Set State restores the sort (aria-sort `ascending`), with `Saved state` / `Set state` console assertions; Print State logs current state.

**grid-lifecycle**
- `grid-ready` — first column is not pinned by default; ticking "Pin first column on load" and reloading pins `name` to the left via the `gridReady` handler.
- `row-data-updated` — initial load fires `firstDataRendered` + `rowDataUpdated` and stamps both timestamps; reloading re-fires `rowDataUpdated` only, leaving `firstDataRendered` (a one-time event) and its timestamp unchanged.

**view-refresh**
- `redraw-rows` — Redraw All Rows recolours every row (via `getRowStyle`); Redraw Top Rows recolours only the top six, leaving lower rows unchanged.
- `refresh-cells` — renders two pinned top/bottom rows; each of the three Scramble & Refresh buttons propagates out-of-band data mutations to the displayed cells.

## Conventions

- `import { expect, test } from '@utils/grid/test-utils'`, `test.agExample` / `test.eachFramework`, `agIdFor` locators.
- One test block per behaviour the doc prose calls out.
- Deterministic waits only — `ensureGridReady` / `waitForGridContent`, `waitForRowAnimations`, and retrying `await expect(...).toPass()` for console/computed reads. No `waitForTimeout`.


StephenCooper (approved):
We could add lots more options to the GridState example but that is beyond the scope of this ticket.

PR #14378, opened 2026-07-09
Adopt shared ag-shared setup-nx-core action

alantreadway:
Companion to ag-charts [#7372](https://github.com/ag-grid/ag-charts/pull/7372), which extracts the Nx CI setup into a shared `setup-nx-core` action in the `ag-shared` subrepo.

- Pull `ag-shared` up to the revision published by that PR: adds `github/actions/setup-nx-core` + the shared `github/actions/tidy-nx-cache`, unions the `.snyk` Nx dev-dependency ignores across all three repos, and picks up the `chartsWatch.config.js` change.
- Rewrite `.github/actions/setup-nx` as a thin wrapper over the shared core, supplying only this repo's cache paths and `hashFiles()` keys.
- Remove the now-redundant local `.github/actions/tidy-nx-cache` (the core uses the shared one, which is behaviourally identical).

No behavioural change to CI other than gaining the core's `next`-branch and release `-success`-tag handling.

PR #14519, opened 2026-07-21
Fix dropped node_modules cache save and prime dependency caches via init

StephenCooper:
## What

Since #14378 nested `setup-nx-core` under the `setup-nx` wrapper, the `actions/cache@v4` post-save hook for node_modules has silently failed on every run with `Input required and not supplied: path`. Root cause is actions/runner#2009: expression contexts (`inputs`/`steps`) are lost in post steps of nested composite actions, so `path: ${{ inputs.node_modules_paths }}` resolves empty. The Nx cache kept saving only because its `path` is a literal.

The last successful node_modules save was 2026-07-05; since then every CI job (PR and `latest`) has fallen back to that stale restore-key cache, wiped it, and run a full `yarn install --ci` (~3 min per job).

## Fix

1. **Explicit save** (`setup-nx-core`): split the `rw` node_modules cache into `actions/cache/restore@v4` + an explicit `actions/cache/save@v4` main step after `yarn install`, where all contexts are available. The save runs only on a non-exact restore; a failed install never caches. The Nx cache deliberately keeps its post hook (literal `path`, and save-on-failure preserves partial task results) with a guard comment against parameterising it.

2. **Prime the cache before consumers run** (from `sdwvit/ci/ci-caching/prime-jobs-and-yarn-mirror`, excluding the yarn download-cache mirror): `lint`/`build`/`docs` now depend on `init`, so the rw Initialise Workflow job saves the node_modules cache before the ro jobs restore it, instead of racing it and always missing on a fresh dependency hash. Yarn installs gain `--prefer-offline` to resolve from the local mirror before hitting the registry.

Also adds a SYNC-LOG entry for consuming repos (`external/ag-shared/docs/SYNC-LOG.md`).

## Verification

Not verifiable locally (composite action). After merge, the `latest` CI "Initialise Workflow" job should emit `Cache saved with key: node_modules-Linux-<hash>` and subsequent runs should report `EXACT_CACHE_HIT: true` and skip the reinstall.


alistairjcbrown (approved):
LGTM :shipit:

alantreadway (approved):

PR #14521, opened 2026-07-21
Restore public-recipes nested workspace to node_modules cache paths

StephenCooper:
Follow-up to #14519. That PR's fix restored the node_modules cache save, which unmasked a second regression from #14378: the setup-nx rewrite dropped `testing/public-recipes/*/node_modules/` from the cache path list (originally added by #14369). `testing/*/node_modules/` does not match the nested workspace, so an exact-hit restore was missing those folders — `yarn check --integrity` correctly rejected the cache ("Some module folders are missing") and jobs fell back to a full reinstall, as seen in the Docs job of run 29819696925.

Restores the one missing path line, matching the pre-#14378 list.

After merge, ro jobs with `EXACT_CACHE_HIT: true` should pass `yarn check --integrity` and skip the reinstall entirely.

### DIFF
--- a/.github/actions/setup-nx/action.yml
+++ b/.github/actions/setup-nx/action.yml
@@ -85,5 +85,5 @@ runs:
                   testing/public-recipes/*/node_modules/
                   documentation/*/node_modules/
                   external/*/node_modules/
-              node_modules_hash: ${{ hashFiles('yarn.lock','patches/*.patch', 'packages/*/package.json', 'plugins/*/package.json', 'community-modules/*/package.json', 'documentation/*/package.json', 'testing/*/package.json', 'external/*/package.json') }}
+              node_modules_hash: ${{ hashFiles('yarn.lock','patches/*.patch', 'packages/*/package.json', 'plugins/*/package.json', 'community-modules/*/package.json', 'documentation/*/package.json', 'testing/*/package.json', 'testing/public-recipes/*/package.json', 'external/*/package.json') }}
               nx_cache_hash: ${{ hashFiles('yarn.lock','patches/*.patch', 'packages/*/package.json', 'plugins/*/package.json', 'documentation/*/package.json', 'testing/*/package.json', 'community-modules/*/package.json') }}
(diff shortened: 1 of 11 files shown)


.github/actions/setup-nx/action.yml
```diff
@@ -82,6 +82,7 @@ runs:
                   plugins/*/node_modules/
                   community-modules/*/node_modules/
                   testing/*/node_modules/
+                  testing/public-recipes/*/node_modules/
```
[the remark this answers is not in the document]
StephenCooper: Fixed in acf146f0d5e — added `testing/public-recipes/*/package.json` to the `node_modules_hash` inputs. Kept the explicit workspace glob rather than `testing/**/package.json`: a recursive glob would sweep in `node_modules` manifests whenever the hash is evaluated on a non-fresh workspace, risking divergent keys between the rw and ro jobs (a silent permanent miss), whereas forgetting a future nested workspace is caught by the integrity check. `nx_cache_hash` left as-is since its key already appends `github.sha`.

alistairjcbrown (approved):
LGTM :shipit:

.github/workflows/module-size-comparison.yml
```diff
@@ -84,12 +84,47 @@ jobs:
...
+    outputs:
+      relevant: ${{ steps.filter.outputs.size }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - name: Detect file changes
+        id: filter
+        uses: dorny/paths-filter@v3
```
alistairjcbrown: Current version is 4.0.2 -- should we lock to a hash (`7b450fff21473bca461d4b92ce414b9d0420d706`) so we know the version running in CI?

StephenCooper: Pinned all third party actions

alistairjcbrown: @StephenCooper Did this one get missed? 😁

PR #14534, opened 2026-07-21
AG-16959 run tests with vitest

SalvatorePreviti:
# AG-16959 — Run tests with a single Vitest workspace

FIX AG-16959

## What

Merge the package unit tests and the behavioural suite into one Vitest workspace so `./behave.sh` runs the whole unit suite from the repo root, no Nx required.

## Why

- **One command for the whole unit suite locally.** Previously `./behave.sh` ran only the behavioural suite; the package tests needed separate `nx test` invocations. Now a single run covers both.
- **Better CI shard distribution.** The old layout sharded only the behavioural suite (shards 1-3) and ran every package test unsharded on shard 0, making shard 0 a bottleneck. Now all unit test files are pooled and split evenly across the shards.
- **One workspace, discoverable in the IDE.** Every project is listed in `vitest.workspace.ts`, so the Vitest IDE extension picks them all up from one file.
- **Less indirection.** Running Vitest directly (local and CI) drops the Nx layer for the test run, giving faster, more predictable feedback.

## Changes

- **Workspace config** — root `vitest.config.ts` holds the runner-global options (reporters, `onConsoleLog`, pool, JUnit output) that Vitest reads only from the root in workspace mode; `vitest.shared.ts` centralises the per-project unit config (TZ pin, jsdom CSS-parse-error filter, `threads` pool, JUnit path). Each project keeps its own `vitest.config.ts`.
- **`vitest.workspace.ts`** — lists every project (unit + node-env tooling) so the IDE discovers them all.
- **`behave.sh`** — runs the workspace from the repo root; defaults to the unit projects (`ag-stack`, `ag-grid-community`, `ag-grid-enterprise`, `locale`, `behavioural`), with `--project <name>` / `--project all` to override.
- **CI** — the `test` job shards the unit projects across shards 1-3 via `vitest --shard`; shard 0 type-checks (`build:test`), runs the node-env tooling projects, `pack:verify`, and `assertNoDebugger`.
- **Cleanup** — removed the stale duplicate keys from the `chartModelMigration` snapshot; dropped now-redundant `vitest` devDependencies from sub-packages (resolved from the root).

## Verify

```bash
./behave.sh                    # whole unit suite
./behave.sh --project behavioural
```

CI shards partition test files deterministically (no overlap, no gaps) and emit `reports/workspace.xml` per shard, merged by the report job.


packages/ag-grid-enterprise/src/charts/__snapshots__/chartModelMigration.test.ts.snap
```diff
@@ -3801,3805 +3801,3 @@ exports[`chartModelMigration > upgradeChartModel > should upgrade 28.0.0 success
   "unlinkChart": false,
 }
 `;
-
-exports[`chartModelMigration upgradeChartModel should upgrade 22.1.0 successfully 1`] = `
```
StephenCooper, line 3805: Looks like we are missing this test case now?

StephenCooper: Looks like just a different format for Jest that can now be removed

.github/workflows/ci.yml
```diff
@@ -229,25 +243,28 @@ jobs:
...
+      - name: Type-check test files
+        # `build:test` (tsc --noEmit on tsconfig.spec) was previously pulled in as the nx `test` target's
+        # dependency; run it explicitly now that the shards above run Vitest directly instead of via Nx.
+        if: matrix.shard == 0
+        id: build-test
+        run: yarn nx ${{ github.event.inputs.nx_command || 'affected' }} -t build:test -c staging --exclude all
+      - name: nx test (node-env tooling projects)
+        # Every affected `test` target except the Vitest workspace unit projects and behavioural (sharding):
+        # docs, ag-website-shared, typedoc-links, generate-example-files, etc.
         if: matrix.shard == 0
```
StephenCooper, line 271: Do we still need special treatment for the different shards?

SalvatorePreviti: Updated the comment to reflect better the reality.
          #   - shard 0: everything that can't be file-sharded — type-check (build:test), the node-env
          #     tooling `test` projects, pack:verify, assertNoDebugger.
          #   - shards 1..3: the Vitest workspace unit projects (ag-stack, ag-grid-community,
          #     ag-grid-enterprise, locale, behavioural), split by file via `vitest --shard`.

StephenCooper (approved):

PR #14565, opened 2026-07-22
fix-skia-canvas-flakey-install

SalvatorePreviti:
- fix issues with skia-canvas not always running post install script as allow-scripts works only in the root project and not in sub packages
- fix a test that was noisy as it was missing wrapping within waitFor / act async reacty code


yarn.lock
```diff
@@ -21678,9 +21678,9 @@ sitemap@^9.0.0:
     arg "^5.0.0"
     sax "^1.4.1"
 
-skia-canvas@^3.0.8:
+skia-canvas@3.0.8:
```
sdwvit, line 21681: ty for pinning

testing/behavioural/package.json
```diff
@@ -11,7 +11,6 @@
     "ag-grid-community": "36.0.0-beta.20260719.2054",
     "ag-grid-enterprise": "36.0.0-beta.20260719.2054",
     "ag-grid-react": "36.0.0-beta.20260719.2054",
-    "skia-canvas": "^3.0.8",
     "@types/react": "^18.3.23",
```
sdwvit, line 14: can we please pin the rest of the packages while we are on it?

SalvatorePreviti: I would not in this PR - we should agree with @seanlandsman  i guess? i pinned the exact version here for security as we do run post install scripts

yarn.lock
SalvatorePreviti: for security as we run post install scripts on this package

testing/behavioural/package.json
seanlandsman: Pinned deps are good for security - I wouldn't be against it, especially for testing related packages

seanlandsman (approved):

PR #14567, opened 2026-07-22
Header Menu Icon Improvement

gportela85:


community-modules/styles/src/internal/themes/alpine/_index.scss
```diff
@@ -232,16 +232,20 @@
     .ag-floating-filter-button-button,
     .ag-filter-toolpanel-expand,
     .ag-chart-menu-icon {
-        &:hover {
+        &:hover,
+        &.ag-has-menu-open {
```
StephenCooper, line 236: Do we not need to update all the themes?

gportela85: Balham does not style icons on hover, and material does not style icons on hover on legacy themes.
see: https://plnkr.co/edit/eCFaafDx6d3ymyfI?open=main.js&preview

PR #14572, opened 2026-07-23
test(docs): remove React fixme for menu initial focus (AG-17934)

StephenCooper:
## Summary

AG-17934 (React menu with custom items never receiving initial keyboard focus) is fixed by `6f61d00fd0a` ("retry menu initial focus after async framework items render"). This removes the now-obsolete `test.fixme` React skip on the **"opening the menu moves keyboard focus into it"** test in the `custom-menu-item` docs example.

The skip was tagged under the umbrella ticket AG-17785, but the behaviour it guarded is precisely the AG-17934 bug.

## Testing

`./docs-e2e.sh "custom-menu-item" --grep "opening the menu moves keyboard focus into it"` — **6 passed** across all frameworks (typescript, vanilla, reactFunctionalTs, reactFunctionalTs_Dev, angular, vue3).

Other AG-17785 fixmes (floating-filter emits, group-total expansion) are unrelated and left in place.


alistairjcbrown (approved):
🎉

PR #14588, opened 2026-07-23
AG-17933 invalid edit value mode block with full row fix

SalvatorePreviti:
# Full-row editing: honour cell-level validation on all stop paths (+ bug fixes and tests)

Fix [AG-16257](https://ag-grid.atlassian.net/browse/AG-16257)
Fix [AG-16390](https://ag-grid.atlassian.net/browse/AG-16390)
Fix [AG-17245](https://ag-grid.atlassian.net/browse/AG-17245)
Fix [AG-17685](https://ag-grid.atlassian.net/browse/AG-17685)
Fix [AG-17688](https://ag-grid.atlassian.net/browse/AG-17688)
Fix [AG-17933](https://ag-grid.atlassian.net/browse/AG-17933)

Grid packages

files: 28 changed
lines added: +913
lines removed: -347
net delta: +566

## New API: `api.getEditValidationErrors()`

```ts
getEditValidationErrors(): ICellEditorValidationError[] | null;
```

**Why it is needed.** `validateEdit()` was the only way to ask about validation, and it is not a read: it repopulates the validation model, restages every open editor's value and rewrites the invalid markers. Observing the state meant perturbing it. This returns what the Grid already holds, same shape, no side effects; `validateEdit()` remains the way to revalidate.

**Why users might need it.** Gating a Save button, rendering an error summary or status-bar count, blocking navigation away from a dirty row - all run per keystroke or per framework render, exactly where a revalidating call is unsafe.

It also unblocked our own tests: `GridRows` snapshots called `validateEdit()`, so a diagram validated the grid it was describing (see Tests).

## Tickets

**AG-16257 - Firefox date editor drops a single-digit segment on Enter.** Firefox withholds a partially-typed segment from `input.value` until blur. The editor now flushes on the stop read (`agFlushInput` → `AgInputDateField.flushPendingSegment()`), guarded to Firefox and the focused input, before validation is populated.

**AG-16390 - Rich Select popup not shown on a partly-visible column.** `afterGuiAttached` used a bare `focus()`; focusing a partly-visible cell scrolls it, and that scroll reaches `showPicker()`'s `bodyScroll → hidePicker` listener a frame later. Now `focus({ preventScroll: true })`, as every other built-in editor already does. Trade-off: a sliver cell is no longer scrolled fully into view (~38px of a 198px editor); the popup service still keeps the picker in the viewport.

**AG-17245 - cell errors ignored on row-scoped stops.** With `fullRow` + `invalidEditValueMode: 'block'`, a cell error failed to block any stop carrying a `rowNode` but no `column` (`api.stopEditing()`, click-away, focus loss); only Enter worked. `hasCellValidation({ rowNode })` answered about the whole grid instead of the row. Now scoped to the position given, via a side-effect-free `EditModelService.hasValidationErrors(position)`.

**AG-17685 - Custom Editors error values are not reset when triggering filter change if previous value is undefined/null** - does not reproduce on the current codebase - fixed in previous changes; tests added.

**AG-17688 - popup editor left uneditable after focus loss in block mode.** Block mode holds an invalid editor open, but a closed popup has no DOM to hold, so the cell stuck in `state: 'editing'` with no editor. `onPopupEditorClosed` now flushes and repopulates validation, then reverts a block-mode invalid edit (except on Escape); mid-batch it goes through `revertCellEdit`, keeping any earlier staged value.

**AG-17933 - `invalidEditValueMode: 'block'` did not work with `editType: 'fullRow'`.** The headline defect, and the one every other change here hangs off: a full-row edit holding an invalid cell was blocked only by Enter. Every other way of ending it committed the value — `api.stopEditing()`, clicking another cell, losing focus to the page, Tab off the last column, a sort or filter, a batch commit, a popup editor closing. Each path decided for itself whether to check validation, and the checks it did make asked a grid-wide question of a row-scoped stop (AG-17245, above). All of them now go through one gate: repopulate the model, ask about the position being stopped, and hold the edit open with focus on the offending cell if it is invalid. The per-path detail is in the bug list below.

## Bugs found while testing and auditing

1. **`api.stopEditing()` and sort/filter committed a row-level-invalid value in block mode** - the stop path checked no validation at all, and sort/filter read the validation maps without repopulating them.
2. **`api.stopEditing()` and `stopEditingWhenCellsLoseFocus` bypassed block mode during a batch** - each had its own inline persist-then-destroy. Both now go through `EditService.stopBatchEditors`.
3. **`commitBatchEdit()` ended the batch on an invalid value.** Now rejected like every other block-mode close, reported by an explicit `stopBlockRejected`, since the `boolean` return is ambiguous (a consumed mid-batch key also returns `false`).
4. **A self-committing editor's invalid value was discarded rather than held.** A picker-style editor calls `params.stopEditing()` on pick, and the editor was then torn down unconditionally.
5. **A popup closing in full-row mode committed the whole row** at whatever half-typed values the siblings held. It now ends only its own cell: the value is staged so the row stop still commits it, or reverted if that cell is invalid in block mode. The revert is decided on that cell's own errors - a sibling's, or a row-level one, holds the row's commit, but neither is fixed by discarding a value validly typed in the popup - and buffered input is flushed first, so a value withheld until `agFlushInput` is not read as the old valid one.
6. **Focus leaked off the invalid row on Tab.** The blocked branch resolved no cell to focus, so focus moved on regardless; it now pins to the first invalid cell, scrolling its column in if virtualised out.
7. **Purging an edit fired no `cellEditingStopped` and left its validation behind** - the sweep deleted the model entry directly. It now stops the editor properly, clears the cell's validation and the row-level error that edit had caused, and reports a correctly shaped cancel event carrying the row's real index and the value the cell was showing.
8. **Removing the edited row never fired `rowEditingStopped`** and left the detached node in the strategy's started-rows set, holding it alive.
9. **A row purged by removing its edited column kept `ag-row-inline-editing`** - nothing restyled the row after the edits went.
10. **Mid-batch Escape fired no `cellEditingStopped`** - the teardown passed `silent: true`. Two existing tests had asserted the unbalanced counts and were corrected.
11. **`bulkEditingStopped` fired once per cell range** against a single `bulkEditingStarted`, so undo reverted only one range.
12. **`cellEditingStopped` hardcoded `valueChanged: false`** whenever the editor component had already been torn down.
13. **`api.isEditing(cellPosition)` returned the grid-wide answer** for any row without a controller, making it scroll-dependent.
14. **A cell closed on its own kept its validation error.** Validation was cleared only wholesale, once the last editor closed, so a cell ended mid-batch - or reverted in place - stayed recorded as invalid while its siblings edited on, suppressing its note hover and counting towards "any error?". Every per-cell close now clears what it closed, and a per-cell revert re-runs the row rule against the value it leaves behind.
15. **Removing the last validation rule mid-edit wedged the row it was holding.** The validation pass returns early when nothing can report an error, so an error recorded while a rule existed outlived the rule: block mode then held the row on a verdict the user had no way to satisfy (drop `getFullRowEditValidationErrors`, or the colDef rule, while its error stands). A recorded error is now reason enough to run the pass, which clears the model and its markers when no rule is left.
16. **A revert-in-place swallowed the next `cellEditingStopped`.** Reverting a cell re-creates its editor, and the re-setup reset the flag recording that `cellEditingStarted` had fired, so the close that followed reported nothing: `api.stopEditing(true)` during a batch (every editor, both edit types) and Escape on a value held by block mode in a popup. The flag now survives a re-setup of the same session.

## Performance and hardening

- **Edit purging polled for departures instead of being told about them.** `_refreshEditCells` swept six events for as long as an edit was open, scanning the row model and rebuilding a Set of every column, and purged only if one of them happened to fire. Departure is now pushed: rows end their own edits on destroy, `refreshCols` reports every column leaving `colsList` (which covers the parked as well as the destroyed - a pivot primary is parked alive), and manual pinned rows push from `PinnedRows.delete`. Each hook bails in O(1) when nothing is being edited. One sweep remains for static pinned rows, which are dropped without being destroyed and without reporting it.
- **The validation probe materialised every rendered cell ctrl, per probe.** The two halves are now one memoised `EditService.hasValidationRules()`: the colDef half is keyed on a new `ColumnModel.colDefsVersion`, and the editor scan bails on `hasOpenEditors()` and is cached until an editor is created, attached or destroyed with its cell. The grids that paid most were those _without_ configured validation, where the answer is always no.
- **Staging every open editor's value allocated per edit, per keystroke.** `_syncFromEditors` walked `getEditPositions()`, which builds an array plus — for each open edit — a fresh position object and an object-rest copy of the whole `EditValue`, to reach the two fields the sync reads. A full-row edit over twenty columns allocated forty objects per keypress, since validation repopulates on each one. It now walks the live edit map with one scratch position, as the flush does.
- **The validation model was repopulated twice per keystroke and per Enter** by adjacent probes in `cellKeyboardListenerFeature`. The read half is split out as `checkValidated`, so only the first call repopulates. (The Tab commit path still populates twice: a focus event fires in between, so the state genuinely can change.)
- **`clearCellValidation` left emptied rows in the map**, and `cellValidations.size` is read as "any error?". It now prunes.
- **`_destroyEditors` tore down every open editor when `edits` was omitted.** Now required.
- **`EditService.hasValidationErrors` read as a query but repopulated the model and touched the DOM.** Renamed `revalidateAndCheck`, leaving the name to the pure model read.
- **The column purge dispatches from inside the column rebuild**, so it now runs last, once `colsList`, the tree, the id map and the derived state have all settled - a `cellEditingStopped` listener never sees a half-rebuilt column model.
- **`CellCtrl` carried edit logic and two callback queues**, all of it shipped whether or not a grid edits. It now delegates: `onPopupEditorClosed`, `onEditorAttached`, `onCompAttached` (which also owns the start-or-continue decision `setComp` used to make) and `onCellDestroyed` all live on `EditService`. Both queues are gone - an array plus up to two closures per editing cell. The focus a cell is owed is one service field, since focus can only be in one place; a start deferred until React mounts the cell is one nullable field on the cell, which keeps it per cell and releases it with the cell, and is read only when set.
- **`EditService.validateEdit` was a pass-through.** Its two callers (the API and `params.validate()`) call `_validateEdit` directly.
- **`api.validateEdit()` validated a value the browser was still withholding.** Every internal path repopulates validation _after_ flushing open editors; the API was the one that did not, so on Firefox a half-typed date segment revalidated at its previous value while the next stop saw the new one. It now flushes first, like every other caller. Trade-off: the Firefox flush is a blur and refocus, so calling `validateEdit()` mid-edit resets that input's segment cursor - the same cost the stop path already pays, now payable during an edit. The flush deliberately stays out of `_validateEdit` itself, which every built-in editor reaches per keystroke through `params.validate()`; that path is why `getEditValidationErrors()` exists.
- **A `cellEditorParams` function hid its validation rules from the probe.** The rules are per-row and unreadable from the colDef, so such a column now counts as configured rather than being skipped.

## Tests

- `cell-editing-validation*.test.ts` covers both edit types and both modes across every stop trigger, with React guards: block reject/hold, revert-mode discard-invalid-keep-valid, re-commit after a fix, undo/redo, the self-committing editor, the purged-then-recommitted batch, a re-entrant commit that must not clear a live rejection, the blocked-Tab focus pin, a row-level-invalid value held by `api.stopEditing()`, a pull-only editor's invalid value discarded rather than committed, a popup revert revalidating the row against the value it left behind, a rule removed while its error was holding the row, and the two batch stop paths that bypassed block mode.
- `cell-editing-purge.test.ts` covers departures that leave the node alive, which a destroy hook alone would miss: entering pivot mode with an edit staged on a primary, that column being editable again after the round trip, and `pinnedTopRowData` replacement - plus the event shape on row removal, silence during grid teardown, the value reported once the cell controller has gone (a computed column, mid-edit), and the row-level error clearing with the edit that caused it while the rest of the row edit carries on.
- `GridRows` is validation-aware (`❌` on invalid cells and rows, cross-checked against DOM markers and row edit classes), reading through `getEditValidationErrors()` so a snapshot observes rather than validates. Nine snapshots had frozen in state the old `validateEdit()` call produced - a value staged only by that call, `❌` on cells a pull-only editor had never reported, and the stale error behind bug 14. The marker cross-check covers every inline editor, whether or not it has a text input, so checkbox, select and custom composite editors are reconciled too (13 snapshot sites in the suite); popup-edited cells are excluded, since a popup's markers render outside the cell with nothing tying them back. A marker means an explicit one (`setCustomValidity`, `aria-invalid`, the `invalid` class): a native `min`/`max` violation fills `validationMessage` with no validation pass behind it. The scan also stops at this grid's own root, so a nested detail grid's editing cells cannot be read as the master's. Its row-level lookups were also keyed by `rowIndex` alone, so a pinned row leaked its editing flag onto the body row at the same index; they now share a `rowKey` helper with `cellKey`.
- The full-row popup close is guarded in both view layers, since React drives `onPopupEditorClosed` through its own `popupEditorComp`: the invalid case reverts only its own cell, a valid value survives the close and commits with the row, a valid cell is kept when only a sibling is invalid or the row itself is, and a buffered popup value is flushed before the close decides. Each fails without its fix.
- Event balance is asserted where a revert-in-place used to swallow the stop: `api.stopEditing(true)` mid-batch in both edit types, and Escape on a popup holding an invalid value. One Ctrl+Enter across two disjoint cell ranges is asserted to be one `bulkEditingStarted`/`bulkEditingStopped` pair, with both ranges taking the value.
- `getEditValidationErrors()` is asserted empty after each close that bypasses the stop pipeline (the batch popup close, a batch `api.stopEditing()`, the full-row popup revert), and against `validateEdit()` on a pull-only editor to pin the read-vs-refresh split. A buffering editor also pins `validateEdit()`'s flush: it reports the typed-but-withheld value as invalid, and records it.
- Every navigation and filter-change test now snapshots the state on both sides of the action it exercises - the typed-but-uncommitted value before the Tab, arrow key, Enter or filter change, and what the grid holds after - rather than the end state alone. That is what makes a blocked navigation legible as a diagram: the same rows, still `🖍️ ❌`, before and after. `cell-editing-validation-filter-change-react.test.tsx` registers `RowApiModule` and `CellApiModule` for it, since `GridRows` reads the grid through the API.
- `ag-grid-enterprise/e2e/rich-select-scroll-picker.spec.ts` opens the picker on a barely-visible column after a horizontal and a vertical scroll, plus a baseline. Needs real layout, so jsdom cannot cover it.

## Unrelated

Fixed two pre-existing flaky Advanced Filter builder tests, both waiting a fixed macrotask for a widget that mounts when it mounts, now polling with `waitFor`: `AdvancedFilterBuilderHarness.setValue` (the value editor) and `selectRichSelectRow` (the picker's virtual rows).

Fixed a pre-existing flaky group-edit helper. `editCell` typed into the editor, then dispatched Enter without rechecking that the editor was still there — and a redraw landing in that window detaches the input, so Enter commits nothing while leaving the grid not editing, which its post-Enter assertion reads as success. The stale value then surfaces as a mismatch several assertions later (`expected 70 to be 140` in `distribute-modes`). It now requires the input to still be attached and a `cellEditingStopped` to arrive, and re-opens the editor otherwise. Forcing the race (tearing the editor down between the last keystroke and Enter) fails six `distribute-modes` tests on the old helper and none on the new one.


StephenCooper (approved):
Great work!

Issue #637, opened 2016-01-12
Grid height problem when using flexbox (ngMaterial)


ninety7:
Hello,
I'm having big issues using agGrid inside ngMaterial flex card. The grid should have a 100% height but somehow it misses setting the correct height and so it shows a scrollbar inside the rows (not desired). Apparently it is not adding the header height to the grid height from what I could see.
I greatly appreciate if you could fix this issue which can be reproduced on firefox and chrome with this plunker:
http://embed.plnkr.co/ivKiUxsyIlDxUxdVKYGO/
Grid height is correctly set when setting print mode on the grid's options, but obviously I don't want to use print mode.

ceolter:
I don't see anything wrong. In the plunker, the grid is showing with 5 or so rows and a scroll.

Using dev tools, I manually change the div height from 100% to 500px, and this works fine also.

ninety7:
Thanks for your quick reply @ceolter. I'm just seeing 3 rows and I am trying to hide the scroll. Any ideas?
<img width="357" alt="screen shot 2016-01-19 at 10 33 42 am" src="https://cloud.githubusercontent.com/assets/5440311/12428217/42341af4-be98-11e5-90b1-824cc96924c3.png">

ceolter:
afraid not :(

ninety7:
Do you need any help on this? I tried to look at the code but there seems to be a very specific architecture that I could not understand

ceolter:
i'm not sure what you are expecting. you don't have any height specified for the grid, but you say you want 100% height. in the dev tools, i set "height: 500px;" for the grid div, and it resizes fine. when it put it to 100%, ie resizes past the screen, but that's normal if you don't have height set appropriately in other parts of your application.

Issue #636, opened 2016-01-11
adding field sorting to setState / getState?


vladimir-ivanov:
Had a request today to save the multi column sorted state to localStorage - so had to extend the setState and getState methods.
Can create a  pull request for the sorting to be added to the columns state if the feature could be potentially useful?

ceolter:
get/set sort state can already be done via it's own api. this isn't part of the main column get/set state as it isn't managed by the columnController (inside my code). i'll keep this issue open as i may refactor to get it responsible, then the main get/set state will do this also.

vladimir-ivanov:
it is just the case of ColumnController having a property columns -> and each column inside that array already contains sort property and is easy to expose (alternatively it should not belong there?)
Feel free to close if it goes against your design principles.
Thank you

ceolter:
no, makes sense. column controller should do it all. just the evolutionary nature of the design is why it's the way it is now.

ceolter:
see "Saving / Restoring State, Sort and Filter" at bottom of this page:
https://www.ag-grid.com/javascript-grid-column-api/

this is the preferred way of getting and saving the sort order.

we could change the design, however given it's the way it is now, changing the design would be a breaking change for people, so am leaving sleeping dogs lie.

PR #14599, opened 2026-07-24
github-triage: auto-chain browser-verify + repro-rebuild after a fresh confirmed bug

sabahang:
## Summary

**Draft — not for merge without Hesam's explicit review/approval.**

Adds two new jobs to `.github/workflows/github-triage-pipeline.yml` so `browser-verify` and `repro-rebuild` fire automatically instead of needing a separate manual `workflow_dispatch` per stage:

- **`browser-verify-chain`** — fires only after a FRESH `triage` (a real `issues` event, or an explicit `workflow_dispatch stage=triage`) records a genuine `triage-confirmed-bug` outcome. Never fires after `resume`/`execute`, and never on a stale or failed run — the `confirmed_bug` output it reads is itself gated on that (see `ag-grid/ag-dev-prompts` [PR #636](https://github.com/ag-grid/ag-dev-prompts/pull/636) / [PR #637](https://github.com/ag-grid/ag-dev-prompts/pull/637), both merged to `canary`).
- **`repro-rebuild-chain`** — fires only after `browser-verify-chain` records a genuine, fresh `reproduces` verdict.

Both re-invoke the SAME composite action with a different `stage` input, via GHA job-graph `needs:` chaining — not a new `repository_dispatch` event type (simpler: no new cross-repo write scope needed, reuses outputs already threaded through the same workflow run).

## ⚠️ Security note — please read before approving

Both new jobs reuse the **same job permissions** as the existing `run` job (`contents: read`, `issues: write`, `packages: read`) — there is **no dedicated minimal-permission job split, no `persist-credentials: false`, no harden-runner block mode** yet. That hardening is deliberately separate, not-yet-landed follow-up work (documented in the action's own `CLAUDE.md`, § "Browser-verify (B2) — security posture"). The threat model: during these stages the agent keeps full Bash/Read/Write tools live (not just the browser), and the job's `GITHUB_TOKEN` + Jira credentials sit in the same process — a prompt-injected repro page could in principle try to exfiltrate them. This is accepted as bounded for manual, human-chosen dispatch (which is all that's happened to date); auto-firing on every confirmed pick with no per-instance human screening is exactly the scenario the deferred hardening is meant to close before this runs at real volume. Landing the chain first, unhardened, was an explicit call this session — flagging it here so it's a genuine decision at review time, not an oversight.

## Test plan
- [x] `actionlint .github/workflows/github-triage-pipeline.yml` — clean.
- [x] Sanity-checked actionlint actually catches the class of bug it's meant to (a `needs.<job>` reference to a job not listed in the job's own `needs:` — the exact mistake caught in the upstream ag-dev-prompts PR's own review).
- [ ] A real dispatch exercising the full chain (triage → browser-verify → repro-rebuild) on a real issue — not yet run; recommend a single controlled test before wider use.

### DIFF
--- a/.github/workflows/github-triage-pipeline.yml
+++ b/.github/workflows/github-triage-pipeline.yml
@@ -119,6 +119,99 @@ jobs:
             contents: read
             issues: write
             packages: read
+        outputs:
+            # NEW — downstream chain jobs need the mapping-ticket key and the
+            # fresh-confirmed-bug signal. See github-triage-pipeline action's own
+            # CLAUDE.md § "Auto-chaining browser-verify / repro-rebuild".
+            issue_key: ${{ steps.pipeline.outputs.issue_key }}
+            confirmed_bug: ${{ steps.pipeline.outputs.confirmed_bug }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  fetch-depth: 1
+            - name: Install @ag-grid/dev-prompts
+              uses: ./external/ag-shared/github/actions/install-ag-dev-prompts
+              with:
+                  channel: ${{ env.DEV_PROMPTS_CHANNEL }}
+            - id: pipeline
+              uses: ./.ag-dev-prompts/node_modules/@ag-grid/dev-prompts/.github/actions/github-triage-pipeline
+              with:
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #14603, opened 2026-07-24
AG-17910 [Calculated Cols] Persist runtime-added calculated columns in grid state

StephenCooper:
### Problem
Runtime-added (dynamic) calculated columns — created via the header-menu dialog rather than declared in `columnDefs` — were lost on a grid-state round-trip. Saving state, recreating the grid, and setting the saved state back did not bring the calculated columns back.

### Fix
Introduces a `userColumns` section in `GridState` to persist columns that exist at runtime but not in `columnDefs`:

- **`gridState.ts`** — new `UserColumnState` types (`UserColumnStateBase`, `CalculatedUserColumnState`, discriminated by `kind: 'calculated'`) and a `userColumns?: UserColumnState[]` field on `GridState`.
- **`stateService.ts`** — saves `userColumns` from the calculated-columns service on `getState()`, and applies them on restore (`initialState` / `api.setState`). The incoming state is **authoritative**: calc cols it lists are recreated and any it omits are removed (an absent/empty section removes them all). When the feature is disabled, provided `userColumns` are preserved untouched so a later re-save into a calc-enabled grid is not lossy. `propertiesToIgnore` can be used to skip the section entirely.
- **`calculatedColumnsService.ts`** — `reconcileUserColumns` serialises/recreates each dynamic calc col and drops any not present in the restored state. The `groupAnchorColId` is always a **stable defined column** (a leaf from `columnDefs`), never another runtime-added column — so restore order is irrelevant and chained calc cols collapse to a shared leaf anchor.

### Docs
- Updated `grid-state` and `calculated-columns` documentation and examples (community + all framework variants) to show state persistence.

### Tests
`testing/behavioural/src/formulas/calculated-columns-state.test.ts` — behavioural specs covering:
- `initialState` round-trip and the `api.setState` re-entrancy path
- **removal / reconcile**: a calc col absent from the applied state is removed; a partial set keeps listed cols and drops the rest
- save-side freshness (getState reflects the current expression)
- non-lossy behaviour when the feature is disabled
- plural / chained calc cols and grouped-anchor restore

All pass under the throw-on-all-severities dev-validation hook.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.rulesync/rules/examples.md
```diff
@@ -45,6 +45,47 @@ All public documentation examples MUST work across all frameworks:
 - Avoid complex DOM manipulation
 - No external library dependencies
 
+### Hand-authored `provided/` framework variants
```
StephenCooper, line 48: Agent didn't automatically update provided examples, it only did the main.ts file. This change teaches it about provided examples.

documentation/ag-grid-docs/tsconfig.examples.json
```diff
@@ -0,0 +1,63 @@
+{
+  // Type-checks the driving `main.ts` of every documentation example (the vanilla source the
```
StephenCooper: Just locking in the current state. Found this file was missing so the example check was not doing what it was expecting.

PR #14604, opened 2026-07-24
AG-17969 - [Column Model] - Fix cell positioning for reused column groups

gportela85:
fix [AG-17969](https://ag-grid.atlassian.net/browse/AG-17969)

### DIFF
--- a/packages/ag-grid-community/src/columns/columnUtils.ts
+++ b/packages/ag-grid-community/src/columns/columnUtils.ts
@@ -22,6 +22,31 @@ export function getWidthOfColsInList(columnList: AgColumn[]): number {
     return width;
 }
 
+/**
+ * Checks whether a column rebuild added, removed, or replaced a displayed column, ignoring order-only changes.
+ *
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/columns/columnUtils.ts
```diff
@@ -22,6 +22,31 @@ export function getWidthOfColsInList(columnList: AgColumn[]): number {
...
 }
 
+/**
+ * Checks whether a column rebuild added, removed, or replaced a displayed column, ignoring order-only changes.
+ *
+ * @param previous The displayed columns before the rebuild.
+ * @param current The displayed columns after the rebuild.
+ * @param previousById The column-id lookup captured before the rebuild, used to compare column instances.
+ */
+export function _hasColumnMembershipChanged(
```
SalvatorePreviti, line 32: are you sure this is needed? I think we have already a mechanism that detect if column list changed, colsListChanged wll be true if anything changed I think to remember

SalvatorePreviti (approved):

PR #14605, opened 2026-07-24
[AG-115] Feat(column-header): make columnHeaderNameChanged a public event

sdwvit:
Follow-up to #14411: makes `columnHeaderNameChanged` a public grid event so apps can react to header renames.

- Moved from `_INTERNAL_EVENTS` to `_PUBLIC_EVENTS`; `ColumnHeaderNameChangedEvent` now extends `ColumnEvent` and adds a `columnGroup` field (set for group renames, `null` otherwise).
- Added `onColumnHeaderNameChanged` GridOptions callback and the Angular/Vue wrapper bindings (React is covered at runtime).
- Documented on the grid-events reference page.
- `columnHeaderEditHighlightChanged` stays internal.

Fix [AG-115](https://ag-grid.atlassian.net/browse/AG-115)


packages/ag-grid-community/src/entities/agColumn.ts
```diff
@@ -851,7 +851,13 @@ export class AgColumn<TValue = any>
         // without filtering by colId; the grid-level event drives the state service and keeps parity
         // with column groups (which have no per-column event bus).
         this.dispatchColEvent('headerNameChanged', source);
-        this.beans.eventSvc.dispatchEvent({ type: 'columnHeaderNameChanged', colId: this.colId });
+        this.beans.eventSvc.dispatchEvent({
+            type: 'columnHeaderNameChanged',
+            column: this,
+            columns: [this],
```
StephenCooper: Why do we have a `columns` property when it is only set to a single item?

sdwvit: could be null probably. I'll check how other events are shaped

sdwvit: removed in favor of null

StephenCooper (approved):

PR #14615, opened 2026-07-27
[AG-115] Fix(column-header): edit popup padding + header tooltip refresh on rename

sdwvit:
Fixes [AG-115](https://ag-grid.atlassian.net/browse/AG-115) QA follow-up  

## Fixes
- **Edit popup padding** — in the deferred column header edit popup the editor grows to fill the fixed-height dialog, so the space above the action buttons was the flex `gap` while the space below was double the container vertical padding. Set `padding-bottom` to match the gap (Theming API CSS + Legacy SCSS).
- **Header tooltip stale after rename** — `tooltipService` captured the header/group display name once at setup, so renaming a header refreshed the header text but not the tooltip. Resolve the name lazily inside `getTooltipValue` so the `columnHeaderNameChanged` refresh surfaces the edited name.

### DIFF
--- a/packages/ag-grid-community/src/tooltip/tooltipService.ts
+++ b/packages/ag-grid-community/src/tooltip/tooltipService.ts
@@ -291,17 +291,22 @@ export class TooltipService extends BeanStub implements NamedBean {
         }
         const location = 'header';
         const headerLocation = 'header';
-        const valueFormatted = this.beans.colNames.getDisplayNameForColumn(column, headerLocation, true);
-        const value = passedValue ?? valueFormatted;
         const tooltipCtrl: ITooltipCtrl = {
             getGui: () => eGui,
             getLocation: () => location,
-            getTooltipValue: () =>
-                passedValue ??
-                colDef?.headerTooltipValueGetter?.(
-                    _addGridCommonParams(gos, { location, colDef, column, value, valueFormatted })
-                ) ??
-                colDef?.headerTooltip,
+            // Resolve the display name on each read so a refresh (e.g. after a header rename) reflects the
+            // current name, rather than serving the name captured when the tooltip was first set up.
+            getTooltipValue: () => {
+                if (passedValue != null) {
+                    return passedValue;
+                }
+                const valueFormatted = this.beans.colNames.getDisplayNameForColumn(column, headerLocation, true);
+                return (
+                    colDef?.headerTooltipValueGetter?.(
+                        _addGridCommonParams(gos, { location, colDef, column, value: valueFormatted, valueFormatted })
+                    ) ?? colDef?.headerTooltip
+                );
+            },
             shouldDisplayTooltip,
             getAdditionalParams: () => ({
                 column,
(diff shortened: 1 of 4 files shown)


testing/behavioural/src/columnHeaderEdit/editable-header-name-tooltip-and-grouping.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import { TestGridsManager, asyncSetTimeout } from '../test-utils';
+
+/**
+ * A renamed header name is stored as an override on the column entity (and the group-id-keyed group
+ * map), so it flows through the same display-name resolver the header tooltip and the row-group
+ * machinery consume. These tests lock in two behaviours not covered by editable-header-name.test.ts:
+ * the header tooltip reflecting a rename, and a leaf rename surviving a row-group/ungroup cycle.
+ */
+describe('Editable header name — tooltips', () => {
```
StephenCooper, line 17: Do we have tests that also validate the editted header name is correctly displayed for each HeaderLocation

export type HeaderLocation =
    | 'chart'
    | 'columnDrop'
    | 'columnToolPanel'
    | 'csv'
    | 'filterToolPanel'
    | 'groupFilter'
    | 'header'
    | 'model'
    | 'advancedFilter'
    | null;

sdwvit: addressing, ty

testing/behavioural/src/columnHeaderEdit/editable-header-name-tooltip-and-grouping.test.ts
```diff
@@ -97,6 +95,43 @@ describe('Editable header name — tooltips', () => {
...
+    const gridMgr = new TestGridsManager({ modules: [AllEnterpriseModule] });
+
+    afterEach(() => {
+        gridMgr.reset();
+        vi.resetAllMocks();
+    });
+
+    // The override is resolved in getHeaderName ahead of the headerValueGetter, independent of the
+    // location passed in, so the edited name must be returned for every consumer location.
+    const locations: Exclude<HeaderLocation, null>[] = [
```
StephenCooper: Sorry, this isn't what I meant. 

I was suggesting that we have a test that ensures that the rendered header in each of the locations is accurate. 

That means having different examples that enable each of these header locations to be validated. i.e one will need integrated charts.

sdwvit: ah ok, gonna add these tests. a bit more difficult task.

packages/ag-grid-community/src/tooltip/tooltipService.ts
```diff
@@ -342,18 +347,23 @@ export class TooltipService extends BeanStub implements NamedBean {
...
-                    _addGridCommonParams(gos, { location, colDef, column, value, valueFormatted })
-                ) ??
-                colDef?.headerTooltip,
+            // Resolve the display name on each read so a refresh (e.g. after a group rename) reflects the
+            // current name, rather than serving the name captured when the tooltip was first set up.
+            getTooltipValue: () => {
+                if (passedValue != null) {
+                    return passedValue;
+                }
+                const valueFormatted = this.beans.colNames.getDisplayNameForColumnGroup(column, headerLocation);
```
StephenCooper: Should we have a helper that removes the duplication here. I think I can only see the `includeAggFunc` param value as the difference.

StephenCooper (approved):
Have approved but if you could check if the code duplication can be resolved before merging. If not its ok to proceed.

PR #14617, opened 2026-07-27
[AG-9664] Fix(pivoting): apply pivotSort from colDef on column creation

sdwvit:
Fixes [AG-9664](https://ag-grid.atlassian.net/browse/AG-9664)

- `colDef.pivotSort` / `initialPivotSort` are now seeded onto the column at creation, so they take effect on initialisation 
- `pivotSort` API docs now show `default: 'asc'` 
- The two SSRM pivoting examples show the pivot / row group panels so the sorting is discoverable

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
@@ -35,11 +35,12 @@ ModuleRegistry.registerModules([
 let gridApi: GridApi<IOlympicData>;
 const gridOptions: GridOptions<IOlympicData> = {
     columnDefs: [
-        { field: 'country', rowGroup: true },
-        { field: 'year', pivot: true }, // pivot on 'year'
-        { field: 'gold', aggFunc: 'sum' },
-        { field: 'silver', aggFunc: 'sum' },
-        { field: 'bronze', aggFunc: 'sum' },
+        { field: 'country', rowGroup: true, enableRowGroup: true },
+        { field: 'sport', enableRowGroup: true },
+        { field: 'year', pivot: true, enablePivot: true }, // pivot on 'year'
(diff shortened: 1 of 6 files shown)


documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
```diff
@@ -52,6 +54,12 @@ const gridOptions: GridOptions<IOlympicData> = {
 
     // enable pivoting
     pivotMode: true,
+
```
StephenCooper, line 58: Does this example fall into the trap of the user not being able to change anything in the rowGroupPanel if we have not enabled row grouping or pivoting?

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/supplying_pivot_result_fields/main.ts
```diff
@@ -46,6 +48,12 @@ const gridOptions: GridOptions<IOlympicData> = {
     // enable pivoting
     pivotMode: true,
 
+    sideBar: {
```
StephenCooper, line 52: Same for this example, can it be interacted with in the sidebar and panel?

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
sdwvit: thanks, fixing it

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/supplying_pivot_result_fields/main.ts
sdwvit: ty, fixing

StephenCooper (approved):

PR #14627, opened 2026-07-27
Test(docs-e2e): fix remaining nightly Documentation Tests failures

sdwvit:
Accounts for all 93 failures in the nightly [Documentation Tests](https://github.com/ag-grid/ag-grid/actions/runs/30235437697) run, and fixes the 30 that were still outstanding.

That run's head SHA is `2a06528` (03:47); #14610 merged at 09:37, **after** it started, so 63 of the 93 were already fixed. I re-ran those against staging to confirm.

| Cause | Tests |
|---|---|
| Already fixed by #14610, which merged after this run started: `row-spanning-simple` (18), `status-bar` (9), `multi-column` (9), `suppress-click-selection` (9), `server-side` (9), `customising-menu-items` (9). Re-ran against staging, all green. | 63 |
| `grid-state` / `set-state`: `age` is 90px wide, so the always-shown menu/filter icons squeeze the sort label to 18px and a centre click hits the menu. Added `clickHeaderToSort`. | 9 |
| `post-sort` / `absolute-sorting`: per-row-await `orderedValues` copies blew the test budget inside `toPass()`. Promoted #14610's single-`evaluate` version to a shared helper and deduped all 4 copies. | 7 |
| `pdf-widths-autosizing` / `pdf-wrapping-grouping-pinning`: `findAllVariables` stored a destructured param as the pattern's source text, so `column` leaked into the generated `useCallback` deps where it is out of scope. | 6 |
| `pdf-export-page-setup`: the generators inline the `rowData` initialiser and drop the declaration, leaving a `rowData.push(...)` loop on a dead name. | 5 |
| `rtl-complex`: the 1x1 stub PNG has a bad IDAT CRC and no IEND chunk; Firefox rejects it. | 2 |
| `column-groups/group-changes`: Angular/Vue reformat long button labels onto their own lines, so the anchored regex never matched. | 1 |
| **Total** | **93** |

Regenerating all 1004 examples with the parser change diffs only those two examples (`}, [column])` -> `}, [])`).

The 95 flaky tests are not test defects: the console errors behind them are 404s, 502s and `MIME type ('text/html')` / Quirks-Mode warnings, i.e. staging serving error pages mid-redeploy. All 146 tests across the flaky-listed examples pass against healthy staging, and the six worst offenders pass at `--repeat-each 3`.

### DIFF
--- a/plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
+++ b/plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
@@ -305,6 +305,23 @@ export function tsNodeIsTopLevelFunction(node: any): boolean {
 }
 
 /**
+ * The names a declaration's binding name introduces into scope: the identifier itself, or every
+ * element of an object/array destructuring pattern.
+ */
+function boundNames(name: ts.BindingName): string[] {
+    if (ts.isObjectBindingPattern(name) || ts.isArrayBindingPattern(name)) {
+        const names: string[] = [];
+        name.elements.forEach((element) => {
+            if (ts.isBindingElement(element)) {
+                names.push(...boundNames(element.name));
+            }
+        });
+        return names;
+    }
+    return [name.getText()];
+}
+
+/**
  * Find all the variables defined in this node tree recursively
  */
 export function findAllVariables(node) {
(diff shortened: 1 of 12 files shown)


StephenCooper (approved):

plugins/ag-grid-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
```diff
@@ -304,6 +304,23 @@ export function tsNodeIsTopLevelFunction(node: any): boolean {
     return false;
 }
 
+/**
+ * The names a declaration's binding name introduces into scope: the identifier itself, or every
+ * element of an object/array destructuring pattern.
+ */
+function boundNames(name: ts.BindingName): string[] {
```
sdwvit, line 311: it is here to fix a bug in the generator. The generator sees a fn like ({ column }) => ... and extracts `column` into a variable, then react example gen picks it up as a useCallback dependency, causing a script error in systemjs, and, as a result, a test timeout.

PR #14641, opened 2026-07-28
[AG-9664] Fix(pivoting): pivot sort state and SSRM pivot result column order

sdwvit:
Fixes [AG-9664](https://ag-grid.atlassian.net/browse/AG-9664) (TC11, TC12, TC13).

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/example.spec.ts
+++ b/documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/example.spec.ts
@@ -1,15 +1,5 @@
 import { expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
-// Header group cells are absolutely positioned, so visual order comes from their `left` offset
-// rather than DOM order. Return the year group headers ordered left-to-right.
-const yearGroupOrder = (page: import('@playwright/test').Page) =>
-    page.locator('.ag-header-group-cell .ag-header-group-text').evaluateAll((els) =>
-        els
-            .map((e) => ({ year: e.textContent!.trim(), left: e.getBoundingClientRect().left }))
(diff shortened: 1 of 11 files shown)


documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/example.spec.ts
```diff
@@ -1,5 +1,15 @@
 import { expect, test, waitForGridContent } from '@utils/grid/test-utils';
 
+// Header group cells are absolutely positioned, so visual order comes from their `left` offset
+// rather than DOM order. Return the year group headers ordered left-to-right.
+const yearGroupOrder = (page: import('@playwright/test').Page) =>
```
StephenCooper: Can we just import the type normally.

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
```diff
@@ -107,20 +108,51 @@ function getServerSideDatasource(server: any): IServerSideDatasource {
     };
 }
 
+let suppliedPivotResultColsKey: string | undefined;
+
 function addPivotResultCols(request: IServerSideGetRowsRequest, response: any, api: GridApi) {
-    // check if pivot colDefs already exist
-    const existingPivotColDefs = api.getPivotResultColumns();
-    if (existingPivotColDefs && existingPivotColDefs.length > 0) {
+    // the sort direction of each pivot chip, so the columns can be ordered to match
```
StephenCooper: Can we check if the behavior is correct that this SSRM has the pivot sort enabled in the UI.

This needs to be opt in so if this is not implemented the end user is not confused by the sort not working.

SalvatorePreviti (approved):

PR #14645, opened 2026-07-28
[AG-9664] Fix(pivoting): client-side pivot column sort in SSRM and for supplied pivot result columns

sdwvit:
Pivot pill sorting in SSRM is now a client-side column reorder instead of a store reset + refetch, and it applies to application-supplied pivot result columns.

- `setPivotResultColumns` resets `pivotSort` to no sort, then sorting reorders the supplied columns.
- `colDef.pivotSort` is now applied on a column-definitions update, not only on creation.
- Fixed `pivotComparator` JSDoc (it is not CSRM-only) and documented the `pivotSort` interaction.

Tests: pivot sort options/properties coverage, SSRM no-refetch and supplied-column ordering, colDef update path.

Fixes [AG-9664](https://ag-grid.atlassian.net/browse/AG-9664)


documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
```diff
@@ -107,6 +107,9 @@ api.applyColumnState({
 Sorting reorders the pivot column groups while preserving any column width and within-group ordering changes
 the user has made.
 
+Supplying the pivot result columns yourself with `setPivotResultColumns` resets `pivotSort` to no sort on every
```
StephenCooper: Does this work for CSRM? Not sure we should be calling this out in the Pivoting docs if we only really promote this function for SSRM.

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
```diff
@@ -163,10 +163,31 @@ function addColDef(
     return res;
 }
 
+// Stable pseudo-random rank for a pivot key, so the scrambled order is the same on every run.
+function pseudoRandomRank(pivotKey: string): number {
```
StephenCooper: We have built in support for this via the example generator. Should just need to use Math.random, and the generator will seed it via the seedrandom lib.

packages/ag-grid-community/src/interfaces/iPivotColDefService.ts
```diff
@@ -3,5 +3,7 @@ import type { ColDef, ColGroupDef } from '../entities/colDef';
 /** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
 export interface IPivotColDefService {
     createColDefsFromFields: (fields: string[]) => (ColDef | ColGroupDef)[];
+    /** Order supplied pivot result colDefs (in place) by the pivot columns' `pivotSort`. */
```
StephenCooper: in place? Does this mutate the app provided columns?

packages/ag-grid-enterprise/src/pivot/pivotApi.ts
```diff
@@ -45,7 +46,29 @@ export function getPivotColumns(beans: BeanCollection): Column[] {
...
+    }
+}
+
+/** The app owns the pivot result column order once it supplies the columns itself, so any `pivotSort` is
+ *  cleared to the explicit "no sort" value rather than left claiming an ordering the grid didn't apply.
+ *  Returns the cleared cols for the caller to announce. */
+function clearPivotSort(beans: BeanCollection): AgColumn[] | null {
+    const pivotCols = beans.pivotColsSvc?.columns ?? [];
+    let cleared: AgColumn[] | null = null;
+    for (let i = 0, len = pivotCols.length; i < len; ++i) {
```
StephenCooper: Can we confirm if this was the desired behaviour. If pivotSort is on a column I thought the idea was to use that as the sort direction when the app provided the columns. As this is an new opt in behaviour they would have specified.

packages/ag-grid-enterprise/src/pivot/pivotResultColsService.ts
```diff
@@ -15,6 +15,37 @@ import type {
 } from 'ag-grid-community';
 import { BeanStub, _buildColumnTree, _destroyColumnTreeAll, _destroyColumnTreeUnused } from 'ag-grid-community';
 
+/** A colDef and its children, in supplied order; children `null` for a leaf col. */
+type ColDefOrder = {
+    def: ColDef | ColGroupDef;
+    children: ColDefOrder[] | null;
+};
+
+const captureColDefOrder = (colDefs: (ColDef | ColGroupDef)[]): ColDefOrder[] => {
```
StephenCooper: Do we need this capture and restore approach if we leave their application provided pivot cols as they are so that it can always then be used as the natural ordering?

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
sdwvit: I don't need random, just something not ordered

sdwvit: would this ag random work that way? may be the same amount of code actually

packages/ag-grid-community/src/interfaces/iPivotColDefService.ts
sdwvit: ah forgot to check, ty for spotting

packages/ag-grid-enterprise/src/pivot/pivotApi.ts
sdwvit: yea-no, applying what you suggested irl

packages/ag-grid-enterprise/src/pivot/pivotResultColsService.ts
sdwvit: exactly, applying it

packages/ag-grid-enterprise/src/pivot/pivotApi.ts
sdwvit: Confirmed, and changed to what you describe: `setPivotResultColumns` no longer touches `pivotSort`. The supplied order is now the natural order that `pivotSort: null` resolves to, and any other direction (including the ascending default) orders the supplied columns. That matches the ticket, where `pivotSort` defaults to `asc` and TC8/TC10 were raised specifically because an explicitly supplied `pivotSort`/`initialPivotSort` must take effect — clearing it here contradicted that.

Dropping the clear also removed the `columnPivotChanged` dispatch, which fixed two of the Codex findings for free (the SSRM ownership clobber and the duplicate tree rebuild).

packages/ag-grid-enterprise/src/pivot/pivotResultColsService.ts
sdwvit: No — removed. The supplied array is now retained as-is and used directly as the natural ordering, exactly as you suggest; `resortPivotResultCols` re-derives the order from it on every change, so "no sort" resolves to the supplied order rather than to the last direction. That deleted `captureColDefOrder`/`restoreColDefOrder` and the `suppliedOrder` field (~60 lines).

packages/ag-grid-community/src/interfaces/iPivotColDefService.ts
sdwvit: Good catch — it did. Fixed: renamed to `orderPivotResultColDefs`, which returns a reordered copy (or the array itself when nothing moves) and never mutates the supplied arrays or defs. Group defs that need reordered children are shallow-cloned with their `groupId` carried over so the built tree still reuses the same group instance. Added a behavioural test asserting the application-owned array is unchanged after a sort.

documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
sdwvit: You are right, and it is removed. Checking it properly: under CSRM the pivot stage regenerates the pivot result columns from the row data on the next refresh, so application-supplied columns are transient there — `setPivotResultColumns` is only meaningful for SSRM. I dropped the paragraph from this page and kept the note on the SSRM pivoting page instead. I also moved the supplied-column ordering tests into the SSRM behavioural file for the same reason.

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
sdwvit: Switched to `Math.random` as you suggested — the generator seeds it via seedrandom, so it stays deterministic, and it replaced ~20 lines of hand-rolled hashing with a 5-line shuffle. I also relaxed the e2e assertion to check the no-sort order holds the same years rather than an exact sequence, since a shuffle has no fixed result.

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -41,6 +41,65 @@ const convertToHeaderNameComparator =
...
+        return null;
+    }
+    const ordered = colDefs.slice().sort(comparators[depth]);
+    let changed = false;
+    for (let i = 0, len = ordered.length; i < len; ++i) {
+        if (ordered[i] !== colDefs[i]) {
+            changed = true;
+        }
+        const groupDef = ordered[i] as ColGroupDef;
+        const children = orderColDefLevel(groupDef.children, depth + 1, comparators);
```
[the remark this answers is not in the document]
sdwvit: Fixed. `orderColDefLevel` now skips entries with no `children` rather than recursing into them, so a leaf's `undefined` children never reach the next level and `colDefs.length` is never read off `undefined`.

Regression added in `pivot-column-sort-ssrm.test.ts` — "pivotSort orders flat supplied pivot columns when pivoting on two columns" — covering two pivot columns with an entirely flat supplied result.

packages/ag-grid-enterprise/src/pivot/pivotColDefService.ts
```diff
@@ -41,6 +41,65 @@ const convertToHeaderNameComparator =
...
+
+/** Whether a level of supplied colDefs represents pivot keys, so is ordered by `depth`'s pivot column.
+ *
+ *  A level of groups always is. A level of leaves only is at the top, where a single pivot column - or
+ *  `removePivotHeaderRowWhenSingleValueColumn` - legitimately flattens the pivot keys to plain colDefs. Deeper
+ *  down, leaves are the measure columns of one pivot key: ordering those by header name would interleave measures
+ *  across pivot keys and destroy their within-key order. A mixed level has no single meaning, so it is left alone. */
+const isPivotKeyLevel = (colDefs: (ColDef | ColGroupDef)[], depth: number): boolean => {
+    const kind = levelKind(colDefs);
+    return kind === 'groups' || (kind === 'leaves' && depth === 0);
```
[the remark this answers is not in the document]
sdwvit: Half addressed, and I've left this thread open for the other half.

**Mixed levels — fixed.** A level mixing pivot groups with a standalone colDef is no longer skipped: the groups now sort among the slots they occupy while plain colDefs pin to their index, so a row total holds its edge instead of the whole level being left alone. Regression added in `pivot-column-sort-ssrm.test.ts`.

**Nested flattened levels — not applied.** I tried dropping the `depth === 0` guard and it breaks the existing `pivot-column-sort-ssrm.test.ts:201`, which documents a supplied structure that nests measure leaves directly under a year group rather than spelling out a second pivot level.

That makes an all-leaves level at `depth < comparators.length` genuinely ambiguous — it is either flattened pivot keys (the `removePivotHeaderRowWhenSingleValueColumn` case you describe) or the measures of a single pivot key (that test). Nothing in the supplied colDef shape distinguishes the two, and treating measures as pivot keys would sort them by header name and interleave them across pivot keys, which is the worse failure.

Left as-is with the reasoning recorded in the `orderColDefLevel` docstring. Happy to revisit if we introduce an explicit signal for a flattened level.

testing/behavioural/src/sorting/pivot-column-sort-options.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    // to the front. Recorded as-is: whether it should hold its position is with QA.
+    test('pivotRowTotals: after sorts the row total along with the pivot groups', async () => {
+        const api = await createGrid({ pivotRowTotals: 'after' });
+        expect(pivotsWithRowTotal(api)).toEqual([...ASC, ROW_TOTAL]);
+
+        api.applyColumnState({ state: [{ colId: 'year', pivotSort: 'desc' }] });
+        await asyncSetTimeout(10);
+        expect(pivotsWithRowTotal(api)).toEqual([ROW_TOTAL, ...DESC]);
+    });
+
```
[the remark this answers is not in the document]
sdwvit: Fixed, though the cause was not the pivot-group comparator.

`pivotGroupKey` keyed a row total (`pivotKeys: []`) and the auto group column identically as `''`. The two shared a bucket in `reRankByPivotGroupOrder`, so the total inherited the auto column's leading rank and sorted to the front. Ungrouped cols are now keyed by colId, so each ranks at its own position and a total holds its configured edge.

The test now expects `[...DESC, ROW_TOTAL]`. Worth noting the `pivotRowTotals: 'before'` case was previously passing only by luck — the same collision happened to place it correctly.

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
```diff
@@ -163,10 +163,21 @@ function addColDef(
     return res;
 }
 
+// The supplied order is the pivot result columns' natural order, used when the YEAR pill in the pivot panel is
+// cycled to no sort. This example supplies the years shuffled so that order is distinguishable from asc/desc.
+function scramblePivotFields(pivotFields: string[]): string[] {
+    return pivotFields
+        .map((field) => ({ field, rank: Math.random() }))
+        .sort((a, b) => a.rank - b.rank)
```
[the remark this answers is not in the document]
sdwvit: Fixed — and the nondeterminism was real, though not for the reason assumed during review.

`Math.random` is not seeded here. The docs runner does `window.agRandom = new Math.seedrandom(SEED)`, and the `new` form returns a standalone prng rather than monkey-patching `Math.random`; nothing else in the repo patches it. So the example genuinely varied run to run.

Switched to `window.agRandom()`, the seeded generator (precedent: `row-sorting/absolute-sorting` and `column-headers-components/header-template`). Rather than hard-code a permutation coupled to the seed and its two warm-up calls, the e2e now asserts the no-sort order is neither ascending nor descending — which is what the example actually claims.

testing/behavioural/src/sorting/pivot-column-sort-ssrm.test.ts
```diff
@@ -17,30 +17,302 @@ describe('SSRM: interactive pivot column sorting (pivotSort)', () => {
...
+        };
+        const api = await createPivotGrid(datasource);
+
+        // Supplied in an order that is neither ascending nor descending, so "no sort" is distinguishable.
+        api.setPivotResultColumns([
+            { groupId: '2004', headerName: '2004', children: [{ colId: '2004_gold', field: '2004_gold' }] },
+            { groupId: '2000', headerName: '2000', children: [{ colId: '2000_gold', field: '2000_gold' }] },
+            { groupId: '2008', headerName: '2008', children: [{ colId: '2008_gold', field: '2008_gold' }] },
+        ]);
+        await asyncSetTimeout(50);
```
StephenCooper: Are these hardcoded timeouts the only way to write these tests. I am concerned they will become flaky.

sdwvit: yeah lets avoid them

StephenCooper:
Nit: `arePivotColsUnchanged` (serverSideRowModel.ts) hand-rolls a length-check + indexed loop that duplicates what the shared `_areEqual<T>(a, b, comparator?)` helper (`packages/ag-stack/src/utils/array.ts`) already does elsewhere in this codebase (e.g. `columnStateUpdateExecutionStrategy.hasPendingChanges`, `columnToolPanel.isGridStateEqual`). Could this be rewritten as:

```ts
private arePivotColsUnchanged(): boolean {
    const oldCols = this.storeParams?.pivotCols;
    if (!oldCols) return false;
    const newCols = this.columnsToValueObjects(this.pivotColsSvc?.columns);
    return _areEqual(oldCols, newCols, (a, b) => a.id === b.id && a.field === b.field && a.aggFunc === b.aggFunc);
}
```

Not blocking, but would remove a hand-rolled duplicate of an existing utility.

documentation/ag-grid-docs/src/content/docs/server-side-model-pivoting/_examples/creating_pivot_result_columns/main.ts
```diff
@@ -163,10 +163,22 @@ function addColDef(
     return res;
 }
 
+// The supplied order is the pivot result columns' natural order, used when the YEAR pill in the pivot panel is
+// cycled to no sort. This example supplies the years shuffled so that order is distinguishable from asc/desc.
+// `window.agRandom` is the docs' seeded generator, so the shuffle is the same on every run - unlike `Math.random`.
+function scramblePivotFields(pivotFields: string[]): string[] {
+    return pivotFields
+        .map((field) => ({ field, rank: window.agRandom() }))
```
StephenCooper: The generator will do this for you but means then the dev looks at the example source code on the website they will just see Math.random().

```suggestion
        .map((field) => ({ field, rank: Math.random() }))
```

sdwvit: ah good to know

SalvatorePreviti (approved):

PR #14649, opened 2026-07-29
AG-14065 add trial license loading state

skewis6:
Fix AG-14065


taktran:
This will be implemented in https://github.com/ag-grid/ag-grid/pull/14646

PR #14663, opened 2026-07-29
AG-17910 Save calculated columns in grid state

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-17910

Calculated columns were not saved in grid state: after saving state, recreating the grid and restoring, user-added calculated columns did not come back, and user edits to declared columns (renames, expression changes, removals) were lost.

Grid state only described columns that `columnDefs` already declared, so there was nowhere to record columns the user created at runtime, or the changes they made to declared ones.

This adds a user-owned column layer that grid state persists alongside the existing column state:

- **Created columns** — calculated columns the user adds are recorded with their properties and group placement, and are rebuilt on restore.
- **Overrides** — user edits to declared columns are stored as a delta against the developer's original `columnDefs` declaration, so reverting an edit persists nothing.
- **Tombstones** — declared columns the user removes are recorded, so the removal survives a restore.

The layer records user choices only; it is not a second route into `columnDefs`. Features register as owners along with the properties their UI can produce, and incoming state is filtered to those properties. With no owner active — feature disabled or the enterprise module absent — entries are preserved and re-saved unchanged but never reach the built columns, so state saved on a grid with the feature is not lossy on one without it. Setting `columnDefs` or resetting column state is the developer reclaiming the columns, and clears the layer.

Fix [AG-17910](https://ag-grid.atlassian.net/browse/AG-17910)


packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -424,25 +437,43 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
         return typeof calculatedColumns === 'object' && calculatedColumns != null ? calculatedColumns : undefined;
     }
 
-    public overrideFor(colDef: ColDef): ColDef | null | undefined {
-        if (!this.isEnabled()) {
-            return undefined;
+    /** Records the user's changes to a `columnDefs`-declared calc col in the user-column layer. Properties
+     *  put back to the declared value are dropped, so reverting an edit persists nothing. */
+    private setDeclaredColOverride(column: AgColumn, nextColDef: ColDef): void {
```
gportela85, line 456: After the first rebuild, `getUserProvidedColDef()` contains the declared definition with the existing override already merged into it. 

Diffing the next edit against that value and replacing the entire map entry means an expression edit followed by a title only edit stores only the title and the expression silently reverts to the original `columnDefs` value. The service needs access to the untouched declared definition, or it must merge new deltas with the existing entry. Not sure which option would be better.

packages/ag-grid-community/src/columns/userColumns/userColumnService.ts
```diff
@@ -0,0 +1,219 @@
...
+                      }
+                    : { colId, properties: toProperties(entry.properties) }
+            );
+        });
+        return state;
+    }
+
+    /** Replaces the layer with `states`. Returns whether anything changed, so the caller can skip the
+     *  column tree rebuild when a re-applied state is identical. */
+    public setState(states: UserColumnState[] | undefined): boolean {
```
gportela85, line 186: Whether an override is accepted depends on colModel.getCol(), not on whether columnDefs declares the column. If the current state "tombstoned" a declared column, applying a later state containing an override for it drops that override because the column is currently absent. 

The inverse collision check for created columns has the same problem and can allow a created column to occupy a currently "tombstoned" declared ID. Resolving ownership against the raw declared definitions should work.

packages/ag-grid-community/src/columns/buildColumnTree.ts
```diff
@@ -214,9 +214,9 @@ export function _buildColumnTree(
      *  `colId` when present (reuse only the same-colId column — a changed colId is a new column even on a
      *  retained colDef ref); without a colId it is the colDef ref, then field/positional in buildKeyedColumn. */
     const buildColumn = (def: ColDef): AgColumn | undefined => {
-        const override = calculatedColsSvc?.overrideFor(def);
+        const override = userColumnSvc.overrideFor(def);
```
gportela85: The core build applies every override unconditionally. With calculatedColumns disabled or the module absent, created entries are merely preserved, but "declared column" overrides still alter columns and tombstones still remove them. When the module is missing, `setColumnDefs` and `resetColumnState` also cannot clear this generic layer because cleanup is delegated to the optional calculatedColumnsService. This contradicts the tested disabled-feature behaviour.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -490,10 +522,89 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+                anchorColId: this.findAnchorInGroup(entry.parentGroupId),
+                instance: null,
+            });
+            changed = true;
+        });
+        return changed;
+    }
+
+    /** A leaf in `groupId` to append a restored column after, so it is built into that group. */
+    private findAnchorInGroup(groupId: string | null | undefined): string | null {
```
gportela85, line 617: adoptUserColumns removes stale dynamic columns from its map before this search, but those columns remain in colDefList until the subsequent rebuild. When setState replaces a grouped calculated column with another one, this reverse scan can choose a column that will be removed as the anchor. The new column then falls back to the top level. Searching only surviving declared/restored columns or attaching directly to the persisted group would be better.

packages/ag-grid-enterprise/src/columnHeaderEdit/columnHeaderEditService.ts
```diff
@@ -63,9 +63,12 @@ export class ColumnHeaderEditService extends BeanStub implements NamedBean, ICol
     }
 
     private isEditable(target: EditTarget): boolean {
-        return isProvidedColumnGroup(target)
-            ? !!target.colGroupDef?.headerNameEditable
-            : !!target.colDef.headerNameEditable;
+        if (isProvidedColumnGroup(target)) {
```
gportela85, line 66: I haven't tested this but wouldn't this make `headerNameEditable` option ineffective for every calculated column?

StephenCooper: This is the intention.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
StephenCooper: Good catch. Fixed in `ab48003104d`: the service now diffs against `userColumnSvc.getDeclaredDef(colId)`, which returns the developer's untouched declaration rather than the built column's merged definition, so successive edits to different properties accumulate.

packages/ag-grid-community/src/columns/userColumns/userColumnService.ts
StephenCooper: Fixed in `ab48003104d`. Ownership resolves against the raw declared definitions via `isDeclared()` now, both here and for the created-column collision check in `createUniqueColId`/`adoptUserColumns`, so tombstoned declared columns keep their identity.

packages/ag-grid-community/src/columns/buildColumnTree.ts
StephenCooper: Fixed in `ab48003104d`. `overrideFor` returns `undefined` when no owner is active, so declared-column overrides and tombstones no longer apply with the module absent, and `clear()` now runs from `setColumnDefs`/`_resetColumnState` in core rather than being delegated to the optional service.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
StephenCooper: Fixed in `ab48003104d`. `findAnchorInGroup` now skips tombstoned entries and only anchors to columns that survive the rebuild (declared or currently-dynamic).

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -478,22 +520,112 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+        }
+        const userColumnSvc = this.userColumnSvc;
+        const dynamicColumns = this.dynamicColumns;
+        let changed = false;
+        const staleColIds: string[] = [];
+        dynamicColumns.forEach((_dc, colId) => {
+            if (userColumnSvc.getEntry(colId) === undefined) {
+                staleColIds.push(colId);
+            }
+        });
```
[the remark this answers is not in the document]
StephenCooper: Fixed in `c112d92099c`.

Confirmed with three behavioural tests before fixing — for a tombstone and for a created entry with no expression the column did stay rendered (`['a', 'calculated_1', 'b']`). The cleanup asked *"is there an entry with this colId?"* when it needed to ask *"is there an entry that **defines** this column?"*, so it now keeps a column only for an entry that describes a created calculated column, via a `describesCalcCol` predicate shared with the adoption guard below it. Applied to `inactiveDynamicColumns` too, so a parked column whose entry no longer describes a calc col can't be resurrected by a later `applyColumnState`.

One correction to the report's severity: the declared-column-override case named here **already behaved correctly**, and that pins the cause rather than widening it. `userColumnService.setState` only stores a non-created entry when `isDeclared(colId)`, and a dynamic calc col is never in `columnDefs` — so that entry is discarded, `getEntry` returns `undefined`, and the old check happened to fire.

That also bounds the real-world reach: deleting a dynamic calc col never writes a tombstone (`removeColumn` deletes the entry outright, since nothing declares the column and so nothing could resurrect it), and a genuinely declared `calculated_1` is refused by the `isDeclared` guard anyway. So a tombstone for a dynamic calc colId only arises from hand-authored, app-mangled or cross-version state. Worth fixing as `setState` contract robustness — it was leaving an unowned column instead of ignoring garbage — but I'd grade it P2 rather than P1.

packages/ag-grid-enterprise/src/calculatedColumns/calculatedColumnsService.ts
```diff
@@ -478,22 +520,112 @@ export class CalculatedColumnsService extends BeanStub implements NamedBean, ICa
...
+                return;
+            }
+            const colDef = this.toCalculatedColDef({ ...properties }, colId);
+            const existing = dynamicColumns.get(colId);
+            if (existing) {
+                if (!_mergedEqual(colDef, existing.colDef)) {
+                    existing.colDef = colDef;
+                    changed = true;
+                }
+                return;
```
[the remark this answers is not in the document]
StephenCooper: Fixed in `c112d92099c`. Agreed this one is a genuine P1 — reproduced purely through the UI.

Adoption now re-seats a held column when the entry names a different group:

```ts
const instance = existing.instance;
if (instance !== null && (entry.parentGroupId ?? null) !== _getParentGroupId(instance)) {
    existing.anchorColId = this.findAnchorInGroup(entry.parentGroupId);
    changed = true;
}
```

Two deliberate constraints:

- **It compares against the column's actual group rather than recomputing unconditionally.** A column added next to a sibling carries a meaningful `anchorColId` that also fixes its display position; recomputing on every adoption would collapse that to `findAnchorInGroup(null) === null` for any top-level column and silently lose the position. Comparing first makes it a no-op when the column is already where the entry says.
- **It skips unbuilt columns** (`instance === null`). `_getParentGroupId(null)` returns `null`, which would read as "currently top-level" and misfire; before the first build `anchorColId` was already derived from the entry, so there is nothing to reconcile.

What made this user-visible is that the same state produced two different layouts: restoring into a *fresh* grid worked, because that goes down the create path where `parentGroupId` is read — only a grid where the column already existed lost the placement.

StephenCooper:
Two further changes pushed that don't correspond to a review comment, so flagging them here for visibility.

### `6db5735866a` — `UserColumnService` internal JSDoc marker

Pre-existing on the branch: `main-internal.test.ts` requires every symbol exported from `main-internal.ts` to carry the internal marker, and the class JSDoc lacked it, so the full unit suite was red. One line, no behaviour change.

### `f6eb9ce28a3` — calculated column colIds are no longer reused

**This one changes observable colIds**, so it's worth a look. `createUniqueColId` allocated the lowest free `calculated_N`, which meant deleting a column freed its colId for the next one added. Anything that outlives the column and refers to it *by id* then bound to an unrelated column. Two confirmed cases, both reproduced before fixing:

- A sort saved against the deleted column attached to the new one holding its colId.
- Worse: a dependent column's expression silently recomputed from unrelated data. Internal expressions store colIds rather than header text, so `[Doubled] + 1` (which evaluated to `11`) started returning **`201`** — `b * 100 + 1` — after its source was deleted and a new column took `calculated_1`. Silently wrong values, no error.

The allocator now tracks the highest index handed out and resumes above it:

```ts
let index = this.lastCalculatedColIndex;
do {
    colId = `calculated_${++index}`;
} while (colModel.getCol(colId) !== undefined || parked.has(colId) || userColumnSvc.getEntry(colId) !== undefined);
this.lastCalculatedColIndex = index;
```

The existing occupancy checks stay — they handle collisions with live, parked and state-restored columns; the mark only stops a *deleted* index coming back. It is per-service, so per-grid.

**Visible consequence:** adding a calculated column, deleting it, then adding another now yields `calculated_2`, not `calculated_1`. Nothing in the docs specifies the numbering, and state carries `parentGroupId` and the expression so the colId is only a handle — but if anything depends on the old numbering, this is where to object.

State restore needed no special handling: restoring `calculated_9` into a fresh grid leaves the mark at 0, so the next add takes `calculated_1` (free, no collision) and monotonicity holds from there.

Three behavioural tests cover it, each converted from a failing repro: deletion not freeing the colId, the stale sort not attaching, and the dangling expression staying `#PARSE!` rather than resolving to unrelated data.

Full `./behave.sh` green (10084 passed, 0 failed), plus format, `build:types`, lint and `build:test`.

gportela85 (approved):

PR #14668, opened 2026-07-30
perf(csrm): identify displayed rows by slot instead of a Set of row ids

SalvatorePreviti:
# perf(csrm): clear stale row tops from the previously-positioned rows

## What

The `map` stage gives every displayed row a top and index, then clears the stale tops left on rows that
dropped out. Finding what to clear was the expensive half: a `Set<string>` of every displayed row id,
then a recursion over the whole node tree testing `Set.has(id)` per node.

**1. Identify displayed rows by slot, not by id.** `setRowTopAndRowIndex` has already given each
displayed row its index, so an array index and a pointer compare replace the set, the N string hashes
filling it, and a hash lookup per node:

```ts
if (rowIndex == null || rowsToDisplay[rowIndex] !== rowNode) {
    rowNode.clearRowTopAndRowIndex();
}
```

**2. Iterate the previously-positioned rows, not the tree.** `setRowTopAndRowIndex` records what it
positions into a buffer and the clear step walks that. O(all nodes) becomes O(previously-displayed), and
the recursion goes with its `changedPath` pruning, its per-node `hasChildren()` and its
`detailNode`/`sibling` probes. Two buffers swap, so nothing is reallocated per refresh.

**3. Stop allocating in the map path.** With formulas active, `doRowsToDisplay` called
`unfilteredRows.filter(...)` every refresh just to reproduce `unfilteredRows`. Soft-filtered rows are
the exception, so the scan now returns `unfilteredRows` itself until one turns up.

## Why it is safe

**The buffer holds exactly the rows that can carry a stale top.** `rowTop` is assigned only in
`RowNode.setRowTop` and `RowNode._destroy`; every other caller passing a non-null top belongs to another
row model (pinned, infinite, SSRM, viewport) or to `RowCtrl`. Within CSRM the sole caller is
`setRowTopAndRowIndex`. Nodes cleared elsewhere only make the buffer a superset.

**Detail rows and group footers are `rowsToDisplay` entries**, so they are recorded like any other row;
the old walk reached them via `detailNode`/`sibling`. Pinned clones hang off `pinnedSibling`, which
neither the old walk nor the buffer touches.

**Destroyed rows are skipped.** `clearRowTopAndRowIndex` leaves `oldRowTop` at the row's last top
(`setRowTop` assigns `oldRowTop = this.rowTop` before nulling) — the anchor `_destroy(fadeOut)` and
`RowCtrl.setAnimateFlags` use to slide the row out from where it was. A second call nulls that anchor.
The old walk never reached these nodes; the buffer holds them, hence the `!rowNode.destroyed` guard.

**The buffer is a copy, not a reference to `rowsToDisplay`.** `sortStage` reuses the previous
`childrenAfterSort` by reference and passes it to `postSortRows`, a user callback documented as mutating
it in place, and the community and formula map paths assign `childrenAfterSort` straight to
`rowsToDisplay`. A callback removing entries would silently leave stale tops. The spare is emptied on
handover so it doesn't pin the previous generation of rows, destroyed ones included.

Slot identity is also stricter than the id comparison it replaces: group ids are deterministic and
regenerated on rebuild, so a stale node could previously alias a displayed one and keep its top. Nodes
with `id == null`, unlookupable in an id-keyed set, were skipped entirely; they are now handled.

## Benchmarks

| Benchmark                                                    | base   | test    | Result      |
| ------------------------------------------------------------ | ------ | ------- | ----------- |
| map-stage › remap with groups collapsed                      | 726.2  | 207,601 | **285.87x** |
| map-stage › remap with groups expanded                       | 396.2  | 1133.8  | **2.86x**   |
| map-stage › remap 20000 displayed rows                       | 1846.7 | 4870.9  | **2.64x**   |
| pivot-transaction › transaction update                       | 3395.7 | 6361.9  | **1.87x**   |
| delta-sort-transaction › applyTransaction (deltaSort: true)  | 102.6  | 158.4   | **1.54x**   |
| grouping-pipelines › grouping from scratch 20k               | 69.27  | 94.41   | **1.36x**   |
| grouping-pipelines › sorting from scratch 15k (5 sorts)      | 31.59  | 40.65   | **1.29x**   |
| grouping-pipelines › update sorting rowData 15k (5 sorts)    | 64.96  | 79.49   | **1.22x**   |
| grouping-pipelines › update grouping rowData 20k             | 114.8  | 138.9   | **1.21x**   |
| calculated-live-preview › keystroke flush 10k, sorted        | 194.4  | 233.3   | **1.20x**   |
| flat-pipelines › immutable update with filter + sort 15k     | 72.24  | 86.42   | **1.20x**   |
| flat-pipelines › delta sort with 30% updates                 | 581.0  | 686.1   | **1.18x**   |
| flat-pipelines › full sort with 30% updates                  | 573.4  | 667.4   | **1.16x**   |
| flat-pipelines › filter + sort 15k                           | 26.76  | 30.82   | **1.15x**   |
| tree-data-parent-id › build from scratch 12k                 | 172.7  | 196.8   | **1.14x**   |
| pivot-aggregation › full refresh                             | 84.62  | 96.40   | **1.14x**   |
| pivot-aggregation › full refresh (pivotComparator)           | 83.24  | 94.24   | **1.13x**   |
| calculated-live-preview › keystroke flush 1k, chained calc   | 104.7  | 118.1   | **1.13x**   |
| grouping-pipelines › toggle text filter on/off 12k           | 91.55  | 103.0   | **1.13x**   |
| grouping-pipelines › immutable update with active filter 12k | 108.5  | 121.6   | **1.12x**   |
| flat-pipelines › immutable update with active filter 10k     | 133.6  | 148.2   | **1.11x**   |
| tree-data-path › build from scratch 21k                      | 14.33  | 15.65   | **1.09x**   |
| flat-pipelines › toggle text filter on/off 10k               | 126.3  | 137.8   | **1.09x**   |
| delta-sort-transaction › applyTransaction (deltaSort: false) | 14.86  | 16.18   | **1.09x**   |
| flat-pipelines › sort 8000 rows                              | 77.83  | 84.05   | **1.08x**   |
| tree-data-path › update rowData 21k                          | 11.54  | 12.34   | **1.07x**   |
| grouping-pipelines › 8 further agg/transaction benches       |        |         | 1.04–1.06x  |

### DIFF
--- a/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
+++ b/packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
@@ -52,6 +52,12 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
     /** Row nodes used for formula calculations when formula feature is active. */
     private formulaRows: RowNode[] = [];
 
+    /** Rows given a row top by the previous pass — the only rows that can hold a stale one. */
+    private positionedRows: RowNode[] = [];
+
+    /** Buffer the current pass fills, swapped into `positionedRows` once the stale tops are cleared. */
+    private positionedRowsNext: RowNode[] = [];
+
     /** The ordered list of row processing stages: group → filter → pivot → aggregate → filterAggregates → sort → flatten. */
     private stages: IRowNodeStage[] | null = null;
 
@@ -311,7 +317,7 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         return minIndex < stagesLen ? stages[minIndex].step : null;
     }
 
-    private setRowTopAndRowIndex(outputDisplayedRowsMapped?: Set<string>): void {
+    private setRowTopAndRowIndex(): void {
         const { beans, rowsToDisplay } = this;
         const defaultRowHeight = beans.environment.getDefaultRowHeight();
         let nextRowTop = 0;
@@ -320,13 +326,10 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
         // with these two layouts.
         const allowEstimate = _isDomLayout(this.gos, 'normal');
 
-        for (let i = 0, len = rowsToDisplay.length; i < len; ++i) {
+        const len = rowsToDisplay.length;
+        const positioned = this.positionedRowsNext;
+        for (let i = 0; i < len; ++i) {
             const rowNode = rowsToDisplay[i];
-
-            const id = rowNode.id;
-            if (id != null) {
-                outputDisplayedRowsMapped?.add(id);
-            }
 
             if (rowNode.rowHeight == null) {
                 const rowHeight = _getRowHeightForNode(beans, rowNode, allowEstimate, defaultRowHeight);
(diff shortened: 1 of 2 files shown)


StephenCooper (approved):
Great improvement!

PR #14680, opened 2026-07-30
AG-18023 Read series label colour from the placement style

alantreadway:
https://ag-grid.atlassian.net/browse/AG-18023

Fix [AG-18023](https://ag-grid.atlassian.net/browse/AG-18023)

### DIFF
--- a/packages/ag-grid-enterprise/src/charts/chartComp/menu/format/fontPanel.ts
+++ b/packages/ag-grid-enterprise/src/charts/chartComp/menu/format/fontPanel.ts
@@ -114,7 +114,9 @@ export class FontPanel extends Component {
         if (params.value == null) {
             const placement = this.chartOptions.getValue<string | undefined>(keyMapper('placement'));
             if (typeof placement === 'string') {
-                const styleKey = placement.startsWith('outside') ? 'outsideStyle' : 'insideStyle';
+                // Mirrors the charts-side predicate: everything that is not `inside-*` is an outside
+                // label, including bar series' `beside-*` placements.
+                const styleKey = placement.startsWith('inside') ? 'insideStyle' : 'outsideStyle';
                 params.value = this.chartOptions.getValue(keyMapper(`${styleKey}.color`));
             }
         }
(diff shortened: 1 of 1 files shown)


peterjrreynolds (approved):

packages/ag-grid-enterprise/src/charts/chartComp/menu/format/fontPanel.ts
```diff
@@ -105,6 +104,23 @@ export class FontPanel extends Component {
...
+        const { chartMenuParamsFactory, keyMapper } = this.params;
+        const params = chartMenuParamsFactory.getDefaultColorPickerParams(keyMapper('color'));
+        // Series labels have no `color` of their own - the effective colour is resolved onto
+        // `insideStyle` / `outsideStyle` according to the label placement. Read the style matching the
+        // current placement so the picker has a value to show. Writes still go to `color`, which the
+        // chart applies to both styles.
+        if (params.value == null) {
+            const placement = this.chartOptions.getValue<string | undefined>(keyMapper('placement'));
+            if (typeof placement === 'string') {
+                const styleKey = placement.startsWith('outside') ? 'outsideStyle' : 'insideStyle';
```
iMoses: there's also `beside-*` for bar series which is an outside label.
I'd have the condition reversed, checking for `startsWith('inside')`

alantreadway: Good catch, thanks - fixed in f62653501ff. Reversed to `startsWith('inside')`, which also matches the charts-side predicate in `barSeries.ts` (`labelPlacement?.startsWith('inside') ? label.insideStyle : label.outsideStyle`), so `beside-*` now resolves to `outsideStyle` in both places.

packages/ag-grid-enterprise/src/charts/chartComp/menu/format/fontPanel.ts
```diff
@@ -105,6 +104,23 @@ export class FontPanel extends Component {
...
 
+    private getColorPickerParams(): ColorPickerParams {
+        const { chartMenuParamsFactory, keyMapper } = this.params;
+        const params = chartMenuParamsFactory.getDefaultColorPickerParams(keyMapper('color'));
+        // Series labels have no `color` of their own - the effective colour is resolved onto
+        // `insideStyle` / `outsideStyle` according to the label placement. Read the style matching the
+        // current placement so the picker has a value to show. Writes still go to `color`, which the
+        // chart applies to both styles.
+        if (params.value == null) {
+            const placement = this.chartOptions.getValue<string | undefined>(keyMapper('placement'));
```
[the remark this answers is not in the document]
alantreadway: Known and deliberate - the scope here was kept to populating the initial value. The only in-panel placement control is the range-bar dropdown, so the stale case is limited to flipping that with the panel open, and it is a stale swatch rather than a wrong write (the picker still writes to `label.color`, which overrides both styles). Happy to add a refresh on the placement select if reviewers want it, but keeping it out of this fix.

iMoses (approved):
LGTM

PR #14687, opened 2026-07-30
[AG-9664] Fix(pivoting): supplied pivot result columns default to no pivot sort

sdwvit:
TC13: an unset `pivotSort` resolved to ascending for every pivot column, so application-supplied pivot result columns (`setPivotResultColumns`) were reordered by header name and their pill showed an ascending indicator instead of keeping the supplied order.

An unset `pivotSort` now resolves to `null` (no sort) while the pivot result columns are application-supplied, and to `'asc'` for the columns the grid generates. The pill cycle, the column state and the docs follow the same default.

Behavioural tests added/updated in `pivot-column-sort-ssrm.test.ts`.

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
@@ -91,9 +91,10 @@ the same way row group columns are sorted from their pills.
 
 `pivotSort` is independent of `sort`: it controls the order of a pivoted column's result columns only, and neither
 direction flows to or from the column's own sort. Pivot columns the grid generates are sorted ascending by default,
-while pivot result columns supplied via [SSRM Pivoting](./server-side-model-pivoting/) default to no sort so their
-supplied order is kept. Clicking a pill cycles through ascending, descending and no sort, where no sort keeps the
-order the columns were generated or supplied in.
+while pivot result columns supplied through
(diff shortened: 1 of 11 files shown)


documentation/ag-grid-docs/src/content/docs/pivoting-column-groups/index.mdoc
```diff
@@ -90,8 +90,10 @@ the same way row group columns are sorted from their pills.
 {% apiDocumentation source="column-properties/properties.json" section="pivoting" names=["pivotSort"] /%}
 
 `pivotSort` is independent of `sort`: it controls the order of a pivoted column's result columns only, and neither
-direction flows to or from the column's own sort. Pivot columns are sorted ascending by default. Clicking a pill
-cycles through ascending, descending and no sort, where no sort keeps the order the columns were generated in.
+direction flows to or from the column's own sort. Pivot columns the grid generates are sorted ascending by default,
+while pivot result columns supplied via [SSRM Pivoting](./server-side-model-pivoting/) default to no sort so their
+supplied order is kept. Clicking a pill cycles through ascending, descending and no sort, where no sort keeps the
```
StephenCooper: I think this just needs a small tweak to point to the exact api. Currently I would think that SSRM does not sort asc by default because you do provided the pivot results column in the getRows callback. 

Link to advance section instead.

SalvatorePreviti (approved):

PR #14694, opened 2026-08-02
[AG-9664] Fix(docs-e2e): correct SSRM pivot spec and skip broken pdf-export specs

sdwvit:
Nightly Documentation Tests have been red since 2026-07-28, in two clusters.

**`creating_pivot_result_columns` (10 failures).** #14645 changed supplied pivot result columns to default to no sort but left this spec asserting ascending. Reworked against the documented behaviour: supplied order and unsorted pill first, then asc → desc → supplied. The first test sorts ascending before its DOM assertions because the example shuffles the year groups and 2000 is otherwise outside the rendered header window.

**`pdf-export*` (64 failures).** These examples throw `(SystemJS) Cannot read properties of undefined (reading 'version')` on grid-staging, so the grid never renders. The config already ignored `**/pdf-export*/**`, but project-level `testIgnore` *replaces* the config-level value rather than merging with it, so chromium/firefox/webkit discarded it. Both lists now share one constant.

Targets `b36.1.0` due to the code freeze on `latest`.

### DIFF
--- a/documentation/ag-grid-docs/playwright.config.ts
+++ b/documentation/ag-grid-docs/playwright.config.ts
@@ -9,6 +9,15 @@ const baseURL = BASE_URL || PREV_URL || PROD_URL || 'https://localhost:4610';
 
 const LOCAL_HOSTNAMES = ['localhost', '127.0.0.1', '[::1]'];
 
+const IGNORED_TESTS = ['**/pdf-export*/**'];
+
+const BROWSER_IGNORED_TESTS = [
+    ...IGNORED_TESTS,
+    '**/async-test/provided/angular/app.component.spec.ts',
+    // page-verification.spec.ts runs in its own dedicated CI job.
+    '**/page-verification.spec.ts',
+];
+
 function isLocalRun(url: string): boolean {
     try {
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/playwright.config.ts
```diff
@@ -35,8 +44,12 @@ if (process.env.FRAMEWORK) {
...
     snapshotPathTemplate: '{testDir}/{testFileDir}/{testFileName}-snapshots/{arg}{ext}',
     testDir: './src/content/docs/',
-    /* PDF Export is not part of the public API yet (release delayed), so its examples cannot load the module. */
-    testIgnore: '**/pdf-export*/**',
+    /**
+     * PDF Export is not part of the public API yet (release delayed), so its examples cannot load the module.
+     * Project-level `testIgnore` replaces this value rather than merging with it, so every project below that
+     * declares its own `testIgnore` must also spread in IGNORED_TESTS.
+     */
+    testIgnore: IGNORED_TESTS,
```
StephenCooper: Maybe we should not have this at the top level. I think that might break the other page-verification job. Just adding this to the project configs below should be enough and safer.

StephenCooper (approved):

PR #14696, opened 2026-08-03
RTI-3421: keep the framework name in the docs page h1, version out

skewis6:
Restores the framework name to the docs page `<h1>`, which AG-17860 moved out into a sibling `div` when it restructured the header into a CSS grid. Search engines only index the heading, so the framework name had stopped counting towards it.

The framework name now sits inside the `<h1>`; `Version X.Y.Z` stays outside it, so it doesn't read as part of the heading. They still share the eyebrow line because the `<h1>` is `display: contents`, which places its children in the header grid directly. The page title is now a marked-up element (`data-page-title`) rather than a bare text node, so it can be given a grid area — the Algolia indexer reads that marker, falling back to the h1's text nodes.

Header layout is unchanged from AG-17860 at every breakpoint, including the controls sitting between the eyebrow and the title on narrow screens.

RTI-3421

### DIFF
--- a/documentation/ag-grid-docs/src/components/docs/components/Header.tsx
+++ b/documentation/ag-grid-docs/src/components/docs/components/Header.tsx
@@ -37,21 +37,23 @@ export const Header: FunctionComponent<Props> = ({
         <header className={styles.docsPageHeader}>
             <div id="top" className={styles.docsPageTitle}>
                 <div className={styles.pageTitleContainer}>
-                    {/* The framework name must stay inside the h1 to count towards it for SEO, and `title`
-                        must stay a direct text node — the Algolia indexer reads only the h1's text nodes. */}
+                    {/* The framework name must stay inside the h1 to count towards it for SEO. The version
+                        must stay outside it, or it would read as part of the heading too — it only sits
+                        beside the framework name because the h1 is `display: contents`, which lets the two
+                        be laid out together from different parents. `data-page-title` is how the Algolia
+                        indexer finds the page title within the heading. */}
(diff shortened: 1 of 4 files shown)


documentation/ag-grid-docs/src/components/docs/components/Header.tsx
```diff
@@ -37,14 +37,21 @@ export const Header: FunctionComponent<Props> = ({
...
+                        must stay a direct text node — the Algolia indexer reads only the h1's text nodes. */}
+                    <h1 className={styles.pageTitle}>
+                        {(!suppressFrameworkHeader || version) && (
+                            <span className={styles.titleMeta}>
+                                {!suppressFrameworkHeader && (
+                                    <span className={styles.headerFramework}>
+                                        {`${getFrameworkDisplayText(framework)} Data Grid`}
+                                    </span>
+                                )}
+                                {version && <span className={styles.version}>{`Version ${version}`}</span>}
```
taktran: Would we want the version indexed as well?

skewis6: @taktran i wasn't sure either... @markdurrant ?

Though moving it out would mean changing the page layout

markdurrant: Confirmed with JSB we only want to combine the `FrameworkDisplayText` & `{title}` not version. 

Ideally without changing the layout ? 🤞 Double ideally, without changing the layout on mobile.

taktran (approved):

PR #14697, opened 2026-08-03
[RTI-3416] Fix(menu): suppress focus ring on menu item activated with no document focus

sdwvit:
Fixes [RTI-3416](https://ag-grid.atlassian.net/browse/RTI-3416).

The first menu item is focused programmatically while the document has no focused element, which Chrome treats as keyboard-driven, so `:focus-visible` drew a focus ring on a mouse-opened menu. Activation in that state now suppresses the ring; keyboard navigation is unaffected.

### DIFF
--- a/community-modules/styles/src/internal/base/parts/_menu.scss
+++ b/community-modules/styles/src/internal/base/parts/_menu.scss
@@ -33,6 +33,10 @@
     }
     @include ag.keyboard-focus((ag-menu-option), 1px);
 
+    :where(.ag-menu-list-no-focus-ring) .ag-menu-option:focus-visible::after {
+        content: none;
+    }
+
     .ag-menu-option-active,
     .ag-compact-menu-option-active {
         background-color: var(--ag-row-hover-color);
--- a/community-modules/styles/src/internal/base/parts/_menu.scss
+++ b/community-modules/styles/src/internal/base/parts/_menu.scss
@@ -33,7 +33,7 @@
     }
     @include ag.keyboard-focus((ag-menu-option), 1px);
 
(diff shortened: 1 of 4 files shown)


packages/ag-grid-enterprise/src/agStack/agMenuList.ts
```diff
@@ -250,13 +250,13 @@ export class AgMenuList<
...
-        // unless focus has since moved to an item or away from the menu entirely.
+        // unless focus has since moved to an item or away from the menu entirely. Activating while
+        // nothing is focused would draw a focus ring on a mouse-opened menu.
         this.itemsReady?.then(() => {
             if (!this.isAlive() || this.activeMenuItem) {
                 return;
             }
-            const activeElement = _getActiveDomElement(this.beans);
-            if (_isNothingFocused(this.beans) || this.getGui().contains(activeElement)) {
+            if (this.getGui().contains(_getActiveDomElement(this.beans))) {
```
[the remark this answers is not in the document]
sdwvit: lies :)

community-modules/styles/src/internal/base/parts/_menu.scss
```diff
@@ -33,6 +33,10 @@
     }
     @include ag.keyboard-focus((ag-menu-option), 1px);
 
+    :where(.ag-menu-list-no-focus-ring) .ag-menu-option:focus-visible::after {
```
gportela85: SASS files don't get the `where` function

sdwvit: it's plain css, but yea, not needed here

gportela85 (approved):

PR #14700, opened 2026-08-03
Fix update-versions-data silently mislabelling every version

taktran:
`nx update-versions-data ag-grid-docs` is currently destructive. `/archive` was
removed and now redirects to `/documentation-archive/`, and the site rejects
non-browser clients with a 403 regardless of path.

Both failures were swallowed. `getWebsiteVersions()` returned an empty list, and
because `noDocs` is derived from "version absent from the archive", every entry
got `noDocs: true` - which filters it out of `VersionsSelector` and the archive
`MajorTable`. Running the target on `b36.1.0` today empties both.

- Read `/documentation-archive/` and match the absolute hrefs it actually serves
  (`href="https://www.ag-grid.com/archive/14.0.0"`); the old regex expected an
  Apache directory listing that no longer serves.
- Send a browser `User-Agent` so the request is not rejected.
- Throw on a non-OK response or an empty version list rather than writing a file
  derived from nothing.
- Exempt the newest entry from the `noDocs` check: the live release is served
  from the main docs site, so it is never in the archive and must not be flagged.
  Without this, `36.1.0` itself would be dropped from the selector.

The second commit is the resulting data refresh, adding the five released
versions missing from the list. Every flag was cross-checked against the live
archive: 36.0.2, 36.0.1 and 35.3.1 are archived (no `noDocs`); 32.3.9, 32.3.8 and
18.1.0 are not (flagged). Nothing lost a flag it should have kept, and `36.1.0`
is untouched.

Same fix as ag-charts and ag-studio, which have the identical script.

### DIFF
--- a/documentation/ag-grid-docs/scripts/versions-data/updateVersionsData.ts
+++ b/documentation/ag-grid-docs/scripts/versions-data/updateVersionsData.ts
@@ -4,7 +4,13 @@ import * as fs from 'fs/promises';
 
 // Absolute path from the root directory
 const VERSIONS_DATA_PATH = 'documentation/ag-grid-docs/src/content/versions/ag-grid-versions.json';
-const WEBSITE_ARCHIVE_URL = 'https://www.ag-grid.com/archive';
+const WEBSITE_ARCHIVE_URL = 'https://www.ag-grid.com/documentation-archive/';
+
+// The archive page is served behind a filter that rejects non-browser clients with
+// a 403, so identify as one. Without this every request fails and the script would
+// conclude that no version has archived docs.
+const BROWSER_USER_AGENT =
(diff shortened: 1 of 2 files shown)


taktran:
Closing in favour of #14705.

This fixed how the archive page was fetched but kept the underlying mistake: `/documentation-archive/` is generated from the very JSON file the script writes (`MajorTable` filters entries by `!entry.noDocs`), so deriving `noDocs` from it is circular and self-reinforcing - once a version is flagged it drops off the page and stays flagged forever. Concretely, the data commit here wrongly flagged 18.1.0, 32.3.8 and 32.3.9, all of which are archived.

#14705 asks the server whether `…/archive/<version>/` exists instead.

PR #14708, opened 2026-08-03
[RTI-3423] Fix(docs): enable grouping in columns tool panel examples

sdwvit:
Examples showing the columns tool panel rendered the row group / pivot / values drop zones but no column could be dragged into them, because `enableRowGroup`/`enablePivot`/`enableValue` were not set.

Adds all three to `defaultColDef` in the reported example and its close siblings, matching the existing `tool-panel-tabs` example on the same page.

Fixes [RTI-3423](https://ag-grid.atlassian.net/browse/RTI-3423)

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/theming-v32-customisation-tool-panels/_examples/column-tool-panel/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/theming-v32-customisation-tool-panels/_examples/column-tool-panel/main.ts
@@ -44,18 +44,27 @@ const gridOptions: GridOptions<IOlympicData> = {
         {
             headerName: 'Athlete',
             children: [
-                { field: 'athlete', minWidth: 170, rowGroup: true },
-                { field: 'age', rowGroup: true },
-                { field: 'country' },
+                { field: 'athlete', minWidth: 170, rowGroup: true, enableRowGroup: true, enablePivot: true },
+                { field: 'age', rowGroup: true, enableRowGroup: true, enablePivot: true },
+                { field: 'country', enableRowGroup: true, enablePivot: true },
             ],
         },
         {
             headerName: 'Event',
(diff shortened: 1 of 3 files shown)


documentation/ag-grid-docs/src/content/docs/theming-v32-customisation-tool-panels/_examples/column-tool-panel/main.ts
```diff
@@ -61,6 +61,9 @@ const gridOptions: GridOptions<IOlympicData> = {
     defaultColDef: {
         editable: true,
         filter: true,
+        enableRowGroup: true,
```
StephenCooper: I would push back against using the defaultColDef as this opens up other strange possible behaviours. 

I think it would be better to be more specific in our example updates. 

But also maybe there is something to be investigated with our cellDataTypes to make this easier out of the box.

StephenCooper:
Can we just update both examples on https://www.ag-grid.com/archive/36.1.0/javascript-data-grid/theming-tool-panels/#top so that they both provide sensible settings via the individual columns. 

Then revert the changes to the other examples (apart from the legacy CTP example)

documentation/ag-grid-docs/src/content/docs/theming-v32-customisation-compactness/_examples/dynamic-height/main.ts
```diff
@@ -41,16 +41,16 @@ function changeSize(value: string) {
 }
 
 const columnDefs: ColDef[] = [
-    { field: 'athlete', minWidth: 170 },
-    { field: 'age' },
-    { field: 'country' },
-    { field: 'year' },
+    { field: 'athlete', minWidth: 170, enableRowGroup: true },
```
sdwvit: see Stephens comment above

StephenCooper (approved):

PR #14709, opened 2026-08-03
RTI-3414-click-while-collapse-animation-loses-focus

SalvatorePreviti:
# fix(focus): clicking a row that is animating out no longer loses focus

FIX RTI-3414

## Problem

Collapse a group with `animateRows` on, then click a row while it animates. Focus is lost, and arrow
keys scroll the viewport instead of moving the focused cell.

Removed rows are destroyed in a first pass only: their `CellCtrl`s and DOM linger for 400ms to fade out,
but their `RowNode`s have already lost `rowTop` and `rowIndex`. A click in that window is handled by a
stale cell, so the grid records `{ rowIndex: null }` and the browser focuses a doomed element.

## Solution

`getControlsForEventTarget` retargets such an event to the row now owning that display slot, resolved
with `getRowIndexAtPixel(oldRowTop)`. A stale row is only reachable through its cells, since the row's
own dom data goes on destroy. `onMouseDown` then calls `preventDefault()` so the browser cannot focus
the dying element, and focuses the live cell instead. Clicks, context menus, cell selection and the
keyboard path follow, as they share that lookup.

That leaves the ctrl live while `mouseEvent.target` still points into the dying row, so three decisions
taken from `target` are skipped when retargeted: `containsWidget` and `lastMouseDownOnDragger`, which
would let a stale checkbox or drag handle suppress the live cell's `cellMouseDown` and its row's
`rowClicked`, and `mouseStayingInsideCell`, whose hover de-dup never matches across two elements.

## Notes

The whole mousedown runs against the retargeted cell, so a cell range starts there too.
`getRowIndexAtPixel` clamps, so clicking below the new end of the grid focuses the last row.

## Tests

`rows/interaction-during-row-animation.test.ts` collapses a group and interacts with a fading row,
covering the retarget and each of the three guards; all four fail without the fix. The React twin covers
the focus path, since `CellCtrl.onMouseEvent` is a twinned controller.

### DIFF
--- a/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
+++ b/packages/ag-grid-community/src/gridBodyComp/rowContainer/rowContainerEventsFeature.ts
@@ -20,6 +20,13 @@ import { _isStopPropagationForAgGrid } from '../../utils/gridEvent';
 import { _isUserSuppressingKeyboardEvent } from '../../utils/keyboardEvent';
 import { _selectAllCells } from '../../utils/selection';
 
+interface EventTargetCtrls {
+    cellCtrl: CellCtrl | null;
+    rowCtrl: RowCtrl | null;
+    /** True when the event came from a row animating out and was moved onto the row now owning its slot. */
+    retargeted?: boolean;
+}
+
 export class RowContainerEventsFeature extends BeanStub {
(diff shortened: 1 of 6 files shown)


StephenCooper:
Moving to Post Release for further considerations and testing.

PR #14724, opened 2026-08-04
Fix(tests): use ControlOrMeta in example specs so they pass on macOS

sdwvit:
Two example specs hard-code `Control`, which does not do the intended thing on macOS:

- `row-sorting/multi-column` — Ctrl+click is a context-menu gesture on macOS, so the secondary sort was never applied. `sortService.ts:45` accepts `ctrlKey || metaKey` for `multiSortKey: 'ctrl'`.
- `formula-editor-component` — select-all is Meta+A, so `Control+a` left the existing formula in place and the typed formula appended.

Both now use `ControlOrMeta`, matching existing usage in `row-selection-multi-row` and `deep-dive` specs. Verified green locally across all six frameworks.


sdwvit:
not sure if we want to merge this, we usually run these in ci, which is linux env

StephenCooper (approved):

PR #14730, opened 2026-08-05
Keep dev-validations comment in generated framework examples

StephenCooper:
The examples generator captures the dev-validations guard via `ts.Node.getText()` (`parser-utils.ts:254`), which skips leading trivia — so the `// Enable extended validations only for development` comment above the `if` statement was silently dropped from every generated React, Angular and Vue example.

Rather than change the capture, the comment is moved inside the guard block in all 1228 example `main.ts` files, where the verbatim source copy preserves it.

Also:
- The vanilla/UMD generator injects its own unguarded call; its comment wording is aligned with the authored one (was `// Development only — remove for production`).
- `removeModuleRegistration`'s strip regex now tolerates the comment sitting inside the block; otherwise the vanilla path would keep the `process.env` guard alongside the injected `agGrid.enableDevValidations()`.

Verified by regenerating all examples (`nx generate-examples ag-grid-docs`, 1004 jobs) and checking the output for reactFunctionalTs, angular, vue3, typescript and vanilla.


alistairjcbrown (approved):
Assuming the change is the files under `plugins/` and everything under `documentation/` is auto generated (spot checked, these looks fine) -- LGTM :shipit:

PR #14735, opened 2026-08-05
faster-checks-script

SalvatorePreviti:
# Add `./checks.sh` — pre-commit gate in one parallel Nx run

`AGENTS.md` documented type-check and lint as per-package commands: five sequential `yarn nx` calls, each re-paying Nx startup and graph computation, with 14 of 16 cores idle. `./checks.sh` runs the same work as a single `nx run-many`.

Measured on 16 cores:

|                                              | Cold  | Warm | CPU  |
| -------------------------------------------- | ----- | ---- | ---- |
| Five sequential `yarn nx` calls (3 packages) | 88.5s | ~10s | 171% |
| `./checks.sh` (default — every project)      | 33.8s | 1.8s | 676% |

The default runs `build:types`, `lint` and `build:test` against every project defining them — 18 projects and 7 dependency tasks. Nx skips projects without a target, so there is no list to maintain and the gate matches CI's `yarn nx lint`. The extra 15 projects cost ~5s cold, filling otherwise-idle cores.

Silent on success, dumps failing output otherwise. That would hide warnings too, so a passing run reports the count and keeps the log; `--warn` prints them with file paths:

```
CHECKS-PASSED (34s) — targets: build:types,lint,build:test | projects: all
  1714 warnings (run with --warn to print them): node_modules/.cache/ag-checks-warnings.log
```

Nx output goes to a file, never a pipe: node flushes pipes asynchronously, so nx exits mid-write and drops most of it — 639 of 2188 lines survived a pipe here. `--verbose` prints the captured log after the run rather than streaming it.

Flags: `--projects`, `--targets`, `--fresh`, `--warn`, `--verbose`.

## Take the public internet out of docs E2E

New `src/utils/grid/test/localSources.ts`, wired in as an auto fixture. A warm run makes no network requests at all: over a representative 138-test slice, a second run refetched 0 of 53 stored responses.

**Example assets come from disk.** Examples fetch data from `https://www.ag-grid.com/example-assets/…` and the same files ship in `public/example-assets`. `./docs-e2e.sh "filtering-with-sort"` (18 tests) goes from 23.4s with 2–4 failures to 5.1s all passing — the failures were React variants that mount last, so their fetch lost the race against the assertion timeout.

**Everything else off-origin is mirrored to disk.** Not an allowlist of known CDNs: that leaves every host nobody thought of on the public internet, which is why `flags.fmcdn.net` and `cdn.cookielaw.org` already had per-spec stubs. Only the build under test is exempt.

Entries are keyed by URL _and_ user agent, so a negotiating host cannot answer the wrong browser — Google Fonts serves `font-stretch: 100%` to one and `normal` to another, which is why it previously had to be excluded from mirroring altogether.

Freshness is the origin's call, evaluated between runs and never during one. The first request for a URL in a worker uses the stored copy if `cache-control` still vouches for it, otherwise revalidates conditionally so a 304 costs a round trip and no body; every later request in that run is served from the in-flight memo, so a run never re-checks what it has already answered. Of 53 entries in the slice above, 36 are `immutable` and only one (Google Fonts CSS) is shorter than a week. A stored copy is discarded only for a 2xx that replaces it — an error or an unreachable origin keeps serving it.

Bodies are stored decoded, because `route.fetch` decompresses and never keeps the encoded bytes, so `content-encoding` and `content-length` are dropped with them. Metadata and body share one file, written to a pid-suffixed name and published by a single `rename`: with a file each, two workers renaming at once can pair one's body with the other's status. The stored byte count is checked on read, so an entry damaged after publication is a miss rather than a corrupt bundle.

A 4xx is negative-cached for the run — the origin's settled answer, and re-asking costs a round trip per test — while a 5xx or a dropped socket is retried, and neither is written to disk. Anything the mirror could not answer is named once per run, with the example page that asked for it.

The sweep is disk housekeeping only, not invalidation: a read renews the entry, so anything still wanted never ages out and only superseded URLs are reclaimed, after 30 days. Orphaned temp files go after an hour.

`playwright-network-cache` is dropped. It was a second store of the same responses differing only in policy, and the TTL refresh it existed for is precisely the network access this removes. `nx clean` still removes its directory so stale copies go.

## Fix the column widths this branch broke

Serving example data from disk made it arrive before the webfont. Examples auto-size columns the first time data renders and those widths stick, so the grid measured text in the fallback font and kept the result — `column-sizing/column-resizing › fitCellToContents` sized `age` to 146 instead of 142 on every cold run.

Waiting inside the test cannot fix it: by the time a test can observe anything, fonts and stylesheets both report loaded and the widths are already stale. The example-assets route instead holds its response until the page can measure text.

Only `fetch` and `xhr` are held. An image or stylesheet delays `load` itself, so holding one while waiting on `load` deadlocks the navigation — `custom-icons/icons-css` paints its icons from `/example-assets/svg-icons/`, and holding those fails 2 of its 12 tests on `page.goto`.

The two routes are also registered in order rather than concurrently. Whichever won that race decided whether disk or the CDN served `/example-assets/`, so a cold run could store assets that ship in the repo, and a later run could ask the mirror for something it had never stored.


StephenCooper (approved):
Looking forward to trying this out!

PR #14768, opened 2026-08-07
Fixed legacy theme class being duplicated

gportela85:
Fix [AG-18110](https://ag-grid.atlassian.net/browse/AG-18110)

### DIFF
--- a/packages/ag-stack/src/core/baseEnvironment.ts
+++ b/packages/ag-stack/src/core/baseEnvironment.ts
@@ -97,13 +97,18 @@ export abstract class BaseEnvironment<
     }
 
     public getStyledRootClasses(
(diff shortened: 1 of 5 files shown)


BernieSumption (approved):
The implementation looks good. The test is React only. Is that correct? The related master-detail issue #14607 (AG-17960) was react-only because of Master/Detail init timing issues, but this seems like it should affect every framework?

gportela85:
> The implementation looks good. The test is React only. Is that correct? The related master-detail issue #14607 (AG-17960) was react-only because of Master/Detail init timing issues, but this seems like it should affect every framework?

We do have a test covering JS only as well. It verifies that the class is added to popups but not to the main wrapper.

PR #14769, opened 2026-08-07
AG-17862 Add markdown twins for the remaining 37 sitemap pages

taktran:
## What

Probing every URL in the production `sitemap-0.xml` for a `.md` sibling found **1472 of 1511 already covered** — all docs pages, across all four frameworks — and **38 non-docs pages with no twin**. This closes that gap, so an agent can append `.md` to any indexable URL (or send `Accept: text/markdown`).

**37 of the 38 got a twin. The 38th, `/privacy/your-choice/`, was removed from the sitemap instead** — see below.

## The registry refactor comes first

The set of twinned pages was hardcoded in **four** places that all had to agree, with nothing to catch an omission:

- `plugins/agDevMarkdownNegotiation.ts` — dev-server negotiation
- `htaccessRules.ts` — the Apache rewrite rule
- `htaccessRules.ts` — the Apache `Vary: Accept` scope
- `agentReadinessFiles.ts` — the prose in `/llms.txt` and `/AGENTS.md`

They now derive from a single `GRID_MARKDOWN_PAGE_GROUPS`. The `llms.txt` / `AGENTS.md` prose states the rule ("append `.md` to any page URL listed in the sitemap") instead of enumerating pages that would drift.

## The 37 twins

Every builder reads the same data its page renders. Where a page had no shared source — the demos, `/niall`, the policy intros, the contact and redirect-stub metadata — the copy is extracted first and the page repointed at it.

| Group | n | Shared source |
| --- | ---: | --- |
| Landing pages + `react-table` | 6 | one builder over the same `landingPages` JSON |
| Legal / policies | 3 | the same `.mdoc` bodies, via `renderMarkdocToMarkdown` |
| Conference sessions | 11 | `SESSIONS` data |
| Bryntum campaigns | 6 | campaign JSON + `decorateBryntumHtml` |
| Roadmap, What's New | 2 | `roadmap.json`, `versions` collection |
| Demos, Niall | 4 | prose extracted to JSON |
| Stubs & tools | 5 | metadata extracted to shared consts |

## A sitemap bug found on the way

`/privacy/your-choice/` (the opt-out confirmation page) is in `getSitemapIgnorePaths()`, so **robots.txt disallows it** — but the sitemap filter only excluded `/contact/failure/` and `/contact/success/`, the identical kind of post-submission page. The site was submitting a URL it simultaneously blocks, which Search Console reports as *"submitted URL blocked by robots.txt"*.

It is now excluded from the sitemap alongside the contact result pages, and has no `.md` twin — matching them. The page still renders from the shared `POLICY_CONTENT` entry, so its copy stays single-sourced.

## Ready for Charts / Studio

Builders those sites will also need — landing pages, policies, What's New, roadmap, contact, framework redirects — live in `external/ag-website-shared/` with injected URL resolvers. Only the grid page list stays grid-local.

One deliberate exception: `markdownPageRegistry.ts` exports the **type only**. The subrepo has no `"type": "module"`, so under plain node (the htaccess harness runs `htaccessRules.ts` through `tsx`) it loads as CJS and named exports vanish. The three-line derivation stays product-side.

## Verification

- `./checks.sh` — **PASSED** (`build:types`, `lint`, `build:test`, all projects)
- Unit — **710 docs + 188 shared** tests pass
- htaccess Apache harness — **6975 pass, 0 fail**
- Post-build coverage check — **1558/1561** sitemap URLs have a twin

The coverage check asserts every sitemap URL has a `.md` in `dist`, with an explicit three-page exclusion list for the `noindex` marketing pages production filters out (`/campaigns/power-of-ag-charts`, `/campaigns/return-to-support`, `/community/lets-cook`). It is guarded against passing vacuously — an earlier version did exactly that, because on a local build *every* page is noindexed — and was verified to fail when a twin is deleted.

## Two fidelity fixes found by reading the generated output

- `htmlInlineToMarkdown` silently dropped `<strong>` / `<b>` / `<em>` / `<code>`, so every twin lost inline emphasis. Now converted, with CommonMark-correct whitespace handling. This also improves the **existing** About, homepage and community twins.
- The campaign twin rendered `section.links`, which the page never renders and which duplicates the anchors already inside `body_html`.

## Known gap

The htaccess harness asserts redirects only (status + `Location`) and cannot send an `Accept` header, so the negotiation rules are verified as *generated text* — representative paths matched and rejected in unit tests — but never executed by Apache. That gap predates this change; extending the harness would need a new column in its TSV format.

### DIFF
--- a/documentation/ag-grid-docs/plugins/agDevMarkdownNegotiation.ts
+++ b/documentation/ag-grid-docs/plugins/agDevMarkdownNegotiation.ts
@@ -1,29 +1,19 @@
 import type { Plugin } from 'vite';
 
 import agDevMarkdownNegotiation from '../../../external/ag-website-shared/plugins/agDevMarkdownNegotiation';
-import { DISABLE_MARKDOWN_DOCS, FRAMEWORKS } from '../src/constants';
+import { DISABLE_MARKDOWN_DOCS } from '../src/constants';
+import { markdownPathPatterns } from '../src/utils/markdownPages';
 
-// A single docs page path, e.g. `/react-data-grid/cell-editing/`. The final
-// segment excludes dots so the `.md` variant itself never matches (no rewrite
-// loop) and the framework landing page (`/react-data-grid/`, which has no `.md`)
-// is left alone.
-const DOCS_PAGE_PATH = new RegExp(`/(?:${FRAMEWORKS.join('|')})-data-grid/[^/.]+/?$`);
-
-// Top-level (non-docs) pages that also ship a `.md` twin. Kept in sync with the same page
-// list in the SE-80 htaccess negotiation rule (see htaccessRules.ts).
-const TOP_LEVEL_MD_PATH = /^\/(?:license-pricing|changelog|pipeline|about|documentation-archive|example)\/?$/;
-
-// The /community landing page and its subpages, each with a `.md` twin.
-const COMMUNITY_MD_PATH = /^\/community(?:\/(?:events|showcase|tools-extensions|media|beyond-the-prompt))?\/?$/;
-
-// SE-80: content-negotiate grid docs pages to their markdown variant in the dev
-// server. Grid supplies its URL shapes; the shared factory holds the mechanism
+// SE-80: content-negotiate grid pages to their markdown variant in the dev server.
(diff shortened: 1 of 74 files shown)


skewis6 (approved):
LGTM

PR #14774, opened 2026-08-07
faster-stabler-e2e-tests-improvements

SalvatorePreviti:
# Docs E2E: stop two flakes, and fail a snapshot that asserted nothing

Follow-up to #14756, rebased on #14773. `./docs-e2e.sh "filter-"` failed 1–8 times per run, always in different specs; none were grid assertions. #14773 has since landed the route-handler half of this independently, so what remains here is the teardown race, the builder-pill flake, and two harness assertions that could pass without asserting.

**Teardown asserting against the destroyed grid.** Examples call `setGridOption('rowData', …)` in a fetch `.then`; if that resolves after teardown, warning #26 lands in `checkForErrorsAndTearDownExample`'s second assertion. New `test/pendingRequests.ts` tracks the page's outstanding `fetch`/`xhr` requests — a font or image says nothing about this race — and teardown settles them first (bounded at 1s, skipped entirely for a test that made none, and entries age out after 5s since an aborted request can go without a terminal event). `requestfinished` fires before the page applies the response, so the settle flushes a page macrotask and only calls the page quiet once that leaves nothing pending. Whatever still slips through is ignored only when a data request really was in flight, and only for `setGridOption()` — with none in flight the same warning is a leaked timer or listener, and still fails. Scoped to the destruction window; the first assertion is unchanged. It holds no reference to the page, because the suite keeps one tracker per describe for the whole worker run and capturing the page there would pin every closed page and its context graph.

**Builder pill picker waited for, not re-clicked.** `AdvancedFilterBuilderHarness.selectPill` clicked, waited a macrotask, then looked for rows — but the virtual list can re-render and detach the pill, swallowing the click. Now retries the click inside a `waitFor` until the list mounts, re-reading the row by its `aria-posinset` each attempt — the one identity that survives a re-render, since the virtual list only renders the rows in view. A time budget rather than a tick count, so it does not shrink under full-suite load. Alongside: `openValueEditor` uses `input.ag-input-field-input` (the old `ag-text-field-input` missed date/number pills), and `selectRichSelectRow`'s "not found" message lists the rows that _were_ present.

**An empty `GridRows` snapshot placeholder passed.** `runSnapshotCheck` wrote a notice to stderr and downgraded an undefined snapshot to `skip-snapshot`, so a `.check()` with nothing in it asserted nothing and went green. It now throws — `'skip-snapshot'` remains how a check is deliberately bypassed — and `--update-grid-rows` fills the placeholder in, inserting between the brackets when there is no template literal to replace. The updater only claims a no-argument call when the receiver is a `GridRows` / `GridColumns` / `FilterDom` construction, so unrelated calls such as `destroyedNodeChecker.check()` are never rewritten; those three names and the three method names are now derived from one registry instead of three hand-maintained lists.

**On top of #14773's `routeGuard`.** Two more Playwright wordings (`Target closed` without the long form, and `Test ended`), and `isAbandoned` exported so `pendingRequests` can reuse it for `page.evaluate` rejecting the same way. The WAF-bypass route in `test-utils.ts` is wrapped too — #14773 covered the three handlers in `localSources` / `localTranspile`, not this one.

**Harness tidy-ups.** `frameworkTest` and `eachFramework` held identical describe scaffolding, so each new per-test fixture had to be added in both; they now share one `describeWithTeardown`. Also drops a dead `if (errors.length > 0)` block after an `expect` that always throws first — it rebound `errors` away from the live array the console listener writes into.

**`checks.sh`** now reprints the failing task ids under `CHECKS-FAILED`, rather than leaving them somewhere up the streamed output.

## Verification

- `./docs-e2e.sh "server-side-model"` (330) and `"filter-"` (823) green post-merge; both re-run to confirm.
- `./behave.sh --project behavioural` — 7877 passed, 25 skipped.


StephenCooper (approved):
Will have to merge these changes into https://github.com/ag-grid/ag-grid/pull/14773 as it looks like this is ready to go.

StephenCooper (approved):

PR #14782, opened 2026-08-10
AG-15299: prepare line-height call sites for absolute px values

skewis6:
Groundwork for [AG-15299](https://ag-grid.atlassian.net/browse/AG-15299). **The token values are not changed yet** — this makes that change safe to do.

Two problems block a straight swap of the eight `--text-lh-*` tokens to px:

1. Only 19 of the 65 `line-height: var(--text-lh-*)` call sites in this repo pair a line-height token with its matching `--text-fs-*`. The rest reference a token belonging to a different step, so changing the token values would silently rewrite them.
2. `_base.scss` sets `html { line-height: var(--text-lh-base) }`. A unitless line-height re-multiplies against each descendant's font-size; an absolute px one does not, so every element inheriting from `html` would collapse onto one fixed length.

So this PR (a) points each mismatched site at the `--text-lh-*` belonging to its own `--text-fs-*`, keeping the scale as the single source of truth, (b) splits `--text-lh-ratio-base` out for the consumers that genuinely need a ratio — the `html` default, and form elements whose font-size comes from the user agent — and (c) renames the remaining size-agnostic ratios to match, so the naming itself says which tokens are safe to inherit.

### This changes rendering — deliberately

Honouring the pairing moves those sites onto the correct step's ratio. Computed font-size and line-height measured for every text element across 20 pages; **62 of 11,478 elements move**:

| where | font-size | line box | shift |
|---|---|---|---|
| campaign eyebrows, hero body, quote blockquote, docs intro paragraph (21 els) | 20px | 28px → 24px | −4px |
| three campaign headings (10 els) | 32px | 35.2px → 48px | +12.8px, largest element +25.6px |
| homepage and pricing quotes (5 els) | 16px | 24px → 22.4px | −1.6px |

The remainder is reflow around those changes plus known homepage animation jitter. Worth a design eye on the campaign headings in particular.

### Why the ratio split matters

With the px tokens applied on top of this branch, 419 of 9,637 elements change. Without the ratio split it is 6,957 — 40px headings drop from a 48px to a 24px line box, and the largest shift is 167px.

### Naming: `--text-lh-ratio-*` vs `--text-lh-<step>`

`--text-lh-tight` and `--text-lh-ultra-tight` are size-agnostic by design and can never become px: `--text-lh-ultra-tight` is applied at both `font-size: 46px` and `font-size: 30px`, and 17 of `--text-lh-tight`'s uses sit on rules with no `font-size` at all (`td`, `.tabs-nav-link`, a bare `p`), where the author cannot know the size to pin. They are renamed to `--text-lh-ratio-tight` / `--text-lh-ratio-ultra-tight`, which leaves one rule:

- `--text-lh-ratio-*` — unitless; use where the font-size is inherited, fluid or unknown.
- `--text-lh-<step>` — paired; valid only alongside its own `--text-fs-<step>`.

The eight scale tokens deliberately keep their names — they are the ones becoming px, so a `ratio` prefix on them would be wrong the moment the values change.

### Notes for review

- Four sites keep a literal value because no token applies: `BryntumCampaign .heroHeading` is `clamp(48px, 5.5vw, 72px)` so its line box must be a ratio; `LicenseSetup .licencePlaceholder` (15px), `ContactResultPage .heroHeading` (36px/48px) and `license-pricing .price` (46px) declare raw font sizes that are not on the scale.
- The docs intro `code` was `font-size: 0.8em` of a 20px parent, i.e. exactly `--text-fs-base` — stated explicitly so the pairing survives the token change.
- The old `--text-lh-tight` / `--text-lh-ultra-tight` names are kept as aliases of the new ones. ag-charts and ag-studio each have 5 files of their own referencing the old names, and would silently fall back to `line-height: normal` when this shared change lands there ahead of their migration. Remove the aliases once both repos are done.
- The token table in `ag-eng/skills/website-css/design-tokens.md` (ag-dev-prompts) also lists the old names.
- The rename is value-preserving by construction; verified by confirming every referenced `--text-lh-*` still resolves to a definition, rather than by re-measuring.
- Commit 1 pinned these sites to px to keep them pixel-identical; commit 3 supersedes that with the token pairing and commit 4 does the rename. Worth reading the commits in order.

### What ag-charts and ag-studio inherit on sync

16 of the changed files are in `external/ag-website-shared/`, so they land in both other products on the next subrepo sync. Eleven are pure renames with identical values; three change a computed line box:

| where | applies to | font-size | line box | shift |
|---|---|---|---|---|
| docs intro paragraph (`article > p:first-child`) | every docs page, both products | 20px | 28px → 24px | −4px |
| pricing page `.quote` (`SocialProof`) | both products | 16px | 24px → 22.4px | −1.6px |
| theme-builder param docs (`ItemDocs`) | ag-studio only | 12px | 14.4px → 20px | +5.6px (+39%) |
| `ContactResultPage .heroHeading` | both products | 36px / 48px | 43.2 → 43px, 57.6 → 58px | sub-pixel, de-fractioning |
| `license-pricing .price` | both products | 46px | 55.2px → 55px | −0.2px |

The docs intro is the widest-reaching, since it is on every documentation page in both products. The theme-builder one is the largest single jump and is studio-only.

Neither repo needs a code change to keep working — the deprecated aliases cover their old-name references. The follow-up work there is: migrate their 5 files each off the old names so the aliases can be deleted, and audit their own line-height call sites for step mismatch (ag-charts has 12, ag-studio 16) before the token values flip to px.

### Still to do on the ticket

Sign-off on the px token values, decisions on two phantom tokens (`--text-lh-relaxed`, `--text-lh-1xl`, both currently rendering as `line-height: normal`), and 13 residual rule patterns where a scale token sits on a re-sized element.

Separately found, not fixed here:

- `CodeHighlight.module.scss` sets `line-height: 0.1em` on code-block lines, computing to 1.5px — the likely cause of the code-block spacing bug reported on the ticket.
- `homepage.module.scss:346` is `list-style: var(--text-lh-ratio-tight)`, a line-height token assigned to `list-style`. Invalid, and dropped by the parser.

### DIFF
--- a/external/ag-website-shared/src/design-system/_root.scss
+++ b/external/ag-website-shared/src/design-system/_root.scss
@@ -61,6 +61,10 @@
     --text-semibold: 600;
     --text-bold: 700;
 
+    // Ratios, for a line-height inherited by a subtree: a length is inherited as-is, so descendants
+    // with a different font-size would keep the ancestor's line box. Pair --text-lh-<step> with its
+    // own --text-fs-<step> only.
+    --text-lh-ratio-base: 1.4;
     --text-lh-tight: 1.2;
     --text-lh-ultra-tight: 1;
 
--- a/external/ag-website-shared/src/design-system/_root.scss
+++ b/external/ag-website-shared/src/design-system/_root.scss
@@ -61,12 +61,15 @@
     --text-semibold: 600;
     --text-bold: 700;
 
-    // Ratios, for a line-height inherited by a subtree: a length is inherited as-is, so descendants
-    // with a different font-size would keep the ancestor's line box. Pair --text-lh-<step> with its
-    // own --text-fs-<step> only.
+    // Ratios: a length is inherited as-is, so use these wherever the font-size is inherited, fluid or
+    // unknown. --text-lh-<step> is paired, and valid only alongside its own --text-fs-<step>.
     --text-lh-ratio-base: 1.4;
-    --text-lh-tight: 1.2;
-    --text-lh-ultra-tight: 1;
+    --text-lh-ratio-tight: 1.2;
+    --text-lh-ratio-ultra-tight: 1;
+
+    // Deprecated: ag-charts and ag-studio still use these names. Remove once both have migrated.
+    --text-lh-tight: var(--text-lh-ratio-tight);
+    --text-lh-ultra-tight: var(--text-lh-ratio-ultra-tight);
 
     --text-fs-2xs: 10px;
     --text-lh-2xs: 1.2;
--- a/documentation/ag-grid-docs/public/styles/bryntum-demo.css
+++ b/documentation/ag-grid-docs/public/styles/bryntum-demo.css
@@ -21,7 +21,7 @@
     margin: 0 0 8px;
     font-size: var(--text-fs-lg);
     font-weight: var(--text-semibold);
-    line-height: 28px;
+    line-height: var(--text-lh-lg);
     color: var(--color-util-brand-600);
 }
 
--- a/documentation/ag-grid-docs/public/styles/bryntum-demo.css
+++ b/documentation/ag-grid-docs/public/styles/bryntum-demo.css
@@ -21,7 +21,7 @@
     margin: 0 0 8px;
     font-size: var(--text-fs-lg);
     font-weight: var(--text-semibold);
-    line-height: var(--text-lh-base);
+    line-height: 28px;
     color: var(--color-util-brand-600);
 }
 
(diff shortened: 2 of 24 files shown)


external/ag-website-shared/src/design-system/_root.scss
```diff
@@ -61,6 +61,10 @@
     --text-semibold: 600;
     --text-bold: 700;
 
+    // Ratios, for a line-height inherited by a subtree: a length is inherited as-is, so descendants
+    // with a different font-size would keep the ancestor's line box. Pair --text-lh-<step> with its
+    // own --text-fs-<step> only.
+    --text-lh-ratio-base: 1.4;
     --text-lh-tight: 1.2;
     --text-lh-ultra-tight: 1;
```
markdurrant: Seems like we'll want to make all of these `--text-lh-ratio-xxx` now?

markdurrant (approved):
All looks good, tested some pages on local. 

One comment on LH ratio vars.

external/ag-website-shared/src/design-system/_root.scss
skewis6: eventually yes, atm charts and studio still need these names. but once the same work has been done on them, we can update this. i'd ideally like to get these all in sync ASAP, so i'm going to wait til a nice sync point to merge this

PR #14800, opened 2026-08-12
AG-16738-filter-options-validation-and-typing

SalvatorePreviti:
# AG-16738: validate `filterOptions`, and type it per filter

FIX [AG-16738](https://ag-grid.atlassian.net/browse/AG-16738)

A `filterOptions` entry the filter cannot use is accepted in silence, and the declared type suggests options
the filter never evaluates.

| scope  | net lines |   lines changed | hunks | files changed |
| ------ | --------: | --------------: | ----: | ------------: |
| source |      +126 | 392 (+259 -133) |    84 |     17 edited |
| tests  |     +1330 |      1330 added |     1 |         1 new |
| docs   |        +0 |       4 (+2 -2) |     3 |      1 edited |

## Breaking bug fixes

No breaking changes. Two fixes are visible to code that depended on the defect.

**1. Every `type` and `filterOptionKey` field admits a Custom Filter Option's `displayKey`**, so a narrower
annotation stops compiling. This covers `ISimpleFilterModel['type']`, the four concrete models, and
`IFilterPlaceholderFunctionParams['filterOptionKey']`. The old type was wrong, not merely tight: the grid already
writes a custom `displayKey` into those fields, at five sites where the narrow type compiles only by casting. The
cost is that a mistyped built-in key is no longer a compile error on a model, because one field cannot be both open
to custom keys and closed to typos; the runtime reports it instead, under `76`. Annotate with the exported
`FilterOptionKey`, or drop the annotation. Runtime is unchanged.

**2. A `FilterOptionDef` carrying only the removed `test` alias is rejected**, with warning `72` naming the
missing property. Nothing calls `test`, so the option was offered and matched no rows. It is not on
`IFilterOptionDef`, so only JS callers see it disappear.

Error `73` is removed, unobservably: an option with no `displayKey` is now dropped before the lookup that raised
it. Warning `72` reads `is missing` rather than `doesn't contain one of`, which described the alias group it no
longer has. Warning `74` changes meaning, from "this column has no filter options" to "your list left none, so the
built-in ones are in use"; its text is unchanged.

## TC1: a malformed or unusable entry is reported

`OptionsFactory` gains a single build pass over the configured list, replacing a loop that mutated the list it
was iterating. Five defects fall out of it, two of them crashes:

| defect                                                                          | before                                                                                     |
| ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ |
| a rejected entry emptied the list                                               | `filterOptions.filter((v) => v === filterOption)` kept only the malformed entry            |
| `displayKey: 'toString'` crashed                                                | custom options were an object literal, so the lookup returned an `Object.prototype` member |
| a hole in the list crashed                                                      | `typeof null` is `'object'`, so `null` passed the string guard                             |
| `defaultOption` was returned unchecked                                          | the dropdown could sit on a value it does not list; now warning `326`                      |
| a built-in key and a `FilterOptionDef` sharing a `displayKey` were both offered | two dropdown rows for one key                                                              |

A list that keeps no option falls back to the built-in one under warning `74`, a column offering nothing being
unable to open its filter at all. `validateModel` checks conditions against the offered list rather than raw
`filterParams.filterOptions`, so a model naming a dropped entry is cleared instead of evaluated, and an O(n·m)
scan becomes a map lookup.

The Text Filter reports warning `76` for a key its matching cannot answer, as the scalar filters already do,
though the two answer it differently: the scalar filters admit the row they cannot judge, the Text Filter
excludes it. It fires at evaluation, blank values included, and only under the default matcher, so a
`textMatcher` supplying its own keys stays trusted, and never for a Custom Filter Option, whose predicate is
skipped for a missing value rather than a bad key. Both report once per key rather than once per row, the
logger formatting its message and
documentation URL before the once-per-message guard discards the duplicate: 593 ms against 32 ms on a
50 000-row pass.

## TC2: per-filter option keys

`ISimpleFilterModelType` splits into `TextFilterOptionKey`, `ScalarFilterOptionKey` and `DateFilterOptionKey`
over a shared `CommonFilterOptionKey`. Each filter's params and model then declare `filterOptions`, `defaultOption`
and `type` in terms of its own union, so a Date preset is no longer suggested on a Text Filter or a
`TextFilterModel`. `ISimpleFilterModelPresetType` is exported, `DateFilterOptionKey` being defined in terms of it.

Those four declarations are restated per filter rather than inherited from one generic
`ISimpleFilterParams<TOptionKey>`, which is what this started as. The generic reads better and is 40 lines shorter,
but it costs the reference documentation: `generate-doc-references` substitutes a parent's type parameter only when
the child is generic too, so `interface ITextFilterParams extends ISimpleFilterParams<TextFilterOptionKey>` renders
`filterOptions` as `(IFilterOptionDef | TOptionKey | CustomFilterOptionKey)[]` on the Text, Number, Date and BigInt
Filter pages, where today it reads `(IFilterOptionDef | ISimpleFilterModelType)[]`. The parameter also buys no
safety to offset that: because `CustomFilterOptionKey` widens both sides to `string`, a child restating
`filterOptions` with another filter's keys compiles clean. Restating is the same information a reader can see.

A custom `displayKey` can be any string, so every one of those unions has to admit one. Writing `| string` would
work and destroy the suggestions: TypeScript reduces `'equals' | 'contains' | string` to plain `string`, and the
editor then offers nothing. Intersecting with an empty object type stops that reduction — `string & Record<never,
never>` is still just `string` to assignability, but it is not *identical* to `string`, so the literals survive
beside it and stay in autocomplete. That is what `CustomFilterOptionKey` is.

`filterPlaceholder` resolves a custom option through its `displayKey`, as the dropdown label does, rather than
handing the callback an unresolved key; the three sites doing that share one `translateFilterOption`.

## Verification

33 behavioural tests in `testing/behavioural/src/filters/filter-options-validation.test.ts`, 24 of them
`FilterDom` snapshots, spanning `OptionsFactory`, `validateModel`, all four handlers and placeholder resolution
on both the popup and the floating filter. A new suite because no existing one owns `filterOptions` validation.

**22 fail with the source reverted and the tests left in place**; the rest are regression cover, including the
`refresh` path and combined AND/OR models. The ticket's own repro,
`filterOptions: ['equals', 'notEqual', 'wrong']`, fails there with `expected '' to contain 'warning #76'`.

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/iSimpleFilter.ts
+++ b/packages/ag-grid-community/src/filter/provided/iSimpleFilter.ts
@@ -23,9 +23,9 @@ export interface ISimpleFilter extends IProvidedFilter, IFloatingFilterParent {
 
 export interface IFilterPlaceholderFunctionParams {
     /**
-     * The filter option key
+     * The filter option key. A Custom Filter Option reports its `displayKey`.
      */
-    filterOptionKey: ISimpleFilterModelType;
+    filterOptionKey: FilterOptionKey;
     /**
      * The filter option name as localised text
      */
@@ -49,10 +49,11 @@ export type SimpleFilterParams<TData = any> = ISimpleFilterParams & IFilterParam
 export interface ISimpleFilterParams extends IProvidedFilterParams {
     /**
      * Array of filter options to present to the user.
+     * A key the filter cannot evaluate is reported when a value is tested against it under the built-in matching.
      */
-    filterOptions?: (IFilterOptionDef | ISimpleFilterModelType)[];
-    /** The default filter option to be selected. */
-    defaultOption?: string;
+    filterOptions?: (IFilterOptionDef | ISimpleFilterModelType | CustomFilterOptionKey)[];
+    /** The default filter option to be selected. Must be one of the offered options. */
+    defaultOption?: ISimpleFilterModelType | CustomFilterOptionKey;
     /**
      * By default, the two conditions are combined using `AND`.
      * You can change this default by setting this property.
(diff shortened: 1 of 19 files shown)


packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -369,8 +369,7 @@ export const AG_GRID_ERRORS = {
     70: ({ newFilter }: { newFilter: any }) =>
         `Grid option \`quickFilterText\` only supports string inputs, received: ${typeof newFilter}` as const,
     71: () => '`debounceMs` is ignored when apply button is present' as const,
-    72: ({ keys }: { keys: string[] }) => [`ignoring \`FilterOptionDef\` as it doesn't contain one of `, keys] as const,
-    73: () => `invalid \`FilterOptionDef\` supplied as it doesn't contain a \`displayKey\`` as const,
```
StephenCooper, line 373: I would like to know if removing an existing code breaks old versions of the grid. I know the version is part of the urlParams but the pages might be built off this list. 

Is there a precedence for this?

StephenCooper (approved):

packages/ag-grid-community/src/filter/provided/iSimpleFilter.ts
```diff
@@ -49,10 +49,11 @@ export type SimpleFilterParams<TData = any> = ISimpleFilterParams & IFilterParam
 export interface ISimpleFilterParams extends IProvidedFilterParams {
     /**
      * Array of filter options to present to the user.
+     * A key the filter cannot evaluate is reported when a value is tested against it under the built-in matching.
      */
-    filterOptions?: (IFilterOptionDef | ISimpleFilterModelType)[];
-    /** The default filter option to be selected. */
-    defaultOption?: string;
+    filterOptions?: (IFilterOptionDef | ISimpleFilterModelType | CustomFilterOptionKey)[];
```
peterjrreynolds, line 54: How can this be `CustomFilterOptionKey` here? Custom filter options are defined as `IFilterOptionDef`

PR #14816, opened 2026-08-13
Cover React variant tests in docs e2e CI and local runs

StephenCooper:
Splits the two jobs `FRAMEWORK` was doing — partitioning CI into per-framework shards, and choosing which React build the sweep runs on — so the sweep economy no longer drops specs that name a variant outright.

Collected tests, chromium:

| Pin | Before | After | |
| --- | --- | --- | --- |
| `reactFunctionalTs` | 1801 | **1805** | the four `test.reactFunctionalTs_Dev` tests, which no CI job selected before |
| local unpinned | 8948 | **8966** | tests naming `reactFunctionalTs`, including all six `reorder-grids` configurations |
| `angular` | 1784 | 1784 | unchanged, as for the other shards |

Nothing is dropped in either direction. The sweep on the React development build stays CI-excluded by design.

Context, not fixed here: https://ag-grid.atlassian.net/browse/AG-18154 — `reorder-grids` reaches the development build through that ticket's `version`/`prod` `loadPageOptions` rather than through the framework variant, so it always ran in CI and the gap there was local only.

### DIFF
--- a/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
+++ b/documentation/ag-grid-docs/src/utils/grid/test-utils.ts
@@ -63,27 +63,62 @@ const ALL_FRAMEWORKS = [
 ] as const;
 type AgFramework = (typeof ALL_FRAMEWORKS)[number];
 
+const PROD_VARIANT_FRAMEWORK = 'reactFunctionalTs';
+
+/**
+ * The framework this run is pinned to, or null for a run that covers them all. CI shards the suite by
+ * framework, one job per value, so a pinned run must declare nothing outside its own shard.
+ */
+function getPinnedFramework(): AgFramework | null {
+    const frameworkFilter = process.env.FRAMEWORK;
+    if (!frameworkFilter) {
+        return null;
+    }
+    const requestedFramework = frameworkFilter as AgFramework;
+    if (!ALL_FRAMEWORKS.includes(requestedFramework)) {
+        throw new Error(
+            `Invalid framework specified in FRAMEWORK environment variable: ${frameworkFilter}. Valid options are: ${ALL_FRAMEWORKS.join(', ')}`
+        );
+    }
+    return requestedFramework;
+}
+
+const PINNED_FRAMEWORK = getPinnedFramework();
+
 /**
  * React is the only framework declared twice - once on the production build, once forcing the development
- * one - so this only ever drops a React variant. That is a sixth of the suite, and locally the development
- * build is the one worth the wall clock, so CI runs both and a local run does not.
+ * one - so running both across every example doubles a sixth of the suite for little return. Only one of
+ * the two is swept: the development build locally, where its warnings are worth the wall clock, and the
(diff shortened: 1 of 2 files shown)


documentation/ag-grid-docs/src/utils/grid/test-utils.ts
```diff
@@ -63,43 +63,89 @@ const ALL_FRAMEWORKS = [
...
 
 // One worker, so a run says once what it is doing rather than once per worker.
-if (process.env.TEST_WORKER_INDEX === '0' && !process.env.FRAMEWORK) {
+if (process.env.TEST_WORKER_INDEX === '0' && !PINNED_FRAMEWORK) {
     // eslint-disable-next-line no-console
     console.log(
         runAllFrameworkVariants
-            ? `Running all framework variants: ${FILTERED_FRAMEWORKS.join(', ')}`
-            : `Skipping the ${PROD_VARIANT_FRAMEWORK} variant - set ALL_FRAMEWORK_VARIANTS=true to include it`
+            ? `Sweeping all framework variants: ${SWEPT_FRAMEWORKS.join(', ')}`
```
sdwvit: is it possible to use different wording? sweeping here is confusing imo

StephenCooper: Updated.

documentation/ag-grid-docs/src/utils/grid/test-utils.ts
```diff
@@ -63,43 +63,89 @@ const ALL_FRAMEWORKS = [
...
-const FILTERED_FRAMEWORKS = getFilteredFrameworks();
+const SWEPT_FRAMEWORKS = getSweptFrameworks();
+
+/**
+ * Whether a test naming this framework outright should be declared. Only the CI shard pinning applies -
+ * declaring a test just to skip it produces a fake "skipped" result and still spins up a browser fixture.
+ */
+function isFrameworkDeclared(agFramework: AgFramework): boolean {
+    if (PINNED_FRAMEWORK === null || agFramework === PINNED_FRAMEWORK) {
+        return true;
```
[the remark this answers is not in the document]
StephenCooper: Good catch on the mismatch, though I've taken it the other way: the documentation was wrong rather than the behaviour.

Gating `PINNED_FRAMEWORK_VARIANTS` on `process.env.CI` would make `--framework reactFunctionalTs` select a different set locally than the CI job of the same name. That divergence is the bug this PR exists to remove — `FRAMEWORK` was already pinned in every CI job, so the `runAllFrameworkVariants` CI check never applied there, and the four `test.reactFunctionalTs_Dev` tests were selected by no job at all. Reproducing a shard locally is worth more than avoiding four extra tests.

What was genuinely wrong is that `--framework`'s help text still implied a pin meant exactly one framework. Fixed in 8296ae86583.

On blast radius: the `_Dev` tests live in three specs (`cell-editors/component-editor`, `react-test/test-tear-down`, `react-test/updating-row-data-without-row-ids`), so a targeted run against any other example is unaffected. `--framework reactFunctionalTs_Dev` still selects only those four if you want them in isolation.

sdwvit (approved):

PR #14826, opened 2026-08-13
github-triage: go live on new issues, gate the labeled trigger, resolve channel from the ticket pin

sabahang:
## Why

Three changes to the GitHub-issue triage stub. One is the go-live switch; the other two are what makes it safe to flip.

**Blast radius of going live is small and read-only.** Triage creates an AITGH mapping ticket and hands back — it makes **no customer-facing write**. Nothing reaches the GitHub issue without a human dragging the ticket `Needs Review → In Progress` on the AITGH board. New-issue volume is also low (1–2/week measured over the last two weeks).

### 1. `issues:` uncommented — the go-live switch

Auto-triage of newly-filed `triage`-labelled issues. Note this is **not retroactive**: GitHub Actions triggers are event-driven, so the ~64 pre-existing open `triage` issues are untouched. Working through those stays a deliberate batch activity.

### 2. The `labeled` trigger is gated on the ADDED label

`labeled` fires on **any** label being added, and the previous condition only checked that the issue *has* `triage`. So stamping `AG-NNNNN`, `waiting-for-repro` or `invalid` on one of those 64 backlog issues would have fired a full triage on it — bypassing the batch-selection rule, and re-triaging issues that already carry a reviewed verdict, which the pipeline overwrites. Now the `labeled` half requires `github.event.label.name == 'triage'`; `opened` is unchanged.

### 3. `resolve-channel` — production and dry-run traffic coexist

The drag's payload comes from a **single** Jira Automation rule, so it cannot carry a per-ticket channel — one globally-configured channel for the whole board. During the pilot that rule sends `canary` unconditionally, so pointing it at `latest` for production would have removed the canary board drag that dry runs use.

Instead the pipeline stamps the channel it ran from onto the mapping ticket as a `channel-canary` label (ag-grid/ag-dev-prompts#815), and this job reads it back: present ⇒ `canary`, absent ⇒ `latest`.

- A **raw REST GET**, not the pipeline bundle — it has to decide *which* channel to install, so it cannot depend on an installed package. `labels` is a system field, so no field-id resolution is needed.
- **Fails SAFE to `latest`.** An unreadable ticket must not silently route a production drag to canary; a dry run losing its pin is the cheaper and more visible failure.
- The issue key goes in via `env:`, never string-interpolated into the script body.
- The job holds `permissions: {}` — it needs no GitHub token at all.

The guarantee is **release-channel continuity, not build immutability**: dist-tags move, so a later drag runs newer code from the same channel. That is deliberate — a version pin would freeze a ticket triaged before a fix onto the broken build.

## Migration — zero-gap, and the Jira rule is the switch

While the Jira rule still sends `"channel":"canary"`, that payload **wins** and behaviour is unchanged; the pin sits inert. Deleting the `channel` field from the rule body (AITGH → Project settings → Automation → *transitioned → In Progress* → Send web request) is what hands routing to the pin. So merging this changes nothing about which channel runs until that edit is made.

## Dependency

Needs `@ag-grid/dev-prompts >= 1.6.229` for the stamping half — already live on `canary` (`1.6.229-20260813120808.054f159f`), promoting to `latest` at 04:00 UTC. The action's `channel` input is **optional**, so this is safe to land before that promotion: until then nothing stamps, every ticket reads as unpinned, and drags resolve exactly as they do today.

## Verification

YAML validated (parses; `resolve-channel` correctly wired into both `preflight` and `run`, which use `needs.resolve-channel.outputs.channel || env.DEV_PROMPTS_CHANNEL`). `preflight` gained `!cancelled()` because `resolve-channel` is skipped on every non-dispatch event, and a skipped dependency would otherwise skip it regardless of its own condition.

Design reference: ag-dev-prompts → `docs/github-triage-pipeline.md`.

### DIFF
--- a/.github/workflows/github-triage-pipeline.yml
+++ b/.github/workflows/github-triage-pipeline.yml
@@ -43,11 +43,20 @@ name: GitHub issue AI triage
 # ============================================================================
 
 on:
-    # Auto-triage of new triage-labelled issues — DISABLED for the initial manual
-    # dogfood (run via workflow_dispatch first). Uncomment to go live once the
-    # dispatch + drag paths are validated.
-    # issues:
-    #     types: [opened, labeled]
+    # Auto-triage of new `triage`-labelled issues — LIVE (this is the go-live switch;
+    # it was commented out for the manual dogfood phase).
+    #
+    # `labeled` is needed as well as `opened` because the `triage` label is
+    # very often added a moment after the issue is filed, not at creation.
+    #
+    # ⚠️ The `labeled` half fires on ANY label being added, so the `run` job's gate
+    # additionally requires the ADDED label to be `triage` specifically. Without
+    # that, stamping any other label (`AG-NNNNN`, `waiting-for-repro`, `invalid`)
+    # on one of the ~64 pre-existing open `triage` issues would fire a full triage
+    # on it — bypassing the batch-selection rule, and re-triaging an issue that may
+    # already carry a reviewed verdict (which the pipeline would overwrite).
+    issues:
+        types: [opened, labeled]
     repository_dispatch:
         types: [gh-triage-resume, gh-triage-manual-resume, gh-triage-manual-browser-verify, gh-triage-manual-repro-rebuild]
     workflow_dispatch:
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #14839, opened 2026-08-14
Speed up integrated-chart behavioural tests

alantreadway:
AG Charts waits out a 500ms first-auto-size timeout on every layout-level update until it has seen a non-zero container size. jsdom reports `0` for `client*`, so `_lastAutoSize` was never set and every integrated-chart behavioural test paid that 500ms on each update.

`mockGridLayout` now reports a size (600x400, configurable) for `gridChartComp`'s chart canvas wrapper — the element AG Charts observes as its container — so the first synchronous size read succeeds and the wait never arms.

`format-panel-options.test.ts`: 136s -> 16.5s. Its timeout drops from 20 minutes to 2. The other chart suites also speed up. Full `./behave.sh` is green (10,408 tests), and the mock is gated on a class that appears nowhere outside integrated charts.

Reported by Salvatore Previti.

### DIFF
--- a/testing/behavioural/src/charts/format-panel-options.test.ts
+++ b/testing/behavioural/src/charts/format-panel-options.test.ts
@@ -455,7 +455,9 @@ describe('chart tool panel options', () => {
               ]
             `);
         },
-        20 * 60 * 1000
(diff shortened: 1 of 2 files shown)


SalvatorePreviti (commented):
we can close this

Issue #14837, opened 2026-08-13
[Bug]: Toolbar does not have an accessible name, a prop, or an API for setting an accessible name


cappscr:
### Link to reproducible scenario

https://codesandbox.io/p/devbox/musing-david-8mpqrq

### Steps to reproduce 
1. Open developer tools inspect the div with the ag-toolbar class and role="toolbar"

### Actual behaviour

No aria-label attribute. The interface for configuring the Toolbar is:

```typescript
/**
 * Configure the [Quick Access Toolbar](https://ag-grid.com/javascript-data-grid/toolbar/)
 */
export type Toolbar = {
    /** Default alignment for items in the toolbar. Defaults to `'left'`. Item-level `alignment` takes precedence. */
    alignment?: 'left' | 'right';
    /** Items to render in the toolbar. Each entry is either a shorthand string identifier or a full item definition object. */
    items: (ToolbarItemDef | ToolbarItemShorthand)[];
};
```

### Expected behaviour

Toolbar config should accept some parameter that sets the `aria-label`

### More information

I already have accessible names on all custom components that I put inside of the toolbar; however, would like to set an accessible name on the toolbar itself. This could be accomplished by using a ref and a useEffect hook; however, I would prefer to have a prop or an api for setting the accessible name built into the component itself.

### Version

35.3.1

### Does the issue occur for a specific framework only?

React

### Is the issue only observable on a specific browser?

All browsers

sabahang:
Hi

Thanks for getting in touch via GitHub.

We are currently not considering feature requests via GitHub and we do not believe there is an underlying issue with this feature from a UX perspective.

We appreciate your time for raising this and hope you can continue to use the current functionality available in AG Grid Community.

Kind regards,

Issue #14517, opened 2026-07-21
[Bug]: Destroyed footer RowNode keeps footer=true with severed sibling; isSelected() throws (crashes React grids using groupTotalRow)


mhsdef:
### Link to reproducible scenario

https://codesandbox.io/p/sandbox/wdsgw8 (rendered app: https://wdsgw8.csb.app)

### Describe the bug

Steps to reproduce
1. Open the sandbox. The grid shows group total rows via `groupTotalRow: "bottom"`.
2. Click the button. It captures a displayed footer `RowNode` (exactly as a mounted `RowCtrl`/cell component holds it), then removes total rows (`setGridOption("groupTotalRow", undefined)` — any path that reaches `_destroyRowNodeFooter` behaves the same, including SSRM store refreshes), then calls `isSelected()` on the captured node.

Actual behaviour

After the destroy, the captured node is left with `footer === true` but `sibling === undefined`, so `isSelected()` throws:

```
TypeError: Cannot read properties of undefined (reading 'isSelected')
```

`_destroyRowNodeFooter` (packages/ag-grid-enterprise/src/aggregation/footerUtils.ts) severs both `sibling` links but leaves the footer node's `footer` flag set, while `RowNode.isSelected()` (packages/ag-grid-community/src/entities/rowNode.ts) dereferences `this.sibling` unconditionally in its footer branch:

```ts
isSelected() {
  if (this.footer) {
    return this.sibling.isSelected();   // sibling is undefined after destroy
  }
  ...
}
```

Expected behaviour

A destroyed footer node should not be left in a state where `isSelected()` throws — e.g. clear `footer` in `_destroyRowNodeFooter`, or null-guard the footer branch of `isSelected()` (a footer is never selectable, so `false` is a correct answer).

More information

This is not just a stale-reference footgun — **AG Grid's own rendering hits it in production with React**. `ag-grid-react` mounts row/cell components asynchronously; when a server-side row model store refreshes while a React commit is in flight, the commit can still mount components for a footer node that was just destroyed. We see two crash paths in production error tracking (AG Grid 36.0.0, React 18), both ending at the same line:

1. Row mount: `RowCtrl.setComp → initialiseRowComp → getInitialRowClasses → rowNode.isSelected()` (checks for `ag-row-selected`).
[...]

PR #14854, opened 2026-08-14
github-triage: unblock auto-triage for community-reported issues

sabahang:
## Why

**Auto-triage has never worked for a genuine community report, and it looked healthy.**

`claude-code-action` refuses to run when the **triggering actor** lacks write access. On an `issues` event that actor is the **issue author** — a community reporter with `read` — so the run dies immediately:

```
##[error]Action failed with error: Actor does not have write permissions to the repository
```

Which means auto-triage only ever worked when a **maintainer** touched the issue. That is the opposite of the feature.

**Why it looked fine:** the first live ticket, AITGH-27 / #14517, was label-triggered by a maintainer, so the *maintainer* was the actor and it passed. The first real community report — AITGH-28 / #14837, author permission `read` — failed on **both** auto-fires (`opened` and `labeled`). The manual resume then succeeded, because the dispatcher was the actor.

## What

Passes `allowed_non_write_users` to the pipeline action, **scoped to the `issues` event alone**:

```yaml
allowed_non_write_users: ${{ github.event_name == 'issues' && '*' || '' }}
```

The board drag, the manual `•••` buttons and `workflow_dispatch` all already carry a write-capable actor, so the gate **stays armed** on those paths at no cost — and stays armed for any trigger added later. Setting the input also makes `claude-code-action` best-effort scrub secrets from subprocess environments.

## The alternative that was researched and rejected

Relaying the `issues` event through a `repository_dispatch` sent with the existing `JIRA_AGENT` GitHub App, so the actor becomes the app (which has write) and no bypass is needed. Rejected because:

- **Identical exposure** — the agent still reads attacker-authored issue text in the same job. It makes the *actor* trusted while the *trigger* is still a stranger's action.
- **Unverifiable dependency** — needs the app installed on `ag-grid` itself with dispatch rights, which cannot be confirmed without org admin.
- **The action deliberately resists it** — `getActorsToCheck` walks the *upstream* run's actor for `workflow_run` events specifically so chaining cannot launder an untrusted actor.

Scoping the flag per-event recovers the one real advantage the relay had: the gate stays armed everywhere else.

## What this does not change

The exposure is the same either way — the agent processes untrusted content in a job holding `issues: write` (which `execute` needs). Reducing *that* is the job split, tracked in ag-dev-prompts `docs/github-triage-deferred-hardening.md` § 1 and deliberately deferred for the pilot.

Note the browser chain jobs are already narrowed to `issues: read` (#14834), so this widens nothing there.

## Dependency

Needs `@ag-grid/dev-prompts >= 1.6.239` (ag-grid/ag-dev-prompts#862) for the passthrough. The action input is **optional and defaults to empty**, so this is safe to land before that promotes — until it does, the input is simply ignored and behaviour is unchanged.

### DIFF
--- a/.github/workflows/github-triage-pipeline.yml
+++ b/.github/workflows/github-triage-pipeline.yml
@@ -263,6 +263,23 @@ jobs:
                   # regardless of preflight's own behaviour.
                   stage: ${{ (github.event.action == 'gh-triage-manual-resume' && 'resume') || (github.event.action == 'gh-triage-manual-browser-verify' && 'browser-verify') || (github.event.action == 'gh-triage-manual-repro-rebuild' && 'repro-rebuild') || needs.preflight.outputs.stage || inputs.stage || 'triage' }}
                   product: grid
+                  # claude-code-action refuses to run when the TRIGGERING ACTOR lacks
+                  # write access — and on an `issues` event that actor is the ISSUE
+                  # AUTHOR, a community reporter with `read`. Auto-triage therefore
+                  # failed for every genuine community report ("Actor does not have
+                  # write permissions to the repository") and only ever worked when a
+                  # maintainer happened to touch the issue, which is the opposite of
+                  # the feature. It looked healthy because the first live ticket was
+                  # label-triggered by a maintainer; the first real community report
+                  # (AITGH-28 / #14837) failed on both auto-fires.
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #14859, opened 2026-08-14
fix-behavioural-tests-speed

SalvatorePreviti:
# Make the unit suite fast, and keep it that way: 183s to 68s

The suite's wall clock was mostly sleep: tests sat out product UX delays no grid option could reach, and the
largest suites were single files, which vitest cannot parallelise within. 664 files and 10,500 tests now run in
68 seconds on 16 cores at ~14x parallel; three fifths of the remaining worker-seconds are per-file load
(importing the grid, building a DOM), which is flat across files.

The worst single test spent 146 of its 146 seconds in AG Charts' auto-size timeout: it sizes from
`clientWidth`/`clientHeight`, which a mocked DOM pins at 0, so `Chart.checkFirstAutoSize()` waited out its 500ms on
every layout update and never laid out. `canvasPolyfill` now reports 600x400 for `.ag-chart-canvas-wrapper`.

## happy-dom replaces jsdom

Every project that wants a DOM runs on `happy-dom`, through the shared `unitProjectTestConfig`: the grid packages
and the behavioural suite. `community-modules/locale` takes the same helper with its `environment: 'node'` knob,
having no DOM to build; the two node-only tooling projects (`ag-grid-docs`, `ag-website-shared`) do not use the
helper at all, so the timeout and reporter settings below are the unit projects', not the whole workspace's.

|                                     |  jsdom | happy-dom |
| ----------------------------------- | -----: | --------: |
| import the module (once per worker) |  403ms |      96ms |
| construct a window (once per file)  | 1.84ms |    0.46ms |
| `getComputedStyle(el)`              | 27.9µs |     0.1µs |
| read one property off the result    | 0.03µs |    0.23µs |

The last two rows are the reason: jsdom resolves the whole cascade per call and returns a materialised declaration,
where happy-dom returns a live one and defers the work to each property read. A grid measures layout constantly and
reads a handful of properties per call, so it paid 28µs to save 0.2µs, and nothing in the suite processes real CSS.

`ag-test-utils/polyfills/` covers happy-dom's gaps: the `DOMTokenList` and `Option` constructors it omits from
`window`, `CompressionStream`, `elementsFromPoint`, and an `innerText` whose setter throws on the null the grid
legitimately assigns. jsdom stays only as an HTML parser for the Algolia index generator and the website sitemap,
and in `testing/angular-tests` (Jest). `./benches.sh --node` uses happy-dom too, detected by user agent so the
default Chromium run is untouched.

## Collapsing hard-coded UX delays: `FAST_TEST_TIMINGS`

`packages/ag-stack/src/fastTestTimings.ts` exports a single `false`; `testing/behavioural/vitest.config.ts` aliases
that module to a `true` copy for `mode: 'test'` only, so E2E, the docs site, the benchmarks and every published
bundle read the real `false`. Eight constants read it, each a delay no grid option can reach:

| constant                                         |  real | under the flag |
| ------------------------------------------------ | ----: | -------------: |
| tooltip show/hide floor (`Math.max(200, delay)`) | 200ms |              0 |
| menu item activation delay                       |  80ms |              0 |
| submenu open delay                               | 300ms |              0 |
| column-drag hold-at-edge tick (scroll, then pin) | 100ms |           20ms |
| date validity-report debounce                    | 500ms |              0 |
| export overlay minimum show time                 | 300ms |          150ms |
| ARIA announcement debounce                       | 200ms |              0 |
| ARIA re-announce gap after blanking              |  50ms |              0 |

`packages/ag-stack/src/fastTestTimings.test.ts` pins the shipped `false`; the behavioural `fastTestTimings.test.ts`
reads the flag back through `ag-stack` to assert the alias is live, since a specifier that stopped matching would
restore every delay silently. Only that one suite guards it, so the mode is stated once. A lifted floor buys nothing on its own, so the
tooltip suites ask for `tooltipShowDelay: 0`; the three tests asserting the delays keep real values, in
`tooltip/tooltip-delays.test.ts`.

## Splitting the suites that serialise

Six suites become sibling files sharing a harness module, each `describe` title suffixed by that file's slice:

| was                                                         | now                                               |
| ----------------------------------------------------------- | ------------------------------------------------- |
| `selection/row-selection.test.ts` (8.1k lines)              | 7 files + `rowSelectionHarness.ts`                |
| `selection/server-side-row-selection.test.ts` (5.0k lines)  | 8 files + `serverSideRowSelectionHarness.ts`      |
| `formulas/calculated-columns{,-state}.test.ts` (6.6k lines) | 15 files + 2 harnesses                            |
| `columnToolPanel/deferred-pivot-mode.test.ts` (2.3k lines)  | 4 files + `deferredPivotModeHarness.ts`           |
| `charts/format-panel-options.test.ts`                       | 5 files by chart family + `formatPanelOptions.ts` |
| `group-cell-renderer/.../exported.test.ts` (csrm + ssrm)    | 3 files each, via `withValues`                    |

The chart suite went from one test walking all 37 chart types under a 20-minute timeout of its own to a `test.each`
case per type. The two `exported` directories held each other's matrix (`csrm/` ran `rowModelType=serverSide`,
`ssrm/` the client-side pair); each now runs the row model its directory names, which is most of the snapshot churn.
The row-selection suites stop blanket-mocking `console.error` and `console.warn` across 222 tests, which dropped
everything the grid reported unread; the server-side ones are legitimately warned that `getRowId` is wanted
(warning 188), so the shared hooks allow that one and fail on any other. Ten leaf tests are renamed: six in the selection suites and one calc-column case, each because the name
described something the test did not do, plus three whose titles said `jsdom` for what is really the absence
of a layout engine.

## `ag-test-utils`

The 67 shared harness files move out of `testing/behavioural/src/test-utils/` into a private `ag-test-utils`
package exporting a barrel and per-module subpaths, with a `project.json` and eslint config so the gate lints it
(Nx skips a project with no `lint` target) and the ban on guessed `asyncSetTimeout` delays reaches it. It carries
its own `tsconfig.json`, which the eslint config names: left to infer one, typescript-eslint finds both this repo
root and `external/ag-shared` and declines to parse the file. The setup
file and the benchmarks import by subpath, so neither drags the barrel's node-only helpers into a browser worker.

`ignoreKnownNoise.ts` moves in with them rather than staying under `testing/shared/vitest/`, which the barrel was
otherwise reaching across a package boundary to re-export. Its `ignoreConsoleLicenseKeyError` is called by the
setup file, by `TestGridsManager` and by the two benchmark entry points, which the benchmark runner reaches
without the suite hooks; the 22 further calls dotted through the React suites were re-installing a filter that
was already in place. `clickMenuOption` joins `openMenuOption` in the package, replacing a copy in each of the
two calculated-column harnesses.

Harness checks that could not fail now can. `isInNestedGrid` compared against the grid's outermost owned element,
which sits above its own `.ag-root-wrapper`, so everything looked nested and the duplicate-row-id check never ran;
it now excludes only another grid's rows, and a row-span anchor's second element is not counted as a duplicate. DOM
order skipped whenever `ensureDomOrder` was set, which `TestGridsManager` sets on every grid. Row-span markers
exclude another grid's spans, so a span inside a detail grid no longer marks the master rows above it, and a span
anchor's second element sorts after the row itself, which is the element carrying the row's classes and aria.
Four validators stopped rejecting valid grids: sort indices count the hidden sorted columns that own them, the
`pivot_` prefix no longer outranks a colId the user declared, a registered string cell renderer may reshape its
cell text as a function one may, and a virtualised set-filter list is read from `aria-setsize` rather than from
the rows that happen to be mounted, whose test now asserts the indeterminate `(Select All)` its title names against
a real snapshot. The leaf-set
comparison had both sides built from the same field, and `assertColumnsSelected` accepted a leftover full-column
range; `assertSelectedCellRanges` accepted any leftover range at all, which was hiding that `addCellRange` appends a
second range under `suppressMultiRanges` and refuses only from the third (warning 93): the test that claimed the
opposite now asserts that contract. The leaf walk reports a cycle rather than overflowing the stack, and the
auto-group column lookup stays behind the ColumnApi guard its constructor computes two lines above. `waitForInput`
merges its 2s default rather than dropping it whenever a caller passes any option, and returns the input-or-textarea
it can actually match. `mockGridLayout` reads a pinned row's prefixed `row-index` (`t-0`), which `parseInt` turned
into a NaN rect for the row and every cell inside it, asserted by a new pinned-rows case; and the scrollbar-probe
test reads `_getScrollbarWidth()` rather than measuring a hand-built copy of the probe, which is what made a drifted
signature show up only as two unrelated fit tests. Answering that probe conclusively is also what puts the suite on
overlay-scrollbar geometry (0 wide, `invisibleScrollbar` true, so the fake scrollbars measure 16), where before it
read 0/0, cached nothing and re-measured on every call. `mockGridLayout`'s axis-mirrored patches share one installer
per axis, and the two lists of virtual-list viewports needing a forced height are one list. The
row-drag harness detaches its listeners after settling rather than before, caps that wait so a lost `rowDragEnd`
fails on an assertion rather than the test timeout, and requires the `rowDragEnter` whose absence it accepted;
synthetic drag events reach the drop container once rather than twice. Snapshot update mode rewrites a `check()`
only when its receiver is one of the snapshot classes, and leaves a template whose backticks share a line with
content. A row-numbers selection test asserts both of the ranges its CTRL-click produces, having named the second
in its title only.

## Vitest 4

`vitest` moves from 2.1.9 to 4.1.10 as a single root dependency, with `@vitest/coverage-v8`, `@vitest/ui` and
`@vitest/browser-playwright`, which is what v4 split the browser provider into. `vite` stays at `~5.4.19`: Vitest 4
carries its own Vite 8 nested, which is what yarn's peer warning reports.

| removed in v3/v4                          | now                                                                 |
| ----------------------------------------- | ------------------------------------------------------------------- |
| `defineWorkspace`                         | `test.projects` in the root config, fed from `vitest.workspace.ts`  |
| `basic` reporter                          | `['default', { summary: false }]`                                   |
| `onFinished(files)` reporter hook         | `onTestRunEnd(testModules)` over `TestModule`/`TestCase`            |
| `poolOptions.forks.{singleFork,execArgv}` | top-level `maxWorkers` and `execArgv`                               |
| `browser.provider: 'playwright'` + `name` | `provider: playwright()` and `instances: [{ browser: 'chromium' }]` |
| `esbuild: { target, jsx }`                | nothing: Vite 8 transforms with oxc, whose defaults already match   |

`pool: 'threads'` is explicit per project, since v4 no longer cascades project-scoped options from the root config;
benchmark mode keeps `forks`, because worker threads reject the `--expose-gc` the bench harness needs. Vite 8's
`configLoader: 'native'` warning is silenced by `VITE_CONFIG_NATIVE_IGNORE_WARNING`, since clearing it needs `.mts`
files `packages/*/tsconfig.spec.json` cannot type-check under `module: commonjs`.

## Test infrastructure

`testing/shared/vitest/` holds the shared helpers, thresholds, setup file, diff config and reporter that previously
sat loose in the repo root. `slow-tests.reporter.ts` prints the slowest tests, the slowest files with a per-test rate
(so volume is distinguishable from a slow test), and the files that spent most of their time off-CPU, from
`performance.eventLoopUtilization()`: usually a timer a fix would give back, though RPC and snapshot writes land
there too, which the table says. `testTimeout` for the unit projects is 12s locally and 60s in CI, and neither a
per-test nor a `describe`-level override escapes it: `output.setup.ts` rejects an out-of-range one before the body
runs, then checks measured duration against a `performance.now` bound taken at load, so `vi.useFakeTimers()` cannot
fake past it. The five overrides are gone. The cap is per test, so hook and collection time sit outside it.

The four gate scripts are six-line shims onto `scripts/gate/`, which carries the run log, the argument parsing, the dispatch
and the help for the flags they share once rather than once per gate. Every run streams stdout and stderr to
`tmp/_<name>-output/<id>/output.log`, its path printed first, colours kept on the console and stripped in the file,
with `result.json`, the command and a status beside it; `--async-status`, `--wait` and `--kill` act on a run by id,
exit 3 meaning still running. The status file is replaced by rename like the `latest` symlink beside it, so a
`--wait` in another terminal cannot read one half-written and call a passing run failed, and `CI` is read for
presence in one place, so the gate and the run it wraps cannot disagree about which timeouts are in play. Nx exits 0
when `--projects`/`--targets` select nothing, so `./checks.sh` reports that as a failure rather than a gate that
passed without running, and every runner a gate invokes is `node_modules/.bin/<runner>` rather than `npx`, the
sharded unit-test job included.

`external/ag-shared/.claude-settings.template.json` gains `Read(.env.local)` and a recursive `Read(credentials/**)` deny
rules. `.claude/settings.json` is rendered from that template, so a rule held only in the generated file does not
survive a render.

Two packages publish `exports` pointing at a `dist/` no test run builds. The unit projects alias
`community-modules/` to source as well as `packages/`, so a test reaches the translations through
`@ag-grid-community/locale` rather than by relative path; and the six Excel export tests import `xlsx` as a
namespace, its ESM build declaring no default export.

## LOC

| scope  | net lines |         lines changed | hunks |             files changed |
| ------ | --------: | --------------------: | ----: | ------------------------: |
| source |       -12 |         146 (+67 -79) |    49 |     27 (+1 -2, 24 edited) |
| tests  |     -4274 | 64218 (+29972 -34246) |  1837 | 692 (+75 -11, 606 edited) |
| docs   |      -417 |        557 (+70 -487) |    23 |                 10 edited |
| other  |      +878 |    3320 (+2099 -1221) |   155 |     23 (+7 -2, 14 edited) |

No test was deleted. The negative test figure is the selection suites' repeated setup snapshots: 25 files carried
622 byte-identical copies of a snapshot taken before the case had done anything. One copy per distinct fixture
remains, and every case still asserts its own end state, all but four of them through a full-grid snapshot. Five cases are new:
`enableClickSelection: 'enableDeselection'` under multi-row server-side selection, three grouped ranges
crossing a group boundary, a country boundary and a collapsed group, and one asserting that server-side
selection warns for a missing `getRowId`, which is what keeps the harness's allowance of warning 188 honest.


gportela85 (approved):

.rulesync/rules/ag-grid.md
```diff
@@ -27,6 +27,10 @@ This file provides guidance to AI Agents when working with code in this reposito
 
 Each script takes `--help`. Full flag reference lives in the guides below rather than here: test and E2E flags in the Testing Guide, benchmark and profiling flags in the Benchmarks Guide.
 
+**Never run `./checks.sh`, `./behave.sh`, `./benches.sh` or `./docs-e2e.sh` in the foreground.** They take minutes, and a foreground call blocks the whole session: the user cannot reach the agent, no other work proceeds, and the wait is dead time repeated at every gate. Start them with the agent harness's background mechanism, which delivers a completion event in a later turn, then keep working. **Never `sleep` to wait for one either** — that is the same block, wearing a different hat.
```
StephenCooper: This might fight with the instructions for the AI workflow which had problems with background tasks. Might need to make sure this is safe and get @alantreadway input.

SalvatorePreviti: I see! Locally it works, maybe we should put an "if running locally and not in a workflow" if that is well defined. But is making a lot of difference - agent stopped for 3, 4 minutes when running tests and stops listening to user requests while waiting, unbearable

SalvatorePreviti: rephrased it

StephenCooper (approved):

PR #14876, opened 2026-08-17
Send an identifying User-Agent on build-time fetches to the live sites

taktran:
## Problem

`nx run ag-grid-docs:preview:production` (and `nx build ag-grid-docs -c production`) fails during the build while generating `robots.txt`: the requests for the charts/studio disallow lists go out with Node's default agent and are not served.

Production builds are the only ones that take this path — `src/pages/robots.txt.ts` only fetches those lists when `getIsProduction()` is true, and staging builds fall into `disallowAll`.

## Changes

- **`BUILD_USER_AGENT`** added to `@ag-website-shared/constants` — it lives in the shared subrepo because both affected fetches do the same thing from different packages.
- **`fetchRobotsDisallow`** (`robotsTxt.ts`) sends it — this is the reported failure.
- **`getSitemapXml`** (`ag-website-shared`) sends it too, and now **throws on a non-ok response**. It fetches `LIVE_SITEMAP_URL` (`https://www.ag-grid.com/sitemap-0.xml` in production) on a cold cache with no `ok` check, so an error response body was being used as the sitemap and written into the `/sitemap` page. `-c production` passes `--clean-cache=true --run-second-build=true`, so the first build always hit this and only the second build's cache hid it. It failed silently where `robots.txt` failed loudly.

## Testing

- `robotsTxt.test.ts` — extended with a `fetchRobotsDisallow` suite: the header is sent, disallow lists from multiple URLs are flattened, and a failed request rejects rather than silently dropping that site's disallow paths.
- `getSitemapXml.test.ts` — new suite covering the cache-miss path: the header is sent, and a failed response throws instead of returning the error page as the sitemap.
- Both suites confirmed red against the unpatched source before the fix.
- Verified against the real endpoints: `/charts/robots-disallow.json`, `/studio/robots-disallow.json` and `/sitemap-0.xml` all fetch successfully with the change and fail without it.
- `./checks.sh` passes.

## Notes

- `yarn nx dev` takes the same fetch path (`getIsDev()` is also true in `robots.txt.ts`), so this fixes `/robots.txt` locally as well — it just failed per-request there instead of failing the build.
- Pointing the URLs at staging is not an option: `charts-staging`/`studio-staging` return 404 for `robots-disallow.json`; only the localhost charts/studio dev servers serve them (the commented-out lines in `.env.dev`).
- Failing soft on the fetch was rejected — it would risk silently shipping a production `robots.txt` missing the charts/studio disallow entries.
- `getSitemapXml.ts` and `constants.ts` are in `external/ag-website-shared`, shared with the charts and studio repos.

### DIFF
--- a/documentation/ag-grid-docs/src/utils/robotsTxt.test.ts
+++ b/documentation/ag-grid-docs/src/utils/robotsTxt.test.ts
@@ -1,4 +1,7 @@
-import { AI_CRAWLERS, productionRobotsTxt } from './robotsTxt';
+import { BUILD_USER_AGENT } from '@ag-website-shared/constants';
+import { vi } from 'vitest';
+
+import { AI_CRAWLERS, fetchRobotsDisallow, productionRobotsTxt } from './robotsTxt';
 
 // Representative Allow/Disallow paths, mirroring what the route feeds in production.
 const ALLOW_PATHS = ['/campaigns/bryntum-gantt/'];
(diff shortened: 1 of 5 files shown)


markdurrant (approved):
LGTM 👍

Issue #889, opened 2016-05-06
typescript react cellrenderer


Fl0r14n:
cannot use reactCellRendererFactory with ColDef since 
cellRenderer?: {
        new (): ICellRenderer;
    } | ICellRendererFunc | string

should also be | Function or change ag-grid-react typings

ceolter:
i have put typings into the ag-grid-react project and release 4.1.1.

if still not working, just cast it to ICellRendererFunc

PR #14885, opened 2026-08-18
feat(github-triage-pipeline): auto-fire a confidence-refresh resume off the evidence chain

sabahang:
## Summary

Adds `confidence-refresh-chain`: after `browser-verify-chain`/`repro-rebuild-chain`
finish, if the original triage's confirmed clean-pick confidence was below the
execute threshold, auto-fires `stage=resume` (`trigger: confidence-refresh`) so
the pick gets re-assessed with whatever evidence the chain gathered — without a
human having to drag first purely to make the pipeline re-read evidence it
already has. No new execution path: `resume` still cannot execute on its own
regardless of what triggered it. Same permission posture as the two existing
browser-stage chain jobs (`issues: read`, never write).

Companion action-side change: `ag-dev-prompts` PR
[#889](https://github.com/ag-grid/ag-dev-prompts/pull/889) (the `confidence_gap`
output + `trigger` input this job consumes). Until that PR is merged and
published, this job simply never fires (the output doesn't exist yet on the
installed channel) — safe no-op ordering either way.

Full design + reflection (including why this is explicitly NOT a guaranteed
confidence bump — the AITGH-37 case that motivated it actually went the other
way) is in the action's own `CLAUDE.md § "Confidence-gap auto-refresh"`.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/github-triage-pipeline.yml'))"` — parses.
- [x] `actionlint .github/workflows/github-triage-pipeline.yml` — zero findings.
- [ ] First live fire on a real below-threshold clean-pick, once ag-dev-prompts#889 is live on whichever channel is in use.

### DIFF
--- a/.github/workflows/github-triage-pipeline.yml
+++ b/.github/workflows/github-triage-pipeline.yml
@@ -260,6 +260,9 @@ jobs:
             # CLAUDE.md § "Auto-chaining browser-verify / repro-rebuild".
             issue_key: ${{ steps.pipeline.outputs.issue_key }}
             confirmed_bug: ${{ steps.pipeline.outputs.confirmed_bug }}
+            # NEW (2026-08-18) — confidence-refresh-chain's own gate. See the
+            # action's CLAUDE.md § "Confidence-gap auto-refresh".
+            confidence_gap: ${{ steps.pipeline.outputs.confidence_gap }}
         steps:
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #14907, opened 2026-08-19
github-triage: self-dispatch the auto-trigger so the triage run carries a bot actor

sabahang:
The `issues` event no longer runs the triage agent directly. It now runs a small
`redispatch` job that re-fires this workflow as a `workflow_dispatch`, so the run that
does the work carries a bot actor which the shared composite action allow-lists.

Auto-triage did not work for community-reported issues on the previous arrangement.

Also removes the now-unreachable `issues` disjunct from the two chain-job gates.

Requires the companion `ag-dev-prompts` change to be on `latest` before this merges.

https://ag-grid.atlassian.net/browse/AG-17369


alantreadway (approved):
LGTM

PR #14909, opened 2026-08-19
AG-18241 number-formatter-filter-inputs

SalvatorePreviti:
FIX AG-18241

A Number Filter column configuring `numberFormatter` showed nothing in its filter inputs while filtering rows, and
its operands were the one place the Advanced Filter ignored a column's own parser and formatter. Reaching either
needs `numberFormatter`, `numberParser` or `bigintFormatter`, so a default grid is unaffected.

## Filter inputs

`NumberFilter` used a text input only for `allowedCharPattern`, so a column with just a `numberFormatter` got an
`input type="number"`, which drops any value outside its own number grammar. The element type now follows either
parameter, and the new `filterInputType` names it outright. An `allowedCharPattern` applies to whichever input
results, narrowing what a `number` input already accepts. `BigIntFilter` never applied `bigintFormatter` to its
inputs, and now does.

Both share the new `TextInputSimpleFilter`, which owns the input pair. An input holds the value the filter wrote
while the text is still the text it wrote, so formatter output is never read back through `numberParser` nor
re-parsed for range validation, and one rendered as empty stands for no value.

A `colDef` refresh changing any of those parameters shows every input again, replacing the elements whose type they
decide and restoring the caret. Text the filter did not write is carried across as typed, so a refresh mid-edit
neither moves the caret nor rounds what is being typed. Replacements are re-validated before the panel reads their
validity, which is what decides whether a condition is complete.

`recreateFloatingFilterInputService` set the recovered value on the destroyed service rather than its replacement,
and now reads the text as typed through the new `FloatingFilterInputService.getInputText`.

`NumberFloatingFilter.convertValue` read its input with `Number`, and now uses the column's `numberParser`. A column
naming none reads with `parseFloat` as its inputs do, and an unreadable operand clears the filter rather than
setting a `NaN` model.

`SimpleFilterModelFormatter` took its value formatter as a constructor argument, so a read-only summary kept the one
it was built with. It is now the overridable `getValueFormatter()`.

`AgInputTextField` wrapped `allowedCharPattern` in a character class unconditionally, so a pattern already written as
one compiled to a two-character match and rejected every keystroke. An already-bracketed pattern is now used as it is.

## Advanced Filter number operands

A number column pairing `numberParser` with `numberFormatter` decides its operands, as `bigintParser` already does
for BigInt: read with the parser in the expression, the builder and the model, and displayed through the formatter.
`Number` stays the default for a column naming only one of them, since an operand the grid cannot write it must not
read. A format is only displayed where the parser reads it back unchanged, so one that rounds leaves the operand
plain rather than rewriting the model on the next read — `bigintFormatter` included, which could already round one
away.

A numeric operand may now be quoted, which is what lets a format hold a space or a bracket. Either quote ends an
operand, so a format holding both kinds is written as the plain number instead. The accepted grammar is a superset
of the previous one, quoting a number having been a validation error, so nothing that parsed before parses
differently.

Two places assumed a number operand is its own input form. `isOperandModelValueEditable` now excludes such a column,
as it already excluded `bigint`, and the Filter Builder edits any formatted operand in a text input rather than the
`AgInputNumberField` that discarded an edit typed as `1,234,567`. The operand pill also follows a change between two
number columns.

An operand the parser cannot read is reported as not a number: a custom `numberParser` returns `null` where `Number`
returns `NaN`, and only the latter was treated as invalid.

## Test runner

The gate scripts gained `--log-tail <n>` and `--log-grep <pattern>`, so trimming a run's output no longer needs a
pipe — which reports the last command's status and hands back a red run as green. They print the selected lines in
place of the quiet digest, work on `--wait` and `--async-status` for a run started elsewhere, and are refused where
there is no log to read (CI, `--no-log`, `--async`, a watch mode) rather than passing the runner's status off as
output nobody got.

## Size

| scope  | net lines |    lines changed | hunks |      files changed |
| ------ | --------: | ---------------: | ----: | -----------------: |
| source |      +214 | 1262 (+738 -524) |   105 | 23 (+1, 22 edited) |
| tests  |     +1369 | 1467 (+1418 -49) |    38 |   8 (+2, 6 edited) |
| docs   |       +20 |      32 (+26 -6) |     7 |           4 edited |
| other  |       +78 |   148 (+113 -35) |    14 |           3 edited |


scripts/gate/run-log.mjs
```diff
@@ -270,7 +271,45 @@ export class RunLog {
...
+                matches = (line) => re.test(line);
+            } catch {
+                // An unusable pattern is a substring the caller typed, not a reason to lose the output.
+                matches = (line) => line.includes(grep);
+            }
+            lines = lines.filter(matches);
+        }
+        const total = lines.length;
+        if (tail != null && total > tail) {
+            lines = lines.slice(-tail);
```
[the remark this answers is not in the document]
SalvatorePreviti: fixed

testing/behavioural/src/filters/filter-behaviour/bigint-filter-custom-parser.test.ts
```diff
@@ -137,37 +165,158 @@ describe('BigInt Filter — custom bigintParser', () => {
...
+                // An apply button holds the typed value in the input, which is what the rebuild has to carry.
+                filterParams: { buttons: ['apply' as const], allowedCharPattern },
+            },
+        ];
+        const api: GridApi = await gridsManager.createGridAndWait('grid4', {
+            columnDefs: columnDefs('\\d'),
+            rowData: [{ val: 255n }, { val: 16n }],
+        });
+
+        await FloatingFilterHarness.get(api, 'val').setValue('16');
```
StephenCooper, line 316: Does the behaviour here change the following item? 

https://ag-grid.atlassian.net/browse/AG-17381

peterjrreynolds (approved):

testing/behavioural/src/filters/filter-behaviour/bigint-filter-custom-parser.test.ts
SalvatorePreviti: No - just added two additional tests to prove it

StephenCooper (approved):

PR #14910, opened 2026-08-19
AG-15983 - AgInput - Allow auto complete for individual features

gportela85:
fix [AG-15983](https://ag-grid.atlassian.net/browse/AG-15983)

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/floating-filters/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/floating-filters/index.mdoc
@@ -70,21 +70,13 @@ The screen shots below show example scenarios where the provided Number floating
 
 ### Controlling Autocomplete on Floating Filters
 
-Browser autocomplete is disabled for grid input fields by default. Set `enableInputAutoComplete` to `true` to enable it for eligible inputs. Inputs that provide grid-owned suggestions, such as Rich Select and Advanced Filter inputs, keep browser autocomplete disabled so the browser UI cannot overlap their suggestion lists.
+Browser autocomplete on grid inputs is controlled globally by `enableInputAutoComplete` and per input with
+`browserAutoComplete` parameters, see [Input Fields](./input-fields/) for the full behaviour and the accepted values.
 
-Individual inputs can override the global behaviour with a `browserAutoComplete` parameter, available on cell editor
-params, provided filter params (`filterParams`, which floating filter inputs inherit), the Set Filter Mini Filter,
-tool panel params, and the Find / Quick Filter toolbar items (via `toolbarItemParams`). The `Text` and `Number`
-floating filters additionally accept `browserAutoComplete` in `floatingFilterComponentParams` (as defined in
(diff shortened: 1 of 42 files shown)


packages/ag-grid-community/src/interfaces/iAutoComplete.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Params for components whose input fields support overriding the browser's
+ * autocomplete/autofill behaviour.
+ * @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time.
```
StephenCooper, line 4: Should this be internal if it is part of the public API surface?

packages/ag-grid-community/src/main-internal.ts
```diff
@@ -312,6 +312,7 @@ export type { IShowValuesAsService } from './interfaces/iShowValuesAsService';
 export type { ISideBar, ISideBarService } from './interfaces/iSideBar';
 export type { SortOption } from './interfaces/iSortOption';
 export { _resolveSortOptions } from './sort/sortOptionUtils';
```
StephenCooper, line 314: As mentioned above I think this should be public

documentation/ag-grid-docs/src/content/docs/floating-filters/index.mdoc
```diff
@@ -72,9 +72,11 @@ The screen shots below show example scenarios where the provided Number floating
 
 Browser autocomplete is disabled for grid input fields by default. Set `enableInputAutoComplete` to `true` to enable it for eligible inputs. Inputs that provide grid-owned suggestions, such as Rich Select and Advanced Filter inputs, keep browser autocomplete disabled so the browser UI cannot overlap their suggestion lists.
 
-The `Text` and `Number` floating filters also support overriding this behaviour for an individual input by passing the
-`browserAutoComplete` parameter in `floatingFilterComponentParams` (as defined in `ITextFloatingFilterParams` and
-`INumberFloatingFilterParams`).
+Individual inputs can override the global behaviour with a `browserAutoComplete` parameter, available on cell editor
```
StephenCooper: It seems like this section of docs has lots more use cases that are not relevant to the floating filter page. 

Might be good to have this documented somewhere that can then be linked to from all these use cases.

packages/ag-grid-community/src/interfaces/iAutoComplete.ts
gportela85: I think this bit should remain private, customers will import that interfaces extending this.

packages/ag-grid-community/src/main-internal.ts
gportela85: same as above

documentation/ag-grid-docs/src/content/docs/floating-filters/index.mdoc
gportela85: done

StephenCooper (approved):

PR #14915, opened 2026-08-20
AG-18194 Inject the Enzuzo policy loader after the inline-script strip

taktran:
## What

Post-deploy page verification reported a new enforced CSP violation against staging — `script-src-elem` blocked `inline` on `/cookies/`, attributed to `app.enzuzo.com/scripts/cookies/<uuid>`. This removes the race that causes it.

Nothing was failing: the CSP report job is non-gating. But the violation is real, and it is intermittent — it appeared in 1 of the 8 post-deploy runs that day, which is why it read as new rather than as a known issue.

## Cause

The vendor policy HTML ships three inline `<script>`s (`window.ezphost = …`, `window.__enzuzo.policy ||= …`, and one calling `eval()` to define a fallback `reboot`). The loader inserts the policy with `Range.createContextualFragment`, which — unlike `innerHTML` — yields **executable** scripts, so they run on insertion.

`enzuzo-policy-tidy.js` patched that method to strip them. That only works if the patch is installed before the loader's policy fetch resolves, and as two sibling `<script src>` tags in `cookies.astro` that ordering was a race:

- **Full page load** — the strip had to win a network round trip against the vendor fetch. Measured against the real bytes, a ~5ms handicap on our script is enough to lose it.
- **Astro client-side navigation** — no ordering to win at all. `ClientRouter` re-inserts every `data-astro-rerun` script via `createElement` + `replaceWith` in one synchronous loop (`astro/dist/transitions/router.js` `runScripts`), which makes both scripts async and drops document order entirely. Reordering the tags would not have fixed this path.

When the strip loses, the three scripts reach the DOM and the CSP correctly refuses them.

## Fix

Remove the ordering dependency instead of trying to win it. The loader is no longer a tag in `cookies.astro` — the embed script injects it once the patch is installed and the `MutationObserver` is watching, so "patch first" is structural rather than a matter of timing, on both load paths and independently of Astro's script-rerun semantics.

`enzuzo-policy-tidy.js` → `enzuzo-policy-embed.js`, since injecting the loader is now its main job.

**No CSP change.** The vendor scripts are still unauthorised; they simply never reach the DOM to need authorisation. Adding hashes for them was the alternative and is worse — it would pin vendor bytes *and* authorise scripts we deliberately don't want to run.

## Verification

Real vendor loader and policy bytes served from localhost under the enforced policy shape, varying only the timing:

| | before | after |
| --- | --- | --- |
| hard load, our script late | 3 inline scripts blocked | 0 |
| hard load, policy late | — | 0 |
| Astro rerun, our script late | 3 inline scripts blocked | 0 |
| Astro rerun, policy late | — | 0 |

In every "after" case the policy still renders in the content wrapper, the loader still lands there, and the vendor marketing section is still stripped.

Also added `enzuzoPolicyEmbed.test.ts` (5 tests, jsdom). The one that matters asserts the patch is active at the moment the loader element is created, so the loader cannot drift back to being a sibling tag unnoticed. `./checks.sh` passes; `./behave.sh --project ag-grid-docs` is clean apart from a pre-existing `getErrorText.test.ts` failure (`Failed to resolve entry for package "ag-stack"`) that reproduces on an unmodified tree in a fresh worktree.

## Notes for review

- **`external/ag-website-shared` is touched**, so ag-charts and ag-studio pick this up on their next subrepo pull. The shared component now points at `/scripts/enzuzo-policy-embed.js`, so **each site needs the renamed script in its own `public/scripts/`** — pulling the component without it leaves `/cookies/` with no policy at all. Worth flagging because the page-verification suite filters 404s out of its annotations, so a missing file would not surface there.
- The loader is now injected rather than parsed, so the preload scanner no longer discovers it and the policy fetch starts marginally later. On a page whose entire content comes from that fetch this seemed the right trade for correctness, but a `<link rel="preload">` could claw it back if we care.
- The pre-existing site-wide `script-src` `eval` violation from the *cookiebar* bundle is untouched and unrelated. For what it's worth, it traces to Enzuzo's own `preferences.manager_poweredby` template going through `new Function`, so it can't be configured away our end — worth a separate ticket to decide whether to suppress it in the report or raise it with the vendor.


markdurrant (approved):
LGTM 👍

PR #14934, opened 2026-08-21
Merge b36.1.0 into latest

seanlandsman:


external/ag-website-shared/src/content/policies/cookies-data-07-08-26.json
```diff
@@ -0,0 +1,330 @@
+{
```
taktran: It feels like this should be deleted along with the `CookiesTable.astro`

taktran (changes requested):

external/ag-website-shared/src/content/policies/cookies-data-07-08-26.json
seanlandsman: fixed

taktran (approved):

PR #14945, opened 2026-08-24
AG-12723 - [Overlay] - Added Skeleton Rows to CSRM

gportela85:
fix [AG-12723](https://ag-grid.atlassian.net/browse/AG-12723)


packages/ag-grid-community/src/focusService.ts
```diff
@@ -697,6 +697,12 @@ export class FocusService extends BeanStub implements NamedBean {
             }
 
             if (column.isSuppressNavigable(rowNode)) {
+                if (_isClientSideLoadingRow(this.gos, rowNode)) {
```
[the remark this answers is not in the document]
gportela85: This is not an issue. `column.isSuppressNavigable()` delegates to `CellNavigationService.isSuppressNavigable()`, which checks for a client side loading row before invoking the user callback.

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
```diff
@@ -1060,13 +1061,14 @@ export class AgGridAngular<TData = any, TColDef extends ColDef<TData> = ColDef<a
...
-    /** Show or hide the loading overlay.
+    /** Show or hide the loading UI.
      * - `true`: the loading overlay is shown.
      * - `false`: the loading overlay is hidden.
+     * - `LoadingOptions`: configure the loading UI.
      * - `undefined`: the grid will automatically show the loading overlay until `rowData` and `columnDefs` are provided. (Client Side Row Model only)
      * @default undefined
      */
-    @Input({ transform: booleanAttribute }) public loading: boolean | undefined = undefined;
+    @Input() public loading: boolean | LoadingOptions | undefined = undefined;
```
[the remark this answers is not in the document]
gportela85: ???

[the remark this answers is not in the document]
gportela85: This is done automatically, not something that we change in the code.

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -199,15 +212,62 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
         }
+        while (loadingRows.length > rowCount) {
+            loadingRows.pop()!._destroy(null);
+            changed = true;
+        }
+        return changed;
+    }
+
+    private getRowsToRender(): RowNode[] {
+        return this.loadingRowsActive ? this.loadingRows : this.rowsToDisplay;
```
SalvatorePreviti: perf - since we know when we update loadingRows and rowsToDisplay and loadingRowsActive, I'd create a new field and set it in the right moments instead of using a method.

This method adds an indirection (method call), 3 fields access and one condition - that has a cost. Adding a new field would have the same exact cost as before instead of constantly checking if loading is active.  It is called in getRowBounds that is a hot method, and other hot parts.

I think that keeping rowsToDisplay or adding a new field that indicates the same concept, something that can be loading rows or real rows, would also make the code simpler

rowsToRender could be a name - so replacing the method with just a field

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -49,6 +57,11 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
     /** The rows mapped to rows to display, during the 'map' stage. */
     private rowsToDisplay: RowNode[] = [];
 
+    /** Display-only placeholders used while client-side data is loading. */
+    private readonly loadingRows: RowNode[] = [];
```
SalvatorePreviti: maybe to indicate that loading rows is not active we can use this field to be null? Boolean is fine though as long as we don't access two fields in the hot paths

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -430,13 +492,16 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
...
     public getRowCount(): number {
-        return this.rowsToDisplay.length;
+        return this.getRowsToRender().length;
     }
 
     /**
      * Returns the number of rows with level === 1
      */
     public getTopLevelRowCount(): number {
+        if (this.loadingRowsActive) {
```
SalvatorePreviti: if we have a final field "rowsOrLoadingRowsToDisplay" this code returns fast and easier

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
SalvatorePreviti: Have also you checked all the other parts in the code where we access rowsToDisplay? Because this could be a problem in other parts. We do read rowsToDisplay in many places

packages/ag-grid-angular/projects/ag-grid-angular/src/lib/ag-grid-angular.component.ts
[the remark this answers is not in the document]
SalvatorePreviti: This seems a potential breaking change though - we need to check this as user might have to change their code then. Usually they assign it to a field but setting loading = false means disable loading screen I think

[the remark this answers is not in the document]
gportela85: Setting `[loading]="false"` works fine with Angular, I have tested it.

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
gportela85: Done 
- Added a cached rowsToRender field and removed getRowsToRender().
- Synchronised it when loading mode changes and when real displayed rows are rebuilt.
- Looked into all rowsToDisplay accesses. Remaining uses intentionally require real processed rows for grouping, formulas, footers, and top-level indexing.

packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
gportela85: `loadingRows` is now `RowNode[] | null` where `null` means inactive and removed `loadingRowsActive`.

SalvatorePreviti (approved):

PR #14953, opened 2026-08-25
SE-125 Demo page SEO copy, a stable page header and framework links

taktran:
Ticket: [SE-125](https://ag-grid.atlassian.net/browse/SE-125)

## Summary

The demo pages were built by a shared layout that produced a `Demo - {name}` title and a bare-word H1 (`Performance`, `Inventory`), with short generic meta descriptions and no indexable text under the H1. The weakness came from one place, so one change fixes every demo page.

`src/content/demos/demos.json` becomes a frontmatter contract. Three of the four grid demo pages already read their copy from that file; this widens it to the contract SE-125 asks for and brings `/example/` — which was hardcoded on `Layout` with `title="Demo - Performance Grid"` — onto it.

**The contract** (`useCase`, `pageType`, `framework`, `features[]`, `hook?`, `seoTitle?`, `seoH1?`, `seoDescription?`, `intro?`, `href`, `githubUrl`, `activeTab`) splits into a descriptive half that always has to be written and a copy half that does not. New `src/components/demos/demoContent.ts` resolves the two, deriving anything omitted:

| Field | Derivation |
| --- | --- |
| H1 | `{useCase} {Grid\|Dashboard} Example` |
| Title | `… Example - {hook} \| AG Grid`, dropping the hook when that would pass 60 characters |
| Meta | A sentence assembled from `useCase`, `pageType`, `features` and `framework`, filled toward 140 characters and capped at 155 |
| Intro | The features in prose, plus where the code is available — excluding the framework the live demo already runs in |

So a demo page added with no `seo*` frontmatter gets a real title, H1, meta description and intro, rather than the old fallback.

**Rendering.** The header markup all four pages repeated — H1, GitHub/video/contact links, description, tab strip — is now `DemosPageHeader.astro`, rendering the H1 from `seoH1` and the new 1–2 sentence intro underneath it. `DemosLayout` gained the `markdownUrl` prop `/example/` needs. OG and Twitter tags needed no work: `Layout.astro` already derives `og:title`/`og:description`/`twitter:*` from the `title`/`description` props, so they now come from the same frontmatter for free.

**Markdown twins.** `buildDemoMarkdown` and `buildExampleMarkdown` read the same resolved copy, so `/example.md` no longer serves `title: "AG Grid Demos"` while the HTML page serves something else. `buildDemoMarkdown` also takes the page path from the contract's `href` instead of assuming `/example-{demo}/`, which is what lets `performance` (`/example/`) into the same shape as the rest.

## Copy

Set as written in the ticket. Two of its own limits are not met by the copy it specifies, so they are flagged here rather than silently edited — the wording is SEO-owned, and the inventory note says to deploy it as written:

| Page | Title | Meta (ticket target 140–155) |
| --- | --- | --- |
| `/example/` | 57 ✅ | 147 ✅ |
| `/example-finance/` | **65 — over the stated 60** | **139** |
| `/example-hr/` | 51 ✅ | 145 ✅ |
| `/example-inventory/` | 54 ✅ | **130** |

Happy to trim the finance title and pad the three short metas if @nick-redding / @carlhendy want that.

## Acceptance criteria

| AC | Status |
| --- | --- |
| The pages serve the titles, H1s and metas in the raw server HTML | ✅ for the four grid demos |
| OG and Twitter tags render from the same frontmatter | ✅ already derived from `title`/`description` in `Layout.astro` |
| Each page has an indexable intro under the H1 | ✅ |
| One H2 per feature shown | ➖ deferred — see below |
| A demo page with no `seo*` frontmatter derives its values rather than falling back to `Demo - {name}` | ✅ `demoContent.ts`, covered by unit tests |
| `/studio/example/` title and meta | ➖ different repo — see below |

## Two things deliberately not in this PR

**The per-feature H2s.** These pages are full-bleed grids: `hideFooter={true}`, and `example.module.scss` sizes the grid from a fixed `--layout-top-header-height`, so there is no scrollable prose region for headings to live in. Putting them anywhere — a section below the grid, a compact strip, visually-hidden text — is a layout and design call, not a copy change, so it wants its own ticket. `features[]` is in the contract and exposed on the resolved content, so the follow-up is markup only.

**`/studio/example/`.** It lives in the ag-studio repo (`packages/ag-studio-docs/src/pages/example.astro`, still `title={'Demo - AG Studio'}`), so its title and meta need a companion PR there, alongside SE-126 for its H1.

## Tests

- **New** `src/components/demos/demoContent.test.ts` — 21 cases. Per demo: no `Demo -` fallback title or H1, an H1 and a real intro, at least one feature, and an `href` matching its demo tab. Plus the derivation exercised directly on a demo that supplies only the descriptive half — heading and title formulas, the hook dropped when the title would overrun, the description inside the window, the framework defaulting to JavaScript.
- **New** four Playwright cases in `example-demos.spec.ts` asserting title, meta description, `og:title`, H1 and intro in the served HTML for each demo, read from the same contract.
- **Updated** the two markdown-twin suites for the renamed fields.

## Verification

- `./behave.sh --project ag-grid-docs` on the three affected suites: 41 passed.
- `ag-grid-docs:lint` clean; `tsc` over the docs project reports nothing new (the project has no `build:types` target and is not type-clean, so `.astro` files are not covered by any gate — worth knowing when reviewing).
- `markdownPages.test.ts` fails on this branch, and fails identically on a clean `latest` — it checks `.md` twins against a stale `dist/` and is unrelated.

## The header the old notes flagged

The intro did overflow the hardcoded `--layout-top-header-height: 180px`, and the header's height varied per demo: the H1 shared a wrapping row with the GitHub/video/contact links, and the copy column was content-sized, so the intro wrapped to a different width and line count on each page. Moving between the demos shifted the page (146-206px of header at 1920px wide, and more below that).

The header is now laid out to a reserved height: heading, links and intro are separate rows in a fixed-width column, with a reserve sized from the tallest wrap each takes, so nothing is clipped and no page can move another's grid. `--layout-top-header-height` is the measured 214px (177px on short screens). Measured across 34 widths from 320px to 2560px at three viewport heights: identical on all four demos, and guarded by a new `Demo page header layout` e2e case.

The frameworks the intro names now link to their own copy of the demo in ag-grid-demos (JavaScript points at that repository's `typescript` directory), in the page and in its markdown twin.

The Playwright cases the old notes could not run have now been executed locally: 7 passed. The install problem is gone, so `./behave.sh --project ag-grid-docs` (934 passed) and `./checks.sh` (CHECKS-PASSED) also ran clean. Fixing a JSON import that stopped `example-demos.spec.ts` loading under Playwright was part of that - the four SEO cases from the first commit were being silently skipped as "No tests found".


markdurrant (approved):
LGTM 👍

PR #14962, opened 2026-08-26
AG-3390 Allow the previewed UTM-webhook listener hash in the site CSP

taktran:
## Summary
- Grid staging was blocking an inline script from the shared GTM container: a capturing-phase version of the UTM-attribution webhook listener, currently trialled in GTM preview (container version "env-55", not yet published to the live container).
- Confirmed root cause by fetching the exact `gtm.js` preview payload from the URL in the CSP violation and diffing it against the live/published container — the only change is a third `true` argument on `addEventListener` (bubble → capture phase), otherwise byte-identical to the already-allowed tag.
- Added the new hash (`GTM_UTM_WEBHOOK_CAPTURING_PHASE_HASH`) alongside the existing `GTM_UTM_WEBHOOK_HASH` in `cspRules.ts` so both the live and previewed tag versions pass CSP. Drop the old hash once the GTM change is published and confirmed unused.

## Test plan
- [x] `cspRules.test.ts` — added a test pinning the new hash in the site scope, and asserting it's absent from `examples`
- [x] `./behave.sh --project ag-grid-docs` — all CSP/htaccess tests pass (901 passed; the 1 unrelated failing suite, `getErrorText.test.ts`, is a pre-existing module-resolution issue in a fresh worktree, confirmed unrelated by reproducing/absence on the existing checkout)
- [x] `.htaccess` snapshot regenerated (`--update`) — diff is exactly the one added hash
- [x] `./checks.sh` — CHECKS-PASSED (type-check, lint, spec type-check)


markdurrant (approved):
LGTM 👍

skewis6 (approved):

PR #14968, opened 2026-08-26
[AG-4093] Feature(columns): continuous auto-sizing

sdwvit:
Adds opt-in `autoSizeStrategy.continuous`, which re-applies the configured strategy whenever the grid changes in a way that affects column widths, plus a `shouldAutoSizeColumns(columns, reason)` callback to veto an individual re-size.

Applies to all three strategies (`fitCellContents`, `fitGridWidth`, `fitProvidedWidth`). Columns whose width is owned by a deliberate user resize or an explicit `applyColumnState` width are held out of repeated sizing; `width`/`initialWidth` stay eligible, and `suppressAutoSize`/`suppressSizeToFit` remain the way to pin a column.

Implements the ownership/coalescing approach rather than the `events: []` API sketched on the ticket. Continuous `fitCellContents` is not supported by the Viewport Row Model; there is no runtime re-registration of the strategy.

Fixes [AG-4093](https://ag-grid.atlassian.net/browse/AG-4093)

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
@@ -178,41 +178,6 @@ In the example below, narrow the columns and then run **Autosize All Columns** f
 
 {% gridExampleRunner title="Auto-Size Actions Respecting the Strategy" name="auto-size-ui-actions" /%}
 
-#### Continuous Auto-Sizing
-
-By default the columns are sized to fit their cell contents once, when the first data is rendered. Set `continuous` to have the grid re-size them whenever the data, the displayed columns, the viewport or the grid size changes.
-
-```{% frameworkTransform=true %}
-const gridOptions = {
-    autoSizeStrategy: {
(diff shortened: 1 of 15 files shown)


documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -195,6 +195,47 @@ Just like Excel, each column can also be auto-resized by double clicking the rig
 
 {% /note %}
 
+## Continuous Auto-Sizing
+
+By default `autoSizeStrategy` is applied once, when the grid first renders. Set `continuous` to have the grid re-apply the configured strategy whenever the grid changes in a way that affects column widths. It works with all three strategies, and the strategy itself is unchanged — only how often it runs.
```
sdwvit:
```suggestion
By default `autoSizeStrategy` is applied once, when the grid first renders. Set `continuous` to have the grid re-apply the configured strategy whenever the grid changes in a way that affects column widths. 
```

SalvatorePreviti (approved):

PR #14973, opened 2026-08-26
AG-10819 custom-filter-feat

SalvatorePreviti:
# AG-10819: Custom Filter Options in the Advanced Filter

FIX AG-10819

A column's `filterParams.filterOptions` reached the Advanced Filter only as a list of built-in keys: a list containing a `FilterOptionDef` was discarded whole, so the column offered every default option and none of its own. It now offers a column's Custom Filter Options and evaluates them with the same `predicate` as the column filter, for options taking 0, 1 or 2 values.

| scope  | net lines |     lines changed | hunks |      files changed |
| ------ | --------: | ----------------: | ----: | -----------------: |
| source |      +391 |  1605 (+998 -607) |   179 | 22 (+1, 21 edited) |
| tests  |     +2370 | 2576 (+2473 -103) |    46 |   7 (+1, 6 edited) |
| docs   |      +317 |         317 added |     5 |   5 (+3, 2 edited) |

## A mixed `filterOptions` list narrows what a column offers

`getActiveOperators` handed back the configured array as the key list, so it bailed on any entry that was not a string: a list mixing built-in keys with `FilterOptionDef` objects narrowed nothing, and the column offered every option for its data type while offering none of its own. Each entry now maps to its key (the string itself, or the definition's `displayKey`), which is what the column configured in the first place.

Narrowing decides what is suggested, not what can be read. What the column offers is matched first, and the rest it can resolve may then lengthen that match, so an expression naming an option the list omits still applies, whether typed or restored from a saved model, and the Builder keeps it rather than clearing the condition. A column whose list was all strings rejected such an expression before; accepting it widens what parses, so nothing that worked stops working. `isOperatorOffered` is left with the Builder's column switch, where the user is choosing rather than restoring.

A key that names no operator here cannot narrow to anything, so a list of only those (the `filterOptions` the column filter understands and this one does not, such as the relative date presets or `inRange`) leaves the built-ins standing, as a list whose every entry was dropped does.

## Options are resolved per column

`AdvancedFilterExpressionService` overlays a column's Custom Filter Options on its data type's built-ins, cached per column, so `getExpressionOperator` and its siblings take the column being filtered. A `displayKey` reusing a built-in key replaces that built-in for that column alone, as it does in the column filter. A Multi Filter reads the list from the first child that declares one.

An entry missing `displayKey`, `displayName` or `predicate` is dropped and reported as warning 72, and where a list keeps nothing usable the built-ins stand. A name is the localised text, then the `displayName`, then the `displayKey`, as a column falls back from `headerName` to its field: an expression is written in that name, so an option that resolves to none is spelled by its key rather than lost. Renaming a header through `applyColumnState` now takes the old display name out of the expression vocabulary.

Shared with the column filter, so the two cannot disagree about one entry: `_classifyFilterOptions` decides what a usable entry is, `_getCustomOptionDisplayName` decides what it is called, and `_getCustomOptionNumberOfInputs` clamps `numberOfInputs` to the declared `0 | 1 | 2`, so `numberOfInputs: -1` hands the predicate no values where it previously handed it one. `_getOwn(table, key)` joins `_hasOwn` in `ag-stack`; both refuse `__proto__`, as `_mergeDeep` and `fieldAccess` already do.

The name fallback reaches the column filter too. A zero-input option whose name resolves to blank is summarised by its `displayKey` in a read-only floating filter and in the Filter Tool Panel, where the summary was previously the empty string.

## An operand region of one or two values

`ColFilterExpressionParser` gains an `OperandsParser` owning the whole operand region. One value parses as before; two are comma-separated, conventionally in brackets, which are optional:

```text
[Age] Even Numbers
[Age] Between (Exclusive) (30, 40)
[Date] Between (Exclusive) ("2012-08-12", "2012-08-30")
```

Each value is quoted by its cell data type on the rules the built-in options already use, and additionally where it contains the `,` separating the pair. Completing a two-value option from the suggestion list opens the bracket, and the first value's quote where that type is quoted; both are now written where a space already follows the caret, which previously suppressed the quote as well.

## `filterTo` on the column filter model

Every `ColumnAdvancedFilterModel` member gains `filterTo`, and `type` widens to accept a `displayKey`. `BooleanAdvancedFilterModel` gains `filter` too.

```js
const advancedFilterModel = { filterType: 'number', colId: 'age', type: 'betweenExclusive', filter: 30, filterTo: 40 };
```

## The AI toolkit schema is generated from the offered operators

`getStructuredSchema` listed each data type's operators by hand, so a Custom Filter Option was not expressible and neither was `filterTo`. The `type` enum now comes from the operators the column offers and the value slots from their arity, which also drops two mirrors that had drifted: a boolean column's enum was missing `blank` / `notBlank`, and a `dateString` column wrote `filterType: 'date'` where its model member is `'dateString'`. Models are keyed on the operator set as well as the data type, so two number columns offering different operators produce `numberAdvancedFilterModel` and `numberAdvancedFilterModel2` rather than one enum naming operators half of them do not offer.

The column filter's own schema, built when the Advanced Filter is off, takes its `type` enum from `_classifyFilterOptions` as well, so an entry the filter drops is no longer offered to a model as a legal option.

## The Builder renders a pill per value

`ConditionPillWrapperComp` holds a list of operand pills, rebuilt when the chosen option's arity changes; a pair takes the column filter's `ariaFilterFromValue` / `ariaFilterToValue` labels. Choosing a different column re-resolves the operator against what that column offers: one it does not offer is cleared, one it offers under another name is relabelled.

A condition short of an operand now holds Apply shut whether or not its row is rendered, and for either value rather than only the first: the check the mounted pill runs is the one the Builder's all-items pass runs, so a two-value option left without its second value cannot be applied from a row scrolled out of view.

### DIFF
--- a/packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
+++ b/packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
@@ -14,7 +14,7 @@ import type {
 import type { ISetFilterParams } from '../interfaces/iSetFilter';
 import type { IBigIntFilterParams } from './provided/bigInt/iBigIntFilter';
 import type { IDateFilterParams } from './provided/date/iDateFilter';
-import type { ISimpleFilterParams } from './provided/iSimpleFilter';
+import type { IFilterOptionDef, ISimpleFilterParams } from './provided/iSimpleFilter';
 import type { INumberFilterParams } from './provided/number/iNumberFilter';
 import type { ITextFilterParams } from './provided/text/iTextFilter';
 
@@ -101,6 +101,31 @@ type FilterParamsDefMap = CheckDataTypes<{
     object: FilterParamCallback<ITextFilterParams, any>;
 }>;
 
+/** One list shared by every boolean column, so it can be recognised as the grid's rather than the column's. */
+const BOOLEAN_FILTER_OPTIONS: (string | IFilterOptionDef)[] = [
+    'empty',
+    {
+        displayKey: 'true',
+        displayName: 'True',
+        predicate: (_filterValues: any[], cellValue: any) => cellValue,
+        numberOfInputs: 0,
+    },
+    {
+        displayKey: 'false',
+        displayName: 'False',
+        predicate: (_filterValues: any[], cellValue: any) => cellValue === false,
+        numberOfInputs: 0,
+    },
+];
+
+/**
+ * Whether the list is the grid's own for the cell data type, rather than options the column author wrote.
+ * @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time.
+ */
+export function _isGridSuppliedFilterOptions(filterOptions: unknown): boolean {
+    return filterOptions === BOOLEAN_FILTER_OPTIONS;
+}
+
 // using an object here to enforce dev to not forget to implement new types as they are added
 const filterParamsForEachDataType: FilterParamsDefMap = {
     number: () => undefined,
@@ -108,21 +133,7 @@ const filterParamsForEachDataType: FilterParamsDefMap = {
     boolean: () => ({
         maxNumConditions: 1,
         debounceMs: 0,
-        filterOptions: [
-            'empty',
-            {
-                displayKey: 'true',
-                displayName: 'True',
-                predicate: (_filterValues: any[], cellValue: any) => cellValue,
-                numberOfInputs: 0,
-            },
-            {
-                displayKey: 'false',
-                displayName: 'False',
-                predicate: (_filterValues: any[], cellValue: any) => cellValue === false,
-                numberOfInputs: 0,
-            },
-        ],
+        filterOptions: BOOLEAN_FILTER_OPTIONS,
     }),
     date: () => ({ isValidDate }),
     dateString: ({ dataTypeDefinition }) => ({
--- a/packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
+++ b/packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
@@ -101,8 +101,7 @@ type FilterParamsDefMap = CheckDataTypes<{
     object: FilterParamCallback<ITextFilterParams, any>;
 }>;
 
-/** One list shared by every boolean column, so it can be recognised as the grid's rather than the column's. */
-const BOOLEAN_FILTER_OPTIONS: (string | IFilterOptionDef)[] = [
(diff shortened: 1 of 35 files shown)


packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
```diff
@@ -101,28 +101,39 @@ type FilterParamsDefMap = CheckDataTypes<{
     object: FilterParamCallback<ITextFilterParams, any>;
 }>;
 
+/** One list shared by every boolean column, so it can be recognised as the grid's rather than the column's. */
+const BOOLEAN_FILTER_OPTIONS: (string | IFilterOptionDef)[] = [
```
gportela85: If every boolean column now receives the same array, modifications through a returned colDef affects all boolean columns, while the identity check still classifies the modified list as grid supplied. This can make Column Filter and Advanced Filter disagree. Right?

packages/ag-grid-enterprise/src/advancedFilter/builder/advancedFilterBuilderComp.ts
```diff
@@ -572,38 +573,28 @@ export class AdvancedFilterBuilderComp extends Component<AdvancedFilterBuilderEv
...
-            if (!operator) {
-                item.valid = false;
-                clearOperator(filterModel);
-                clearOperand(filterModel);
-                continue;
-            }
-            if (operator.numOperands > 0 && !_exists((filterModel as any).filter)) {
+                delete (filterModel as PartialColumnFilterModel).colId;
+                clearCondition(filterModel);
+            } else if (!this.advFilterExpSvc.getExpressionOperator(baseCellDataType, filterModel.type, column)) {
```
gportela85: Is it possible for a custom option to change from one input to two after a coDef refresh? If it is, a virtualized column condition can remain valid without filterTo. Only mounted components perform the complete operand check, so Apply can become enabled for an incomplete model.

gportela85 (approved):

packages/ag-grid-enterprise/src/advancedFilter/builder/advancedFilterBuilderComp.ts
SalvatorePreviti: Yes found an issue - `hasEveryOperand` extracted so the all-items pass and the mounted pill share one check, now covering both operands. Fix coming

packages/ag-grid-community/src/filter/filterDataTypeUtils.ts
SalvatorePreviti: hardening it

PR #14978, opened 2026-08-27
AG-3390 Allow the contact form's inline script hash in the site CSP

taktran:
Staging reported a CSP violation blocking the contact form's inline script (`sha256-D3cdipua6lhS2IQ0W0AlSNVVsS+2b/sXycSE8m8PkxY=`), which wasn't yet in the `script-src` allowlist. Adds it as an opaque hash constant, following the existing pattern for other unowned inline scripts (e.g. `GTM_ZOOMINFO_HASH`).


skewis6 (approved):
lgtm

PR #15000, opened 2026-08-28
AG-11294 date-filter-process-blank

SalvatorePreviti:
# Blank values in the number, bigint and date filters

FIX AG-11294
FIX AG-18279

A cell on a number, bigint or date column is now one of three things. Missing (`null`, `undefined`, an empty
string, or only spaces) answers `blank`, and the `includeBlanksIn*` params decide whether a comparison admits it.
Unreadable (present but not the column's type, such as `abc`) answers `notBlank`, is excluded from comparisons
except negations, and no param reaches it. Readable is compared. The column filter and the Advanced Filter now
give the same answer for every combination of the two.

| scope  | net lines |  lines changed | hunks |    files changed |
| ------ | --------: | -------------: | ----: | ---------------: |
| source |       +32 |   90 (+61 -29) |    28 |        10 edited |
| tests  |      +354 | 442 (+398 -44) |    30 | 6 (+1, 5 edited) |
| docs   |        +8 |    18 (+13 -5) |     9 |         4 edited |

## Behaviour changes

Filter models are unchanged, so there is no migration step. Text columns, the Quick Filter and the floating
filter are unaffected. What changes is which rows pass, for data that was being read as something it is not.

- **An empty or whitespace-only string is missing.** A number column read it as the number zero, so one cell
  answered `blank` and also answered less than 5, greater than -5, and not equal to 0. A bigint or date column
  rejected it at its validity gate instead, so it answered `notBlank`. Fewer rows now pass a comparison, more
  pass `blank`, and the `includeBlanksIn*` params can put them back, which they previously could not do for data
  that uses an empty string rather than `null`.
- **A whitespace-only value on a `dateString` column moves from unreadable to missing**, so it no longer passes
  `notEqual`.
- **A bigint cell is parsed before it is compared.** `equals 10` matches a row holding `10n`, `'10'` or `10`,
  where it previously matched only the `bigint`; `notEqual 10` no longer matches rows that are equal to 10.
- **The Advanced Filter's `blank` and `notBlank` answer the same way for a boolean column**, and
  `BooleanAdvancedFilterModelType` gains `'blank' | 'notBlank'`, which the runtime, the autocomplete and the
  documentation already offered.

## Blank is decided before the value is read ([AG-11294](https://ag-grid.atlassian.net/browse/AG-11294))

`SimpleFilterHandler` gains an overridable `isNullValue`. The text filter keeps `cellValue == null`, because an
empty string is a value it can match; `ScalarFilterHandler` widens it to `_isBlank`. Deciding this before the
per-type validity gate is what lets a bigint or date column tell a missing value from an unreadable one, and what
puts both within reach of `includeBlanksIn*`.

`isBlank` becomes `_isBlank` and is exported from `main-internal` alongside a new `_isNotBlank`, so the Advanced
Filter's operators share one definition with the column filters rather than carrying inlined copies.

## The two engines agree ([AG-18279](https://ag-grid.atlassian.net/browse/AG-18279))

`ScalarFilterExpressionOperators` treats a blank value as absent on the same terms, and the `bigint` data type
gets a `valueConverter` of `_parseBigIntOrNull` rather than falling through to identity, which also routes an
unreadable bigint into the guard that already exists for unreadable dates.

`testing/behavioural/src/filters/blank-and-unreadable-values.test.ts` drives every combination of the three
states through both engines and asserts the rows each returns, so a divergence fails rather than only a
disagreement.

## Performance

`_isBlank` runs per row per condition on the scalar path. A first character in the open interval `(32, 160)`
cannot be whitespace to `trim`, so one `codePointAt` settles the common case:

| cell values                       | before   | after    |
| --------------------------------- | -------- | -------- |
| date strings                      | 7.52 ns  | 5.75 ns  |
| text                              | 6.93 ns  | 5.22 ns  |
| strings beginning with whitespace | 10.96 ns | 13.28 ns |

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
+++ b/packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
@@ -42,8 +42,19 @@ export function removeItems<T>(items: T[], startPosition: number, deleteCount?: 
     return deleteCount == null ? items.splice(startPosition) : items.splice(startPosition, deleteCount);
 }
 
-export function isBlank<V>(cellValue: V) {
-    return cellValue == null || (typeof cellValue === 'string' && cellValue.trim().length === 0);
+/** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
+export function _isBlank<V>(cellValue: V): boolean {
+    if (typeof cellValue === 'string') {
+        // No code point in 33..159 is whitespace to `trim`, so one read settles the common case.
+        const first = cellValue.codePointAt(0) ?? 0;
+        return first > 32 && first < 160 ? false : !cellValue.trim();
+    }
+    return cellValue == null;
+}
+
+/** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
+export function _hasValue<V>(cellValue: V): boolean {
+    return !_isBlank(cellValue);
 }
 
 export function getDefaultJoinOperator(defaultJoinOperator?: JoinOperator): JoinOperator {
--- a/documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
+++ b/documentation/ag-grid-docs/src/content/docs/filter-date/index.mdoc
@@ -320,7 +320,7 @@ Applying the Date Filter is described in more detail in the following sections:
 
 ## Blank Cells
 
-If the row data contains blanks (i.e. `null` or `undefined`), by default the row won't be included in filter results. To change this, use the filter params `includeBlanksInEquals`, `includeBlanksInNotEqual`, `includeBlanksInLessThan`, `includeBlanksInGreaterThan` and `includeBlanksInRange`. For example, the code snippet below configures a filter to include `null` for equals, but not for less than, greater than or in range (between):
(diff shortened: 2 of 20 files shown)


packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -42,8 +42,19 @@ export function removeItems<T>(items: T[], startPosition: number, deleteCount?:
...
+    if (typeof cellValue === 'string') {
+        // No code point in 33..159 is whitespace to `trim`, so one read settles the common case.
+        const first = cellValue.charCodeAt(0);
+        return first > 32 && first < 160 ? false : !cellValue.trim();
+    }
+    return cellValue == null;
+}
+
+/** @internal AG_GRID_INTERNAL - Not for public use. Can change / be removed at any time. */
+export function _isNotBlank<V>(cellValue: V): boolean {
```
sdwvit:
```suggestion
export function _hasValue<V>(cellValue: V): boolean {
```
? 
not a huge fan of the word 'not' in the name

sdwvit (approved):

packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
SalvatorePreviti: done

PR #15006, opened 2026-08-28
AG-10029 advanced filter builtin ranges

SalvatorePreviti:
# AG-10029: Built-in Range and Relative Date Options in the Advanced Filter

FIX AG-10029

The Advanced Filter had no operator for `inRange` or for the Date Filter's relative options, so a Number or Date column could not be filtered on a range through it at all, and a column naming those keys in `filterParams.filterOptions` narrowed to nothing. It now offers `between` on every scalar Cell Data Type, and the 22 relative date options on the date ones.

| scope  | net lines |   lines changed | hunks |      files changed |
| ------ | --------: | --------------: | ----: | -----------------: |
| source |       +79 | 707 (+393 -314) |    44 | 11 (+1, 10 edited) |
| locale |       +32 |        32 added |    32 |          32 edited |
| tests  |      +635 |  747 (+691 -56) |    47 |   8 (+1, 7 edited) |
| docs   |      +223 |   225 (+224 -1) |     8 |   4 (+3, 1 edited) |

## `between` on every scalar type

`ScalarFilterExpressionOperators` gains `inRange`, taking two values through the operand grammar that already serves a Custom Filter Option of the same arity, and positioned above `is blank`. It is offered by default, as it is in the column filter, for `number`, `bigint`, `date`, `dateString`, `dateTime` and `dateTimeString`.

```text
[Age] between (21, 38)
[Date] between ("2026-07-01", "2026-08-01")
```

Evaluation follows `ScalarFilterHandler`'s range comparison: exclusive of both ends unless `inRangeInclusive`, and blank cells match only under `includeBlanksInRange`. Both params join the set the evaluator reads off `filterParams`. Two differences from the column filter, both of which the Advanced Filter already had for its other scalar options: `filterParams.comparator` is not honoured, and a reversed range is applied and matches nothing rather than being refused, which is what the ticket asks for.

## Relative date options, offered where the column asks for them

The 22 options `ISimpleFilterModelPresetType` names take no value and are evaluated against the same half-open range the Date Filter computes. The range functions, their cache and the map naming them move out of `DateFilterHandler` into `relativeDateRanges.ts`, so the two filters cannot disagree about what `Last Year` means and neither reaches the other's module; the option stays relative rather than being resolved to fixed dates when it is applied.

```text
[Date] Last 90 Days
```

`PRESET_DATE_FILTER_TYPES` comes off that map, and `zeroInputTypes` in `simpleFilterUtils` is built from it rather than repeating the 22 keys.

None of them is a default, as none is in the Date Filter, so `DataTypeFilterExpressionOperators` gains `defaultOperators`: what a column offers where it narrows nothing, as distinct from everything it can resolve. A date column suggests one only where its `filterOptions` names it, while an expression naming one it left out still parses, which is the rule the rest of the filter's options already follow.

## Both are named as the column filter names them

`between` gets an `advancedFilterInRange` entry beside the other operators', lower case as they are, seeded in every locale from the `filterSummaryInRange` the column filter already summarises the option with. `Today` and `Last 90 Days` read the Date Filter's own entries, so the two filters cannot disagree about what an option is called. The key in the filter model is the column filter's too, so `type: 'inRange'` with `filterTo`, and `type: 'lastYear'`.

`ScalarAdvancedFilterModelType` gains `'inRange'`, and the four date members of `ColumnAdvancedFilterModel` take a new `DateAdvancedFilterModelType`, which is that union plus the relative options. The ticket records that no API change is needed; these two are what lets the model express the options it now produces, and widening an accepted union breaks no caller.

## Open: the casing of the relative date names

AG-10029 asks for lowercase labels. `between` is lower case, from the summary entry that already carries it. The 22 relative options are not: they read the Date Filter's own names (`Today`, `Last 90 Days`), which have no lower-case entry, so matching them would need either a locale-aware lowercasing, wrong wherever nouns are capitalised, or 22 new translated keys.

### DIFF
--- a/packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
+++ b/packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
@@ -4,6 +4,7 @@ import type { Column } from '../../interfaces/iColumn';
 import type { FilterInputCallbackParams } from '../../interfaces/iFilter';
 import type { LogService } from '../../validation/logService';
 import type { FilterLocaleTextKey } from '../filterLocaleText';
+import { PRESET_DATE_FILTER_TYPES } from './date/relativeDateRanges';
 import type { FilterOptionKey, IFilterOptionDef, ISimpleFilterModelType, JoinOperator, Tuple } from './iSimpleFilter';
 import type { OptionsFactory } from './optionsFactory';
 
@@ -94,28 +95,7 @@ const zeroInputTypes: ReadonlySet<string> = new Set<ISimpleFilterModelType>([
     'empty',
     'notBlank',
     'blank',
-    'today',
-    'yesterday',
-    'tomorrow',
-    'thisWeek',
-    'lastWeek',
-    'nextWeek',
-    'thisMonth',
-    'lastMonth',
-    'nextMonth',
-    'thisQuarter',
-    'lastQuarter',
-    'nextQuarter',
-    'thisYear',
-    'lastYear',
-    'nextYear',
-    'yearToDate',
-    'last7Days',
-    'last30Days',
-    'last90Days',
-    'last6Months',
-    'last12Months',
-    'last24Months',
+    ...PRESET_DATE_FILTER_TYPES,
 ]);
 
 /** An entry missing any of these cannot be offered; they are listed so the warning can name the missing one. */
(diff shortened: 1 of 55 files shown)


packages/ag-grid-community/src/filter/provided/simpleFilterUtils.ts
```diff
@@ -94,28 +95,7 @@ const zeroInputTypes: ReadonlySet<string> = new Set<ISimpleFilterModelType>([
     'empty',
     'notBlank',
     'blank',
-    'today',
```
sdwvit, line 97: there may have been a reason these are spelled out, but I don't remember anymore, so lets roll with it

sdwvit (approved):

PR #15032, opened 2026-09-01
AG-8950 pre refactor

SalvatorePreviti:
# Advanced Filter autocomplete and operator lookup

Groundwork for Set Filter support in the Advanced Filter. None of it is Set Filter specific, so it lands
first.

| scope  | net lines | lines changed | hunks | files changed |
| ------ | --------: | ------------: | ----: | ------------: |
| source |        +1 | 181 (+91 -90) |    35 |      4 edited |
| tests  |       +80 |      80 added |     4 |      1 edited |

## The autocomplete search was quadratic in the matches

`AgAutocompleteList.runSearch` built an array of every entry's text, filtered that to the matching strings,
then filtered the entries by testing each one's text for membership of the match array. That is
O(entries x matches), on every keystroke. It is now a single pass returning the entries to show and the row to
suggest together.

One keystroke in the column autocomplete, headless Chromium, where the entries are the filterable columns:

| entries | keystroke            |   before |  after |
| ------: | -------------------- | -------: | -----: |
|   2 000 | matches every entry  |   2.55ms | 0.30ms |
|   2 000 | narrows to a handful |   0.23ms | 0.17ms |
|  20 000 | matches every entry  | 245.93ms | 0.84ms |
|  20 000 | narrows to a handful |   2.73ms | 0.56ms |

Ten times the entries cost 96x before and 2.8x after.

Three list scans go with it. `AgVirtualList` hands its component creator the row index it already has, rather
than the creator recovering it with `indexOf`, and the list records where its selection sits instead of
searching for it in `getActiveOptionId`, which `onMouseMove` reaches at pointer rate.

The suggestion rule is unchanged and now has a test: the shortest entry starting with the search string,
otherwise the shortest holding it.

## `useFuzzySearch` is dead

`AgAutocompleteList` has three construction sites (`formulaInputAutocompleteFeature`, `calculatedColumnForm`,
`agAutocomplete`), none of which sets the flag, and the class is exported from no entry point. Its sibling
`useStartsWithSearch` is set by the formula input, so only this arm is dead. It put a Levenshtein pass in front
of the two cheap search arms on every keystroke.

`AgAutocomplete.validationMessage`, assigned by `validate` and read nowhere, goes with it.

## Operator tables are keyed by column, then data type

`columnExpressionOperators` was an object of data types holding a `WeakMap` each. It is now one `WeakMap` from
column to a record keyed by data type.

The data type stays in the key rather than collapsing to `WeakMap<AgColumn, ColumnOperators>`.
`getModelOperator` resolves against `model.filterType`, which is public API, so a model restored from storage
or set through `setAdvancedFilterModel` can name a data type the column no longer has, and a single-slot cache
would hand that entry back to the column's own lookup.

`createColumnOperators` builds its narrowed key list in one loop, instead of materialising the offered keys, a
filtered copy and a closure.

### DIFF
--- a/packages/ag-grid-enterprise/src/advancedFilter/autocomplete/agAutocomplete.ts
+++ b/packages/ag-grid-enterprise/src/advancedFilter/autocomplete/agAutocomplete.ts
@@ -66,7 +66,6 @@ export class AgAutocomplete extends Component<AgAutocompleteEvent> {
     private autocompleteListParams: AutocompleteListParams;
     private lastPosition: number = 0;
     private valid: boolean = true;
-    private validationMessage: string | null;
     private listAriaLabel: string;
     private listGenerator?: (value: string | null, position: number) => AutocompleteListParams;
     private validator?: (value: string | null) => string | null;
@@ -248,7 +247,7 @@ export class AgAutocomplete extends Component<AgAutocompleteEvent> {
         if (!this.validator) {
             return;
         }
-        const validationMessage = (this.validationMessage = this.validator(value));
+        const validationMessage = this.validator(value);
         this.eAutocompleteInput.getInputElement().setCustomValidity(validationMessage ?? '');
         this.valid = !validationMessage;
         this.dispatchLocalEvent<AutocompleteValidChangedEvent>({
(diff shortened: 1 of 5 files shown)


packages/ag-grid-enterprise/src/advancedFilter/autocomplete/agAutocomplete.ts
```diff
@@ -248,7 +247,7 @@ export class AgAutocomplete extends Component<AgAutocompleteEvent> {
         if (!this.validator) {
             return;
         }
-        const validationMessage = (this.validationMessage = this.validator(value));
+        const validationMessage = this.validator(value);
```
sdwvit, line 250: lost this.validationMessage = validationMessage?

sdwvit (approved):

packages/ag-grid-enterprise/src/advancedFilter/autocomplete/agAutocomplete.ts
SalvatorePreviti: that field was unused, removed it - was written and never read

PR #15033, opened 2026-09-01
AG-18255 Rework docs markdown actions into an AI Options dropdown

stgrgic:
https://ag-grid.atlassian.net/browse/AG-18255

Replaces the Copy-as-Markdown split button on docs pages with a single **AI Options** dropdown that matches the framework selector beside it (same trigger treatment, whole button opens the menu).

## What changed

- Menu items with icon + title + description: **Open in Claude**, **Open in ChatGPT**, **Copy Markdown**, **View as Markdown**, and a new **Copy Prompt** action that copies the page-read prompt for use in any AI tool.
- New 16x16 inline SVGs (`ai-sparkle`, `claude-logo`, `chatgpt-logo`, `markdown-logo`, `copy-document`) imported via `?react`. The SVGO-minified exports all used `a`/`b` def ids, which would collide once inlined on the same page, so ids are renamed per icon (reference renames only, artwork untouched).
- Dark mode repaints the hard-coded black glyphs white via `path[fill='#000']`; the Claude/AI-sparkle brand colours are left alone.
- **Copy Markdown** hands the pending fetch to `ClipboardItem` so the clipboard write starts inside the click's user activation — Safari rejects a plain `writeText` after an `await` (with a `writeText` fallback for browsers without `ClipboardItem`).
- **View as Markdown** carries `data-astro-reload` so the view-transitions client router doesn't intercept a navigation to a non-HTML resource.
- The trigger keeps a stable "AI Options" accessible name; Copied/Copy failed feedback flashes on the visible label (width reserved, no layout shift) and is announced from a visually hidden live region.

## Notes for reviewers

- Analytics: same `markdownActions` Plausible event, but the action label `Copy as Markdown` is now `Copy Markdown`, and `Copy Prompt` is new — dashboards filtering the old string will see the rename.
- Verified on the local docs dev server in light/dark mode, desktop/mobile: menu rendering, `.md` link navigation, copy feedback, keyboard/select behaviour via Radix, and computed dark-mode icon fills. `./checks.sh` green.

Fix [AG-18255](https://ag-grid.atlassian.net/browse/AG-18255)


taktran (approved):
LGTM

PR #15038, opened 2026-09-02
AG-18371: Privacy policy copy updates

skewis6:
Replaces the privacy policy page copy with the v3 document from the ticket, and realigns the consent copy next to every user-submission form.

- Policy body goes from 11 to 12 sections: the old combined section 2 splits into "What Data Do We Collect About You" and a new "Purposes Of Processing And Lawful Bases". Section 2 keeps its `data-what-why` anchor so inbound deep links still resolve.
- The data-processing consent checkbox is replaced by a processing notice above the submit button, and the marketing consent wording now covers AG Grid and Bryntum in one checkbox.
- Processing is no longer consent-based, so the field is dropped from both submissions rather than auto-filled on the website — the blanket `true` is being applied in Salesforce instead, so it stays website-agnostic.

Needs checking before release:
- The source document has `[DATE]` as its Last Updated value; **2 September 2026** is assumed (in `policyContent.ts` and one test assertion).
- Staging submit on the contact and trial-licence forms, since the field is no longer sent. The trial-licence path goes via the `CreateLeadForTrialLK` cloud function, which isn't in this repo — worth confirming it doesn't require the dropped key.
- Source-document defects transcribed as-is: one stale `info@ag-grid.com` in section 5 where the rest uses `privacy@ag-grid.com`, and a near-duplicate opt-out paragraph. One "Brytum" typo was corrected to "Bryntum".
- Copy shared via `ag-website-shared`, so it reaches ag-charts and ag-studio on the next subrepo push.

### DIFF
--- a/external/ag-website-shared/src/constants.ts
+++ b/external/ag-website-shared/src/constants.ts
@@ -12,13 +12,11 @@ export const RECAPTCHA_URL = 'https://www.google.com/recaptcha/api.js';
  */
 export const CONSENT_FIELD_IDS = {
     default: {
-        dataProcessingConsentId: '00NS900000KbwBF',
         marketingEmailConsentId: '00NS900000KbwBI',
         emailTrackingConsentId: '00NS900000KbwBG',
         franceOrItalyId: '00NS900000KbwBH',
     },
     production: {
-        dataProcessingConsentId: '00NQ500000IfxTS',
         marketingEmailConsentId: '00NQ500000JApXt',
         emailTrackingConsentId: '00NQ500000JApXs',
         franceOrItalyId: '00NQ500000JD5zq',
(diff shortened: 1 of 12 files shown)


external/ag-website-shared/src/constants.ts
```diff
@@ -12,13 +12,11 @@ export const RECAPTCHA_URL = 'https://www.google.com/recaptcha/api.js';
  */
 export const CONSENT_FIELD_IDS = {
     default: {
-        dataProcessingConsentId: '00NS900000KbwBF',
```
taktran, line 15: Is this no longer required for the form?

taktran (approved):
Just a q

external/ag-website-shared/src/constants.ts
skewis6: No, spoke with Joe and this is going to automated as `true` on the SF side

PR #15042, opened 2026-09-02
[AG-18235] Feat(grid-state): add quick filter text to grid state

sdwvit:
Adds a top-level `quickFilter` section to `GridState`, captured from and restored into the `quickFilterText` grid option. Also resyncs the toolbar's quick filter input when the option changes from outside it.

Capture and restore are gated on the Quick Access Toolbar being configured with `agQuickFilterToolbarItem`. Without that item the grid option is the only source, and state neither captures nor clears it.

A restore through `api.setState()` reports `api` as the `filterChanged` source (AC 5.1.8); a user's own edit in the toolbar input still reports `quickFilter`.

An `api` restore resets what it omits, so `setState()` with no `quickFilter` section clears the quick filter — unless the toolbar item is absent.

`./checks.sh`, `./behave.sh` and `./docs-e2e.sh --grep grid-state` green.


SalvatorePreviti (approved):

documentation/ag-grid-docs/src/content/docs/grid-state/index.mdoc
```diff
@@ -51,6 +51,10 @@ The grid state is designed to be serialisable, so any functions will be stripped
 
 {% interfaceDocumentation interfaceName="GridState" overrideSrc="grid-state/grid-state.json" /%}
 
+The `quickFilter` section only takes part in grid state when the [Quick Access Toolbar](./toolbar/) is configured with `agQuickFilterToolbarItem`. Without that item the [Quick Filter](./filter-quick/) is driven by the `quickFilterText` grid option alone, and is neither captured nor restored.
```
StephenCooper: I would like to try and have this only added via the JSDoc so it is explained there without the need for this extra paragraph

documentation/ag-grid-docs/src/content/docs/grid-state/index.mdoc
```diff
@@ -51,6 +51,10 @@ The grid state is designed to be serialisable, so any functions will be stripped
...
+
+When the section does apply, restoring it writes the text back into the `quickFilterText` grid option, so the restored state takes precedence over a value set in the grid options. If the provided state has no `quickFilter` section, the grid option is used as provided. A restore through `api.setState()` reports `api` as the source on the resulting [Filter Changed](./filtering-overview/) event, where a user's own edit in the toolbar input reports `quickFilter`.
```
StephenCooper: I would not include this level of detail, especially as it should be a very rare case for both the toolbar and gridOption to be used at the same time.

documentation/ag-grid-docs/src/content/docs/grid-state/index.mdoc
```diff
@@ -84,6 +88,8 @@ The best way to restore grid state is via initial state as described above. Howe
 
 It is possible to maintain the existing state for individual state contents by passing a second argument to `setState` that contains the top-level properties to ignore. E.g. `api.setState(state, ['filter'])` will maintain the existing filter state in the grid.
 
+Anything the provided state omits is reset rather than left as it is. For example, a state without a `quickFilter` section clears the [Quick Filter](./filter-quick/) — unless the toolbar has no `agQuickFilterToolbarItem`, in which case the `quickFilterText` grid option is left untouched.
```
StephenCooper: I think this extra sentence is helpful but I would update to 

Anything the provided state omits is reset rather than left as is. For example, a state without a `quickFilter` section clears the quick filter value in the toolbar.

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -323,6 +329,12 @@ export class StateService extends BeanStub implements NamedBean {
             this.updateGroupExpansionState();
         };
         const updateFilterState = () => updateCachedState('filter', this.getFilterState());
+        // A case-only edit (`abc` -> `ABC`) parses to the same filter, so it dispatches no `filterChanged`.
+        this.addManagedPropertyListener('quickFilterText', () =>
+            updateCachedState('quickFilter', this.getQuickFilterState())
+        );
+        // Capture is gated on the toolbar's quick filter item, so ownership can change on its own.
+        this.addManagedPropertyListener('toolbar', () => updateCachedState('quickFilter', this.getQuickFilterState()));
```
StephenCooper: Is this actually required? We are always checking the presence of the toolbar item so when the quickFilterText is changed won't it just start updating the state if a toolbar is added?

packages/ag-grid-enterprise/src/toolbar/providedItems/quickFilterToolbarItem.ts
```diff
@@ -57,15 +57,29 @@ export class QuickFilterToolbarItem extends Component implements IToolbarItemCom
...
+        this.addManagedPropertyListener('quickFilterText', ({ currentValue }) => syncInput(currentValue));
+
+        const quickFilterSvc = this.beans.quickFilter;
+        if (quickFilterSvc) {
+            // A restore of the value the option already holds changes nothing to listen to above, but
+            // still has to beat a pending edit.
+            this.addManagedListeners(quickFilterSvc, {
+                quickFilterStateApplied: () => syncInput(this.gos.get('quickFilterText')),
+            });
+        }
```
StephenCooper: I think this is overkill. Its such an edge case to try and support. I am guessing its probably only really going to happen under test given the debounce timing.

documentation/ag-grid-docs/src/content/docs/grid-state/index.mdoc
sdwvit: fixed, ty

packages/ag-grid-enterprise/src/toolbar/providedItems/quickFilterToolbarItem.ts
sdwvit: fixed, ty

PR #15043, opened 2026-09-02
[AG-18393] Docs(grid-state): list the missing state sections

sdwvit:
https://ag-grid.atlassian.net/browse/AG-18393
Fix AG-18393

The Grid State page's "State Contents" list omitted nine sections that `GridState` captures, and described `userColumns` as only covering added or modified Calculated Columns. Prose only — no examples or code touched. Split out of #15042 so the feature diff stays readable.

Also adds a rulesync rule scoped to `gridState.ts` / `stateService.ts`: the list is hand-maintained and nothing fails when it drifts, so the rule attaches when either file is edited.

### DIFF
--- a/.rulesync/rules/grid-state.md
+++ b/.rulesync/rules/grid-state.md
@@ -1,0 +1,28 @@
+---
+targets: ['*']
+description: 'Keeping the Grid State docs page in sync with the GridState interface'
+globs:
+    [
+        'packages/ag-grid-community/src/interfaces/gridState.ts',
+        'packages/ag-grid-community/src/misc/state/stateService.ts',
+    ]
+---
+
+# Grid State
+
+## The docs page is part of the change
+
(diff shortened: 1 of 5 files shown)


.rulesync/rules/grid-state.md
```diff
@@ -0,0 +1,28 @@
+---
+targets: ['*']
+description: 'Keeping the Grid State docs page in sync with the GridState interface'
+globs:
+    [
```
StephenCooper, line 5: This is an interesting pattern to the stale docs. 

However, I am wondering if we should just improve the JsDocs so that the interface component can be the source of truth. This way it will stay in sync and we will remove duplication on that docs page. 

We can always add links via the interface component. Then when a new property is added it will automatically appear and if the overrides file is not updated it will just be missing a link.

sdwvit: How do you make it enforce a line in the markdown?

StephenCooper (approved):

PR #15046, opened 2026-09-02
[AG-18395] Feat(grid-state): Find state

sdwvit:
Fix [AG-18395](https://ag-grid.atlassian.net/browse/AG-18395)
Fix [AG-18447](https://ag-grid.atlassian.net/browse/AG-18447)

Adds `GridState.find` (`searchValue`, `activeMatch`), captured and restored for the Client-Side Row Model. Highlighting needs nothing extra — it derives from `findSearchValue` — and scroll is already covered by the `scroll` section, which is restored after find so a saved scroll position wins.

Also fixes the Find toolbar input not following an external `findSearchValue` change (the same resync bug as #15042 fixes for Quick Filter).

Overlaps #15042 on the grid-state examples: both add a `toolbar` to the same eight files, so whichever merges second needs the two `items` arrays merged into one.


documentation/ag-grid-docs/src/content/docs/grid-state/index.mdoc
```diff
@@ -47,6 +47,10 @@ The grid state is made up of the sections below, each of which can be provided o
 
 The state also contains the grid version number. When applying state with older version numbers, any old state properties will be automatically migrated to the current format.
 
+The `find` section only takes part in grid state when the [Quick Access Toolbar](./toolbar/) is configured with `agFindToolbarItem`. Without that item [Find](./find/) is driven by the `findSearchValue` grid option alone, and is neither captured nor restored.
```
StephenCooper: I would see if we can not add this to the docs page and just have it appear via the JsDoc for the findState. 

The second paragraph gives too much detail about the implementation that a dev will see once they start using it. I also don't think we need to document the case we are not expecting to happen with a mixture of gridOption and gridState.

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -391,6 +391,24 @@ export class StateService extends BeanStub implements NamedBean {
...
 
+        // Runs after the pivot columns and the deferred filter state: the active match is an ordinal
+        // over the visible columns and the filtered rows, so those have to be in place to resolve it
+        // against the same match list it was captured from. Runs before the scroll and pagination
+        // restores: going to the active match scrolls and pages to it, and the captured position is
+        // the one the user was actually left on.
+        if (shouldSetState('find', findState)) {
+            this.setFindState(findState, source);
+            const savedPage = paginationState?.page;
+            if (findState?.activeMatch != null && savedPage != null && !ignoreSet?.has('pagination')) {
```
StephenCooper: Is this to stop it changing the page? Does this mean we are setting the pagination page twice?

sdwvit: the saved page wins over find's navigation, so this is deliberate. but let me see if I can avoid calling it twice.

sdwvit: fixed, check pls

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -417,7 +428,10 @@ export class StateService extends BeanStub implements NamedBean {
                 }
             },
             bodyScrollEnd: () => updateCachedState('scroll', this.getScrollState()),
+            findChanged: () => updateCachedState('find', this.getFindState()),
         });
+        // Capture is gated on the toolbar's find item, so ownership can change without a Find event.
+        this.addManagedPropertyListener('toolbar', () => updateCachedState('find', this.getFindState()));
```
StephenCooper: I would say this is overkill.

packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -801,6 +815,31 @@ export class StateService extends BeanStub implements NamedBean {
         rangeSvc.setCellRanges(cellRanges);
     }
 
+    /**
+     * Find is only state-managed when the Quick Access Toolbar owns an input for it; otherwise the
+     * `findSearchValue` grid option is the only source and state leaves it alone.
+     */
+    private isFindStateManaged(): boolean {
```
StephenCooper: I would create a single method that find and quick filter can share.

packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -171,7 +172,7 @@ export class FindService extends BeanStub implements NamedBean, IFindService {
             this.goToInContainer(topMatches, match, 0);
             return;
         }
-        if (match <= centerNumMatches) {
+        if (match <= topNumMatches + centerNumMatches) {
```
StephenCooper: What is this change required for?

packages/ag-grid-enterprise/src/find/findService.ts
```diff
@@ -296,6 +297,43 @@ export class FindService extends BeanStub implements NamedBean, IFindService {
     }
 
     // updates all the matches
+    public getState(): FindState | undefined {
```
StephenCooper, line 300: There are a lot of comments in this code. Seems like more than the rest of the file.

StephenCooper (commented):
Some questions.

packages/ag-grid-enterprise/src/find/findService.ts
sdwvit: it is a fix for an out of scope issue. will raise a ticket

sdwvit: basically when you have pinner rows, stepping through matches will go out of bounds due to a wrong calculation

sdwvit: I've pulled it out to AG-18482 to keep this PR to the state feature. Reverted in 919c21f4e59

StephenCooper (approved):

PR #15050, opened 2026-09-02
AG-18391 validate-between-value-ordering

SalvatorePreviti:
# AG-18391: Validate Between value ordering in the Advanced Filter

FIX AG-18391

The Advanced Filter accepted `is between` with its two values in the wrong order. `[Date] is between ("2027-09-01", "2025-01-01")` applied, matched no row, and gave the user nothing to tell that apart from a filter that simply found nothing. The column filters refuse such a pair, so the same range was accepted in one filter and rejected in the other.

| scope  | net lines |  lines changed | hunks | files changed |
| ------ | --------: | -------------: | ----: | ------------: |
| source |      +140 | 302 (+221 -81) |    67 |     13 edited |
| tests  |      +299 | 337 (+318 -19) |    14 |      5 edited |
| docs   |        +0 |   42 (+21 -21) |     3 |      1 edited |

## The rule is the column filter's own ([AG-18391](https://ag-grid.atlassian.net/browse/AG-18391))

`_getValidityMessageKey` in `simpleFilterUtils` was already the Number and BigInt filters' definition of the ordering rule and of which message states it, and the Date Filter carried a second copy of the same rule for its own four messages. The two are now one function, keyed on whether the bounds are dates, and the Advanced Filter reads it as well. An out-of-order pair therefore reports whichever message the equivalent column filter selects: `Must be greater than 38` on a number column, `Date must be after 2024-11-30` on a date one, and the inclusive wording of each where the column sets `inRangeInclusive`.

An `is between` whose first value is not below its second is invalid: the expression is rejected, the Apply button stays shut, and the grid keeps whatever filter it already had. Where the column sets `inRangeInclusive`, a pair of equal values is a valid exact match, as it is in the column filter; without it both bounds are exclusive, so an equal pair is rejected too. This reaches an `advancedFilterModel` set through the API as well as one typed or built by hand.

Both surfaces validate. The expression parser reports the error against the second value's own span, and the Builder marks the condition invalid and shuts Apply while the pair is out of order.

`getConditionValidationMessage` is now the one answer to "why can this condition not be applied", replacing the pair of checks each of the Builder's two validity sites used to spell out for itself: the condition's own component, and the pass over every item, which is the only one a row the virtual list has not mounted gets. That pass now decides validity rather than only withdrawing it, so a condition its column stops rejecting recovers without the user scrolling it into view, and the Apply button reports the condition's own message instead of claiming it is unfinished.

## One field for what an option takes, replacing an arity and a flag

`FilterExpressionOperator` carried `numOperands: number`. It now carries `operands: 'none' | 'one' | 'range'` instead, with a module-level `OPERAND_COUNT` for the few sites that genuinely want a number. The rule fires on `range`, so nothing has to agree separately about which arities are ordered, and `OperatorParser` drops the count it used to cache beside the operator.

Any option taking two values is a `range`, a Custom Filter Option declaring `numberOfInputs: 2` included. That is what the column filters already do: their gate is the arity alone, so a custom two-value option's pair is order-validated there today. Treating it as anything else in the Advanced Filter would put the two filters back into disagreement on the case this ticket exists to close.

## The `is between` docs row listed the wrong data types

The table of Advanced Filter options omitted `bigint` from every scalar row, while the page's own BigInt section describes the support in detail and `ScalarFilterExpressionOperators<bigint>` supplies those options. The nine affected rows now list it.


testing/behavioural/src/filters/advanced-filter/advanced-filter-builtin-ranges.test.ts
```diff
@@ -171,21 +214,450 @@ describe('Advanced Filter - built-in range and relative date options', () => {
...
             const api = await gridsManager.createGridAndWait('grid1', opts());
+            const af = AdvancedFilterHarness.get(api);
 
-            await AdvancedFilterHarness.get(api).applyExpression('[Age] is between (38, 21)');
+            await af.applyExpression('[Age] is between (38, 21)');
             await asyncSetTimeout(0);
+            await new FilterDom(api, 'reversed number pair').checkFilterDom(`
+                ADVANCED FILTER
+                input: "[Age] is between (38, 21)"
+                valid: false — Expression has an error. Must be greater than 38 - 21.
```
StephenCooper: Is this the correct error message? "Must be greater than" Do we not have an error about invalid range?

SalvatorePreviti: I pushed some changes to make the messaging clearer in the AF, but, I don't think we need a specific invalid range message / validation, that would add quite a lot of complexity and would need some more design input, I think showing in the from "Must be less than 21" and in the to "Must be greater than 38" for example fits the purpose if you agree

StephenCooper (approved):

PR #15061, opened 2026-09-03
[AG-18402] Fix(column-sizing): debounce viewport-driven continuous auto-size

sdwvit:
Continuous `fitCellContents` auto-sizing re-sized columns once per animation frame while scrolling, because `viewportChanged` fires as new rows render — the grid visibly jitters, more so with `animateColumnResizing`.

Debounces the three viewport triggers (`viewportChanged`, `virtualColumnsChanged`, horizontal `bodyScroll`) by 150ms, matching `SCROLL_END_TIMEOUT`, so a scroll re-sizes once it settles. Data, column and grid-size triggers keep the existing per-frame path.

https://ag-grid.atlassian.net/browse/AG-18402

### DIFF
--- a/packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
+++ b/packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
@@ -42,10 +42,11 @@ const UI_MENU_SOURCES: ReadonlySet<ColumnEventType> = new Set(['columnMenu', 'co
 type AutoSizeReason = AutoSizeColumnsTriggerParams['reason'];
 
 /**
- * Matches `SCROLL_END_TIMEOUT`, so viewport re-sizes settle when the grid considers scrolling finished
- * rather than jittering once per frame through a scroll.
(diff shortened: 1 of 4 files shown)


packages/ag-grid-community/src/columnAutosize/columnAutosizeService.ts
```diff
@@ -41,6 +41,12 @@ const UI_MENU_SOURCES: ReadonlySet<ColumnEventType> = new Set(['columnMenu', 'co
 
 type AutoSizeReason = AutoSizeColumnsTriggerParams['reason'];
 
+/**
+ * Matches `SCROLL_END_TIMEOUT`, so viewport re-sizes settle when the grid considers scrolling finished
+ * rather than jittering once per frame through a scroll.
+ */
+const CONTINUOUS_VIEWPORT_DEBOUNCE = 150;
```
StephenCooper: Should this also be applied to `fitGridWidth` when the size of the grid is being changed. I think resizing leads to the same jitter in the columns.

StephenCooper (approved):

PR #15062, opened 2026-09-03
AG-18408 Add a continuous auto-sizing example to the column sizing docs

sdwvit:
Adds a `continuous-auto-size-in-action` example to the Continuous Auto-Sizing section of the column sizing docs: ~10 columns under three column groups, the full olympic-winners dataset and pagination, with `fitCellContents` + `continuous: true` so the columns re-fit as you page through.

https://ag-grid.atlassian.net/browse/AG-18408

### DIFF
--- a/documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/continuous-auto-size-in-action/main.ts
+++ b/documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/continuous-auto-size-in-action/main.ts
@@ -41,9 +41,6 @@ const gridOptions: GridOptions<IOlympicData> = {
         type: 'fitCellContents',
         continuous: true,
     },
-    // `fitCellContents` can only measure the columns that are rendered, so every column has to be
-    // rendered for all ten to re-fit around each page's content rather than only once scrolled into view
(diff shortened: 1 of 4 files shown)


documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/continuous-auto-size-in-action/main.ts
```diff
@@ -0,0 +1,57 @@
...
+    pagination: true,
+    paginationPageSize: 20,
+    paginationPageSizeSelector: [20, 50, 100],
+    autoSizeStrategy: {
+        type: 'fitCellContents',
+        continuous: true,
+    },
+    // `fitCellContents` can only measure the columns that are rendered, so every column has to be
+    // rendered for all ten to re-fit around each page's content rather than only once scrolled into view
+    suppressColumnVirtualisation: true,
```
StephenCooper: I would like to avoid adding this flag as it can lead to performance issues in users applications. 

The continous flag is meant to be able to help with exactly this case of when columns are virtualised.

StephenCooper (approved):

PR #15069, opened 2026-09-03
AG-13042 Warn when legacy theme variables are set under the Theming API

sdwvit:
The Theming API does not read the v32 legacy theme variables, so setting one is silently ignored.

This is what made the QA report on AG-13042 look like a row-model bug. The `compact` preset sets `--ag-grid-size` (v36 never reads it) and `--ag-font-size: 10px` (clamped by `rowHeight = max(iconSize, cellFontSize) + spacing * 3.25`), so the row height cannot move. Reproduced in a real browser at 42 / 58 / 42px in all four row models including the CSRM baseline, so no row model is at fault — `--ag-spacing: 3px` shrinks rows correctly in SSRM, IRM and VPRM.

Warning #331 now reports the ignored variables and names the replacement for each. Limited to 9 variables whose Theming API equivalent is unambiguous; none appear anywhere in v36 source, so there are no false positives. Fires once per grid, at init and on non-theme styles changes.

Verified in a real browser on both the init path and the reporter’s class-toggle gesture. `ag-website-shared:lint` fails on this branch and on a clean `latest` alike (ESLint 10 / `globals` whitespace incompatibility), unrelated to this change.

Known limit: a compact-only toggle changes nothing measurable — the complaint itself — so the runtime warning may only surface once `large` is also tried. The init path, which is how migrated apps set these, is reliable.

### DIFF
--- a/packages/ag-grid-community/src/environment.ts
+++ b/packages/ag-grid-community/src/environment.ts
@@ -47,9 +47,6 @@ export class Environment
     >
     implements NamedBean
 {
-    /** Whether the Theming API is the active styling system, so legacy variables would be ignored. */
-    private themingApiActive = false;
-
     protected override initVariables(): void {
         this.addManagedPropertyListener('rowHeight', () => this.refreshRowHeightVariable());
         this.getSizeEl(ROW_HEIGHT);
@@ -119,9 +116,9 @@ export class Environment
         if (change === 'rowBorderWidth') {
             this.refreshRowBorderWidthVariable();
         }
-        // Catches variables a class swap introduces after grid creation. 'theme' is excluded
-        // because it fires before postProcessThemeChange updates themingApiActive.
-        if (change !== 'theme' && this.themingApiActive) {
+        // catches variables a class swap introduces after grid creation; 'theme' is covered by
+        // postProcessThemeChange, which runs after the new theme is in place
+        if (change !== 'theme') {
             this.checkLegacyThemeVariables();
         }
         super.fireStylesChangedEvent(change);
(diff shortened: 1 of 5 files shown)


packages/ag-grid-community/src/environment.ts
```diff
@@ -139,6 +165,38 @@ export class Environment
...
+    /**
+     * Warns about legacy theme variables set on the grid while the Theming API is active. The
+     * Theming API never reads them, so they are silently ignored - most visibly with
+     * `--ag-grid-size`, which applications set to change grid density and which does nothing.
+     * Only reported when the legacy stylesheet is absent, as loading it is already an error.
+     *
+     * Reported at most once per grid: every theme change re-runs this, and an application that
+     * animates a theme parameter would otherwise repeat the same warning on each frame.
+     */
+    private warnOnLegacyOnlyVariables(): void {
```
StephenCooper: If we are going to do this I would have it be part of the ValidationModule so that it can be tree shaken out for production builds.

packages/ag-grid-community/src/validation/validationService.ts
```diff
@@ -107,6 +109,25 @@ export class ValidationService extends BeanStub implements NamedBean {
...
+    /**
+     * Warns about legacy theme variables set on the grid while the Theming API is active. The
+     * Theming API never reads them, so they are silently ignored - most visibly with
+     * `--ag-grid-size`, which applications set to change grid density and which does nothing.
+     *
+     * Reported at most once per grid: every theme change re-runs this, and an application that
+     * animates a theme parameter would otherwise repeat the same warning on each frame.
+     */
+    public checkLegacyThemeVariables(eRootDiv: HTMLElement): void {
+        if (this.legacyVariablesReported) {
```
StephenCooper: You shouldn't need this flag. I would expect that `warn` already de-duplicates and won't log them again. 

Can we try and reduce the comments across the PR too.

packages/ag-grid-community/src/environment.ts
```diff
@@ -47,6 +47,9 @@ export class Environment
     >
     implements NamedBean
 {
+    /** Whether the Theming API is the active styling system, so legacy variables would be ignored. */
+    private themingApiActive = false;
```
StephenCooper: I am not a fan of this flag. Do we actually need it here. Can the checkLegacyThemeVariables method not inspect the theme gridOption for if its not legacy?

StephenCooper (approved):

PR #15075, opened 2026-09-03
AG-8950-advanced-set-filters

SalvatorePreviti:
# AG-8950: Set Filter support in the Advanced Filter

FIX AG-8950

A column configured with a Set Filter fell back to text filtering in the Advanced Filter: there was no way to filter on a list of distinct values. This adds `is any of` and `is none of` to every Set Filter column, driven by that column's own Set Filter rather than a parallel implementation of one.

| scope  | net lines |     lines changed | hunks |      files changed |
| ------ | --------: | ----------------: | ----: | -----------------: |
| source |     +1892 | 2498 (+2195 -303) |   204 | 34 (+6, 28 edited) |
| locale |      +128 |         128 added |    64 |          32 edited |
| tests  |     +4127 |   4137 (+4132 -5) |    20 |  15 (+8, 7 edited) |
| docs   |      +226 |         226 added |    10 |   6 (+3, 3 edited) |
| other  |       +25 |          25 added |     2 |           2 edited |

## Typing changes

`ColumnAdvancedFilterModel` gains `SetAdvancedFilterModel`, which filters on `values`. It declares `filter` / `filterTo` as `never`, so reading either off the union still yields `string | number | undefined`. A `switch` over `filterType` that asserts it has handled every member takes a `set` case.

## Grammar and model

```text
[Country] is any of ["Australia", "Italy"]
[Country] is none of ["Australia", "Italy"]
[Country] is any of ["Europe" › "Italy"]
```

```js
{ filterType: 'set', colId: 'country', type: 'isAnyOf', values: ['Australia', 'Italy'] }
```

Parentheses are read in place of the brackets, and quotes only where a value would otherwise read as something else; a blank is written `(Blanks)`. A Tree List value is a path, written `›` and also read as `>` or `/`; a bare value is read as a path first and otherwise as the text taken whole, so `[2024/01]` still names the value it looks like.

`values` holds Set Filter keys. Where `filterParams.valueFormatter` is set the expression shows and matches formatted values while the model keeps the keys, and where two keys format alike the first keeps that text and the rest are written as their key. A model applied before the values arrive writes its keys and is rewritten once they do.

The set options are added to whatever the cell data type already offers, so `filterParams.filterOptions` must name `isAnyOf` and `isNoneOf` to keep them.

## Behaviour

`AdvancedFilterSetService` drives a `SetFilterHandler` per column, so key creation, case folding, array cells, tree data and grouping paths are decided in one place and an expression matches exactly the rows the equivalent Set Filter selection matches. `keyCreator`, `valueFormatter`, `cellRenderer`, `treeList`, `treeListPathGetter`, `treeListFormatter`, `caseSensitive`, `filterParams.values`, `colDef.keyCreator` and `colDef.filterValueGetter` all apply without being re-declared.

Typing inside a value list suggests the column's values, in its own order and with its blank entry; no opening quote is needed, chosen values drop out and deleting one returns it. For a Tree List the list starts at the top-level groups with their remaining child counts, and typing searches the whole hierarchy, naming a match by the rest of its path with the groups above the leaf drawn back.

A value is judged only once written to a close, and one naming nothing reports `Value not found` and disables Apply. That fault alone does not invalidate the expression: an applied one goes on filtering when the data stops holding one of its values, unlike the Set Filter, which drops it. An expression is text the author wrote.

The Builder's value pill opens the column's Set Filter itself, showing what is chosen and shortening past three values. Retargeting the condition onto another column, or changing its option, drops `values` and `filterType` with it.

## AI toolkit schema

The feature schema groups a column's operators by operand kind rather than count, so a list option emits a schema declaring `values` and two Set Filter columns of different data types share one `setAdvancedFilterModel` definition.

## Value autocomplete cost at high cardinality

Measured over 20,000 rows with one distinct value per row, against the same column's Set Filter mini filter in the same run. Steady state: the list is built once and cached, and each figure is a mean per two keystrokes.

|                                  | Advanced Filter | Set Filter mini filter |
| -------------------------------- | --------------: | ---------------------: |
| keystroke matching every value   |          1.13ms |                 2.61ms |
| keystroke narrowing to a handful |          0.86ms |                 2.27ms |
| reopening the value list         |          0.56ms |                        |


StephenCooper (approved):
Happy to approve.

PR #15079, opened 2026-09-03
[TouchEvents] - Code Cleanup

gportela85:
Replaces the per-element TouchListener widget and the tooltip touch coordinator with a single delegated TouchGesturesService per grid, built on Pointer Events. Consumers register long-press interest with a priority (normal beats fallback, deepest element wins), and each registration declares the same eligibility its listener enforces, so a gesture is only claimed where it will actually be handled. An idle grid holds one document listener; move/up/cancel/contextmenu handlers attach per gesture, and the compatibility mouse events a handled long press produces (mousedown/mouseup/click) are suppressed wholesale until the next real interaction, preventing ghost clicks on menus that open under the finger.

**Behaviour preserved**
- Context menu and column menu take the long press over tooltips on all platforms — the iOS-only mock path is gone, and a native contextmenu during a touch gesture opens the grid menu exactly once.
- Long-press tooltips survive finger lift (touch pointerleave is ignored) and dismiss on tap-elsewhere.
- Double-tap-to-autosize on the resize grip and the iPad double-tap-to-edit emulation are reimplemented on the new service.
- suppressTouch is honoured per gesture, including runtime toggles.

**Breaking changes**

TouchListener (and its TapEvent/DoubleTapEvent/LongTapEvent/TouchListenerEvent/TouchListenerOptions types) and TouchShowContextMenuParam are no longer exported; EventShowContextMenuParams always carries a mouseEvent. The dead touchEvent parameters were removed through the context-menu API. Documented in the 36.2 upgrade guide; components inside the grid opt out of gestures by stopping pointerdown propagation.


packages/ag-grid-community/src/misc/touchGesturesService.ts
```diff
@@ -0,0 +1,386 @@
...
+            abandon(cancelled);
+            if (cancelled) {
+                return;
+            }
+            const now = Date.now();
+            if (now - lastTapTime <= DOUBLE_TAP_MILLISECONDS) {
+                lastTapTime = 0; // a triple tap must not count as two double taps
+                onDoubleTap(startEvent);
+            } else {
+                lastTapTime = now;
```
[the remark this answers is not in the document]
gportela85: fixed

packages/ag-grid-community/src/misc/touchGesturesService.ts
```diff
@@ -0,0 +1,399 @@
...
+        if (!gesture || !this.isSameTarget(gesture.startEvent.target, event.target)) {
+            return;
+        }
+
+        // the browser's own context menu never belongs over a grid long press
+        event.preventDefault();
+
+        const registration = this.resolveLongPressRegistration(gesture);
+        if (!registration) {
+            return;
```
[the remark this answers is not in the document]
gportela85: fixed

packages/ag-grid-community/src/misc/touchService.ts
```diff
@@ -76,43 +75,27 @@ export class TouchService extends BeanStub implements NamedBean {
     }
 
     public setupForHeader(comp: AgColumnHeader): void {
-        const { gos, sortSvc, menuSvc } = this.beans;
+        const { gos, menuSvc, touchGesturesSvc } = this.beans;
 
         if (gos.get('suppressTouch')) {
             return;
         }
```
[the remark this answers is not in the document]
gportela85: fixed

packages/ag-grid-community/src/widgets/touchListener.ts
```diff
@@ -1,361 +0,0 @@
...
-    shouldHandleLongTap?: () => boolean;
-    /** Cede the long press to any competing listener, owning it only when no other candidate can. */
-    yieldsLongTap?: boolean;
-    /** Whether this listener should observe this touch gesture. */
-    shouldTrackTouch?: (event: TouchEvent) => boolean;
-    /** Resolves the element represented by a delegated long-press listener. */
-    getLongTapTarget?: (event: TouchEvent, touchStart: Touch) => Element | undefined;
-}
-
-export class TouchListener implements IEventEmitter<TouchListenerEvent> {
```
StephenCooper, line 89: To avoid this being a breaking change we could just leave this in the codebase until v37? If we never reference it we should still get the bundle size savings but avoid the breaking change.

We can mark it deprecated and internal and also the same for all the interfaces?

packages/ag-grid-enterprise/src/agStack/agMenuItemComponent.ts
```diff
@@ -234,9 +234,6 @@ export class AgMenuItemComponent<
     private addListeners(eGui: HTMLElement, params?: IMenuConfigParams): void {
         if (!params?.suppressClick) {
             this.addManagedElementListeners(eGui, { click: (e) => this.onItemSelected(e!) });
-            const touchListener = new TouchListener(eGui, true);
```
StephenCooper, line 237: Does this not need a replacement because changes elsewhere mean instead of a Tap the click event is fired?

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelColumnGroupComp.ts
```diff
@@ -330,17 +342,13 @@ export class ToolPanelColumnGroupComp extends Component {
...
-        const { beans, eGroupClosedIcon, eGroupOpenedIcon, eColumnGroupIcons } = this;
+        const { beans, eGroupClosedIcon, eGroupOpenedIcon } = this;
         eGroupClosedIcon.appendChild(_createIcon('columnSelectClosed', beans, null));
         eGroupOpenedIcon.appendChild(_createIcon('columnSelectOpen', beans, null));
 
         const listener = this.onExpandOrContractClicked.bind(this);
         this.addManagedElementListeners(eGroupClosedIcon, { click: listener });
         this.addManagedElementListeners(eGroupOpenedIcon, { click: listener });
-
-        const touchListener = new TouchListener(eColumnGroupIcons, true);
```
StephenCooper, line 341: Same here, assuming click is fired instead now?

packages/ag-stack/src/tooltip/baseTooltipStateManager.ts
```diff
@@ -68,7 +67,7 @@ export abstract class BaseTooltipStateManager<
 
     private state = TooltipStates.NOTHING;
```
StephenCooper, line 69: We need to loop in the Studio team about the potential impact they might see.

documentation/ag-grid-docs/src/content/docs/component-cell-renderer/index.mdoc
```diff
@@ -323,7 +323,7 @@ By default when a cell is clicked on, the grid will perform actions including:
 
 - Starting editing, if [Editing](./cell-editing/) is enabled.
 
-This behaviour may not be desirable for custom cell components, e.g. if they contain interactive elements. In this situation, the grid can be prevented from handling the mouse event (`'click'`, `'dblclick'`, `'mousedown'` or `'touchstart'`), by passing the `suppressMouseEventHandling` callback to `cellRendererParams`.
+This behaviour may not be desirable for custom cell components, e.g. if they contain interactive elements. In this situation, the grid can be prevented from handling the pointer event (`'click'`, `'dblclick'` or `'pointerdown'`), by passing the `suppressMouseEventHandling` callback to `cellRendererParams`.
```
StephenCooper, line 326: This sounds like a breaking change if applications have done this.

StephenCooper (changes requested):
We need to discuss these changes futher.

packages/ag-grid-enterprise/src/agStack/agMenuItemComponent.ts
gportela85: We don't need a replacement here. `TouchListener(..., true)` prevented the browser-generated click. Now the tap callback is manually called onItemSelected() instead.
Without that `touchend.preventDefault()`, modern touch browsers generate the normal click, handled at by the click listener above.

packages/ag-grid-enterprise/src/columnToolPanel/toolPanelColumnGroupComp.ts
gportela85: same as the case above

packages/ag-stack/src/tooltip/baseTooltipStateManager.ts
gportela85: Yes, we should probably involve @peterjrreynolds in reviewing this.

documentation/ag-grid-docs/src/content/docs/component-cell-renderer/index.mdoc
gportela85: Well, not really, this PR did not introduce this runtime change. The preference for pointerdown was introduced in commit `58363a53e3d` and shipped in 35.1.0. This PR only corrects the stale documentation and JSDoc.

Something @SalvatorePreviti can double check?

packages/ag-stack/src/tooltip/baseTooltipStateManager.ts
peterjrreynolds: We don't override any of the tooltip stuff, but we will need to test the interactions

PR #15092, opened 2026-09-04
AG-18371: Privacy policy copy updates (#15038)

markdurrant:
* AG-18371 [Website] - Updated form consent copy for the new privacy policy

Replaces the data-processing consent checkbox with a standing processing notice above the submit button, and revises the marketing consent wording to cover both AG Grid and Bryntum in a single combined checkbox.

Processing is no longer consent-based, so the data-processing field is dropped from both payloads rather than being auto-filled: the Salesforce web-to-lead post omits its consent field, and the trial-licence API payload omits dataProcessingConsent. Both need a staging submit before release in case either treats the field as required.

Also realigns the Mailchimp newsletter form on the Beyond the Prompt page, which carried its own inline copy of the same two strings.

The privacy policy page copy itself is not in this commit — it depends on the document attached to the ticket.

* AG-18371 [Website] - Updated the privacy policy page to v3

Replaces the policy body, effective-date block and intro copy with the v3 document, transcribed verbatim and verified word-for-word against the source.

The body grows from 11 to 12 numbered sections: the old combined section 2 is split into "What Data Do We Collect About You" and a new "Purposes Of Processing And Lawful Bases" that sets out each purpose against its lawful basis. New copy covers the Bryntum joint-controller relationship, third-party firmographic data providers, and expanded disclosure, transfer and data-subject-right lists. The contact address becomes privacy@ag-grid.com, and the date line becomes "Last Updated".

Section 2 keeps the data-what-why anchor so existing inbound deep links still resolve, and the page nav gains the new section.

Two deviations from the source document, both flagged for review: the date is the placeholder "[DATE]" in the source, and one instance of "Brytum" is transcribed as "Bryntum".

* AG-18371 [Website] - Narrowed the consent label map to its known keys

The map is only ever read through literal property names, so `Record<string, string>` was wide enough to let a mistyped key compile and render an empty consent label. `satisfies` keeps the value constraint while narrowing the keys, matching the pattern already used for POLICY_CONTENT.

* AG-18371 [Website] - Revised the marketing consent wording

Opens with the opt-in rather than a first-person statement, and trims the unsubscribe sentence. Applied to the shared consent label and to the Mailchimp newsletter form, which carries its own copy of the same string.

### DIFF
--- a/external/ag-website-shared/src/components/contact-form/ContactForm.tsx
+++ b/external/ag-website-shared/src/components/contact-form/ContactForm.tsx
@@ -1,14 +1,6 @@
 import { ConsentCheckbox } from '@ag-website-shared/components/consent-fields/ConsentCheckbox';
-<<<<<<< HEAD
-import {
-    CONSENT_LABELS,
-    DATA_PROCESSING_CONSENT_REQUIRED,
-} from '@ag-website-shared/components/consent-fields/consentMessages';
-=======
 import { ProcessingNotice } from '@ag-website-shared/components/consent-fields/ProcessingNotice';
 import { CONSENT_LABELS } from '@ag-website-shared/components/consent-fields/consentMessages';
-import type { CaptchaTicker } from '@ag-website-shared/components/contact-form/initCaptcha';
->>>>>>> fdf48f00bd4 (AG-18371: Privacy policy copy updates (#15038))
 import { initCaptcha } from '@ag-website-shared/components/contact-form/initCaptcha';
 import { Icon } from '@ag-website-shared/components/icon/Icon';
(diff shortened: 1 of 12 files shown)


external/ag-website-shared/src/components/contact-form/ContactForm.tsx
```diff
@@ -1,8 +1,14 @@
 import { ConsentCheckbox } from '@ag-website-shared/components/consent-fields/ConsentCheckbox';
+<<<<<<< HEAD
```
taktran: This is a merge mistake

taktran (approved):
LGTM

PR #15107, opened 2026-09-07
AG-17659 fix ref data editing

SalvatorePreviti:
FIX AG-17659

# Cell editing: an untouched editor no longer writes

Opening a cell editor, typing nothing and committing could change the underlying value. The editor is seeded with
one representation of the value and commits another, while the grid decides whether anything changed by comparing
the committed value against the original. That comparison is only a valid proxy while the two representations
agree.

| scope  | net lines |  lines changed | hunks | files changed |
| ------ | --------: | -------------: | ----: | ------------: |
| source |       +66 | 176 (+121 -55) |    54 |     12 edited |
| tests  |      +695 |  697 (+696 -1) |     7 |      4 edited |

## `refData` no longer formats the editor implicitly ([AG-17659](https://ag-grid.atlassian.net/browse/AG-17659))

`TextCellEditorInput.getStartValue` seeded the editor with the mapped label whenever a column carried `refData`,
whether or not the application asked for it. Nothing reverses that mapping on commit, so the label was written
back into the data as though it were a key: a double-click followed by a click elsewhere turned `bw` into
`Burlywood`, the cell then rendered blank, and a Set Filter listed one empty entry per corrupted key.

Only an explicit `useFormatter: true` now seeds the formatted value. That is the documented mechanism for it, and
the one an application pairs with a `valueParser`. The reference data documentation already describes this
behaviour: with a Text Cell Editor and `refData`, codes must be entered directly.

**Breaking:** a column combining `refData`, a text editor and a `valueParser` that maps labels back to keys, but
without `useFormatter: true`, was relying on the implicit formatting. Such a column must now set
`useFormatter: true` to keep the editor showing labels.

## An untouched editor commits its original value

`SimpleCellEditor` records the text the widget holds once seeded, and `getValue()` returns the original value
untouched while the input still holds it. The seed is read back off the widget rather than taken from
`getStartValue()`, because the widget normalises what it is given.

That normalisation is load-bearing. `agNumberCellEditor` with `cellEditorParams: { precision: 2 }` truncated
`5.6789` to `5.67` when an editor was opened and committed without being touched, and dispatched no
`cellValueChanged`, so nothing could observe it. `agDateCellEditor` reported a change on every untouched commit,
because it builds a fresh `Date` and the comparison is by identity.

A `valueParser` no longer runs when the input text still matches the seed, so a parser used to normalise a value
on every commit will not fire for an edit that changed nothing.

## The remaining provided editors carry the same guard

Five editors do not extend `SimpleCellEditor` and each wrote a different value when an editor was opened and
closed without being touched. None of them dispatched `cellValueChanged`, so nothing could observe the write.

- `agLargeTextCellEditor` and the enterprise formula cell editor ran the column `valueParser` over the stored
  value. The formula editor still commits formula text verbatim, which never went through that parser.
- `agCheckboxCellEditor` read the value back off a widget that cannot hold `null`, so a stored `null` was
  narrowed to `undefined`.
- `agSelectCellEditor` shows `values[0]` when the stored value is not one of the options, which is a display
  fallback; committing an untouched editor wrote that option over the stored value.
- `agRichSelectCellEditor` ran a supplied `cellEditorParams.parseValue` over the stored value.

`agSetEditValue` is the `setDataValue(..., 'edit')` path, so it receives a value that is already a model value.
Every provided editor now commits it verbatim; the formula cell editor previously put it through the column
`valueParser`, which parsed an already-parsed value. A value pushed into an open `agRichSelectCellEditor` also
keeps `undefined` rather than committing the `null` its widget stores.

## Custom validation grades the value that will be committed

A `cellEditorParams.getValidationErrors` callback is handed the value a commit would write. Each editor derived
that value from its own live reading of the widget, so on an untouched editor the callback graded something the
commit would not keep: with a `valueParser` that uppercases, `abc` was validated as `ABC`; a number column with
`precision: 2` validated `5.67` while `5.6789` was preserved; a date string column validated its reparsed result.

The number and date editors' own `min` and `max` checks continue to read the live widget, because those grade
what is being typed rather than what would be committed.

`agNumberCellEditor` reads its widget with `getValue(true)` when deriving the committed value. The widget
withholds a value while the input is invalid, and raising a validation error is itself what marks it invalid, so
without that flag the callback graded `undefined` from the first error onwards.

## `min` and `max` are enforced on a date editor using `includeTime`

`includeTime` switches the editor's input to `datetime-local`, where `valueAsDate` is defined to return `null`.
`agDateCellEditor` read that property and gated its whole `min`/`max` block on the result being a `Date`, so both
options were silently unenforced and a custom `getValidationErrors` callback received `null` for the value. The
editor now parses the input's value string, which is what `agDateStringCellEditor` already did.

### DIFF
--- a/packages/ag-grid-community/src/edit/cellEditors/checkboxCellEditor.ts
+++ b/packages/ag-grid-community/src/edit/cellEditors/checkboxCellEditor.ts
@@ -23,6 +23,8 @@ export class CheckboxCellEditor extends AgAbstractCellEditor<ICellEditorParams<a
     }
 
     protected readonly eEditor: GridCheckbox = RefPlaceholder;
+    /** Read off the widget, which cannot hold `null`, so an untouched commit would narrow it to `undefined`. */
+    private seededValue: boolean | undefined | this = this;
 
     public initialiseEditor(params: ICellEditorParams<any, boolean>): void {
         this.agSetEditValue(params.value);
@@ -39,11 +41,13 @@ export class CheckboxCellEditor extends AgAbstractCellEditor<ICellEditorParams<a
         this.params.value = value;
         const isSelected = value ?? undefined;
         this.eEditor.setValue(isSelected, true);
+        this.seededValue = this.eEditor.getValue();
         this.setAriaLabel(isSelected);
     }
 
-    public getValue(): boolean | undefined {
-        return this.eEditor.getValue();
+    public getValue(): boolean | null | undefined {
+        const value = this.eEditor.getValue();
+        return value === this.seededValue ? this.params.value : value;
     }
 
     public focusIn(): void {
--- a/packages/ag-grid-community/src/edit/cellEditors/dateCellEditor.ts
+++ b/packages/ag-grid-community/src/edit/cellEditors/dateCellEditor.ts
@@ -1,5 +1,5 @@
 import type { LocaleTextFunc } from 'ag-stack';
-import { _exists, _serialiseDate } from 'ag-stack';
+import { _exists, _parseDateTimeFromString, _serialiseDate } from 'ag-stack';
 
 import { AgInputDateFieldSelector } from '../../agWidgets/agInputDateField';
 import type { DataTypeService } from '../../columns/dataTypeService';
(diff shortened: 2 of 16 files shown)


packages/ag-grid-community/src/edit/cellEditors/checkboxCellEditor.ts
```diff
@@ -23,6 +23,8 @@ export class CheckboxCellEditor extends AgAbstractCellEditor<ICellEditorParams<a
     }
 
     protected readonly eEditor: GridCheckbox = RefPlaceholder;
+    /** Read off the widget, which cannot hold `null`, so an untouched commit would narrow it to `undefined`. */
+    private seededValue: boolean | undefined | this = this;
```
StephenCooper, line 27: So assigning to `this` is just for Typescript? Or does it serve some other purpose?

StephenCooper (approved):

packages/ag-grid-community/src/edit/cellEditors/checkboxCellEditor.ts
SalvatorePreviti: I am using "this" as a special indicator of "not seeded" so I can compare === this means not seeded - is definitely a trick to avoid adding a new symbol for example, and is not the first time or place we use it, we used in another few places.

PR #15110, opened 2026-09-07
[AG-4093] Feature(column-sizing): make scroll-driven continuous auto-size opt-in

sdwvit:
Scrolling is now its own category of continuous auto-size trigger, excluded by default.

`fitCellContents` registers the scroll triggers only when `shouldAutoSizeColumns` is provided, so a grid without the callback never re-sizes mid-scroll. The reason keeps its public name, `'viewportChanged'` — return `true` for it to opt in.

Fix [AG-4093](https://ag-grid.atlassian.net/browse/AG-4093)


documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -205,17 +205,19 @@ const gridOptions = {
...
-- The width-distribution strategies, `fitGridWidth` and `fitProvidedWidth`, are arithmetic over the current column set, so scrolling cannot change their result. They re-run on new data, a page change, and a scrollbar appearing or disappearing — the changes to the width there is to share out.
+- `fitCellContents` re-runs on any change to the row data, because that changes what there is to measure.
+- The width-distribution strategies, `fitGridWidth` and `fitProvidedWidth`, are arithmetic over the current column set. They re-run on new data, a page change, and a scrollbar appearing or disappearing — the changes to the width there is to share out.
+
+Scrolling is a category of its own, reported as `'viewportChanged'`, and is excluded by default. Scrolling brings new cell content into view, so it can change what `fitCellContents` measures, but re-fitting columns under the pointer mid-scroll is rarely wanted. To opt in, provide `shouldAutoSizeColumns` and return `true` for that reason; without the callback the grid never re-sizes on a scroll. The width-distribution strategies ignore scrolling entirely, as it cannot change their result.
```
StephenCooper: I think these can be made clearer. Focus on what the user needs to know. Currently it feels a bit low level.

sdwvit: updated, ty

StephenCooper (approved):

documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -205,19 +205,23 @@ const gridOptions = {
         type: 'fitCellContents',
         continuous: true,
         shouldAutoSizeColumns: (params) => {
-            // skip re-sizing while the user is scrolling through the data
-            return params.reason !== 'viewportChanged';
+            // opt into re-sizing as the user scrolls, but only on the first page
+            return params.reason !== 'viewportChanged' || params.api.paginationGetCurrentPage() === 0;
```
StephenCooper: This snippet is actually not the simplest to read. Also does it represent a real use case. As this is the only code snippet lets aim to make it simple and match a real use case.

documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -205,19 +205,23 @@ const gridOptions = {
...
 
-- `fitCellContents` re-runs on any change to the row data, and on viewport changes, because both change what there is to measure — scrolling is what brings new cell content into view.
-- The width-distribution strategies, `fitGridWidth` and `fitProvidedWidth`, are arithmetic over the current column set, so scrolling cannot change their result. They re-run on new data, a page change, and a scrollbar appearing or disappearing — the changes to the width there is to share out.
+- `fitCellContents` updates as row data or displayed columns change.
+- `fitGridWidth` and `fitProvidedWidth` update when the available grid width or displayed columns change, including changes caused by pagination or scrollbars.
 
-Triggers arriving in the same frame are coalesced into a single re-size. Viewport and grid-size changes are debounced on top of that, so scrolling and resizing the grid re-size once the gesture settles rather than on every frame.
+Scrolling does not resize columns by default. When using `fitCellContents`, you can opt into resizing as new rows are scrolled into view by providing `shouldAutoSizeColumns` and allowing the `'viewportChanged'` reason.
```
StephenCooper: We need to be clear that scrolling here is vertical scrolling. 

Horizontal scrolling will cause columns to resize by default. I believe this is called out in the new example that you have added in a different PR

StephenCooper (changes requested):

documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
sdwvit: ty, addressed

documentation/ag-grid-docs/src/content/docs/column-sizing/index.mdoc
```diff
@@ -204,20 +204,22 @@ const gridOptions = {
     autoSizeStrategy: {
         type: 'fitCellContents',
         continuous: true,
-        shouldAutoSizeColumns: (params) => {
-            // skip re-sizing while the user is scrolling through the data
-            return params.reason !== 'viewportChanged';
-        }
+        // re-size as the user scrolls, as well as on data and column changes
```
sdwvit: this snipped is before the explanation why this is needed.

sdwvit: maybe it should show how to use the cb.

StephenCooper (approved):

PR #15131, opened 2026-09-08
AG-8950 qa-fixes

SalvatorePreviti:
# AG-8950: Advanced Filter set operators, tree lists, list brackets and picker width

FIX AG-8950

Four defects QA raised against Set Filter support in the Advanced Filter, plus `enableSetOperators`, which
decides which columns offer the two set options, and a suggestion list that showed one column's options for
another.

| scope  | net lines |     lines changed | hunks |     files changed |
| ------ | --------: | ----------------: | ----: | ----------------: |
| source |       +73 |   799 (+436 -363) |   150 |         26 edited |
| locale |       +64 |       78 (+71 -7) |    32 |         32 edited |
| tests  |      +913 | 1269 (+1091 -178) |   119 | 10 (-1, 9 edited) |
| docs   |       +45 |     109 (+77 -32) |    26 |         10 edited |
| other  |        -9 |         9 removed |     1 |          1 edited |

## `is any of` is offered where the column asks for it ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950))

`filterParams.enableSetOperators` decides which columns offer the two options, whatever filter they have. It
joins `IProvidedFilterParams`, so every provided filter's column can carry it, and it is inherited through
`defaultColDef`. Unset, a column falls back to what its own filter implies, which is what a column with a Set
Filter did before.

| `SetFilter` module | `colDef.filter`          | `enableSetOperators` | offers them | `filterParams` are a Set Filter's |
| ------------------ | ------------------------ | -------------------- | :---------: | :-------------------------------: |
| registered         | `'agSetColumnFilter'`    | unset                |     yes     |                yes                |
| registered         | `'agSetColumnFilter'`    | `false`              |     no      |                yes                |
| registered         | `true`                   | unset                |     yes     |                yes                |
| registered         | `true`                   | `false`              |     no      |                yes                |
| registered         | `'agNumberColumnFilter'` | unset                |     no      |                no                 |
| registered         | `'agNumberColumnFilter'` | `true`               |     yes     |                no                 |
| registered         | `'agMultiColumnFilter'`  | unset                |     no      |                no                 |
| registered         | `'agMultiColumnFilter'`  | `true`               |     yes     |     its Set Filter child's        |
| not registered     | any                      | any                  |     no      |                no                 |

The last two columns are the two questions one predicate had been answering together, now `offersSetOperators`
and `hasSetFilter`. The second is what says a `comparator` orders a value list rather than comparing dates, and
a `textFormatter` formats that list rather than a comparison. A `filterOptions` list naming `isAnyOf` or
`isNoneOf` offers them whatever `enableSetOperators` says, being the more specific statement of what a column
offers. `suppressSetFilterByDefault: true` moves the `filter: true` rows onto the `'agNumberColumnFilter'` ones.
The UMD bundle registers every module, so a UMD grid is always on the `registered` rows.

A column offering them without having a Set Filter keeps its own `filterParams` to itself: those were written
for the filter it has, and a Date Filter's `comparator` reaching a value list is called with two cell values.
A Multi Filter is the exception, its Set Filter child's `filterParams` being a value list's: `values`,
`valueFormatter`, `treeList` and `cellRenderer` reach the list from there. `enableSetOperators` joins
`IMultiFilterParams` as well as `IProvidedFilterParams`, so a Multi Filter column can name it.

## An option only the Advanced Filter has stays out of the column filter's dropdown ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950))

One `filterOptions` list is read by both filters, so a column naming `isAnyOf` for the Advanced Filter was
offering it in its own dropdown as well, under its bare key and matching nothing. `isAnyOf`, `isNoneOf`, `true`
and `false` are now withheld from the column filter, which has no operator for any of them: the boolean pair
reaches it only as the definitions its Cell Data Type supplies. The exclusion applies to a bare key alone, so a
Custom Filter Option under one of those `displayKey`s is offered and evaluated like any other.

Those four keys are also assignable to `filterOptions` now, having been rejected by a type that only named the
column filter's own options. The set is derived, `Exclude<SetAdvancedFilterModelType | BooleanAdvancedFilterModelType,
ISimpleFilterModelType>`, so neither reader can gain an option the other never hears of. It is exported as
`AdvancedFilterOnlyOptionKey`.

## The suggestion list follows the column, not just its data type ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950))

Two columns of one Cell Data Type offering different options showed whichever was asked for first. The list is
rebuilt when its token changes and the token named only the data type, so moving between two text columns
re-searched the open list instead of rebuilding it. Reachable with `filterParams.filterOptions` alone.

Writing a chosen option no longer reads its Cell Data Type back out of that token. The token names the open
popup's list while the column it is paired with comes from the current parse, so how many values an option takes
was being decided from two sources at once; both now come from the parse.

## A Set Filter value list must be enclosed in square brackets ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950)) TC18

`[Country] is any of "Armenia", "Azerbaijan"` parsed as a two value list and applied. Only
`[Country] is any of ["Armenia", "Azerbaijan"]` does now; anything else reports `Missing opening square bracket`
and applies nothing. Parentheses are no longer read as list brackets either, which is what stopped
`is any of (Blanks)` reading the blank's own label as a list holding `Blanks`.

The list's brackets get two locale keys of their own, `advancedFilterValidationMissingListStartBracket` and
`advancedFilterValidationMissingListEndBracket`. `MissingEndBracket` had been reported for both the list `]` and
the round `)` of a group or a range, a distinction no language with separate words for the two shapes can
translate; it and `ExtraEndBracket` are now always the round one. That made four translations correctable:
`fr-FR` said _quotation mark_, `el-GR` used the word for a mounting bracket, `pt-BR` said _square_ for a
round-only message, and `hu-HU` was plural for one bracket.

## A tree list is offered flat, and a path is one value ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950)) TC8

The autocomplete offered one level of the hierarchy at a time and wrote each segment as a value of its own,
`["2018" › "12" › "15"]`. It now offers every path in the column as one flat list, parent segments
de-emphasised, and writes the whole path as one value, `["2018 > 12 > 15"]`. Groups are no longer offered, since
a group is not a value. Choosing the option writes `[` alone rather than `["`, so a value can be searched for
without typing an opening quote.

A path is written with `>` and drawn with `›`, so the text an author reads back is one they can retype. Both are
read as separators, inside quotes as well as outside; `/` is not, so a value spelling itself `2024/01` names that
value. The segment-by-segment spelling `["Europe" > "Italy"]` is still read, and is what a segment containing a
separator of its own needs.

Where the column supplies a `filterParams.cellRenderer`, the renderer draws the leaf and the parent segments are
drawn beside it: offered flat, several leaves share a label.

This removes the drill-down machinery: per-level children, group child counts, the per-keystroke relative-path
flattening, and the group-count row with its CSS in both the Theming API and the legacy themes.

## One writer decides how a chosen value is spelled ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950))

The autocomplete wrote a chosen value itself while a stored model went through the path writer, so the two could
disagree: choosing `Slash / Group › 2024/01` wrote text that reads back as a four segment path naming nothing.
Both now go through the same writer, and a chosen entry carries the path it stands for rather than being looked
up by the text its row displays, which two values can share.

## Retargeting an option puts the caret inside the operand it takes ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950)) TC3

Choosing an option over a condition that already had an operand left the caret before the opening bracket, so
the next value chosen landed outside the list: `is any of ["Afghanistan"]` retargeted and then added to gave
`is any of "Algeria", ["Afghanistan"]`. The caret now lands inside the operand, and one sitting at the very start
of a written value inserts before it rather than overwriting it.

An option that takes a bracket of its own opens one only over a different bracket, so a range's `(` no longer
reads as a list already there. An option taking a bare value opens no quote over an operand already beginning
with a bracket or a quote.

## A Builder picker no longer grows on every redraw ([AG-8950](https://ag-grid.atlassian.net/browse/AG-8950)) TC10

The Builder's option picker widened a step at a time until it reached the dialog width. A row is laid out
`min-width: 100%`, so once the picker had grown, the width its rows wanted came back as the picker's own, and
each pass added the border and scrollbar allowance again. The list now reports only what overflows the space it
has, measured against the container the rows stretch to rather than the integer-rounded `clientWidth`, whose
rounding is why this showed up at fractional device pixel ratios.


documentation/ag-grid-docs/src/content/docs/filter-advanced/index.mdoc
```diff
@@ -358,7 +381,7 @@ A `displayKey` is resolved against the column being filtered, so different colum
...
 
-On a Set Filter column, `is any of` and `is none of` come from the column's filter rather than from its Cell Data Type, so they join the standard options for that Cell Data Type. A `filterOptions` list replaces those options in full, so it must name `isAnyOf` and `isNoneOf` for the column to keep offering them.
+On a Set Filter column, `is any of` and `is none of` come from the column's filter rather than from its Cell Data Type, so they join the standard options for that Cell Data Type. A `filterOptions` list replaces those options in full, so it must name `isAnyOf` and `isNoneOf` for the column to keep offering them. Naming them is also what offers them on a column with any other filter, which then resolves its values as [Set Filters](#set-filters) describes while its own filter is unchanged.
```
StephenCooper: Seems like this is missing the flag. (Just incase we keep this approach)

packages/ag-grid-enterprise/src/advancedFilter/set/advancedFilterSetService.ts
```diff
@@ -163,15 +165,28 @@ export class AdvancedFilterSetService extends BeanStub<'valuesChanged'> implemen
...
-        // Read from the definition alone, so asking does not instantiate anything.
+        // An option list naming them is the most specific statement there is, so it outranks the flag.
+        if (namesSetOperator(column)) {
+            return true;
+        }
+        // What the column asks for, then what its own filter implies.
+        return column.colDef.filterParams?.enableSetOperators ?? this.isSetFilterDef(column);
+    }
+
+    /** Whether the column's own filter is a Set Filter, so its `filterParams` are a list's and not a comparison's. */
```
StephenCooper, line 181: Is this where we need support for a set filter within a multi filter or within a agSelectableColumnFilter

documentation/ag-grid-docs/src/content/docs/filter-advanced/index.mdoc
SalvatorePreviti: added — it now links enableSetOperators and notes that a list naming the options outranks it.

packages/ag-grid-enterprise/src/advancedFilter/set/advancedFilterSetService.ts
SalvatorePreviti: Multi Filter is handled (the value list comes off its Set Filter child); agSelectableColumnFilter can't reach here, as the Advanced Filter refuses the Filters Tool Panel (warning 213).

StephenCooper (approved):

PR #15133, opened 2026-09-08
AG-3390 Mark shared Icon svgs as decorative for accessibility

taktran:
## Summary

Lighthouse's Agentic Browsing check flags **Accessibility tree is not well-formed** on https://www.ag-grid.com/react-data-grid/getting-started/. The single failing node is the inline svg inside the example runner's CodeSandbox button (`OpenInCTA`): `<svg role="img">` with no alternative text.

The wrapping button already has an `aria-label`, but the svg file itself declared `role="img"` with an empty `<title/>`, so axe treats it as an unnamed image.

## Changes

- `Icon.tsx` now renders every icon with `aria-hidden="true"` and `focusable="false"`. The component never provides an accessible name of its own (the surrounding button/link/text does), so all icons are decorative. Carbon icons already did this; the custom `?react` svgs did not.
- Removed the bare `role="img"` (and empty `<title/>`) from `codesandbox.svg` and `figma-logo.svg`, the only two inline svgs that declared it.

## Verification

- Prettier and ESLint pass on the changed file.
- Note: `ag-charts` shares `ag-website-shared`, so the same change likely needs porting there.


markdurrant (approved):
LGTM 👍

skewis6 (approved):

PR #15150, opened 2026-09-09
AG-16831 Improve the most frequently visited error pages

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-16831

Fix [AG-16831](https://ag-grid.atlassian.net/browse/AG-16831)

### DIFF
--- a/documentation/ag-grid-docs/src/content/errors/101.mdoc
+++ b/documentation/ag-grid-docs/src/content/errors/101.mdoc
@@ -15,6 +15,10 @@ The grid tells two cases apart here. Naming a built-in component whose module is
 
 Names are resolved from the `components` grid option. The name in the map must match the string on the column exactly.
 
+```js
+import { StatusRenderer } from './StatusRenderer';
+```
+
 ```{% frameworkTransform=true spaceBetweenProperties=true %}
 const gridOptions = {
     components: {
--- a/documentation/ag-grid-docs/src/content/errors/101.mdoc
+++ b/documentation/ag-grid-docs/src/content/errors/101.mdoc
@@ -1,0 +1,49 @@
+---
+description: A component was referenced by name, but that name is not in the grid's registered components.
+---
+
+## Component Name Not Registered
+
+A column or grid option asked for a component by name — {% errorParam name="propertyName" fallback="cellRenderer" /%} was set to the string {% errorParam name="componentName" fallback="StatusRenderer" /%} — and the grid could not find that name when it came to render. The cell falls back to the grid's default component.
+
+This message means the name is neither one of the grid's own components nor one you have registered. It is worth knowing that the grid tells those two cases apart: if you name a _built-in_ component whose module is not registered, you get [error #200](./errors/200) naming the module instead. So arriving here means the grid believes this is your component.
+
+That leaves two causes, and the message's `Did you mean` suggestions usually settle which:
+
+- **A typo, or a mismatch in case.** Names are matched exactly, so `statusRenderer` and `StatusRenderer` are different components.
+- **The component was never registered**, or was registered on a different grid than the one rendering this column.
+
+### Register it by name
+
+Names are resolved from the `components` grid option, which maps each name to its implementation. The name in the map must match the string on the column exactly.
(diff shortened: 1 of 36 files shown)


documentation/ag-grid-docs/src/content/errors/101.mdoc
```diff
@@ -0,0 +1,51 @@
...
+## Component Name Not Registered
+
+A column or grid option asked for a component by name, setting {% errorParam name="propertyName" fallback="cellRenderer" /%} to the string {% errorParam name="componentName" fallback="StatusRenderer" /%}, and the grid could not find that name when it came to render. The cell falls back to the grid's default component.
+
+The grid tells two cases apart here. Naming a built-in component whose module is not registered raises [error #200](./errors/200) with the module to register, so arriving at #101 means the grid believes this is your own component. That leaves two causes, and the message's `Did you mean` suggestions usually settle which:
+
+- A typo, or a mismatch in case. Names are matched exactly, so `statusRenderer` and `StatusRenderer` are different components.
+- The component was never registered, or was registered on a different grid than the one rendering this column.
+
+### Register it by name
```
StephenCooper, line 14: Vue is an exception here. It does not support by name.  

https://www.ag-grid.com/vue-data-grid/components/#registering-custom-components

documentation/ag-grid-docs/src/content/errors/101.mdoc
```diff
@@ -0,0 +1,51 @@
+---
+description: A component was referenced by name, but that name is not in the grid's registered components.
+---
+
+## Component Name Not Registered
+
+A column or grid option asked for a component by name, setting {% errorParam name="propertyName" fallback="cellRenderer" /%} to the string {% errorParam name="componentName" fallback="StatusRenderer" /%}, and the grid could not find that name when it came to render. The cell falls back to the grid's default component.
```
StephenCooper: Can we change the name of StatusRenderer to something easier to note its a placeholder such as. `MyCellRenderer`

documentation/ag-grid-docs/src/content/errors/240.mdoc
```diff
@@ -0,0 +1,42 @@
+---
+description: The theme grid option was given something other than a Theming API theme object or the string 'legacy'.
+---
+
+## Invalid Theme Grid Option
+
+The `theme` grid option accepts a Theming API theme object, or the string `legacy`. It was given {% errorParam name="theme" fallback="ag-theme-alpine" /%}, which is neither, so the grid cannot apply it.
```
StephenCooper, line 7: When are these fallbacks visible? Is it likely that a user will actually see them and then get confused that it might be mentioning something that they have not added?

documentation/ag-grid-docs/src/content/errors/101.mdoc
alistairjcbrown: I've split the section: JS/React/Angular keep the `components` grid option, and Vue now shows `app.component` / the local `components` option / `defineExpose`

documentation/ag-grid-docs/src/content/errors/240.mdoc
alistairjcbrown: The fallback ships in the static HTML and is replaced client-side only when the URL actually carries that parameter, so it's visible in two situations:
* permanently shown when the parameter is missing (a hand-written or truncated link, or JS blocked)
* and for a quick flash before hydration when the parameter is present

> Is it likely that a user will actually see them and then get confused

The missing-parameter warning appears immediately above the guidance. My hope was that it was better to show something here, even if it can't slot in the specific, than to hide the error entirely.

❓ What do you think?

## Example

### Proposed by this branch
image

### In Production currently
image

StephenCooper (approved):

documentation/ag-grid-docs/src/content/errors/240.mdoc
StephenCooper: That is am improvement so happy to proceed.

StephenCooper (approved):

PR #15172, opened 2026-09-10
AG-18472 Remove the column delay render reveal failsafe

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-18472

Fix [AG-18472](https://ag-grid.atlassian.net/browse/AG-18472)


packages/ag-grid-community/src/validation/errorMessages/errorText.ts
```diff
@@ -977,7 +975,7 @@ export type ErrorId = keyof ErrorMap;
  *
  * @knipIgnore Read by the docs site's error-page route
  */
-export const MAX_ERROR_ID = 334;
+export const MAX_ERROR_ID = 333;
```
StephenCooper, line 978: Safe to drop the error id as this error was not publicly released yet.

PR #15188, opened 2026-09-11
AG-18512 Fix the failing nightly Documentation Tests

sdwvit:
Fixes the five specs that went red in [the nightly run](https://github.com/ag-grid/ag-grid/actions/runs/34555731841).

**Example bug (user-visible):** `continuous-auto-size-fit-grid-width`'s Add/Remove Column and Add/Remove Rows buttons throw `gridApi is not defined` in React, Angular and Vue on the live docs. The generator lifts only the functions the HTML calls into the component; these delegated to a helper that stayed at module scope. Handlers now call the api directly.

**Test fixes:**
- Both continuous auto-size specs sampled a width until it changed, which can read mid-animation or between debounced passes. They now wait for the grid's finished `columnResized` event plus the Web Animations transition.
- The rows test counted rendered `.ag-row`s, which saturates under virtualisation; it measures the scrollable row extent instead.
- The two server-side specs pinned a `row-index` captured before the block resolved. Re-read inside the retry.
- `example-demos` waited for `load` on the heaviest demo and blew the timeout in Firefox; now `domcontentloaded` + `document.fonts.ready`.
- The Advanced Filter set-filters layout tests asked for `[Sport] is any of`, but the docs split gave the Set Filter to Athlete and left Sport on a Text Filter. The "grows" case also ran at 570px, where the row overflows and the pill is pinned to its 152px floor.

Adds `armGridEvent`/`withGridEvent` for waiting on an event armed before the action that triggers it; it listens via `updateGridOptions` since `api.addEventListener` needs `EventApiModule`.

Verified locally: `./docs-e2e.sh -g "continuous-auto-size|set-filters|block-load-debounce|cache-configurations|Demo page"` — 112 passed, all frameworks, chromium.

**Not fixed here:** `filtering/provided-filters` → "Set Filter restricts the Country column to a single selection" fails on webkit + reactFunctionalTs only (all three retries). The mini-filtered item never appears and I couldn't establish the cause.


documentation/ag-grid-docs/src/content/docs/column-sizing/_examples/continuous-auto-size-fit-grid-width/main.ts
```diff
@@ -62,30 +62,35 @@ const gridOptions: GridOptions<IRow> = {
     animateColumnResizing: true,
 };
 
-function setColumnCount(count: number) {
-    columnCount = Math.min(MAX_COLUMNS, Math.max(MIN_COLUMNS, count));
-    gridApi!.setGridOption('columnDefs', buildColumnDefs(columnCount));
+// Each handler must touch `gridApi` directly. Generation lifts the functions the HTML calls into
+// the component, where the api is in scope; a shared helper one level down stays at module scope
+// and its `gridApi` resolves to nothing in React, Angular and Vue.
```
StephenCooper: We don't want to have this comment in the public example code.

documentation/ag-grid-docs/src/utils/grid/test/waitForGridEvent.ts
```diff
@@ -0,0 +1,85 @@
...
+ * and docs examples do not register it. It therefore occupies the example's own `on<Event>`
+ * callback while armed, so avoid it for an event the example under test handles.
+ *
+ * ```ts
+ * const resized = await armGridEvent(page, 'columnResized', { finished: true, source: 'sizeColumnsToFit' });
+ * await page.locator('button.add-column-button').click();
+ * await resized.wait();
+ * ```
+ */
+export async function armGridEvent(
```
StephenCooper: Do we need to export this if its only used via the `withGridEvent` helper?

StephenCooper (approved):

PR #15198, opened 2026-09-11
AG-3390: align the chevrons and centre the docs header dropdowns

skewis6:
Aligns the chevrons on the docs header's AI Options and framework dropdowns, and centres both triggers' contents vertically.

https://ag-grid.atlassian.net/browse/AG-3390

Fix [AG-3390](https://ag-grid.atlassian.net/browse/AG-3390)


taktran (changes requested):
This will need to go into `b36.2.0`

PR #15206, opened 2026-09-14
Test(docs-e2e): cover the Editing examples' documented behaviour

sdwvit:
Audited every example under the Editing nav group against its docs copy and filled the assertions no spec exercised, as #15204 did for Accessories. 45 of 56 specs extended; green on all frameworks.

Fix [RTI-3471](https://ag-grid.atlassian.net/browse/RTI-3471)


documentation/ag-grid-docs/src/content/docs/provided-cell-editors-text/_examples/text-editor/example.spec.ts
```diff
@@ -51,4 +51,22 @@ test.agExample(import.meta, () => {
...
         await expect(cell).toContainText('abcdefghijklmnopqrst');
     });
+
+    // The docs list `useFormatter` alongside `maxLength`, but this example leaves it unset (the
+    // default, false). The observable consequence is that the editor is seeded with the raw value,
+    // not the '£ ' formatted one shown by the cell renderer.
+    test.eachFramework('the editor is seeded with the raw, unformatted value', async ({ page, agIdFor }) => {
+        await ensureGridReady(page);
+
+        // The value column is randomised, so read the rendered text rather than assuming a value.
```
StephenCooper, line 61: Our examples should be stable under test based on the SeedRandom that we have in the example generator. 

Seems like the skill is not aware of this and so could be writing more explicit tests perhaps?

StephenCooper (approved):

PR #15218, opened 2026-09-14
RTI-3444 Show selection count status bar in Select All Mode example

StephenCooper:
https://ag-grid.atlassian.net/browse/RTI-3444

Fix [RTI-3444](https://ag-grid.atlassian.net/browse/RTI-3444)


documentation/ag-grid-docs/src/content/docs/row-selection-multi-row/_examples/header-checkbox/example.spec.ts
```diff
@@ -1,5 +1,8 @@
 import { ensureGridReady, expect, test, waitForRowAnimations } from '@utils/grid/test-utils';
 
+const ROWS_IN_DATA_SET = 31;
+const GYMNASTICS_ROWS_IN_DATA_SET = 4;
```
StephenCooper, line 4: Missing test case for current page. That should be added.

sdwvit (approved):

PR #15221, opened 2026-09-14
[RTI-3475] Register date/number filter modules in tree data examples

alistairjcbrown:
https://ag-grid.atlassian.net/browse/RTI-3475
Fix RTI-3475


documentation/ag-grid-docs/src/content/docs/tree-data-opening-groups/_examples/expand-collapse-api/main.ts
```diff
@@ -16,7 +18,14 @@ if (process.env.NODE_ENV !== 'production') {
     enableDevValidations();
 }
 
-ModuleRegistry.registerModules([RowApiModule, ClientSideRowModelModule, TreeDataModule, TextFilterModule]);
+ModuleRegistry.registerModules([
```
StephenCooper, line 21: Can we get the example.spec.ts files updated so that they actually open these filters in a way that would have shown these were broken. They should probably set the filters and validate the filtering is correct for tree data while at it.

documentation/ag-grid-docs/src/utils/grid/test/groupRows.ts
```diff
@@ -0,0 +1,25 @@
+import type { Page } from '@playwright/test';
+import { expect } from 'playwright/test';
+
+/** The row whose group cell contains `name`. */
+export const groupRow = (page: Page, name: string) =>
```
StephenCooper: Does our testing not support this out of the box already? 

https://www.ag-grid.com/vue-data-grid/testing/#test-ids

StephenCooper (approved):

PR #15238, opened 2026-09-14
[RTI-3472] Refresh group rows when their aggregates change

alistairjcbrown:
# Refresh the cells that derive from aggregates

FIX RTI-3472
FIX RTI-3485

`setAggData` fires `cellChanged` per aggregated column, and `CellCtrl.onCellChanged` refreshes only the cell
whose own column the event names. A column reading those totals through a `valueGetter` never heard about the
change and kept the value it was first rendered with.

| scope  | net lines |   lines changed | hunks | files changed |
| ------ | --------: | --------------: | ----: | ------------: |
| source |       +82 | 372 (+227 -145) |    63 |     13 edited |
| tests  |      +713 |       713 added |     9 |         9 new |
| docs   |       +32 |        32 added |     2 |      2 edited |

## Cells deriving from aggregates repaint when a total moves ([RTI-3472](https://ag-grid.atlassian.net/browse/RTI-3472))

`AggregationStage.refreshAggregateDependentCells` repaints the rendered cells whose value can depend on an
aggregate: a `valueGetter` reading a total, or a Show Values As mode. Which rows are affected cannot be known
without running the getters, since one may read any ancestor's `aggData`, so it narrows by column and sweeps
every rendered row with `force: false`. It runs after the row redraw, so a row the pass created has a ctrl to
refresh. Both aggregation entry points reach it: the refresh pipeline, and the edit path from `endDeferred`,
which names the rows it already refreshed so they are skipped.

Show Values As had a narrower copy of this, removed here along with its sort and filter listener, which was a
second pass over rows the refresh pipeline already swept.

## Show Values As works without a grouping module

`ShowValuesAsModule` depended on `SharedAggregationModule`, which does not provide `AggregationStage` and so
leaves nothing to write `aggData`. Registering it without a grouping module was legal and left every cell
blank. It now depends on `AggregationModule`.

## Aggregate cell-changed events for a React row mounted a task late ([RTI-3485](https://ag-grid.atlassian.net/browse/RTI-3485))

`NormalRowFeature` registered its cell-ctrl listeners in `initialiseComp`, which React defers by a task, so a
model change landing in between was dropped and the row rendered the values its cells were constructed with.
Registration moves to `postConstruct`, which also stops the listeners stacking: `initialiseComp` runs once per
container, so a row with pinned columns registered them up to three times.

## Print layout no longer orphans the pinned controllers

Print layout flows every column through the centre lane, and `NormalRowFeature` replaced the two pinned lists
with empty ones without destroying the controllers they held, leaving them alive and unreachable for the life
of the row. They are destroyed now.

Behavioural cover for print layout comes with it, in both view layers, since it had none.

## A row rendered twice refreshes both copies

`refreshRowByNode` resolved the node through `getRowCtrlByNode`, which returns at most one controller, while a
group row can be rendered both as a sticky header and at its own index. It now refreshes the sticky copy too,
which also makes the post-edit sweep's row exclusion sound.

`FullWidthRowFeature` assigned its notes feature on every `initialiseComp` without releasing the previous one,
leaking up to two per embedded full-width row.

## Aggregate cell-changed events cost no allocation per row

`fireAggDataChangedEvents` took `Object.keys` of the new and old `aggData` on every group row with listeners,
then resolved each key back through `colModel.colsById`. The aggregation stage already holds the value columns
it keyed the result by, so it passes them down and the per-row loop walks that array directly. Detecting a
removed key still needs the old keys, but only on a pass where the value-column set changed, which
`valueColsSvc.columns` answers by identity.

## Flattened cell ctrls are cached

`NormalRowFeature.getAllCellCtrls` rebuilt its result on every call, allocating an array per row whenever any
column is pinned. It is read far more often than the lists change, so the result is cached and dropped when a
container list is replaced.


SalvatorePreviti:
moved to https://github.com/ag-grid/ag-grid/pull/15250

PR #15249, opened 2026-09-15
[AG-17609] Fix(columns): infer cellDataType from the first row holding a value

sdwvit:
Inference read row 0 only, so a column whose leading row was empty resolved to `cellDataType: false` and never recovered. It now scans rows until a value is found, and reads the arriving rows off the row model during an update so transaction-populated grids infer too.

Behaviour change: columns that previously fell back to `false` by accident now get typed editors, filters, formatters and comparators. `tree-data-edit-aggregation.test.ts` needed `NumberEditorModule` registered for that reason.

Fix [AG-17609](https://ag-grid.atlassian.net/browse/AG-17609)


packages/ag-grid-community/src/clientSideRowModel/clientSideRowModel.ts
```diff
@@ -53,6 +53,9 @@ export class ClientSideRowModel extends BeanStub implements IClientSideRowModel,
     /** True when grouping or tree data is active. Updated after grouping stage runs. */
     public hierarchical: boolean = false;
 
+    /** Row data of an update in progress, only set whilst `rowDataUpdateStarted` is dispatched */
+    public _updatingRowData: any[] | null = null;
```
StephenCooper: Do we need the new `_updatingRowData` field on `ClientSideRowModel` / `IClientSideRowModel`? Setting it around the dispatch and then casting `this.beans.rowModel` to read it back in the listener feels like a workaround.

`rowDataUpdateStarted` is an internal event (it's in `_INTERNAL_EVENTS` in `eventTypes.ts`, `RowDataUpdateStartedEvent` isn't exported from `main.ts`, and `DataTypeService` is its only listener), so we're free to change its payload. Suggest replacing `firstRowData: TData | null` with `rowData: TData[] | null` on the event, dispatching `data ?? null` from `ClientSideNodeManager`, and reading `event.rowData` in the listener. That drops the field, the set/unset, and the cast.

StephenCooper (approved):

PR #15251, opened 2026-09-15
[AG-18531] Feat(columns): infer cellDataType under every row model

sdwvit:
Fix [AG-18531](https://ag-grid.atlassian.net/browse/AG-18531).

`cellDataType` inference was gated on the Client-Side Row Model. It now runs under the Server-Side, Infinite and Viewport row models: each calls `dataTypeSvc.onRowsReceived()`, replacing the CSRM-only `rowDataUpdateStarted` trigger. Resolution stays one-shot, and a response with no rows leaves columns pending rather than writing them off.

Under SSRM row grouping, inference waits for a block of **leaf** rows. Group rows carry only their group key and aggregated values, so resolving off them would make every leaf-populated column permanently `false`.

Two things to flag for review:

- Group rows still count as "the grid has rows", because `isPendingInference` also queues filter model updates — leaving it set until a group is expanded stranded `setFilterModel` on grouped grids. Queued autosize operations release on the first rows for the same reason, which deviates from AC 6.2.2's wording.
- `filterDataTypeUtils` previously discarded `filterParams` when it was a function, so a Set Filter's declared `values` was lost once a data type applied. Fixed here (own commit) as typed SSRM columns expose it; it also changes CSRM behaviour and may want its own release-note line.

ACs 4.3 and 6.1.3 say "the first batch of rows" and should be amended to "leaf rows" to match.

Stacked on #15249 (AG-17609), which is a precondition.


sdwvit:
postponed till 37

PR #15269, opened 2026-09-17
ci(gh-triage): wire the construct-repro manual-trigger button

sabahang:
Adds the `gh-triage-manual-construct-repro` `repository_dispatch` event to the GitHub-issue AI-triage stub and routes it to `stage=construct-repro`, following the same pattern as the four existing manual AITGH buttons (Resume Analysis / Verify Repro / Rebuild Repro Link / Force Full Retriage).

This is the consumer-side wiring for a new, manual-dispatch-only `construct-repro` stage shipped in `@ag-grid/dev-prompts`'s `github-triage-pipeline` action — it fires only when no verified reproduction exists on a ticket, and (where constructible) builds one from scratch rather than forking an existing one. Like the other manual buttons, it bypasses the `→ In Progress` drag's routing entirely and is never auto-chained.

The corresponding Jira Automation "Manual trigger" rule for the fifth AITGH button still needs to be added by a project admin, same as every prior button in this list.


alantreadway (approved):
LGTM

PR #15280, opened 2026-09-17
AG-3390: Fix the announcement banner close button after a client-side navigation

skewis6:
The dismiss handler bound its click listener to the close button once, on load. Astro's `ClientRouter` swaps `<body>` on every client-side navigation, so the incoming page's banner is a fresh element with no listener and the already-loaded script is not re-executed — the close button only worked on the page you first hard-loaded.

Re-bind on `astro:page-load`, which fires once per navigation and on the initial load. This is the pattern the inline scripts in `inlineScripts.ts` already use for the same problem (`astro:after-swap`). The handler is a single hoisted function reference, so re-binding the same button is a no-op, and the banner id is read per click since a swap restores `<html>` to the incoming page's server-rendered attributes.

Verified against the live site with the patched script injected: dismissal works after one and after two client-side navigations, still works on a hard load, stays dismissed across a further navigation, and the banner body still follows its link.


taktran (commented):
As spoken about yesterday, I'm not sure we want to fix it with a global `document.addEventListener`, or whether we could hook into the astro swap event for view transitions

taktran (approved):

PR #15290, opened 2026-09-18
AG-18578 Re-pin test-case repros to staging when an agent PR merges

alantreadway:
Adds a `pull_request: closed` job to the AI Workflow stub that, when an agent PR merges, re-points the ticket's PR-pinned repro Plunkers from the per-PR gh-pages preview to the staging CDN. The job is gated on a merged `ghabot-*` head, writes only the test-case register's pin keys, and always exits 0.

https://ag-grid.atlassian.net/browse/AG-18578


alantreadway:
/pr-review

This job runs on a PR that has already merged, so its check gates nothing. A visible failed job is the only signal that a re-pin did not happen (the action posts no comment), and step-level tolerance would hide a bootstrap outage. Job-level continue-on-error keeps the workflow run green, which is the intent; the wording matches the ag-charts stub already on latest.

StephenCooper (approved):

PR #15303, opened 2026-09-21
AG-17765 Guard grid state against dropped ColumnState fields

StephenCooper:
https://ag-grid.atlassian.net/browse/AG-17765

Fix [AG-17765](https://ag-grid.atlassian.net/browse/AG-17765)


packages/ag-grid-community/src/misc/state/stateService.ts
```diff
@@ -489,9 +489,39 @@ export class StateService extends BeanStub implements NamedBean {
...
+            pinned: reset(shouldSetColumnPinningState),
+            hide: reset(shouldSetColumnVisibilityState),
+            flex: reset(shouldSetColumnSizingState),
+            headerName: reset(shouldSetHeaderNameState),
+            // `null` pivotSort would clear the colDef default ('asc') on every pivoted column; width/sortType are no-ops.
+            // Including to enable type checking to guard against missing properties in the future
+            sortType: undefined,
+            width: undefined,
+            pivotSort: undefined,
+        } satisfies Record<keyof ColumnStateParams, unknown>;
```
StephenCooper, line 523: Pulled all the existing values into the declaration. It is safe to define the properties as undefined as opposed to previously the default object just did not included them. undefined is documented as behaving like ignore in the column state.

testing/behavioural/src/grid-state/column-state-coverage.test.ts
```diff
@@ -47,16 +54,38 @@ const GRID_STATE_LOCATION = {
...
+        test.each(fields)('%s', (_field, [read, expected]) => {
+            const api = gridsManager.createGrid('myGrid', {
+                columnDefs: COLUMN_STATE.map(({ colId }) => ({ colId, field: colId })),
+                rowData: [],
+            });
+            api.setState(convertColumnState(COLUMN_STATE, true));
+
+            expect(read(api.getState())).toEqual(expected);
+        });
     });
```
alistairjcbrown, line 90: Do we need a reset test to cover the new function?
> `null` resets a field on columns the state omits; `undefined` leaves it untouched.

alistairjcbrown (approved):
One non-blocking comment, LGTM :shipit:

testing/behavioural/src/grid-state/column-state-coverage.test.ts
StephenCooper: Added in `790fa4a4f3a`: a full initial state clears the colDef defaults of every section on omitted columns, a partial one (`partialColumnState`) clears only the sections it carries, and `pivotSort` is left alone in both cases.

PR #15316, opened 2026-09-22
AG-18571 [Accessibility] Master grids should have treegrid role

gportela85:
fix [AG-18571](https://ag-grid.atlassian.net/browse/AG-18571)


packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
```diff
@@ -109,9 +109,19 @@ export class MasterDetailService extends BeanStub implements NamedBean, IMasterD
...
             row.master = newMaster;
             row.dispatchRowEvent('masterChanged');
+            this.dispatchLocalEvent({ type: 'masterChanged' });
         }
     }
 
+    public hasExpandableMasterRows(): boolean {
+        const rootNode = this.beans.rowModel.rootNode;
+        return !!(
+            rootNode?._leafs?.some(function isExpandableMaster(row) {
```
StephenCooper: Is there a reason this is not an arrow function?

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyStore.ts
```diff
@@ -431,6 +431,12 @@ export class LazyStore extends BeanStub implements IServerSideStore {
         });
     }
 
+    public hasExpandableMasterRows(): boolean {
+        return !!this.cache.getNodes().find(function hasExpandableMaster({ node }) {
```
StephenCooper: This looks like the same logic as above but there is as using `some` and here it is using `find`. Should these be sharing the logic?

packages/ag-grid-enterprise/src/masterDetail/masterDetailService.ts
gportela85: Nope, changed to arrow

packages/ag-grid-enterprise/src/serverSideRowModel/stores/lazy/lazyStore.ts
gportela85: I'd keep the dupe to avoid the serverSideRowModel module importing parts of the masterDatail module.

StephenCooper (approved):

PR #15332, opened 2026-09-23
AG-18617 Gate Space selection on enableClickSelection in AG Studio

alistairjcbrown:
https://ag-grid.atlassian.net/browse/AG-18617

Fix [AG-18617](https://ag-grid.atlassian.net/browse/AG-18617)


packages/ag-grid-community/src/selection/baseSelectionService.ts
```diff
@@ -414,13 +402,27 @@ export abstract class BaseSelectionService extends BeanStub {
...
             };
         }
     }
+
+    /** Whether Space must obey `enableClickSelection`. On a selection checkbox cell it acts like clicking the checkbox. */
+    private isSpaceKeyClickGated(node: RowNode, source: SelectionEventSourceType, column?: AgColumn): boolean {
+        return (
+            source === 'spaceKey' &&
+            _isInternalFeatureFlagEnabled(this.beans, 'spaceKeyFollowsClickSelection') &&
+            !(column && this.isCellCheckboxSelection(column, node))
```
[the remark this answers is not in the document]
alistairjcbrown: Fixed in 26982963e8e. The exemption now looks for a visible, enabled `.ag-selection-checkbox` in the focused cell or row, so it covers the auto group column, full-width group rows and custom renderers without repeating the rules for where checkboxes are drawn. Tests added for both group placements and for a disabled checkbox.

packages/ag-grid-community/src/selection/baseSelectionService.ts
```diff
@@ -292,11 +293,16 @@ export abstract class BaseSelectionService extends BeanStub {
...
+        const isClickGated = isRowClicked || this.isSpaceKeyClickGated(node, source, column);
 
-        if (isRowClicked && !(enableClickSelection || enableDeselection)) {
+        if (isClickGated && !(enableClickSelection || enableDeselection)) {
             return null;
         }
 
+        // whether `enableClickSelection` forbids moving the node to `newValue`
+        const isBlockedByClickSelection = (newValue: boolean) =>
+            isClickGated && (newValue ? !enableClickSelection : !enableDeselection);
```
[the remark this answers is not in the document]
alistairjcbrown: Leaving this as is on purpose. Space follows the same gates a row click does, and a click has the same gaps under `'enableSelection'` / `'enableDeselection'` (Shift-click range, filteredDescendants groups). Making Space stricter would make the two disagree, and changing click would change behaviour outside Studio. With `enableClickSelection: false`, which is what Studio uses, every path is already blocked by the early return.

packages/ag-grid-community/src/selection/baseSelectionService.ts
```diff
@@ -414,13 +403,51 @@ export abstract class BaseSelectionService extends BeanStub {
...
+        return (
+            source === 'spaceKey' &&
+            _isInternalFeatureFlagEnabled(this.beans, 'spaceKeyFollowsClickSelection') &&
+            !hasVisibleSelectionCheckbox(target)
+        );
+    }
+}
+
+/** Reads the DOM so every checkbox placement is covered without repeating the rules that decide where they go. */
+function hasVisibleSelectionCheckbox(element: EventTarget | null): boolean {
```
StephenCooper: This code feels very different to the normal patterns we follow? Is this really required?

documentation/ag-grid-docs/src/content/docs/row-selection-multi-row/index.mdoc
```diff
@@ -160,7 +160,7 @@ The checkbox column is sized statically, and is therefore not affected by [Auto-
 
 ## Enable Click Selection & Deselection
 
-The `rowSelection.enableClickSelection` property configures whether a row's selection state will be impacted when the row is clicked, or when {% kbd "␣ Space" /%} is pressed while the row is focused.
+The `rowSelection.enableClickSelection` property configures whether a row's selection state will be impacted when the row is clicked.
```
StephenCooper: This is correcting the docs to match the current behaviour? We are not changing the behaviour for normal usecase?

alistairjcbrown: Yes, the docs were wrong. `enableClickSelection` has never affected how space works for selection. We state the opposite in https://www.ag-grid.com/javascript-data-grid/row-selection-single-row/#enable-click-selection--deselection
> Note that deselection is still possible using the `␣ Space` key or when checkboxes are enabled by clicking a selected checkbox.

It looks like it was a hangover from `suppressRowDeselection` which did affect space, but was deprecated in v32.2 in favour of `enableClickSelection`

packages/ag-grid-community/src/selection/baseSelectionService.ts
alistairjcbrown: It ended up going this direction to cover dealing with  the auto-group column and full-width group support. Let me see if there's a nicer way of doing this using the model data.

